PR #21, opened 2014-09-14
This add fullscreen support on android devices 4.4 and later

digitaldan:
Signed-off-by: Dan Cunningham dan@digitaldan.com (github: digitaldan)


cribskip:
Just a note aside:
Fullscreen mode is left as soon as the menu was pressed - no action but reopening restores the fullscreen mode. Just a glitch ;)

Issue #5, opened 2014-05-29
Microphone widget


grzech1983:
Hello,

can you consider adding to habdroid microphone widget which will allow telling commands directly after pressing this widget without entering habdroid app?

Regards

cribskip:
This issue may be closed, I guess.

teichsta:
yep, seems so :)

PR #28, opened 2014-11-02
Refactoring of microphone widget from activity into own service

dominicdesu:
This pull request fixes the following issues mentioned in #5 :
- HABdroid is not opened anymore when the microphone widget is used
- Text overlay (Toast) is displayed with the recognized command
- Voice recognition works regardless of whether HABdroid is active on the background or not

It does not address the following suggestions (maybe a seperate ticket can be created for these):
- Do not include two on-click functions (execute voice recognition & openHAB) in the widget
- Enable widget for lock screen

A few words about the implementation:
Previously, the voice command functionality was included in the OpenHABMainActivity, which was called by an Intent. That's probably why the HABdroid GUI was always opened, when the voice widget was used. I couldn't find a way to execute voice recognition without the GUI being opened. Extracting the voice command functionality into an service (OpenHABVoiceService.java) solved this problem however.

The OpenHABVoiceService is called from two places: The microphone menu icon inside HABdroid, and the microphone widget.
- In case the service is called from the menu icon, the openHAB URL can be passed in the Intent.
- In case the service is called from the widget, it uses the OpenHABTracker to determine the URL of the openHAB Server. This was a little bit tricky because OpenHABTracker uses asynchronous autodiscovery if the openHAB URL is not configured. (For this reason, the ContinuingIntentService buffers the voice commands until the openHAB URL is discovered.)

I also had to adjust the AsyncServiceResolver a little (besides converting tabs to spaces) because it is called indirectly also from the service. This cast would then not work anymore: ((Activity) mCtx).runOnUiThread(), as the mCtx is not an activity.

I'd be happy to hear about any feedback.

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -57,6 +57,10 @@
             android:label="@string/title_activity_openhabinfo" >
         </activity>
         <activity android:name="de.duenndns.ssl.MemorizingActivity" />
+        <service
+            android:name=".core.OpenHABVoiceService"
+            android:exported="false">
+        </service>
         <receiver
             android:name="org.openhab.habdroid.core.GcmBroadcastReceiver"
             android:permission="com.google.android.c2dm.permission.SEND">
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
@@ -1,0 +1,218 @@
+/**
+ * Copyright (c) 2010-2014, openHAB.org and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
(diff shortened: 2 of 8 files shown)


cribskip:
Works very well, thanks for your efforts.
Tested:
  - Habdroid running in background
  - habdroid not running in background
  - with 3G / WIFI
  - recognition from app & widget

Very nice is the new toast on recognition +1

Just a note aside:
Fullscreen mode is left as soon as the menu was pressed - no action but reopening restores the fullscreen mode. Just a glitch ;)

dominicdesu:
Hi, thanks a lot for the feedback and your extensive tests!

I don't understand your side note: how do you activate fullscreen and what
menu do you press?

Am 28.01.2015 22:26 schrieb "cribskip" <notifications@github.com>:
>
> Works very well, thanks for your efforts.
> Tested:
>
> Habdroid running in background
> habdroid not running in background
> with 3G / WIFI
> recognition from app & widget
>
> Very nice is the new toast on recognition +1
>
> Just a note aside:
> Fullscreen mode is left as soon as the menu was pressed - no action but
reopening restores the fullscreen mode. Just a glitch ;)

cribskip:
I'm sorry - in my crazy brain this got mixed up with #21. This Issue/PR is really just great.

PR #55, opened 2015-02-06
Indicate that HABDroid also works on devices without GPS

### LINKED ISSUE #33: On my Gigaset SL930 the app habDroid/openHAB is not shown in Google Play Store
I have a new home phone, which I want to use as remote control for openHAB. Unlike on my other Android devices, the new Gigaset SL930A does not show habDroid in the Google Play Store. On my PC I see the SL930A shown as one of my devices in gray with a message that habDROID is not suitable for this device. I do not see why. It has no mobile connection, but a steady WiFi connection and Android 4.04. more specifications can be found here: 
http://www.gigaset.com/en_UK/shop/gigaset-sl930a.html
Can this be solved ?

Gert

dominicdesu:
This pull request should fix #33 . I cannot test it because I don't have a device without GPS.
Problem is that HABDroid uses permission ACCESS_FINE_LOCATION. This implies (see http://developer.android.com/guide/topics/manifest/uses-feature-element.html#permissions) that the android.hardware.location.gps feature is required to run HABDroid. As this is not the case, this feature should be marked as required false.

@belovictor: For which feature of HABDroid is ACCESS_FINE_LOCATION necessary? If it is not necessary, it would make more sense to remove this permission request. Let me know then I could update the pull request.

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -22,6 +22,9 @@
     <uses-permission android:name="org.openhab.habdroid.gcm.permission.C2D_MESSAGE" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-feature
+        android:name="android.hardware.location.gps"
+        android:required="false" />
+    <uses-feature
         android:name="android.hardware.nfc"
         android:required="false" />
     <application


cribskip:
Tested on my el-cheapo old chinese crap tab, seems to work nice.

@dominicdesu I think my.openhab will integrate location-based services as the website already shows 'last location' for devices.

belovictor:
@cribskip is right. There is infrastructure and code in the place already to provide location based features  with mobile apps and my.openHAB but it is not yet merged into Android and iOS apps. Thanks for this fix!

Issue #56, opened 2015-02-08
Atmosphere sessions running systems out of memory


digitaldan:
hey @belovictor !  can you take a look at the this comment https://github.com/openhab/openhab/issues/1961#issuecomment-73418700 ?  Basically on system where items are changing every second we are running out of memory on the server due to too many atmosphere sessions.  I see there was a change to alternate long polling requests, is there another solution?  I'm just trying to patch this together until OH 2 comes out and we can switch out atmosphere for something else.  Thanks!

digitaldan:
@belovictor have you had a chance to look at this yet?  I was hoping to try and fix this over the weekend. Thanks!

belovictor:
This change makes a non-long-polling request after every long-polling request is finished. This was the only way I managed to make it work at all. Without that I never got updates from openHAB for changing items basically. I can change this back to normal, but will it work on the openHAB side?

digitaldan:
It works for me ok, as a reference I use the demo setup with some additional rules that turn everything on and off every second.  Let me get a clean environment setup and do some more testing to really make sure.  Is the app in the store up to date with master?

belovictor:
This is fixed in the linked pull request.

PR #62, opened 2015-02-21
Change to use long polling after we successfully have a connection, set ...

digitaldan:
...our atmopshere to null to make sure we don't reuse one from a previous session.  This is in regards to #56 .  @belovictor I have been running this all week and have not had any issues, could you try out and let me know.  Someone was kind enough to do the work to bring Atmopshere 2.2.x in OH, so I will be testing against that as well this weekend, hopefully this clears a few of our problems up!


digitaldan:
Hey @belovictor , I have been testing this out on a few devices for the last week without any issues.  Can you give it a try?

PR #64, opened 2015-02-25
Ability to add widget to lockscreen

cribskip:
This small change enables the users to add the widget to the lock screen.

First pull for me so please comment if anything's wrong, thanks :)


cribskip:
Google's Way is not to support widgets on lock screen anymore (5.0+). So... closing

PR #65, opened 2015-02-25
Fixes for I18N

cribskip:
Fixes some missing translations for German and i18n-ized preferences menu.


cribskip:
Integrated in #66

PR #66, opened 2015-02-26
Better UX with settings

cribskip:
Hi,

this PR does the following:
* Moves 'clear image cache' from the main menu to the settings menu
* Groups/Restructures the settings menu for better UX (disable URL settings when using demo mode...)
* Fixed a minor problem with I18N

As reviewed by my parents, the settings are now understandable ;-)

Cheers,
Sascha


8666:
Clear images cache is my most used command. Why u r hiding it?

cribskip:
Thanks for the feedback. I've put it into settings as it is more of a setting (see f.e. Browsers) than a normal operation as cache should be clever enough (and in most cases is) to be self-organizing. 

Are you having trouble in your setup when changing icons and it's not updating itself? 

Cheers

cribskip:
Don't waste your time on this - my device environment is completely messed up. I'll fix and resubmit the pull request.

PR #69, opened 2015-03-15
Handle volume Up/Down buttons

eschava:
Volume up/down buttons are handled by first slider/setpoint widget on current frame


cribskip:
Does this harmonate together with the dynamic sitemap handling? 

Thanks for the feature, thought about it also several times😊

eschava:
Unfortunately I cannot find how dynamic visibility is handled by android app. Maybe hidden items are just not present in the frame so it should work OK in this case

kaikreuzer:
Dynamic visibility is handled on server side, not on client side. So for HABDroid, they hidden widgets are just as if they didn't exist.

belovictor:
I've integrated this feature manually, thank you very much for your contribution!

Issue #87, opened 2015-04-04
dark theme is missing


Human:
Commit 4ba5c6dc1fb7c693b76c52c602520a45808bbf6f removed the dark theme and the ability to choose it.

belovictor:
This was done by intension cause I think that our app looks ugly in dark material design. Do you think it looks fine?

Human:
I personally use it at night and for a dedicated home theater control device (old phone, mounted on the wall). I think having the choice is better than not having it.

belovictor:
Well, I'm not against it. Can you post screenshots of how Dark Material looks like?

Human:
![screenshot_2015-04-05-08-56-05](https://cloud.githubusercontent.com/assets/40866/6997039/d899691c-db78-11e4-92f5-4db9bff6a44d.png)

![screenshot_2015-04-05-08-56-11](https://cloud.githubusercontent.com/assets/40866/6997040/e14149fe-db78-11e4-9141-d51273e3d74a.png)

belovictor:
Sorry for bothering, but can you post a screenshot of first page in demo mode?

Human:
![screenshot_2015-04-05-12-27-40](https://cloud.githubusercontent.com/assets/40866/6997600/3d471fd8-db8f-11e4-9b62-6015dc8ba152.png)

cribskip:
Separators should be colored differently IMHO

kaikreuzer:
Yeah, possibly black separators might be nice. I'd also vote to change the orange theme to something less flashy - possibly some grey shades might be a good fit, if this is possible.
In general, I agree that a black theme is a valuable option to have.

belovictor:
Do you mean changing orange for dark theme or for both of them?

kaikreuzer:
Only for dark theme.

teichsta:
> Yeah, possibly black separators might be nice. I'd also vote to change the orange theme to something less flashy

robnielsen:
You can find material color palette at http://www.google.com/design/spec/style/color.html#color-color-palette

belovictor:
As pull request is merged we can now take dev snapshot and discuss the colors.

kaikreuzer:
> You can find material color palettes at http://www.google.com/design/spec/style/color.html#color-color-palette

I'd go for grey for the dark theme.

belovictor:
I'm afraid that grey will break visual identity cause orange is a natural color of openHAB :-(

[...]

PR #88, opened 2015-04-04
Theme choice

Human:
For issue #87


Human:
I have no idea why it keeps including those previous commits, sorry. I'll see if I can clear that up.

PR #91, opened 2015-04-04
Fixed a build error Android Studio on Windows

robnielsen:
Encountered a problem similar to http://stackoverflow.com/questions/21645961/android-studio-processdebugresources-failed. Fixed by upgrading to the latest version of build tools for 21


belovictor:
Thanks!

Issue #93, opened 2015-04-05
Clientside audio playback


tobiasamon:
Is there a way to play audio streams on the client side at the moment? For example to play a sound in habdroid when an event on the server is fired?
##

cribskip:
I was working on TTS triggered by openhab but lost my path. Currently, there is no such thing

tobiasamon:
So is the OpenHAB server able to notify any client (and also habdroid) that a sound has to be played?

belovictor:
No, there is currently no standard way to do this in openHAB Android app. The only thing which comes to my mind is to use notifications through my.openHAB together with some app which would pronounce text of those notifications, like this one:
https://play.google.com/store/apps/details?id=com.pilot51.voicenotify&hl=en
Then you can send notification from openHAB through my.openHAB to your Android device and it will be spoken by this app.

cribskip:
I just drafted together a prototype for you. It enables habdroid to read aloud what came as a notification if the messages starts with "speak ".

Didn't yet create a pull request as I'd first like to discuss this one. Also, your initial request was to play "an audio stream" - can you specify your request a little bit to me? Would you like to playSound a sound from openhab (like in /sounds) to your habdroid?

tobiasamon:
My idea was that habdroid (and also any browser) should be able to play a file from a remote server (the openhab server). Audio files would have to be accessible through the webserver. Habdroid and browser should play the audio file when they receive a notification.

If you share your prototype, I could check if it fits my needs and maybe we can include it in habdroid.

kaikreuzer:
This sounds like a bigger topic that might actually fit well into https://bugs.eclipse.org/bugs/show_bug.cgi?id=463802 - i.e. habdroid could be a sink for audio streams that can be used within rules. I hence wouldn't necessarily see this in openHAB 1.x.

cribskip:
So, after a few years, I guess this should be made possible by using the webaudio backend just like HABpanel and Paper UI.

If I find some time, I'll look into this and post a PR.

cribskip:
Just investigating on how the WebAudio works for habpanel:

1. Subscribe to <OH URL>/rest/events
2. Listening for topic "smarthome/webaudio/playurl"
3. Downloading the URL (payload) and playing the WAV

Example of message (TTS to webaudio sink):
```
event: message
[...]

PR #94, opened 2015-04-06
Feature read notifications

cribskip:
This will read aloud notifications sent by sendNotification, partially enabling what #93 requested.
If media volume is muted, this will not read what's sent to respect user's privacy.

usage: sendNotifaction("speak Somebody opened the frontdoor.")
Text after "speak" will be spoken and this will be displayed in the notification.

Note aside: I stumbled across that GCM introduces a whitespace in the beginning of a notification.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/GcmIntentService.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/GcmIntentService.java
@@ -22,6 +22,11 @@ import android.media.RingtoneManager;
 import android.net.Uri;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
+import java.util.Locale;
+import java.util.concurrent.TimeUnit;
+
+import android.speech.tts.TextToSpeech;
+import android.speech.tts.TextToSpeech.OnInitListener;
 import android.support.v4.app.NotificationCompat;
 import android.util.Log;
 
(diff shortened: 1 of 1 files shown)


belovictor:
This is cool, I would propose to change it a little bit though. Let's use another event for spoken notifications. I will create a separate method in my.openHAB bundle and in the cloud which will be forwarded to Android app and will be spoken instead of showing in notifications. GCM messages from my.OH have a field named 'type' which has value of 'notification'. We can use 'speak' instead. Can you rewrite the code for that?

cribskip:
Sounds nice, will do when I find time. 

Maybe two functions are sensible: speak and show-and-speak. This would prevent double transmits and save traffic for mobile data plans.

Btw, I saw sendNotification supports icons but habdroid doesn't. Shall I create an issue?

belovictor:
No, not yet. It's not clear yet where it should load those icons from :-)

cribskip:
I updated my branch and quickly tested with the emulator, worked just fine.

cribskip:
I will have to rebuild everything, syncing turned out something about gradle.
Shall I create a new pull request when done?

Sascha

digitaldan:
@cribskip  My guess is that the work needed in my.openhab has been put on hold.  What would you like to do with this PR?

digitaldan:
Closing due to inactivity.

PR #99, opened 2015-04-16
created custom thumb image for the handle of switchCompat widget

curlyel:
Created a a custom thumb image replacing the image which represents the handle of the switch widget. This image doubles the width of the widget to please handling it. See attached screen copy from android studio...
layout-2015-04-16-163853


belovictor:
Thanks, but we would like to stick with material design
http://www.google.com/design/spec/components/sliders.html#sliders-continuous-slider
Would be nice if you do a slider showing percentage though! :-)

Issue #90, opened 2015-04-04
make switch touch targets full-screen


Human:
For simple binary switches, the "button" can actually be the full width and height of the cell containing the icon, text, and button.
##

curlyel:
I'd like to encourage you improving the size of the controls too. I'd love to see wider sliders and buttons (at least on tablets in landscape mode). A seven inch screen is big enough to control my home usually. But in recent versions (using 1.6.0.4 currently) the switch controls are way too small for my finger tips ;-)

btw.: Thanks for removing the GPS requirement. It maked habdroid available for my android desk phone.

belovictor:
@Human , I don't clearly understand what do you mean by this description :-)

cribskip:
@belovictor IMHO he'd like to be able to touch either the icon, label or switch and trigger the switch of the item. Also, sliders should stretch to allow more precise control, especially on bigger screens.

Human:
Actually, I'd delegate the touch detection to the whole row.

curlyel:
I've just created a custom image representing the switch handle of binary switch widget. This image actually doubles the width of the widget. 

@belovictor I'd appreciate if you could check it.

Here is a screen copy out of android studio how the binary switch will look like if merged:
![layout-2015-04-16-163853](https://cloud.githubusercontent.com/assets/11849305/7183542/5a079e0a-e458-11e4-90a1-e3cf8e22f1e8.png)

Human:
In the attached image, the row I outlined in yellow would be the button for the tivo, not just the logo, and not just the switch itself. If you're interested, I'll see if I have time this weekend to submit a patch to make that change. (I've made it once before locally.)

![openhab](https://cloud.githubusercontent.com/assets/40866/7212643/77912848-e538-11e4-915f-95a78713c48f.png)

curlyel:
@Human Using the whole row for toggling an item slightly breaks the ui concept of the app from my point of view. Touching the row usually opens a sub pane or group item. When trying to send a command then you have to touch or draw directly at the widget's graphic currently. This is quite intuitive. 

If changing that behaviour to the whole row you may toggle an item accidentally when just trying to scroll the screen. Depending on the item this may be more or less severe.

Therefore I'd prefer to stick a the current behaviour.

curlyel:
[...]

PR #100, opened 2015-04-18
the entire row containing a switch acts as the switch

Human:


belovictor:
I think this is not really a good idea cause I wanted to use clicks on items for openHAB 2.0 functionality (configuration, group assignment and so on)...

xsnrg:
Given the comment and reasoning for not clicking the whole bar, this could probably be closed.

zacharyyy:
@belovictor, why not do the configuration, assignment, etc in a contextual popup menu when a row is long pressed?

Then a single click on a row could act as a switch.

digitaldan:
@Human as discussed in #90 I think this makes sense, are you able to fix the conflicts and re-submit?

kaikreuzer:
> why not do the configuration, assignment, etc in a contextual popup menu when a row is long pressed?

I agree (and that's what MD UX asks for as well).

digitaldan:
So I cleaned up the conflicts, and i think I see what the issue is, the way this is listening for widget presses is not compatible with how the rest of the app needs to listen for long widget presses (for the extras features mentioned above). This means switch widgets would lack the extra functionality that all other widgets have (we would be taking away functionality).   It's probably not hard to make it compatible, but will definitely require some more work.

digitaldan:
Closing due to inactivity

thjunge11:
I have the problem that the OnTouchListener does not work with any Android device that does not have touch input (like Android fire TV), so the App is kind of useless on these devices as you are not able to operate the swichh elements (and the others).

One could attach an OnClickListener in parallel to a onTouchListener to every clickable type and ignore the event if (isInTouchMode() == true).

What do you think? I would like to participate in coding here.

PR #110, opened 2015-05-27
Fix typos; Add missing translation

schinken:


schinken:
Hm actually it's not my change that breaks the build ;) https://buildhive.cloudbees.com/job/openhab/job/openhab.android/15/console

PR #144, opened 2015-11-22
Clearing recycled chart image

TheNetStriker:
I found an issue when loading chart images. If it takes some time for the image to load the image from a recycled image view is displayed. After some time the correct image loads, but it’s confusing because it
looks like the image has already been loaded. I fixed the problem by setting the image drawable to null before loading the image url.

I also added multidex support because I couldn’t build the app without it and Android Studio removed two unused imports.

I’ve already tested this and is seams to work very good. Please merge this to your app.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/OpenHABNotification.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/OpenHABNotification.java
@@ -11,6 +11,7 @@ import org.json.JSONObject;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.TimeZone;
 
 public class OpenHABNotification {
     private String mMessage;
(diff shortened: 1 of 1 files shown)


CristianSarac:
Since `OpenHABWidgetAdapter.java` need a lot of refactoring this fix is OK for now

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -486,6 +484,8 @@ public void onStopTrackingTouch(SeekBar seekBar) {
    		break;
     	case TYPE_CHART:
     		MySmartImageView chartImage = (MySmartImageView)widgetView.findViewById(R.id.chartimage);
+    		//Always clear the drawable so no images from recycled views appear
+    		chartImage.setImageDrawable(null);
```
digitaldan: This seems reasonable, but the PR contains other changes not related to the issue.  Can you submit only this change?  We will hopefully have the multidex supported merged soon.  Also please sign your commit message according to https://github.com/openhab/openhab-distro/blob/master/CONTRIBUTING.md .  Thanks!

TheNetStriker:
Sorry, something didn't work when re-commiting the changes. I've had to create a new pull request: https://github.com/openhab/openhab.android/pull/167

PR #154, opened 2016-03-05
153 color items state

### LINKED ISSUE #153: Switch associated to a color item is OFF when ON is expected
I have this item:

```
Color LampeTeleviseur "Lampe TV" {hue="1"}
```

And this in my sitemap:

```
        Text label="Eclairage" {
            Switch item=LampeTeleviseur
        }
```

When I start HABdroid and the physical color light is ON, the switch is in a wrong state OFF.
Then consider that the color light is OFF and the switch in HABdroid is OFF. I switch it to ON in HABdroid, the light switch ON as expected but one second after, the switch in HABdroid switch to OFF (while the physical light remains ON).
The same item and sitemap are working well in Classic UI.

openHAB 1.8.1
HABdroid 1.7.0.1
##

hazzeh:
Fixes #153 
Also update Android gradle plugin to version 1.5.0

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -40,4 +40,6 @@ dependencies {
     compile 'com.google.code.gson:gson:2.3.1'
     compile 'com.github.shell-software:fab:1.1.0'
     compile 'com.android.support:multidex:1.0.1'
+    testCompile 'junit:junit:4.12'
+    testCompile 'org.mockito:mockito-core:1.10.19'
 }
--- a/mobile/src/main/java/org/openhab/habdroid/model/OpenHABItem.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/OpenHABItem.java
@@ -23,6 +23,9 @@ import org.json.JSONObject;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 /**
  * This is a class to hold basic information about openHAB Item.
  */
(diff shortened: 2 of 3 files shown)


digitaldan:
HI @hazzeh, this adds HSB support  for android, do you know if the openHAB web client or IOS app supports HSB as well? I actually don't know as I have never tried to use hsb values.

hazzeh:
Hi,
The webui (openhab2) supports the HSB values (for color items at least). I've not tried IOS since I don't have access to one.
The author of #153 mentioned that openhab 1.8 classic web ui works with HSB.

Also I added a test dependency to junit and mockito, I hope that is ok?

digitaldan:
Thanks for the clarification, I think the unit test dependencies should be ok since there is not another testing framework in place, also for reference this is what google suggests to include as well  http://developer.android.com/training/testing/start/index.html  .  Thanks for including a test class, that's awesome.    I'll pull this down shortly to test.

hazzeh:
I saw, that I missed to add testInstrumentationRunner in build.gradle, if you wait to pull it, I can update my pullrequest.
Otherwise I can add it in the future.

kaikreuzer:
> I actually don't know as I have never tried to use hsb values.

To clarify: Actually, the REST API (which is used by Android&iOS) ONLY supports HSB (and no RGB) values for color states/commands.

digitaldan:
I will try and pull this down tonight and test, thanks!

digitaldan:
Just wanted to comment, I have not forgot about this ;-) I have a few other things on my plate this week, but my plan is to test and merge this by the weekend.

hazzeh:
No worries. Btw I think that the sdk needs to be updated on the jenkins build slaves.
https://openhab.ci.cloudbees.com/job/HABDroid/79/console

digitaldan:
Thanks @hazzeh for pointing that out.  I'll put that on my list of things to fix this week.

PR #158, opened 2016-03-15
MultiDex support

slawekjaranowski:
Fix #147, #155

PR #159, opened 2016-03-16
Refresh improvment

slawekjaranowski:
I hope this help some of refresh issues.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetListFragment.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetListFragment.java
@@ -404,45 +404,32 @@ public class OpenHABWidgetListFragment extends ListFragment {
      * @return      void
      */
     public void processContent(String responseString, boolean longPolling) {
+
+        Log.d(TAG, "processContent() " + this.displayPageUrl);
+        Log.d(TAG, "isAdded = " + isAdded());
+        Log.d(TAG, "responseString.length() = " + (responseString != null ? responseString.length()  : -1));
+
+        // We can receive empty response, probably when no items was changed
+        // so we needn't process it
+        if (responseString == null || responseString.length() == 0) {
+            showPage(displayPageUrl, true);
+            return;
+        }
+
         // As we change the page we need to stop all videos on current page
         // before going to the new page. This is quite dirty, but is the only
         // way to do that...
-        Log.d(TAG, "processContent() " + this.displayPageUrl);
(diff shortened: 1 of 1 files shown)


slawekjaranowski:
#158 is still needed

digitaldan:
Hi again, before I pull down and test this, what is the refresh issue that this was solving?  Is there a bug opened that this solves?  Thanks.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetListFragment.java
```diff
@@ -412,49 +412,49 @@ public void processContent(String responseString, boolean longPolling) {
...
         // If openHAB verion = 1 get page from XML
-        if (mActivity.getOpenHABVersion() == 1) {
-            DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
-            try {
-                DocumentBuilder builder = dbf.newDocumentBuilder();
-                Document document = builder.parse(new InputSource(new StringReader(responseString)));
-                if (document != null) {
-                    Node rootNode = document.getFirstChild();
-                    openHABWidgetDataSource.setSourceNode(rootNode);
+        if (responseString != null && responseString.length() > 0) {
```
digitaldan: If the response string is null or empty what happens?  Line 433 handles the case where the xml document is null and I believe tries to reconnect. should something similar happen here as well?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetListFragment.java
```diff
@@ -412,49 +412,49 @@ public void processContent(String responseString, boolean longPolling) {
...
+            if (mActivity.getOpenHABVersion() == 1) {
+                DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+                try {
+                    DocumentBuilder builder = dbf.newDocumentBuilder();
+                    Document document = builder.parse(new InputSource(new StringReader(responseString)));
```
slawekjaranowski: When we have empty response "builder.parse" throws exception, so next statement isn't executed

slawekjaranowski:
I simplify my change ... and add some comments in code.

There is no issue for this, when I was testing my other change I had some stack trace form time to time  - caused by empty response from OH.

In this case I think it is ok to stop processing of refresh page.
Maybe it is issue somewhere else.

digitaldan:
Thanks @slawekjaranowski , I'll pull this down in a bit and test.

PR #162, opened 2016-03-20
Refreshed images should have authorization.

slawekjaranowski:
When image or chart is refreshed should has authorization.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/MySmartImageView.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/MySmartImageView.java
@@ -17,19 +17,43 @@ import android.util.Log;
 
 import com.loopj.android.image.SmartImageView;
 
+import java.lang.ref.WeakReference;
+import java.sql.Ref;
 import java.util.Timer;
 import java.util.TimerTask;
 
 public class MySmartImageView extends SmartImageView {
-	private String myImageUrl;
-	private Timer imageRefreshTimer;
-	
-	boolean useImageCache = true;
 
-	public MySmartImageView(Context context) {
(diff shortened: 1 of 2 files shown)


digitaldan:
Hi @slawekjaranowski , thanks (again).  I'll try and get this pulled down soon for testing.  From looking at the changes you made, authentication was being set on the initial request but not when the refresh handler was triggering a request, is this correct?

slawekjaranowski:
Yes, only first request has authorization, requests from handler don't have it.

digitaldan:
hi @slawekjaranowski, looks good so far. I found one issue where the refresh timer didn't get canceled when my device auto locked (the rest of the app did pause), and it ended up running all night.  I'm not sure if this is an existing issue or related to your changes, do you have any thoughts?  My guess is it's an existing issue, but can you take another look just to be sure?

slawekjaranowski:
Hi, I think that is existing issue `cancelRefresh` method in `MySmartImageView` is called from `OpenHABWidgetAdapter`.

I don't see way to recreate `imageRefreshTimer` from class `MySmartImageView`.

One place where I see that application don't stop after `onPause` method is `OpenHABWidgetListFragment`.
`OpenHABWidgetListFragment.onPause` call `mRequestHandle.cancel` but when  request is during processing can cause some interrupt errors which is handled by `mRequestHandle.onFailure` or `processContent` and call again `showPage` which create new  `mRequestHandle`.
And after this processing is not stopped.

digitaldan:
Ok thanks, I'll take a look. It might be as simple as checking if the mRequestHandle is null or canceled when calling show page in the onFailure callback.

PR #165, opened 2016-03-28
Enable display of proper rollershutter control for rollershutter groups.

rmichalak:
Thanks to https://github.com/eclipse/smarthome/pull/1253 development we now can distinguish between switch and rollershutter item and display proper control widget for the group.

Bug: https://github.com/openhab/openhab.android/issues/141
Signed-off-by: Robert Michalak <rbrt.michalak@gmail.com>

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/OpenHABNFCActionList.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/OpenHABNFCActionList.java
@@ -39,7 +39,8 @@ public class OpenHABNFCActionList {
 					actionCommands.add("OFF");
 					actionNames.add("Toggle");
 					actionCommands.add("TOGGLE");
-				} else if (openHABWidget.getItem().getType().equals("RollershutterItem")) {
+				} else if ("RollershutterItem".equals(openHABWidget.getItem().getType()) ||
+						("Rollershutter".equals(openHABWidget.getItem().getGroupType()))) {
 					actionNames.add("Up");
 					actionCommands.add("UP");
 					actionNames.add("Down");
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/model/OpenHABNFCActionList.java
```diff
@@ -39,7 +39,8 @@ public OpenHABNFCActionList(OpenHABWidget openHABWidget) {
 					actionCommands.add("OFF");
 					actionNames.add("Toggle");
 					actionCommands.add("TOGGLE");
-				} else if (openHABWidget.getItem().getType().equals("RollershutterItem")) {
+				} else if ((openHABWidget.getItem().getType().equals("RollershutterItem")) ||
```
slawekjaranowski: consistently this line should be:

    "RollershutterItem".equals(openHABWidget.getItem().getType())

slawekjaranowski:
Does this work with 1.x version of OpenHab?

rmichalak:
Modified code according to your comment. The code works with modifed smarthome runtime, which I believe is used from 1.9 version of openhab. For older versions of OH, 'groupType' field will not be present in the REST API response and thus standard switch control will be displayed for rollershutter groups (exactly the same behaviour as it is now, without this fix).

digitaldan:
Hi @rmichalak , I will try and review in the next few days, thanks!

digitaldan:
Just wanted to comment, I have not forgot about this ;-) I have a few other things on my plate this week, but my plan is  review this by the weekend.  Thanks.

PR #167, opened 2016-04-07
Clearing recycled chart image

TheNetStriker:
I found an issue when loading chart images. If it takes some time for
the image to load the image from a recycled image view is displayed.
After some time the correct image loads, but it’s confusing because it
looks like the image has already been loaded. I fixed the problem by
setting the image drawable to null before loading the image url.

I also added multidex support because I couldn’t build the app without
it and Android Studio removed two unused imports.

I’ve already tested this and is seams to work very good. Please merge
this to your app.

Signed-off-by: David Masshardt <david@masshardt.ch> (github:
TheNetStriker)


digitaldan:
Thanks @TheNetStriker, I will test it here shortly.

digitaldan:
Looks good, thanks!

PR #172, opened 2016-04-20
Fix lint problems

### LINKED ISSUE #171: Lint errors causes build failure
When building with "gradlew build". There are some new lint errors.

1. 
A removed activity is still present in the manifest. OpenHABInfoActivity which was changed to a fragment in #161.

2. 
The log TAG exceeds 23 characters. There is a new lint rule that checks for long log tags to comply with the log api: http://developer.android.com/reference/android/util/Log.html.

1 can be corrected by removing the entry from the manifest. In the same time maybe changing the log tag to current class name?

2 can be "solved" by adding
lintOptions {disable 'LongLogTag'} to build.gradle.
But imo it is better to fix the root cause of the problem: Shortening the log tag. 
The following two classes have to long log tag:
ConnectivityChangeReceiver and NotificationDeletedBroadcastReceiver

Any suggestions how to shorten the tag (or class as well)?

On another note. There is a lot of classes prepended with OpenHAB, that can be shortened to OH to save some characters ;-)

hazzeh:
Fixes #171 
Also specify android wearable support version
This will also fix the "AAPT: No resource found that matches the given name" shown in https://openhab.ci.cloudbees.com/job/HABDroid/98/console

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/ConnectivityChangeReceiver.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/ConnectivityChangeReceiver.java
@@ -16,7 +16,7 @@ import android.os.Bundle;
 import android.util.Log;
 
 public class ConnectivityChangeReceiver extends BroadcastReceiver {
-    private static final String TAG = "ConnectivityChangeReceiver";
+    private static final String TAG = ConnectivityChangeReceiver.class.getSimpleName();
 
     @Override
     public void onReceive(Context context, Intent intent) {
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -63,10 +63,6 @@
             android:name="org.openhab.habdroid.ui.OpenHABWriteTagActivity"
             android:label="@string/title_activity_openhabwritetag" >
         </activity>
-        <activity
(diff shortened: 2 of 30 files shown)


mobile/src/main/java/org/openhab/habdroid/core/ConnectivityChangeReceiver.java
```diff
@@ -16,7 +16,7 @@
 import android.util.Log;
 
 public class ConnectivityChangeReceiver extends BroadcastReceiver {
-    private static final String TAG = "ConnectivityChangeReceiver";
+    private static final String TAG = ConnectivityChangeReceiver.class.getSimpleName();
```
hazzeh: Maybe this should be done in all classes for consistency?

zacharyyy: I think so. The lint warnings about the TAG being too long will go away and if you refactor the class name, the tag will be updated too.

hazzeh: I've pushed an updated commit now.
All classes that uses logging has been updated.

digitaldan:
Thank for the PR , I'll test this asap.

Issue #174, opened 2016-04-22
Entering invalid local url causes OpenHAB app to die


JosSchering1:
Changing local url from https://10.70.6.14:8443 to https://10.70.6.14:84438 causes OpenHAB to die and shows the error "Unfortunately, OpenHAB has stopped."
The reason is an unhandled exception in method checkUrlReachability in OpenHABTracker.java. 
The IllegalArgumentException exception is unhandled and causes the OpenHAB to die.

JosSchering1:
Installed HABDroid #108 from https://openhab.ci.cloudbees.com/job/HABDroid/lastSuccessfulBuild/artifact/mobile/build/outputs/apk/  which has the change in it and confirmed that entering an invalid local URL is handled now. The Openhab app is not falling over.

PR #175, opened 2016-04-22
Entering invalid local url causes OpenHAB app to die #174

JosSchering1:
Entering local url with port 84438 was causing unhandled IllegalArgumentException error in a background thread causing the app to vanish

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHABTracker.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHABTracker.java
@@ -203,10 +203,7 @@ public class OpenHABTracker implements AsyncServiceResolverListener {
                         Log.d(TAG, "Socket connected");
                         s.close();
                         return true;
-                    } catch (MalformedURLException e) {
-                        Log.e(TAG, e.getMessage());
-                        return false;
-                    } catch (IOException e) {
+                    } catch (Exception e) {
                         Log.e(TAG, e.getMessage());
                         return false;
                     }
--- a/mobile/src/main/java/org/openhab/habdroid/ui/AnchorWebViewClient.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/AnchorWebViewClient.java
@@ -10,10 +10,17 @@
 
 package org.openhab.habdroid.ui;
 
+import android.content.Context;
+import android.content.SharedPreferences;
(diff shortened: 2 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/core/OpenHABTracker.java
```diff
@@ -203,10 +203,7 @@ protected Boolean doInBackground(String... strings) {
                         Log.d(TAG, "Socket connected");
                         s.close();
                         return true;
-                    } catch (MalformedURLException e) {
-                        Log.e(TAG, e.getMessage());
-                        return false;
-                    } catch (IOException e) {
+                    } catch (Exception e) {
```
digitaldan: can this be more clearly caught using java 7 syntax like `catch (MalformedURLException | IOException e)` rather than defaulting to a generic exception?

digitaldan:
Hi @JosSchering1 thank you for the PR!  I'll pull this down shortly to test.  Besides reviewing my one comment, would you please sign and squash the commit as per  https://github.com/openhab/openhab.android/blob/master/CONTRIBUTING.md ?  Thanks again.

mobile/src/main/java/org/openhab/habdroid/core/OpenHABTracker.java
JosSchering1: My first attempt was to explicit catch IllegalArgumentException, which was raised, but I decided to catch all exceptions. Two reasons, no different logic is in place for the different exceptions. Secondly, more important, is that if another exception is raised and not handled, the background thread dies. Unhnandled exceptions in threads will bring the app down.

JosSchering1: I don't know how to do the Signed-off-by part and cannot do it in next 2 days because I cannot access my dev machine.

digitaldan: I'm fine with the general exception handling then.  The signing requirement can be done by amending your last commit ( https://help.github.com/articles/changing-a-commit-message/ ) and adding one extra line like ```Signed-off-by: Joe Smith <joe.smith@email.com> (github: github_handle)``` .   I'm going to test this today, thanks!

PR #195, opened 2016-06-06
Fixes issue where volume buttons could crash app.

digitaldan:


Br3nda:
Have you consider using something like this instead:
```
        if(openHABWidgetAdapter != null)
            return openHABWidgetAdapter.onVolumeDown();
        return false;
```

Br3nda:
or even swap it around for safety:  `if(null != openHABWidgetAdapter)`

digitaldan:
@Br3nda  yes that would be much cleaner ;-)  I hastily made this fix, but decided to just make a PR rather then merge so I could come back to it and clean it up, didn't want to forget about it.

PR #203, opened 2016-06-15
Sort the sitemaps in drawer by label.

Br3nda:
I have >30 sitemaps. It's difficult to find the one i'm looking for, because the sitemaps in the drawer are completely unsorted.

This patch sorts the Openhab1 sitemaps by label.


digitaldan:
Thanks!

PR #204, opened 2016-06-15
small style clean up for hard coded floats.

Br3nda:
small style clean up for hard coded floats.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/OpenHABItem.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/OpenHABItem.java
@@ -148,11 +148,11 @@ public class OpenHABItem {
 		Float result;
 		// For uninitialized/null state return zero
 		if (state == null) {
-			result = new Float(0);
+			result = 0f;
 		} else if ("ON".equals(state)) {
-			result = new Float(100);
+			result = 100f;
 		} else if ("OFF".equals(state)) {
-			result = new Float(0);
+			result = 0f;
 		} else {
 			try {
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/model/OpenHABItem.java
```diff
@@ -148,11 +148,11 @@ public Float getStateAsFloat() {
 		Float result;
 		// For uninitialized/null state return zero
 		if (state == null) {
-			result = new Float(0);
+			result = 0f;
```
watou: This also has the benefit of resulting in `Float.valueOf(0f)` being generated by the compiler when autoboxing the native `float`, where the JavaDoc says "[valueOf] should generally be used in preference to the constructor Float(float), as this method is likely to yield significantly better space and time performance by caching frequently requested values.".

Br3nda: also, the android-studio linter will stop glaring yellow at those lines.

digitaldan:
Thanks again for the PR!

mobile/src/main/java/org/openhab/habdroid/model/OpenHABItem.java
digitaldan: Should line 164 also receive this change?

digitaldan:
Hi I have not forgotten about your PR's !  I'm currently trying to get our cloudbee builds back up, as soon as that happens I will merge, thanks!

PR #206, opened 2016-06-17
Upped gradle from 1.5.0 to 2.1.2

Br3nda:
Every open of this project in android studio, i'm prompted to upgrade gradle.

The gradle conf/scripts are compatible and work in my testing - here's a patch to up the gradle version. 

This will allow Instant Run, which is all fun and time saving while doing dev. 
http://tools.android.com/tech-docs/instant-run


hazzeh:
I think that it can be good to take the opportunity to also update the gradle version in the wrapper to latest (2.14) in this PR.

Br3nda:
I presume you mean this
`distributionUrl=https\://services.gradle.org/distributions/gradle-${gradleVersion}-bin.zip`
within gradle/wrapper/gradle-wrapper.properties

The gradle docs say:
> String distributionUrl
> If not set, the download URL is the default for the specified Wrapper.getGradleVersion(). 
https://docs.gradle.org/current/dsl/org.gradle.api.tasks.wrapper.Wrapper.html

The default behaviour will mean we no longer need to maintain the gradle version in 2 places in our codebase. 

I can't see why it was explicitly set back in Jan 2015.  Commit message says
> Android Studio 1.0.2 required a gradle upgrade; the gradle upgrade required changing 'runProguard' to 'minifyEnabled'
So probably not an explicit choice back then

i propose we remove the whole line and use the default behaviour.

hazzeh:
I think that the documentation refer to the "wrapper task" which generates the wrapper scripts (inc. gradle-wrapper.properties). The wrapper jar needs to know where to download the distribution, and that is only done in gradle-wrapper.properties. 
Easiest is to change the distributionUrl and run ./gradle wrapper. Optionally you can run "gradlew wrapper" to generate new startscripts and fetch latest jar, but it is not necessary.

Br3nda:
Updated - and re-ran `gradlew wrapper`,. This updated the jar file, which was tracked in git so checking that in too.

PR #212, opened 2016-06-27
Removed unused imports

Br3nda:
There's plenty of unused imports in the java. Removed using the power of android-studio, and re-tested.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/HABDroid.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/HABDroid.java
@@ -9,8 +9,6 @@
 
 package org.openhab.habdroid.core;
 
-import android.app.Activity;
-import android.app.Application;
 import android.support.multidex.MultiDexApplication;
 import android.util.Log;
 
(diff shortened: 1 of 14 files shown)


digitaldan:
Thanks again, I'm currently trying to get our cloudbee builds back up, as soon as that happens I will take a look.

Br3nda:
BTW: I threw this at Travis-CI and it liked it.
https://travis-ci.org/Br3nda/openhab.android

digitaldan:
Our cloudbees android sdk is not in a happy place, I fixed this a few months ago, but this time its not installing the correct build tools and platform sdk versions that the project needs.  Unfortunately the only way to do this is by running a build job that contains the android sdk command line you want to run .  This has two problems, first is the cli was not really designed well, and second it's like working with a distant satellite, responses take minutes as Jenkins has to find a build server and do all of its stuff before you get a result to the command :-( I'm hoping to fix it soon, in the mean time I will just test locally and merge, but I have been pushing the store builds directly from our build server, so this is also a priority! Thanks for your patience.

PR #225, opened 2016-09-17
Fix display of charts for group items

dominicdesu:
Fixes the display of charts in case a group item is to be dislayed (see https://community.openhab.org/t/no-rrd4j-persistence-items-available-via-rest-access-and-in-habdroid/14189)

Signed-off-by: Dominic Lerbs <dominicdesu@hallojapan.de>


digitaldan:
LGTM, thanks!

PR #226, opened 2016-09-19
Fixes #218

### LINKED ISSUE #218: Notifications are not showing complete message
The text of longer notifications is currently not shown completely in the notifications view. The text is cut of with ... instead.
Would be nice if one could see the complete text, either in the notifications view directly or at least by clicking on a single notification in a new popup.

Example on Nexus 4:
screenshot 30 08 2016 21_56_27
##

CristianSarac:
Fixes #218

### DIFF
--- a/mobile/src/main/res/layout/openhabnotificationlist_item.xml
+++ b/mobile/src/main/res/layout/openhabnotificationlist_item.xml
@@ -32,8 +32,7 @@
                 android:layout_centerVertical="true"
                 android:layout_toRightOf="@+id/notificationImage"
                 android:layout_height="wrap_content"
-                android:singleLine="true"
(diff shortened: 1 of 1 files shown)


digitaldan:
Hi @CristianSarac , sorry for the very long delay! LGTM, do you foresee any issues if the messages where very very long?  Not sure if this is an actual use case.

CristianSarac:
The only issue i can foresee is that, if some  notifications would have a longer message then others then the  list wouldn't be symmetric meaning that one would be bigger( in order to fit the message) and others smaller is size depending on the size of the message. 
I hope it makes sense

digitaldan:
I will go ahead and merge soon, thanks again for the PR!

PR #237, opened 2016-10-06
Check for null values when comparing sitemap labels, Fixes #236

### LINKED ISSUE #236: SItemaps without labels cause NPE and application crash
The compare functionality at https://github.com/openhab/openhab.android/blob/e7e7dc83576657cc4ce8b1f943c37d2bb5afb000/mobile/src/main/java/org/openhab/habdroid/util/Util.java#L92 is not checking if the label is null and causing a NPE for some users. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38164249-sitemaps-without-labels-cause-npe-and-application-crash?utm_campaign=plugin&utm_content=tracker%2F968858&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F968858&utm_medium=issues&utm_source=github).
</bountysource-plugin>

digitaldan:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/Util.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Util.java
@@ -89,6 +89,12 @@ public class Util {
         Collections.sort(sitemapList, new Comparator<OpenHABSitemap>() {
             @Override
             public int compare(OpenHABSitemap sitemap1, OpenHABSitemap sitemap2) {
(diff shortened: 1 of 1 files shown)


Br3nda:
I think we need a regression test for this too.

PR #256, opened 2016-12-13
Support for icons in svg format

hazzeh:
Add support for icons in svg format. 
Add use a bugfixed copy of androidsvg from https://github.com/BigBadaboom/androidsvg.
When androidsvg is corrected the copied classes can be replaces with a dependency to the official library instead.

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -35,6 +35,10 @@ repositories {
         name "OpenHAB"
         url "http://repository-openhab.forge.cloudbees.com/release/"
     }
+    maven {
+        url 'https://jitpack.io'
+    }
+
 }
 
 dependencies {
@@ -52,6 +56,8 @@ dependencies {
     compile 'com.loopj.android:android-async-http:1.4.9'
     compile 'com.loopj:android-smart-image-view:1.0.0'
     compile 'com.github.shell-software:fab:1.1.0'
+
+    compile 'com.github.hazzeh:androidsvg:fix-npe'
 
     testCompile 'org.mockito:mockito-core:1.10.19'
     testCompile 'junit:junit:4.12'
--- a/mobile/src/main/java/org/openhab/habdroid/model/OpenHAB2Widget.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/OpenHAB2Widget.java
@@ -7,17 +7,22 @@ import org.json.JSONObject;
 import java.util.ArrayList;
 
 public class OpenHAB2Widget extends OpenHABWidget {
+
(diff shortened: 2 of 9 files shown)


digitaldan:
Sorry for the late response, I'm planning on spending time reviewing PR's this week, will get to it soon.

mobile/src/main/java/com/caverock/androidsvg/IntegerParser.java
```diff
@@ -0,0 +1,177 @@
+/*
```
digitaldan: external classes should be included as a dependency.

hazzeh: Hi, not sure what you mean. The problem is that the current version of androidsvg is missing a pending bugfix (https://github.com/BigBadaboom/androidsvg/pull/99) which is needed to work in habdroid. The copy I added is patched.
I guess it is possible to create a separate gradle project which contains the patched version, but I tried to follow the pattern used for de.duenndns.ssl.

digitaldan: >but I tried to follow the pattern used for de.duenndns.ssl

This is an issue which is on our list of things to fix, going forward we need to avoid this pattern
>Hi, not sure what you mean.

These classes were not created by us, not in our namespace and not under the EPL.   If there is a fix they have not updated yet, we can fork the repo and create a lib to include until such a time that they can fix it and we can use a gradle dependency.

digitaldan:
Been testing this on a few devices and it looks good to me!  If we could just get the 3rd party code as a dependency I would want to do a public release soon!

mobile/src/main/java/com/caverock/androidsvg/IntegerParser.java
hazzeh: Ok, thanks. I'll look into it next week how I can solve the dependency issue.

digitaldan:
Sorry for the delay, there is a merge conflict due to a change from another PR, could you please resole?  I will go ahead and merge when thats finished, thanks!

hazzeh:
Now the merge conflict is resolved.

cribskip:
Hi,

would be great if it this could be merged and published.
I've been testing the build of the hazzeh/openhab.android 255-svg-support branch for about two days now and it's working for me.

KR,
Sascha

digitaldan:
Hi all, sorry for the very long absence, been trying to fix our myopenhab issues.  LGTM, i will merge and get a release prepared.

PR #259, opened 2016-12-26
updated german translation

mueller-ma:


kaikreuzer:
Danke :-)

mueller-ma:
Is it possible for future strings to add them to all strings.xml files? An English string in the german strings.xml is better (easier to translate) than a missing string

PR #266, opened 2017-02-02
Improve WebView compatibility by enabling DOM storage

wborn:
By default the WebView is not 100% HTML5 compatible. This PR enables [DOM storage](https://en.wikipedia.org/wiki/Web_storage) which is for instance required when you want to create a WebView showing a [Grafana](https://community.openhab.org/t/influxdb-grafana-persistence-and-graphing) panel with tooltip support (JavaScript).


digitaldan:
Thanks!

Issue #117, opened 2015-06-03
Habdroid: NFC does not work with current app version 1.6.2


Schaussi:
With previous versions the NFC feature worked perfectly.
Now when I lay my mobile on an NFC button, it recognizes it - Habdroid opens but only with an empty screen.
##

cribskip:
Thanks for reporting.

I just fired up android studio and thought this might be introduced by my change #71 but as testing shows  I can't reproduce.

Environment:
  - HTC One M7 Android 5.0.1
  - TAG freshly written to toggle item (lamp)
  - Habdroid from Playstore (1.7.0.1 I guess)

Test procedure:
  - Selected item by long-press selected toggle
  - on home screen, presented TAG to phone
  - habdroid prints a "connecting to local..." toast and executes the command just fine. ON, OFF, ON, OFF :)

Just read you report again and thought: habdroid should not pop up anymore. there were some changes to disable this behavior.

Please check your installed version and please post your environment (phone, android version, habdroid version).

Thanks
Sascha

Schaussi:
Hi Sascha,

thank you for your answer.
Actually, I did not know I can use NFC tags with Habdroid to execute e.g. switch commands directly.
I currently only use NFC to access rooms, e.g. I have an NFC tag in the kitchen which should bring the Openhab room "kitchen" to front.

So to write data to the NFC tag I accessed the room in Habdroid and then chose "Write NFC Tag".
Is maybe this feature not supported any more?

- Habdroid version: the current one from Play Store
- Openhab version: 1.6.2
- Phone: Samsung GT-9195 (Galaxy S4 Mini) with Cyanogenmod 11, Android 4.4.4

cribskip:
Hi,

thanks for the further details - now I can confirm: opening sitemap pages by NFC TAG is currently not working.

**May be** related to #71

belovictor:
@cribskip , did you do any further investigations to fix sitemap NFC processing?

cribskip:
hadn't had time to do so yet. AFAIK only sitemap opening is concerned, TAGs with commands work fine. @Schaussi can you confirm?

Schaussi:
Yes I can confirm this.

Testato:
Sitemap bug confirmed here too:

OpenHab 1.7.1
Habdroid from store
LG-G3 on Lollipop

cribskip:
[...]

PR #267, opened 2017-02-06
References #117 - restore NFC functionality to Switch items

syphr42:
This change restores the ability to select an NFC option after a long press on a Switch item. 

The type is currently checked to match "SwitchItem" but the actual type value is "Switch" so I modified the if logic to check for both variations.

Issue #117 has a lot of comments, so I'm not sure if this will complete it. At the least, it is part of the solution.

Signed-off-by: Gregory Moyer <moyerg@syphr.com>


digitaldan:
Thanks!

PR #268, opened 2017-02-16
Update to latest API+tools+gradle+misc.

fat-tire:
WHAT THIS DOES:

* Runs in API 14+
* Targets API 25, buildtools 25.0.2
* Uses latest gradle & updated libraries
* Includes multiwindows support for Chromebook/Nougat
* Adds build timestamp to app info in settings
* Use AppCompatActivity everywhere
* Minor lint warning fixes (there are many more unaddressed)

PULL-REQUEST NOTES:

* DSA is not supported by Google after API 23.  Use EC instead I guess (see the [commit](https://github.com/fat-tire/openhab.android/commit/b19240f4cd359e24ab34dfefb9afe8a3d3ea666d#diff-b8963974f554d6698e36b7b2a0fa790aR151) for how this was handled).
* Has only been tested on API 14 emulator but seems to run fine.
* Do not manually request fine location or other permissionw w/new runtime permissions system.
  I can't find where it is invoked and would rather do it in-context.
  Was this permission requested for future use perhaps?
* For fun (and to help me make sure I was running the latest) I added a timestamp to the last settings item.
* Please test and look for any issues
* If nothing else, this should save someone some time in updating the build.

That's it!  Enjoy!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0'
+        classpath 'com.android.tools.build:gradle:2.4.0-alpha3'
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.3'
+        classpath 'com.android.tools.build:gradle:2.3.0-beta4'
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0-rc1'
+        classpath 'com.android.tools.build:gradle:2.3.0'
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0-beta4'
+        classpath 'com.android.tools.build:gradle:2.3.0-rc1'
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.4.1-all.zip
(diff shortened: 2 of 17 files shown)


fat-tire:
Update:  Updated play services to the new 10.2.0 from today,

fat-tire:
Update:  Updated gradle and play-services library.  Also tested with an actual openhab instalationl and it works as far as I can tell.

fat-tire:
Updated support libraries to today's 25.2.0 release.

fat-tire:
Gradle 3.4 is out and updated too :) :1st_place_medal:

fat-tire:
...and the android plugin, now at 2.3.0

fat-tire:
Updated to latest libraries + reconciled with recent changes to openhab

digitaldan:
@fat-tire thanks for the updated PR. Question, you stated this needs testing, are there parts of this change that you are concerned about?

fat-tire:
Quick note-

I don't have a wear device, so not sure if the support libraries are still needed, but if they are, I can add in

     compile 'com.google.android.support:wearable:2.0.1'
     provided 'com.google.android.wearable:wearable:2.0.1'

to `wear/build.gradle`

There is no issue building without these lines.  But I think that's the solution jic and I can add it in if someone with wear notices an issue.  I'm hoping it's sufficient to use:

     compile 'com.google.android.gms:play-services-wearable:10.2.1'

fat-tire:
Oh, and I didn't see your question above about testing.

Aside from the updates to the libraries and tools, I did make a few minor "upgrades" such as:

* supporting API 14 instead of 15.  This was possible due to some support classes (AppCompatActivity, etc.) I switched to using.  I have tested them on the emulator and they seem to work fine, but I guess I was hoping someone with an actual Ice Cream Sandwich physical device might try it.

* For Marshmallow and newer, I [used](https://github.com/openhab/openhab.android/pull/268/files#diff-b8963974f554d6698e36b7b2a0fa790aR141) the preferred elliptical curve key types in the non-deprecated KeyChain method [choosePrivateKeyAlias](https://developer.android.com/reference/android/security/KeyChain.html#choosePrivateKeyAlias).  The "DSA" keytype has been [deprecated](https://www.gentoo.org/support/news-items/2015-08-13-openssh-weak-keys.html) with EC keytypes replacing it for "M".  While it *should* work, I haven't really tested it, so I was asking for a second pair of eyes to look it over and verify it's correct.

* Finally, as I noted above, I don't have an android wear device to test that nothing broke by updating the library to the gms play-services library.  If there is a problem on older wear devices, I think the solution would be to add those two lines above, but FWIW I was able to build without them, so I don't *think* they're needed.

FWIW I tested w/my own openhab server and it works amazingly great.  I'm really impressed with this app!  Nice job!

digitaldan:
@fat-tire I had to revert this due to multiple build issues.  First is my local build is failing on the linter test with "The SDK platform-tools version (24.0.3) is too old to check APIs compiled with API 25; please update" , this is for the mobile target.  The second issue is with our build server, its not accepting the license properly for the updated android libs and then fails.  This is a shortcoming of our cloudbees setup and not your doing, but I need to fix it first so we can have other PR's continue to build properly on it.   Could you look at the linter error?  I will keep working on cloudbees.

fat-tire:
Sure thing.  Let me take a look at the linter error now.

fat-tire:
So this looks like it may be related to using an older Android Studio.  See:

http://stackoverflow.com/questions/40173850/the-sdk-platform-tools-version-24-0-4-is-too-old-to-check-apis-compiled-with-a

A few solutions are proposed, including clearing your build caches as well as updating AS.  I'm using a preview of Android Studio 2.4, though I'm wondering if just updating to the released 2.3 would be sufficient as II'm guessing 2.3 should be able to build API 25 no problem...

I've updated this PR to include two changes:

* Move the buildToolsVersion to the first line of the android{} block in gradle.build (someone said it made a difference) and

* Update the gradle plugin to 2.4.0-a3 which maybe will fix this as well.

fat-tire:
I guess I can't update a PR once it's been accepted.  So I created #281 to try...  Let me know if it works.  I just built it so it seems okay on my end.

PR #275, opened 2017-03-25
Replace loopj with okhttp

### LINKED ISSUE #238: HTTPS SNI support
Setup a second site on my https server and found that my Android App v4.4 would not connect to my OpenHAB site.  

Still able to access the site via FireFox on my NoteBook.  

Was not able to find any information on SNI support in the Android App. Might be an idea to document lack of HTTPS SNI support, until implemented.  

Also did not get an error or indication of progress in the Android App, which was very frustrating, thing it might have been my SSL Tunnel Terminator, reverse proxy or remote connection was down.
##

rusitschka:
This fixes #238 lack of SNI support. Mjpeg stream and self-signed certificates should work but need testing.


digitaldan:
Thanks for this!  I'll hopefully get to this soon

digitaldan:
Hi, when testing this  the app crashed when it received a 401 not authorized response from the server
```
                                                                    --------- beginning of crash
03-29 09:59:26.134 8873-8873/org.openhab.habdroid E/AndroidRuntime: FATAL EXCEPTION: main
                                                                    Process: org.openhab.habdroid, PID: 8873
                                                                    java.lang.NullPointerException: Attempt to get length of null array
                                                                        at java.lang.StringFactory.newStringFromBytes(StringFactory.java:63)
                                                                        at org.openhab.habdroid.util.MyAsyncHttpClient$3.onFailure(MyAsyncHttpClient.java:167)
                                                                        at org.openhab.habdroid.util.MyAsyncHttpClient$2$1.run(MyAsyncHttpClient.java:134)
                                                                        at android.os.Looper.loop(Looper.java:148)
                                                                        at android.app.ActivityThread.main(ActivityThread.java:5417)
                                                                        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                                                                        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

rusitschka:
Hi Dan - I fixed the issue and also the resulting endless redirects. Can you try it again?

digitaldan:
So I'm noticing that bad responses from the server are not being handled and the call will not complete until a timeout event happens.  So if there is a 404 for an image, the the call will not return until a timeout error is thrown.  This is causing the app to appear to hang when loading pages.  This is also the case for other errors as well (like 401).

rusitschka:
MyWebImage was using HttpConnection - I ported this to MyAsyncHttpClient, too.

I didn't experience the timeouts though - can you give a repro? Maybe there is a different issue with those. For example related to localURL vs. remoteUrl or the like.

digitaldan:
With the latest commit, errors look like they are being handled ok, but I am seeing other issues as well. As i navigate through my sitemap it seems like requests are not being canceled properly as we navigate from page to page, after a few clicks the pages seem to hang for a while and I can see in the logs requests continuing to be made from other sitemap fragments that I am no longer viewing.  This is what I have observed, I have not actually stepped through the code or done extensive testing to confirm.

rusitschka:
The problem was probably the callbacks being called for canceled calls. The atmosphere handler of OpenHABWidgetListFragment now checks if a call was canceled and only if not so, it performs retries/reconnects.

digitaldan:
>The problem was probably the callbacks being called for canceled calls

Networking works much better now on my local instance.   The next issue I see is that when using authentication (like myopenhab.org) it works in OpenHABMainActivity and I am able to pull down the list of sitemaps, but when the OpenHABWidgetListFragment tries to request the sitemap it's getting a 401 error.

rusitschka:
onAttach was called too soon. Initializing MyAsyncHttpClient in onActivityCreated solved this issue for me - can you test?

digitaldan:
Sitemaps look good now, but clicking on the notifications menu from the top left hamburger menu is printing a request failure, my guess is it's the same auth issue?

rusitschka:
I just signed up with myopenhab and my openhab2 instance shows as "online". I also have my first notification on myopenhab. However, I'm not seeing a "Notifications" entry in the hamburger menu. So currently I'm unable to repro this. Is there a trick to get the notifications menu?

digitaldan:
hmm,  if you are connected to myopenhab.org then you should see a notifications entry in the top left menu along with discovery and bindings.  The code that gets for notifications c is OpenHABNotificationFragment

rusitschka:
I tried using https://myopenhab.org as remote URL and my email+password as credentials but habdroid gives me a timeout - how can I connect to myopenhab.org?

digitaldan:
That should be it, do you logs show anything?

rusitschka:
Maybe it's due to the v1 vs. v2 detection via /rest/bindings. I'm getting

04-01 07:05:13.456 16264-16264/org.openhab.habdroid D/OpenHABTracker: Connecting to remote URL https://myopenhab.org/
04-01 07:05:38.434 16264-16264/org.openhab.habdroid D/OpenHABMainActivity: openHAB version 1 - got error java.net.SocketTimeoutException: timeout accessing https://myopenhab.org/rest/bindings
04-01 07:05:38.435 16264-16264/org.openhab.habdroid D/OpenHABMainActivity: Loading sitemap list from https://myopenhab.org/rest/sitemaps
04-01 07:06:09.798 16264-16264/org.openhab.habdroid E/OpenHABMainActivity: class java.net.SocketTimeoutException
04-01 07:06:09.799 16264-16264/org.openhab.habdroid E/OpenHABMainActivity: timeout

I didn't change the detection code and guess myopenhab.org runs v2 already - does this work with the existing client?

rusitschka:
I just tried upstream/master and it's showing the exact same behavior with notifications. https://myopenhab.org/rest/bindings and https://myopenhab.org/rest/sitemaps times out with a 504 Gateway Time-out  nginx/1.10.0 (Ubuntu). I guess this is a server issue and happens for both loopj and okhttp3. How do we progress?

rusitschka:
The timeout is gone after I switched to "Notifications&Items" instead of "Notifications only" in the configuration of the OpenHAB Cloud add-on. Now when connected to the remote URL I see all notifications from https://myopenhab.org inside habdroid! Only when connected to the local URL notifications cannot be loaded - which is the expected behavior, right?

Can you check this? What was the exact problem you ran into with notifications?

digitaldan:
I added a little logging to that class, it looks like the URL formed for the request  is "https://myopenhab.org//api/v1/notifications?limit=20"  with a double "/" after the base url so our NGINX is returning a 404, removing the prefixing slash solves this for me.  I'm not sure what could have changed but its an easy enough fix.  I'll try and look at it a little more today if I can before making any changes.

digitaldan:
Sorry for the huge delay, if you don't have any objections I'm going to merge this today.  I'll look at the issue with my last comment and submit another PR for it.

rusitschka:
Hi Dan, sure go ahead :-)

digitaldan:
Sorry again for the delay, our build server has been not working this week, but is back now.

I am getting an error when running locally, when it connects to my local system using service discovery (i did not enter my oh url) I'm getting this message " SSL Handshake Failed.  Maybe you need a valid client certificate", which means we are catching a SSLHandshakeException in OpenHABMainActivity.  I have ignore invalid certificates ignored in preferences.

digitaldan:
@FlorianSW I am wondering if you can help out here, @rusitschka has been patiently waiting for this to get accepted, unfortunately in #288 you created MySyncHttpClient which subclasses SyncHttpClient which is no longer available.  I was hoping you would be able to review this PR and submit a fix to use the okhttp3 which are intended to fix a number of issues that we have.  As soon as I get though all the PR's on android, I can come back to reviewing openhab-cloud!

mobile/src/main/java/org/openhab/habdroid/util/MyWebImage.java
```diff
@@ -16,34 +16,23 @@
...
-import javax.net.ssl.TrustManager;
-import javax.net.ssl.X509TrustManager;
+import java.util.HashMap;
+import java.util.Map;
+
+import okhttp3.Call;
+import okhttp3.Headers;
 
 public class MyWebImage implements SmartImage {
+    private static final String TAG = "MyWebImage";
```
FlorianSW: I would propose MyWebImage.class.getSimpleName().

mobile/src/main/java/org/openhab/habdroid/util/MyWebImage.java
```diff
@@ -16,34 +16,23 @@
...
-import javax.net.ssl.SSLSocketFactory;
-import javax.net.ssl.TrustManager;
-import javax.net.ssl.X509TrustManager;
+import java.util.HashMap;
+import java.util.Map;
+
+import okhttp3.Call;
+import okhttp3.Headers;
 
 public class MyWebImage implements SmartImage {
```
FlorianSW: The class can be package-private btw.

digitaldan:
>Unfortunately, I can't update the PR here, so what should we do now? :)

Hi @FlorianSW you can submit a pr to @rusitschka 's repo (master branch).  @rusitschka can you take a look at the comments?

FlorianSW:
Here it is:
https://github.com/rusitschka/openhab.android/pull/1

The important commit (apart from the merge of the latest master) is:
https://github.com/rusitschka/openhab.android/pull/1/commits/186cb3c452d6c880aca42881c6867a28c10ca8be

digitaldan:
Thanks!

FlorianSW:
Some of my comments are probably still valid, @rusitschka if you want: Would you like to take a look? Thanks in advance! :)

digitaldan:
@rusitschka  I was just doing some testing and realized that we are no longer able to use self generated SSL certs, despite the options in our settings.  Since the app defaults to connecting to a local OH instance which advertises SSL (which new installs do), this will break it for a lot of users.

mueller-ma:
@rusitschka The trust manager for self signed certs doesn't seem to ask the user if the certificate should be trusted.

PR #276, opened 2017-04-02
More consistent strings

mueller-ma:
The String is used as a "oposite" of connecting to a remote url, so it should be changed to "local". This already happend in at least the German and Spannish translation.


digitaldan:
Thanks

PR #277, opened 2017-04-02
Make a few strings non-translatable

mueller-ma:
app_name, openhab_service_type, openhab_demo_url should always be the same.

There seems to be unnecessary ">" in the xml files. Do they have any meaning?


digitaldan:
>There seems to be unnecessary ">" in the xml files. Do they have any meaning?

You are correct, I would think not, but I also don't understand how they are not throwing a ton of errors during the build phase, let me take a look before I merge.  Thanks again for the PR.

digitaldan:
thanks!

PR #282, opened 2017-04-15
remove messages at startup, fix #258

### LINKED ISSUE #258: Remove info texts on startup
Hi,

Everytime I start habdroid I get two info messages:
- connecting to configured url
- voice recognizer not present...

In my opinion it would be better, when the first messages shows up once after changing url, the second as a popup with "never show this again" and "ok" button

mueller-ma:
IMO these messages are not needed:
1. When I am on WIFI and see my sitemap, it connected most likely direct and when I am not on WIFI and I see my sitemap, it connected via Cloud Companion. It still logs the messages and gives and error when connection failed.
2. If the voice recognizer is not available the microphone icon is not present

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHABTracker.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHABTracker.java
@@ -95,7 +95,7 @@ public class OpenHABTracker implements AsyncServiceResolverListener {
                     // If remote URL is configured
                     if (mOpenHABUrl.length() > 0) {
                         Log.d(TAG, "Connecting to remote URL " + mOpenHABUrl);
-                        openHABTracked(mOpenHABUrl, mCtx.getString(R.string.info_conn_rem_url));
+                        openHABTracked(mOpenHABUrl, null);
                     } else {
                         openHABError(mCtx.getString(R.string.error_no_url));
                     }
@@ -108,8 +108,8 @@ public class OpenHABTracker implements AsyncServiceResolverListener {
                     if (mOpenHABUrl.length() > 0) {
                         // Check if configured local URL is reachable
                         if (checkUrlReachability(mOpenHABUrl)) {
-                            Log.d(TAG, "Connecting to directly configured URL = " + mOpenHABUrl);
-                            openHABTracked(mOpenHABUrl, mCtx.getString(R.string.info_conn_url));
+                            Log.d(TAG, "Connecting to local URL = " + mOpenHABUrl);
+                            openHABTracked(mOpenHABUrl, null);
                             return;
                             // If local URL is not reachable go with remote URL
                         } else {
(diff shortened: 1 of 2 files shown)


digitaldan:
Hi can you resolve the conflicts?  Thanks!

mueller-ma:
@digitaldan Done

mueller-ma:
Dont merge it. The app crashes now when starting the second time

mueller-ma:
@digitaldan I am confused with Android Studio. I built the same version again and it runs without any problems

digitaldan:
I'll test this out as soon as a figure out our build server issues.

digitaldan:
Build server is finally fixed, looking at this now.

mobile/src/main/java/org/openhab/habdroid/core/OpenHABTracker.java
```diff
@@ -95,7 +95,7 @@ public void start() {
                     // If remote URL is configured
                     if (mOpenHABUrl.length() > 0) {
                         Log.d(TAG, "Connecting to remote URL " + mOpenHABUrl);
-                        openHABTracked(mOpenHABUrl, mCtx.getString(R.string.info_conn_rem_url));
```
digitaldan: What is the purpose for changing this?

mueller-ma: It removes the "Connected to remote URL" toast when starting the app

digitaldan: ok, so why do we not want to give feedback to the user that the app is trying to connect to a URL?   I have seen where the app tries the local URL (which is not available) then tries the remote URL,  I think it's good that we give some feedback especially b/c in this case it adds a delay to loading a sitemap. Or is this something else?

mueller-ma: I find the notifications anoying. When I'm in my wifi it connects to local url, if not to the remote url. I don't need feedback for that.
Every time I open the app I get these notifications and I get the "voice recognizer..." every time I rotate the phone

digitaldan: I agree with the voice recognizer , my guess is that it is a very rarely used feature and probably extra noise, but I don't have an issue with the connecting messages.  I don't think I have heard others complain about it so before we make this kind of change I would prefer we have a chorus of feedback saying this is undesirable.  I actually prefer to know if I'm connecting locally or remotely.

mueller-ma: The voice recognizer problem appears only when the google app is not available

digitaldan: Can we modify the PR to not complain about the voice recognizer but leave the other messages?

mueller-ma: So split it in two PRs?

digitaldan: 2 pr's would be good.  I'm not sure we want to change the local/remote notifications however.

mueller-ma: See #303

PR #292, opened 2017-05-14
Set default theme to light

mueller-ma:
Light themes are always default. The light theme uses the openHAB "orange", but the dark theme is just grey.

This also fixes the issue, that no radio button in the theme selection is checked after installation.


digitaldan:
Awesome!

PR #299, opened 2017-05-20
Change dark theme

mueller-ma:
I changed the primary color in the dark theme to the openhab orange
screenshot_2017-05-20-11-52-57
screenshot_2017-05-20-12-34-09

### DIFF
--- a/mobile/src/main/res/drawable/material_divider_dark.xml
+++ b/mobile/src/main/res/drawable/material_divider_dark.xml
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <shape xmlns:android="http://schemas.android.com/apk/res/android">
-    <gradient android:angle="90" android:startColor="#242f35" android:endColor="#1b2328"/>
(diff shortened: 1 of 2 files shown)


digitaldan:
I always like to get @kaikreuzer opinions on UI changes like this, as he always has a (valid) opinion :-)  @kaikreuzer for your reference this changes the "Dark" theme from using a Light/Med Grey header to OH orange.  I'm not sure how many are using the dark theme on things like wall tablets where maybe the orange would not be welcomed.  Thoughts?

mueller-ma:
@digitaldan I have thought about wall mounted tablets:
1. I think something like habpanel is more suitable for tablets.
2. We could add more themes with several material colors

mueller-ma:
Maybe one for each of these: https://material.io/guidelines/style/color.html#

kaikreuzer:
Afair, @belovictor chose the grey header at the time as he wanted something "not bright" for the dark theme - the orange color somehow undermines this idea. So I think it would be the better idea to introduce additional themes instead of changing existing ones.

mueller-ma:
You're right. I wanted to keep the "identity" with the orange, but it is too bright. I will close this.

PR #303, opened 2017-05-26
Remove voice recognizer error

mueller-ma:
If the voice recognizer is not available the microphone icon is not present, so IMO the error message is not necessary
as discused in #282


digitaldan:
Thanks!

Issue #290, opened 2017-05-04
Display of selection elements


lolodomo:
When the current value is not one of the mappings, the displayed value should be an empty string.

Addtionally, valuecolor tag should be handled.

mueller-ma:
@lolodomo Which version of openHAB are you running?
I tested with oh2 and if no mapping for a value is specified nothing is displayed, both in the App and Basic UI.
Valuecolor tags are also working.

mueller-ma:
@lolodomo I didnd see the "selection". Is it possible in Basic UI to use valuecolor and transformation in select elements?

lolodomo:
I am talking about selection element in sitemap with a mapping.
In HABdroid, when you open the selection, you have the mapped values but the selected value when the selection is closed is the non mapped value.
In basic UI, I recently fixed these 2 problems (display of mapped value + handling of valuecolor).

lolodomo:
Look at #https://github.com/eclipse/smarthome/pull/3381 and https://github.com/eclipse/smarthome/pull/3368

mueller-ma:
Item:
````
Number foo2
````

Sitemap:
````
sitemap test2 label="Test"
{
    Frame label="Indoor Climate Control" {
        Selection item=foo2 icon="climate" mappings=[0="Off",1="Heat",2="Cool",3="Auto"]
    }
}
````
HabDroid:
<img src="https://cloud.githubusercontent.com/assets/22525368/26069168/48267252-39a0-11e7-8478-c6f87dee66bb.png" width=500>
<img src="https://cloud.githubusercontent.com/assets/22525368/26069182/4a30e37a-39a0-11e7-81a8-3fcc04c33fbc.png" width=500>

**Comment by lolodomo** [comment]
I suppressed my second message which was totally wrong !
You have to read again my initial issue description. What is not working is when the current value is not a value present in the mappings.
If I take your example, be sure that your item value is 2 (cool) and suppress from the mappings the value "2". You will see that the selection displays a value, probably the first in the list, so probably "Off" in your case.The expected behavior is to have no value displayed.

**Comment by mueller-ma** [comment]
Now I got it. What about the color issue?

**Comment by lolodomo** [comment]
Color is not working too. Try for example to update your sitemap like this:
```
Selection item=foo2 icon="climate" mappings=[0="Off",1="Heat",2="Cool",3="Auto"] valuecolor=[==2="red"]
```
We will see that the text "Cool" remains in black rather than red, when foo2 item value is 2.

[...]

PR #306, opened 2017-06-03
Send an update of a selection item, even if the state does not change

### LINKED ISSUE #305: Tapping the selected element of a selection item again does not send the command
Expect the following setup:
openHAB instance with a sitemap with at least one selection element (e.g. for a roller shutter). The sitemap has, for example, the following entry for the roller shutter:
`Selection item=Pseudo mappings=[0="0", 10="10", 20="20"]`

Now, if you access the sitemap with the basic UI, the default value (e.g. after a restart of openHAB or if no value was provided so far) of the possible selection values is nothing, so the item value is empty. While, when you access the sitemap with the app, the default value is the first selection value (0).

If the user then clicks on the 0 value in the app again, the command (that the field was updated), is not send to openHAB, as the value hasn't changed for the app, even if the value is not set so far. This can also be a problem where the user even wants to receive an update of the item, if the same value is selected again (which is the case in the basic UI). The app should at least behave like the basic UI in the browser and always send an update of the item state, if the user selects a value, even if it's the same one.

FlorianSW:
This streamlines the behaviour of the app and of the basic UI, which now
behave in the same way. The selection item now always sends an update to
the openHAB item, no matter, if the state changes or not, the user just
needs to select one item.

Fixes #305

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -23,7 +23,6 @@ import android.view.ViewGroup;
 import android.view.WindowManager;
 import android.webkit.WebView;
 import android.widget.AdapterView;
-import android.widget.AdapterView.OnItemSelectedListener;
 import android.widget.ArrayAdapter;
 import android.widget.Button;
 import android.widget.ImageButton;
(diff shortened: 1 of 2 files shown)


lolodomo:
Does it fix my issue #290 too ? Including handling of valuecolor ?

FlorianSW:
Not really, the underlying problem is different, however, this PR could probably solve a symptom of the problem (that you need to change to another value first, and then change back to the wanted, pre-selected one).

digitaldan:
LGTM!

PR #313, opened 2017-06-10
Observe the setting, that all certificates should be trusted

### LINKED ISSUE #310: Can not connect to local OH using self signed certificate
With recent http upgrade we can no longer connect to a local OH instance running a self signed certificate, despite what options are selected in the settings.

@rusitschka can you take a look?

FlorianSW:
To trust self-signed certificates, there's a setting, which allows the
user to bypass the certificate trust check. Until now, the setting was not
observed anymore since the change to okhttp3, which is fixed with this
commit.

Fixes #310


digitaldan:
Thanks!  I'm sooting for a Sunday release, so i will test this soon.

digitaldan:
LGTM

PR #318, opened 2017-06-19
Update build to API 26 latest + library updates.

fat-tire:
A quick update for libraries and tools to target the newest "O" stuff from Google as well as other libraries.

Note that due to a (reported fixed) bug in aapt2, the <layout> tag has been temporarily commented out of the manifest.  Not a big deal, but should be put back once the fix appears (scheduled for alpha5) for smarter multi-window support on chromebooks.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,10 @@ If you want to contribute to Android application we are here to help you to set 
 development environment. openHAB Android app is developed using Android studio and also can be
 build with maven.
 
-- Download and install [Android Studio](http://developer.android.com/sdk/installing/studio.html)
-- After installation launch Android Studio, download and install Android SDK version 4.0.3 (Tools ->
-Android -> SDK Manager)
-- Check out the latest code from github
-- Open the project in Android Studio (File -> Open, select project folder)
+- Download and install [Android Studio](http://developer.android.com/sdk/installing/studio.html).  Should you feel adventurous, preview versions of Android Studio are also available [here](https://developer.android.com/studio/preview/index.html#linux-canary-bundle).
+- After installation, launch Android Studio.  Next, select **Tools** from the main menu, then **Android**, and choose **SDK Manager**.  Install the latest Android build tools, SDKs, and libraries (SDK Tools >= 26.0.0 and SDK Platform-Tools >=26.0.2).
+- Download the latest OpenHAB source code from github with the *git* command:  `git clone https://github.com/openhab/openhab.android.git` or download the source from [here](https://github.com/openhab/openhab.android/archive/master.zip).
+- Open the project in Android Studio (Select **File** from the main menu, then choose **Open**.  Select the OpenHAB project folder and choose **OK**.)
 
 You are ready to contribute!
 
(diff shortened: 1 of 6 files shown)


mueller-ma:
How stable are these libaries? Shouldn't we wait until they are out of beta?

And can you update the readme?

README.md
```diff
@@ -31,7 +30,7 @@ Before producing any amount of code please have a look at [contribution guidelin
 ## Trademark Disclaimer
 
 Product names, logos, brands and other trademarks referred to within the openHAB website are the
-property of their respective trademark holders. These trademark holders are not affiliated with
-openHAB or our website. They do not sponsor or endorse our materials.
+property of their respective trademark holders.  These trademark holders are not affiliated with
```
mueller-ma: Why add spaces here?

README.md
```diff
@@ -31,7 +30,7 @@ Before producing any amount of code please have a look at [contribution guidelin
...
+openHAB or our website.  They do not sponsor or endorse our materials.
 
-Google Play and the Google Play logo are trademarks of Google Inc.
+Google Play&trade; and the Google Play&trade; logo are trademarks of Google Inc.
```
mueller-ma: Please revert. It is exactly what Google wants: https://play.google.com/intl/en_us/badges/

fat-tire:
Not sure if the README is what you wanted, but anyway...  sure you can wait until everything is out of beta.  I just built using the latest stuff for fun and it seems to work, so figured may as well share it with anyone else wanting to do the same.  If nothing else this may save a little time

Cheers!

PS you're right https://play.google.com/intl/en_us/badges/ I'll fix that.  Spaces between sentences were added for style.  I can get rid of it too but it looks wrong to me with a single space.  I guess it's controversial either way https://en.wikipedia.org/wiki/Sentence_spacing#Controversy

build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.4.0-alpha7'
+        classpath 'com.android.tools.build:gradle:3.0.0-alpha4'
```
digitaldan: I actually just downgraded to 2.3.3 which is the current stable version, beta versions expire after some amount of time (40 days?) and its too hard to keep up to date.

digitaldan:
>Shouldn't we wait until they are out of beta?

I also would like to keep to stable versions.  While I have found that the betas are genuinely of stable quality, trying to keep up with the cadence of beta releases is difficult, and makes our automated build tougher.

fat-tire:
Okay, I'll withdraw this PR...  maybe too much of it is beta for your taste, which is fine by me.  Just made the changes on my own build and figured you or others might want 'em...

mueller-ma:
@fat-tire Can you make a PR with the Readme update?

mueller-ma:
@fat-tire Since API 26 is stable now, can you bump again it to 26? Will help me with #345

PR #331, opened 2017-07-15
Fix NullPointerException when no viewBox is part of SVG

### LINKED ISSUE #330: Latest PlayStore Version crashed after Launch
After the newest Update it starts up and then shows the sitemap. After about 1 second it just exits.

FlorianSW:
SVG#getDocumentViewBox() could return null, if no viewBox is present in
the provided SVG. In this case, assume, that the SVG size is 16x16px
(probably not the best values, but if the the viewBox is missing, it's
already a pretty weird case and the behaviour is more or less undefined,
this now just prevents the app from crashing).

Fixes #330


digitaldan:
LGTM

PR #339, opened 2017-08-08
Restore SSL Client Certificate Authentication feature

### LINKED ISSUE #328: SSL Client Certificate authentication doesn't work anymore
Apparently since the latest update (6.jul) the client certificate authentication does not work anymore on my setup.
I use a self signed CA with a self signed server certificate and user certificates. The whole setup was working smoothly for ~a year.
Then my phone died and I had to setup a new one. Since then my authentication fails. Either "bad request" when I ignore Certificate validation during handshake - or "SSL handshake failed you need a valid user certificate"
I believe my setup is all right as my Browser is able to connect .

slawekjaranowski:
Fix #328

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -67,4 +67,9 @@ dependencies {
     testCompile 'org.mockito:mockito-core:2.7.6'
     testCompile 'junit:junit:4.12'
     testCompile 'org.json:json:20160810'
+
+    // PowerMock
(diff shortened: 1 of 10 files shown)


digitaldan:
Awesome!  I will review this weekend.

digitaldan:
LGTM, I have not tried the functionality, so I will trust that it work.  thanks for the fix!

PR #340, opened 2017-08-14
Updates SVG library to current master (using git commit hash) Fixes #338

### LINKED ISSUE #338: Android app crashes with some SVG icons
Hello,

i use newest app from Play Store.
if i use <fan> or <fan_box> and SVG, then app crashes if i open a group with this item.
If i switch to PNG, it works.

digitaldan:

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -61,8 +61,8 @@ dependencies {
     compile 'com.squareup.okhttp3:okhttp:3.6.0'
     compile 'com.loopj:android-smart-image-view:1.0.0'
     compile 'com.github.shell-software:fab:1.1.2'
-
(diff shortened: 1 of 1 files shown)


mueller-ma:
Is there a reason not to use latest master?

digitaldan:
>Is there a reason not to use latest master?

everytime we build we would get master, and since I don't think many of us follow that project, I'm afraid this could lead to us having unknown and untested code in every build.   My thought was that we should be using a versioned tag of the library we know works, since they have not published a versioned release in a while, a git hash is the next best thing (since this is all a release tag points too).  If there is a strong argument the other way, I would be happy to hear it.

mueller-ma:
Good argument

PR #341, opened 2017-08-14
Bump to next version for release

digitaldan:

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="org.openhab.habdroid"
-    android:versionCode="24"
-    android:versionName="1.8.0.7" >
+    android:versionCode="25"
+    android:versionName="1.8.0.8" >
     <uses-permission android:name="android.permission.INTERNET" />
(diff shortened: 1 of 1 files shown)


mueller-ma:
Sometimes version numbers are bumped to a much higher version number. Openhab server 1.x looks/is outdated and this way version <2 might sound old. Have you thought about openhab (app) > 2.0.0?

digitaldan:
>Have you thought about openhab (app) > 2.0.0?
 I have, but I would hope a 2.x version would be significantly different then 1.x, however that might not happen anytime soon :-( 

@kaikreuzer  thoughts here?

kaikreuzer:
Well, I would agree that updating only the 4th digit for years feels as if there hasn't been ANY progress. Not sure if this reflects the reality - if so, somthing needs to be done anyhow ;-)

In general I would be ok to go for version 2 to reflect that it is actually compatible with a 2.x runtime. And the version of an app is not very prominent anyhow, so I doubt that anybody expects a whole new app just because the version goes to 2.x...

But the decision is up to the maintainers of this repo 😎

digitaldan:
sounds good, let's bump the release to 2.0 soon then.  FYI, I already have this version rolling out ( 10% of users to start) so it's too late for this release.

mueller-ma:
2.0 sounds like a very awesome new feature was released in the version. Go with 2.1?

kaikreuzer:
> so it's too late for this release.

Then let's put something awesome in the next version and call it 2.0 😆

PR #346, opened 2017-08-22
Use popup for selection items

### LINKED ISSUE #344: Implement selection items as pop up menu
In basic ui a selection is implemented as a pop up window, on the app as dropdown menu which is very small and has a variable width

mueller-ma:
This has several advantages:
1. Consistent width
2. Same behavior in app and basic ui
3. Better for tablets and large smartphones
Fixes #344 

Couldn't make a screenshot of how it is now, but here is a screenshot after:
screenshot_2017-08-22-18-31-27

Maybe we should add a radio button.


digitaldan:
Thanks, I agree this sounds like a better solution.  I will review shortly.

digitaldan:
LGTM

PR #349, opened 2017-08-23
[WIP] About screen

mueller-ma:
This adds an empty about screen. What should I add?

- [x] App name and icon
- [ ] version
- [ ] contributors
- [x] License
- [x] Github, Website and Forum links


digitaldan:
>App name and icon

Sounds reasonable
>version

as long as it pulls this dynamically
>contributors

This might be hard to keep up to date
>License

Sure
>Github, Website and Forum links

again sounds reasonable

An about page seems fine to have, as long as it does not require frequent updating, so version info should be pulled from the xml manifest file and I would avoid things like authors.

mueller-ma:
I can do the static stuff for sure, but may need help for dynamic stuff.
Any design suggestions? Should I use simple links, text buttons or image buttons (might be a copyright problem, e.g. github logo)

mueller-ma:
Right now (Year is autogenerated):
screenshot_2017-08-29-18-52-50

digitaldan:
I think this is a great start, and better then what we have (which is nothing ).

PR #352, opened 2017-08-28
Replace hardcoded strings with strings res

mueller-ma:


mobile/src/main/res/values-de/strings.xml
```diff
@@ -103,4 +103,9 @@
     <!-- drawer -->
     <string name="drawer_open">Sitemap-Bereich öffnen</string>
     <string name="drawer_close">Sitemap-Bereich schließen</string>
+    <string name="minus">-</string>
```
digitaldan: Do you need to specify  "+" "-" "settings_openhab_icon_format_png"  "settings_openhab_icon_format_svg" here?  Or will this be inherited from mobile/src/main/res/values/strings.xml ? I'm asking because I'm not sure either.

mueller-ma: The problem is: I dont know every language supported by Android. If only one language is using a different sign for "plus" or different letter for PNG, we cannot make it non-translatable.
It is not needed to add it to the german translation, but it reduces lint errors

digitaldan: understood, i'm fine with that

digitaldan:
LGTM, @mueller-ma if you could resolve the conflicts I will go ahead and merge.   Also can you send me an email to dan@digitaldan.com ?  I wanted to run something by you not related to this PR.

mueller-ma:
Done

Issue #343, opened 2017-08-16
Integrate Geofence-based Presence


jjhuff:
I'm in the process of migrating from SmartThings, and I'm missing how easy geofence-based presence detection was on ST. With OH2, I need to setup MQTT, secure it, expose it to the 'net, AND all my family's devices need to have another app installed and configured. 

This seems like ideal functionality to just integrate into the Android (and iOS) app.

I'm thinking something like:
1. A setting to enable it (prompt for location permission at that point)
2. Select an item to control
3. Pick a location and radius (unless we can get that from the server?)
4. Setup the geofence using Android's geofencing (better battery life -- we don't need to completely track the user!)
5. Use the REST API to update the item when entering/exiting

I'm happy to do the work, but I wanted to discuss the approach, design, concerns before diving in.

mueller-ma:
> MQTT, secure it, expose it to the 'net, AND all my family's devices need to have another app installed and configured.

I wouldnt know how to do this without exposing some interface to the internet. You will need at least openhab or openhab cloud companion to connect to.

> unless we can get that from the server?

Technicaly you could use homekit tags in some way. See comment of @digitaldan https://github.com/openhab/openhab.android/issues/334#issuecomment-316101305

> Use the REST API to update the item when entering/exiting
The easiest solution would be to have a simple switch which is on, when a person is inside the radius. Maybe you can also transmit a timestamp for entered/exited and last update.

Also have a look at #269 which is quite similar

jjhuff:
> I wouldnt know how to do this without exposing some interface to the internet. You will need at least openhab or openhab cloud companion to connect to.
Yup, but fewer is better. If one was only concerned about local usage, the network plugin with an expiration could be an option. I could also see potential of a more limited 'presence' channel offered by openhabcloud.

> Technicaly you could use homekit tags in some way. See comment of @digitaldan #334 (comment)
Suggestion: tags named "hostname-isInRadius1" or "wifiMac-isInRadius1" or similar uniq information.
I'm not sure what you're thinking here, but I don't know much about HomeKit or it's OH integration

> Also have a look at #269 which is quite similar
Ahh, yes it does!

I'll see what I can put together.

mueller-ma:
We should get some thoughts of @digitaldan about this feature in generell and from @kaikreuzer about homekit tags usage in clients

digitaldan:
Geo tagging would be a great feature for sure. 

>Select an item to control

>Use the REST API to update the item when entering/exiting

[...]

PR #355, opened 2017-08-29
Implement geofence-based presence

jjhuff:
This implements #343 

* Added prefs.
* If Lat/Lng aren't set, it fetches `/services/org.eclipse.smarthome.core.i18nprovider/config` to get `location`. I'd be totally in favor of fetching that another way if there is one. Maybe a new core service to expose for location?
* Request permissions dynamically if they weren't already granted
* Setup geofences (on boot too!)
* Update item state on enter/exit

I don't do tons of Java or Android, so please tell me what can be improved!

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/GeofenceBroadcastReceiver.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/GeofenceBroadcastReceiver.java
@@ -16,10 +16,16 @@ public class GeofenceBroadcastReceiver extends WakefulBroadcastReceiver {
 
     @Override
     public void onReceive(Context context, Intent intent) {
-        Log.i(TAG, "onReceive");
-        ComponentName comp = new ComponentName(context.getPackageName(),
-                GeofenceIntentService.class.getName());
-        startWakefulService(context, (intent.setComponent(comp)));
+        Log.d(TAG, "onReceive");
+        if(intent.getAction()!= null && intent.getAction().equals(Intent.ACTION_BOOT_COMPLETED)){
+            // Did we just boot? Go register.
+            Intent serviceIntent = new Intent(context , GeofenceRegistrationService.class);
+            startWakefulService(context, serviceIntent);
+        } else {
+            ComponentName comp = new ComponentName(context.getPackageName(),
+                    GeofenceIntentService.class.getName());
+            startWakefulService(context, (intent.setComponent(comp)));
+        }
         setResultCode(Activity.RESULT_OK);
     }
 }
--- a/mobile/src/main/java/org/openhab/habdroid/core/GeofenceBroadcastReceiver.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/GeofenceBroadcastReceiver.java
@@ -1,0 +1,25 @@
+package org.openhab.habdroid.core;
+
+import android.app.Activity;
+import android.content.ComponentName;
+import android.content.Context;
+import android.content.Intent;
+import android.support.v4.content.WakefulBroadcastReceiver;
+import android.util.Log;
+
+/**
+ * Created by jjhuff on 8/18/17.
+ */
+
+public class GeofenceBroadcastReceiver extends WakefulBroadcastReceiver {
+    private static final String TAG = GeofenceBroadcastReceiver.class.getSimpleName();
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        Log.i(TAG, "onReceive");
+        ComponentName comp = new ComponentName(context.getPackageName(),
+                GeofenceIntentService.class.getName());
+        startWakefulService(context, (intent.setComponent(comp)));
+        setResultCode(Activity.RESULT_OK);
+    }
+}
--- a/mobile/src/main/java/org/openhab/habdroid/core/GeofenceIntentService.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/GeofenceIntentService.java
@@ -1,16 +1,11 @@
 package org.openhab.habdroid.core;
 
-import android.app.IntentService;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.content.Context;
 import android.content.Intent;
-import android.content.SharedPreferences;
 import android.graphics.BitmapFactory;
 import android.graphics.Color;
-import android.os.Handler;
-import android.os.Looper;
-import android.preference.PreferenceManager;
 import android.support.v4.app.NotificationCompat;
 import android.support.v4.app.TaskStackBuilder;
 import android.text.TextUtils;
@@ -22,14 +17,9 @@ import com.google.android.gms.location.GeofencingEvent;
 import org.openhab.habdroid.ui.OpenHABMainActivity;
 import org.openhab.habdroid.R;
 import org.openhab.habdroid.util.Constants;
-import org.openhab.habdroid.util.MyAsyncHttpClient;
-import org.openhab.habdroid.util.MyHttpClient;
 
 import java.util.ArrayList;
 import java.util.List;
-
-import okhttp3.Call;
-import okhttp3.Headers;
 
 /**
  * Created by jjhuff on 8/17/17.
@@ -42,7 +32,6 @@ public class GeofenceIntentService extends OpenHABIntentService {
         super(TAG);
     }
 
-
     /**
      * Handles incoming intents.
      * @param intent sent by Location Services. This Intent is provided to Location
@@ -50,11 +39,9 @@ public class GeofenceIntentService extends OpenHABIntentService {
      */
     @Override
     protected void processIntent(Intent intent) {
-        SharedPreferences settings = PreferenceManager.getDefaultSharedPreferences(this);
-
         GeofencingEvent geofencingEvent = GeofencingEvent.fromIntent(intent);
         if (geofencingEvent.hasError()) {
-            Log.e(TAG, "Error:"+geofencingEvent.getErrorCode());
+            Log.e(TAG, "Error:" + geofencingEvent.getErrorCode());
             return;
         }
 
@@ -69,16 +56,16 @@ public class GeofenceIntentService extends OpenHABIntentService {
                 triggeringGeofences);
 
         // Log the transition details.
-        Log.i(TAG, geofenceTransitionDetails);
+        Log.d(TAG, geofenceTransitionDetails);
 
         // Send the notification if enabled
-        boolean notificationsEnabled = settings.getBoolean(Constants.PREFERENCE_PRESENCE_DEBUG, false);
+        boolean notificationsEnabled = mSettings.getBoolean(Constants.PREFERENCE_PRESENCE_DEBUG, false);
         if (notificationsEnabled) {
             sendNotification(geofenceTransitionDetails);
         }
 
         // Send command to presence item
-        String item = settings.getString(Constants.PREFERENCE_PRESENCE_ITEM, null);
+        String item = mSettings.getString(Constants.PREFERENCE_PRESENCE_ITEM, null);
         if (!item.isEmpty()) {
             String command;
             if (geofenceTransition == Geofence.GEOFENCE_TRANSITION_ENTER ||
--- a/mobile/src/main/java/org/openhab/habdroid/core/GeofenceIntentService.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/GeofenceIntentService.java
@@ -5,8 +5,12 @@ import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.content.Context;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.graphics.BitmapFactory;
 import android.graphics.Color;
+import android.os.Handler;
+import android.os.Looper;
+import android.preference.PreferenceManager;
 import android.support.v4.app.NotificationCompat;
 import android.support.v4.app.TaskStackBuilder;
 import android.text.TextUtils;
@@ -17,25 +21,27 @@ import com.google.android.gms.location.GeofencingEvent;
 
 import org.openhab.habdroid.ui.OpenHABMainActivity;
 import org.openhab.habdroid.R;
+import org.openhab.habdroid.util.Constants;
+import org.openhab.habdroid.util.MyAsyncHttpClient;
+import org.openhab.habdroid.util.MyHttpClient;
 
 import java.util.ArrayList;
 import java.util.List;
+
+import okhttp3.Call;
+import okhttp3.Headers;
 
 /**
  * Created by jjhuff on 8/17/17.
  */
 
-public class GeofenceIntentService extends IntentService {
+public class GeofenceIntentService extends OpenHABIntentService {
     private static final String TAG = GeofenceIntentService.class.getSimpleName();
 
-    /**
-     * This constructor is required, and calls the super IntentService(String)
-     * constructor with the name for a worker thread.
-     */
     public GeofenceIntentService() {
-        // Use the TAG to name the worker thread.
         super(TAG);
     }
+
 
     /**
      * Handles incoming intents.
@@ -43,7 +49,7 @@ public class GeofenceIntentService extends IntentService {
      *               Services (inside a PendingIntent) when addGeofences() is called.
      */
     @Override
-    protected void onHandleIntent(Intent intent) {
+    protected void processIntent(Intent intent) {
         GeofencingEvent geofencingEvent = GeofencingEvent.fromIntent(intent);
         if (geofencingEvent.hasError()) {
             Log.e(TAG, "Error:"+geofencingEvent.getErrorCode());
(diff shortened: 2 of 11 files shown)


mobile/src/main/java/org/openhab/habdroid/core/GeofenceIntentService.java
```diff
@@ -0,0 +1,168 @@
...
+
+        // Get a PendingIntent containing the entire back stack.
+        PendingIntent notificationPendingIntent =
+                stackBuilder.getPendingIntent(0, PendingIntent.FLAG_UPDATE_CURRENT);
+
+        // Get a notification builder that's compatible with platform versions >= 4
+        NotificationCompat.Builder builder = new NotificationCompat.Builder(this);
+
+        // Define the notification settings.
+        builder.setSmallIcon(R.drawable.ic_launcher)
```
mueller-ma: Usually apps use their icon with a background in notifications

mobile/src/main/java/org/openhab/habdroid/core/GeofenceIntentService.java
```diff
@@ -0,0 +1,168 @@
...
+                // In a real app, you may want to use a library like Volley
+                // to decode the Bitmap.
+                .setLargeIcon(BitmapFactory.decodeResource(getResources(),
+                        R.drawable.ic_launcher))
+                .setColor(Color.RED)
```
mueller-ma: Why red?

mobile/src/main/java/org/openhab/habdroid/core/GeofenceIntentService.java
```diff
@@ -0,0 +1,168 @@
...
+                .setContentTitle(notificationDetails)
+                .setContentText("Geofence transition")
```
mueller-ma: Please use strings.xml ressources for this.

mobile/src/main/java/org/openhab/habdroid/core/GeofenceIntentService.java
```diff
@@ -0,0 +1,168 @@
...
+    private String getTransitionString(int transitionType) {
+        switch (transitionType) {
+            case Geofence.GEOFENCE_TRANSITION_ENTER:
+                return "GEOFENCE_TRANSITION_ENTER";
+            case Geofence.GEOFENCE_TRANSITION_EXIT:
+                return "GEOFENCE_TRANSITION_EXIT";
+            case Geofence.GEOFENCE_TRANSITION_DWELL:
+                return "GEOFENCE_TRANSITION_DWELL";
+            default:
+                return "Unknown type"+transitionType;
```
mueller-ma: Is this visible to the user? If so please use strings.xml ressources for this.

mobile/src/main/java/org/openhab/habdroid/core/GeofenceRegistrationService.java
```diff
@@ -0,0 +1,118 @@
...
+    public void onConnected(Bundle connectionHint) {
+        Log.d(TAG, "onConnected");
+
+        float lat;
+        float lng;
+        try {
+            lat = Float.parseFloat(mSettings.getString(Constants.PREFERENCE_PRESENCE_LAT, "0"));
+            lng = Float.parseFloat(mSettings.getString(Constants.PREFERENCE_PRESENCE_LNG, "0"));
+        } catch(NumberFormatException e) {
+            Log.i(TAG, "Invalid lat/lng");
```
mueller-ma: Does the user get feedback that lat/lng is invalid?

mobile/src/main/java/org/openhab/habdroid/core/OpenHABIntentService.java
```diff
@@ -0,0 +1,207 @@
...
+        Log.d(TAG, "onError(): " + error);
+        stopSelf();
+    }
+
+
+    protected void processIntent(Intent data) {
+        Log.d(TAG, "processIntent()");
+        if (mOpenHABBaseUrl == null) {
+            Log.w(TAG, "Couldn't determine OpenHAB URL");
+            showToast("Couldn't determine OpenHAB URL");
```
mueller-ma: Please use strings.xml ressources for this.

mobile/src/main/java/org/openhab/habdroid/core/OpenHABIntentService.java
```diff
@@ -0,0 +1,207 @@
...
+                        command, "text/plain;charset=UTF-8", new MyHttpClient.ResponseHandler() {
+                            @Override
+                            public void onSuccess(Call call, int statusCode, Headers headers, byte[] responseBody) {
+                                Log.d(TAG, "Command was sent successfully");
+                            }
+
+                            @Override
+                            public void onFailure(Call call, int statusCode, Headers headers, byte[] responseBody, Throwable error) {
+                                if (statusCode == 404) {
+                                    showToast("Item " + itemName + " doesn't exist");
```
mueller-ma: Please use strings.xml ressources for this.

mobile/src/main/java/org/openhab/habdroid/core/OpenHABIntentService.java
```diff
@@ -0,0 +1,207 @@
...
+                                } else {
+                                    showToast("Error " + statusCode + " while sending " + command + " to " + itemName);
```
mueller-ma: Please use strings.xml ressources for this.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -58,6 +58,40 @@
             android:summary="@string/settings_openhab_sslhost_summary"
             android:title="@string/settings_openhab_sslhost" />
     </PreferenceCategory>
+    <PreferenceCategory android:title="Presence">
+        <CheckBoxPreference
+            android:defaultValue="false"
+            android:key="default_openhab_presence_enable"
+            android:summary="Enable location-based presence reporting"
```
mueller-ma: Please use strings.xml ressources for this.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -58,6 +58,40 @@
...
+            android:title="Enable" />
+        <EditTextPreference
+            android:defaultValue="isPresent"
+            android:dependency="default_openhab_presence_enable"
+            android:inputType="textNoSuggestions"
+            android:key="default_openhab_presence_item"
+            android:summary=""
+            android:title="Presence Item" />
```
mueller-ma: Please use strings.xml ressources for this.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -58,6 +58,40 @@
...
+        <EditTextPreference
+            android:defaultValue=""
+            android:dependency="default_openhab_presence_enable"
+            android:inputType="numberDecimal"
+            android:key="default_openhab_presence_lat"
+            android:summary=""
+            android:title="Latitude" />
```
mueller-ma: Please use strings.xml ressources for this.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -58,6 +58,40 @@
...
+        <EditTextPreference
+            android:defaultValue=""
+            android:dependency="default_openhab_presence_enable"
+            android:inputType="numberDecimal"
+            android:key="default_openhab_presence_lng"
+            android:summary=""
+            android:title="Longitude" />
```
mueller-ma: Please use strings.xml ressources for this.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -58,6 +58,40 @@
...
+        <CheckBoxPreference
+            android:defaultValue="false"
+            android:dependency="default_openhab_presence_enable"
+            android:key="default_openhab_presence_debug"
+            android:summary="Enable debug notifications"
+            android:title="Debugging" />
```
mueller-ma: Please use strings.xml ressources for this.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -58,6 +58,40 @@
...
+            android:title="Presence Item" />
+        <EditTextPreference
+            android:defaultValue=""
+            android:dependency="default_openhab_presence_enable"
+            android:inputType="numberDecimal"
+            android:key="default_openhab_presence_lat"
+            android:summary=""
+            android:title="Latitude" />
+        <EditTextPreference
+            android:defaultValue=""
```
mueller-ma: You could replace "" with "@string/empty_string"

digitaldan:
Hi @jjhuff , thanks for the PR, and thanks @mueller-ma for the initial review.  I need to find a large block of time to get through all the android PR's this week.

mobile/src/main/res/xml/preferences.xml
jjhuff: Not entire clear what you meant, but I just removed the default.

mobile/src/main/java/org/openhab/habdroid/core/GeofenceRegistrationService.java
jjhuff: They do now:)

mobile/src/main/java/org/openhab/habdroid/core/GeofenceIntentService.java
jjhuff: Indeed! This was lifted from elsewhere. Fixed.

jjhuff:
Thanks for the review @mueller-ma. I think I fixed everything you flagged.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -58,6 +58,38 @@
             android:summary="@string/settings_openhab_sslhost_summary"
             android:title="@string/settings_openhab_sslhost" />
     </PreferenceCategory>
+    <PreferenceCategory android:title="Presence">
```
mueller-ma: Hardcoded string

mobile/src/main/res/xml/preferences.xml
```diff
@@ -58,6 +58,38 @@
...
+            android:dependency="default_openhab_presence_enable"
+            android:inputType="numberDecimal"
+            android:key="default_openhab_presence_lat"
+            android:summary=""
+            android:title="@string/presence_pref_lat_title" />
+        <EditTextPreference
+            android:dependency="default_openhab_presence_enable"
+            android:inputType="numberDecimal"
+            android:key="default_openhab_presence_lng"
+            android:summary=""
```
mueller-ma: Can you display the selected value as summary?

jjhuff: That's a good idea, done!

digitaldan:
@jjhuff I just had a chance to run this. Couple of suggestions.  My thoughts for this functionality would be that you select a location from a map (using google map api) and then select and item to switch on when you enter it, off when you exit it.  I was not expecting the user to manually enter a single lat/lon or the actual item name.   Would it be possible to let the user select a map for the location, and also fetch a list of items (/rest/items) from the users server to select?  Even better, supporting multiple geofences -> items would be awesome.

jjhuff:
@digitaldan I think both of those would be great, but I wanted to start somewhere

* Selecting items: This might be a little thorny, but I need to dig into the preferences code some more. I'm not sure we'd always have server info available (especially when initially configuring the app).

* As for the map: I think the best option is for the server to provide the location(s) in most (maybe all) cases. I implemented that by pulling `location` from `i18nprovider`, but I'd prefer a better option. Failing all of that, I agree 100% that selecting from a map is preferable to entering LatLng.

digitaldan:
>Selecting items: This might be a little thorny, but I need to dig into the preferences code some more. I'm not sure we'd always have server info available (especially when initially configuring the app).

What I would like to see is this not in the preferences menu, but in the main menu where we already have "write nfc tag", so something like "Geo Fences".  when selecting this it pops up a window showing your saved geofences , or a button/link to create a new one.  

For selecting an item, If you are not connected to a server then you would have an empty dropdown menu, but could still type into the field (if we think its an issue) , selecting from the dropdown would populate that item name in the field. 

>As for the map: I think the best option is for the server to provide the location(s) in most (maybe all) cases. I implemented that by pulling location from i18nprovider, but I'd prefer a better option. Failing all of that, I agree 100% that selecting from a map is preferable to entering LatLng.

So, I think your point of view here is that this is meant for geofencing your home, but that is only one use case.  For example, I have a geofence when I drive into my neighborhood and another one for when I get home.  I also have a geo fence for work.   Tying this to a single location seems a bit limiting, and not what I would expect from a geo fence feature.

Btw, I love this feature!  We want to bump our app version to 2.0 soon and this could be a great reason to do so.

jjhuff:
> What I would like to see is this not in the preferences menu, but in the main menu where we already have

Gotcha. Yup, that solves a lot of problems.

>  Tying this to a single location seems a bit limiting, and not what I would expect from a geo fence feature.

100% agree! However, my main point is that I think that the fences should be defined/stored on the server rather than every single client. Perhaps I have multiple devices, a family, or employees.

Proposal 1:
What if we made a service that does nothing but export a list of geofences similar to how the OwnTracks stuff is configured

Proposal 2:
Perhaps better, configure all of it in the item. Then the app just needs to select a list of geofence items to update!
```
Switch  PresenceBen_PhoneHome   "Ben @ Home"   { geofence="-47.456,122.366,100" }
Switch  PresenceBen_PhoneWork   "Ben @ Work"   { geofence="-47.134,122.655,100" }
```
I suppose you could also have a config file to assign symbolic names to replace the lat,lng,radius values.

digitaldan:
Apologies for the very late reply, I am catching up on PR's today.  

So i had to think about this, I still would prefer that a user selects a location and radius from a map, then selects an item.  I think this would be amazingly simple and be used frequently.

I am however trying to also see it from your POV, and so keeping an open mind.  

 For this project, we always have to keep in mind that it's the official Android app for openHAB.  This means we need to keep design focused on complimenting the openHAB server and its built in functionality.  So I'm generally opposed to requiring users to have to install a new binding unless it would absolutely be necessary .  I also believe we can not overload basic OH concepts like items or sitemaps to meet bespoke functionality.  

That being said, we do have the concept of a Location item.  If the user were to select a location item and another item to switch on or off I have a couple questions on how this would work.

1) how would we select and store the geofence radius? 
2) what happens when the user changes this on the server?  

>100% agree! However, my main point is that I think that the fences should be defined/stored on the server rather than every single client. Perhaps I have multiple devices, a family, or employees.

My opinion is that this over complicates the feature, but again its' just my opinion.  

>Proposal 1:
What if we made a service that does nothing but export a list of geofences similar to how the OwnTracks stuff is configured

The android app should work without any special configuration on the server that is specific to the app, i would like to avoid the user having to install a binding unless there was no alternative.

>Proposal 2:
Perhaps better, configure all of it in the item. Then the app just needs to select a list of geofence items to update!

I think we are overloading what an item is, so I don't think we can do that in this app. 

I'm hoping you are still interested in developing this feature!  I think it would be awesome.

lolodomo:
Just an idea: why not having something similar to the voice recognition, that is the Android app updating a location item with the current phone position ? Then the user could use this position in rules and ser switch for example when the phone position is around a particular position.
Of course, the item name has to be a setting in the Android app because we must consider and differenciate several phones.

With which frequence the app will update the position ? With what impact on phone battery ?

digitaldan:
>With which frequence the app will update the position ? With what impact on phone battery ?

I think that's the issue, not only would our app now wake up in the background for location changes events, but it also tries to wake up the radio and make REST calls at some frequent interval.  I was hoping not to have to deal with possible battery issues with our app.

jjhuff:
I apologize for being MIA -- new job sucked away all of my free time!

How are people feeling about where this sits? I see an approval (I missed that earlier). I'm happy to clean up the merge conflicts if there's still overall interest in the feature.

I agree that a map would be a logical next step, but I'm also not super keen to draw this out.

Thoughts?

digitaldan:
>Having the data stored on client isnt very practical, especially since you dont have a map to pick the location.

Maybe I don't understand, apps that I use that do geo fencing pop up a google map for me to select a location and radius, so we can have client access to a map, yes? 

Also on Android you have to register for entry/exits events to geo-fence locations, which implies the client does need to store the coordinates and some item to trigger locally.  In fact, I believe on each reboot of the device, apps need to listen for a boot complete action (or something) and re-register its geofences.

jjhuff:
FWIW, I'm actually a big fan of having the geofences come entirely from the server -- I'm just not sure of the best mechanism to make that happen.

mueller-ma:
Maybe as state of a string item, but thats not the best solution

mueller-ma:
I really like proposal 2 of @jjhuff, but that would need some work on the server, wouldn't it?

mueller-ma:
Triggering travis

kubawolanin:
Looking forward to test this!
Is there anything missing on this PR (apart frorm the conflicts) to have it merged to the master branch?
Best regards

mueller-ma:
Currently you have to enter the location's coordinates. It would be a lot nicer to fetch them from the server.

mobile/src/main/java/org/openhab/habdroid/core/GeofenceBroadcastReceiver.java
```diff
@@ -0,0 +1,31 @@
...
+    public void onReceive(Context context, Intent intent) {
+        Log.d(TAG, "onReceive");
+        if(intent.getAction()!= null && intent.getAction().equals(Intent.ACTION_BOOT_COMPLETED)){
+            // Did we just boot? Go register.
+            Intent serviceIntent = new Intent(context , GeofenceRegistrationService.class);
+            startWakefulService(context, serviceIntent);
+        } else {
+            ComponentName comp = new ComponentName(context.getPackageName(),
+                    GeofenceIntentService.class.getName());
+            startWakefulService(context, (intent.setComponent(comp)));
```
maniac103: The if and the else block are 100% equivalent.

lolodomo:
I just propose to implement a new geofence service in the server explaning how any UI could interact with.
https://github.com/eclipse/smarthome/issues/5066
I am waiting for your comments.

lolodomo:
My proposal has the advantage to have the geofences defined on server side. Using Paper UI, it will be easy to select the locations insde a map (coming feature in openHAB).
By the way, the Android app could then display all the geofences in a map too, to help the user select the one to use.
The other difference is that the computation whether device is inside geofence or not, is done by the server, the UI just has to call the server with its current position, the selected geofence and an item to update.
WDYT ?

maniac103:
I see that (computation being done server side) as a problem though. When doing it client side, one can leverage OS features (here: Play Services) when determining geofence status. When doing it server side, there's no other option than regularly polling location, which likely is less efficient.

lolodomo:
No problem, we can have a service mode where the geofence status is determined by the client. In this case, the client only call the geofence service when it detects a change of status; the service will not compute the status again but consider the status provided in parameter.

lolodomo:
In case the app only call the server when it enters or leaves a geo-fence, we have no chance to track the position of the person inside openHAB server. Maybe the service could be called by the Android app every XX minutes if the position has changed in order to set a location item. XX should be configurable.

maniac103:
A geofence is not the same thing as location tracking. The former only provides an in/out information, the latter provides the actual location. We could probably support both modes, but the latter rather sounds like an extension of Location items.

lolodomo:
Location tracking could already be done by any client UI simply by updating a particular location item. That is true that no additional service is required for that.

I thought about location tracking because if we have a service that receives the position for geofence computation, it is easy to track position at the same time. But that would be just an option of the service.

lolodomo:
At least the service could allow sharing geofences. It could be used to save on server side new geofences defined by a client.

kaikreuzer:
> @kaikreuzer rejected my proposal

... sounds a bit hard, I would say I questioned the suggested way to implement it.

But I think implementing this in the client side only isn't too bad as the precision and size of the fence might anyhow depend on the OS.

mueller-ma:
Superseded by #914

PR #358, opened 2017-09-01
For security reasons, SSL hostname checks should be enabled by default

mueller-ma:


maniac103:
This change looks incomplete to me. AFAICT, not only the prefs side of this needs to have their default value changed, but also the code that accesses the setting via SharedPreferences.

Issue #300, opened 2017-05-20
Theme issues


mueller-ma:
- Selection items are in bright themes black and in dark themes white and do not look like normal buttons that are mostly grey. See #537
- Version entry in settings is greyed out. (Saw this on a Oreo AVD, but not on a real 7 device) Edit: The version field in the settings has been removed.
- Menu isn't themed (Fix in #361)

![screenshot_2017-05-20-11-52-43](https://cloud.githubusercontent.com/assets/22525368/26275123/cf0d436a-3d59-11e7-92ce-b5fe9fd7e2b6.png)

mueller-ma:
@digitaldan Can you explain how the themes work here? Android Studio says the theme HabDroid.black isn't used anywhere and in the preview the menu is themed

digitaldan:
Sorry I missed this.  I have no idea, have not done any development around themes.

mueller-ma:
Selection items are in bright themes black and in dark themes white and do not look like normal buttons that are mostly grey

lolodomo:
@mueller-ma : isn't it fixed now ? I see now the right menu dark when the theme is dark.

mueller-ma:
The first point isnt fixed.

PR #361, opened 2017-09-07
Add basicui-like themes

mueller-ma:
This PR adds a bright and a dark theme that have the same colorset than the basic ui.
I also changed the button color from colorPrimary to colorAccent. This does not make a difference in the old themes, but in the basic ui theme the buttons are pink.
If the method setActivityTheme failes to resolve the string, it defaults to the default theme. (That happend once, fix is in #292, but it may happen again and than we are save)
I replace some hex codes in themes.xml with color tags.

#300 also applies here.

screenshot_1504780800
screenshot_1504781287
screenshot_1504781314


mueller-ma:
Can be merged.
Tested on several devices.

mueller-ma:
Here is some feedback: https://community.openhab.org/t/basic-ui-like-themes-for-android-app/34208

digitaldan:
LGTM!

mueller-ma:
People in the poll voted for "I really like it, the bright one should be the default theme", but I see a problem. The openhab color is orange, so I woulnd change the icon color, but than we would have a orange icon that starts a blue/pink app.
So I would leave the orange theme as default.

PR #364, opened 2017-09-08
Minor improvements to items with buttons

mueller-ma:
- Colorpicker, setpoint and rollershutter item buttons have the same size now.
- Icon of colorpicker button changed to item used in basic ui


digitaldan:
so  mobile/src/main/res/drawable-hdpi/colorwheel.png is always showing the nearly transparent version here, which I don't think we want.  The basicUI is probably setting some transparent value when it displays this image..

mueller-ma:
I have taken it from the google material repo and it is always white with transparent background.

digitaldan:
Sorry, i just checked and realized the basicUI is just using the dark version of this, for some reason I thought it was altering its color/transparency based on the item value.

PR #371, opened 2017-09-15
Add splash screen

mueller-ma:
This PR adds a screen with the openhab logo on startup.

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -36,14 +36,14 @@
         android:roundIcon="@drawable/openhabicon_material_round"
         android:label="@string/app_name"
         android:resizeableActivity="true"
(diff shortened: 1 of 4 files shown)


maniac103:
https://medium.com/@ssaurel/create-a-splash-screen-on-android-the-right-way-93d6fb444857

(There's also similar posts by Android developers themselves, but I seem to be unable to find them at the moment)

An intermediate activity has the downside of
- longer load times
- unexpected navigation if you navigate away from the splash screen

maniac103:
Oops, I should have read that link that I quickly dug out - it also uses that intermediate activity. I'll look further, I'm pretty sure there's a way without that.

mueller-ma:
@maniac103 Can you have a look at my latest commit? I dont see the splash screen comming up

mueller-ma:
Thanks. Can you see my commit?

PR #372, opened 2017-09-15
Display images as PNG when icon format is set to SVG

### LINKED ISSUE #286: Charts not shown when Icon format=SVG
Hi,

with openhab2 and habdroid 1.8.0.6, when setting "Icon format" is set to "SVG" charts are not rendered anymore.
When changing the setting backto PNG, charts continue to function.

Best,
Sascha

mueller-ma:
The problem is that all images (icons, charts and external images) are
rendered in the same format and the svg libary will fail if it gets a
png.

At this point we dont really have to read the preferences since
mobile/src/main/java/org/openhab/habdroid/model/OpenHAB2Widget.java Line
21 inserts format=(format here) in the url and we can parse this url.

Fix #286

I'm not sure if that is the best solution.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/MyWebImage.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/MyWebImage.java
@@ -69,7 +69,10 @@ public class MyWebImage implements SmartImage {
             	bitmap = webImageCache.get(url);
             if(bitmap == null) {
             	Log.i("MyWebImage", "Cache for " + url + " is empty, getting image");
-                final String iconFormat = PreferenceManager.getDefaultSharedPreferences(context).getString("iconFormatType","PNG");
+                String iconFormat = "PNG";
+                if (url.contains("format=SVG")) {
+                    iconFormat = "SVG";
+                }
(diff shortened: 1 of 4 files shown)


mueller-ma:
I didnt clear the cache while testing. It only renders png

digitaldan:
Hmm, so how would a user decide to use PNG vs SVG?  I did not think this was a system wide setting, but specific to each UI.

mueller-ma:
The problem is: charts are only available in png, so when they are tried to interpreted as svg, it fails and nothing is shown. After this commit, icons will still be in svg, when the user wants it.

digitaldan:
Does this affect more than just charts though?  Would this not apply to every icon?  When testing this the client now pulls every icon in my sitemap in PNG format instead of SVG, so how would I use SVG images now?

mueller-ma:
The settings are not changed. When a icon is downloaded it builts the url and add "&format=SVG|PNG" depending on the setting, but for charts there is never added a format attribute.
So when the URL contains format=SVG than render as svg, else png

digitaldan:
Sorry, i did not read your changes correctly and also forgot I wiped my settings testing another PR, which is why I was downloading all PNG files.

PR #373, opened 2017-09-16
Decrease chart resolution so text size is increased

mueller-ma:
It is not possible to request a chart with higher font size so the only way is to decrease the resoltution. Idk if slightly unsharp charts are better than not readable charts.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -506,8 +506,8 @@ public class OpenHABWidgetAdapter extends ArrayAdapter<OpenHABWidget> {
             ViewGroup.LayoutParams chartLayoutParams = chartImage.getLayoutParams();
             chartLayoutParams.height = (int) (screenWidth/2);
             chartImage.setLayoutParams(chartLayoutParams);
-            chartUrl += "&w=" + String.valueOf(screenWidth);
-            chartUrl += "&h=" + String.valueOf(screenWidth/2);
+            chartUrl += "&w=" + String.valueOf(screenWidth/1.3);
+            chartUrl += "&h=" + String.valueOf(screenWidth/2.6);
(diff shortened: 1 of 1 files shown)


hreichert:
:thumbsup: for this change.
Especially on hi-res smartphones the text is very small.
On my OnePlus 3 in portrait mode the text height is only 2 millimeter or so.

mueller-ma:
We will have to see how it looks on high/low res smartphones/tablets. Maybe we need to parse dpi, screen size, etc to have a good experience on all devices.
However we will need a parameter like '&fontsize=10px' when getting a chart

hreichert:
@mueller-ma I'm currently working on some enhancements to the ESH DefaultChartProvider (support for themes / different looks). You made me think of something like a "scale" parameter...

mueller-ma:
https://github.com/eclipse/smarthome/issues/4284

Also a "hide-legend" might be a good idea. E.g. tab on a chart to hide the legend if its too big.

hreichert:
[Charts] Enhancements to Charts: Themes, FontScale:
eclipse/smarthome#4288

Your "hide-legend" suggestion is also a good idea.

mueller-ma:
This PR was just a quick workaround. Now the text scaling option has been merged in ESH, but it will need some time until it is in openhab 2.2. So for now I would suggest to merge it and when oh2.2 is out I undo
 the change and come up with a correct fix

digitaldan:
>So for now I would suggest to merge it and when oh2.2 is out I undo the change and come up with a correct fix

I worry that changes like this get lost and never undone.  Since we have lived with it this long, can we not fix this the right way now?  Then when users upgrade to 2.2 this will already be fixed for them.

mueller-ma:
> I worry that changes like this get lost and never undone.

It won't get lost :)
The better solution is ready, but I cannot test it, since my oh is on stable. I will upload it and maybe @hreichert can test it.

hreichert:
@mueller-ma Yes, sure!

mueller-ma:
@hreichert See #386

PR #375, opened 2017-09-16
Display correct icon for dimmer items controlled by a switch

### LINKED ISSUE #307: Wrong icon displayed for a Switch element linked to a Dimmer item
Here is the item:
```
Dimmer LampeChambre "Lumière dans la chambre [%d %%]" <dimmablelight> {mios="unit:house,device:510/service/Dimming1/LoadLevelStatus", autoupdate="false"}
```
Here is a part of the sitemap:
```
		Text item=GEclairage {
			Switch item=LampeChambre icon="light"
		}
```
The switch state is perfectly updated (OFF or ON) but the icon is always unchanged, a light ON, even when the light status is OFF.

mueller-ma:
Fix #307

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/OpenHAB2Widget.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/OpenHAB2Widget.java
@@ -20,13 +20,18 @@ public class OpenHAB2Widget extends OpenHABWidget {
         String itemState;
         if (widgetItem != null) {
             itemState = widgetItem.getState();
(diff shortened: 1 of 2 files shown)


lolodomo:
Please consider any value greater than 0 as ON and not only 100.

mueller-ma:
Can a switch set a dimmer item to e.g. 10?

lolodomo:
No but a dimmer item can be set to 10 for example.

digitaldan:
LGTM

PR #378, opened 2017-09-24
Fix not-showing Notifications menu

### LINKED ISSUE #356: Notifications missing
This was reported at https://community.openhab.org/t/habdroid-notifications-tab-missing/33471

```
I no longer have access to the Notifications tab on my Hamburger/Flyout menu in openHAB Android 1.8.0.8 925). My flyout now has Sitemaps, Discovery, and Bindings, but no more access to Notifications.

I still see all my notifications on the https://myopenhab.org/notifications interface, but I can no longer see them on Android at all. I use the sendLogNotification(“blah blah blah”) command pretty heavily so I can keep an eye on what’s going on, but now I can’t see it from Android anymore.
```

FlorianSW:
This was broken by
4ae13d59da2a402c34d50899ca8b25ca6bff85f5 as mIsMyOpenHAB was not set
anymore (and I thought it was not used anywhere).

With this commit, the app will now try to acquire the NotificationSettings
object and shows the Notification menu entry only, if that succeeds (which
means, that the remote URL is used).

Fixes #356


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -1289,19 +1288,15 @@ private void loadDrawerItems() {
             }
             mDrawerItemList.add(OpenHABDrawerItem.dividerItem());
         }
-//        mDrawerItemList.add(OpenHABDrawerItem.menuItem("Favorites", getResources().getDrawable(R.drawable.ic_star_grey600_36dp)));
```
mueller-ma: I think this line is the only "usage" of ic_star_grey600_36dp. Can you let this line here or also remove the star?

FlorianSW: As I don't like commented out code, as we can always go back in the history of git and look at the code, if we need to, I'll remove the star :)

digitaldan:
Thanks

PR #380, opened 2017-09-24
Open Notifications fragment when notification is selected

### LINKED ISSUE #379: Pressing a notification just opens the openHAB app
When a notification arrives, it will create a notification element in the notification bar with the message (which is also stripped when it's too long). When the user presses this notification, the only action taken is to open the openHAB app.

Instead of doing that, the openHAB app should be opened with the Notification tab opened by default, or the message should be visible as a full text string (e.g. in an "OK-able" message window) when the notifications tab is not available, e.g. because you're connected to the local network, see also #253).

FlorianSW:
If a notification from the systems notification bar is selected, the app
now opens the notifications fragment, if it is available currently
(openHAB remote url with username and password is used).

It also shows an alert box with the message content of the selected
message, if the message was added to the intent (should always happen).
This will happen regardless if the notification fragment is opened or not
to show the full message to the user (again).

Fixes #379


digitaldan:
LGTM, thanks!

Issue #381, opened 2017-09-26
Support for Android Wear 2.0


DanielMalmgren:
Since upgrading my watch to Wear 2.0 (and thus getting the play store directly in the watch) I have got an item for openHAB in the category "Apps on mobile" in my watch's play store, with a choice to install the app on the watch. However, clicking on install doesn't seem to work, it just says "installing" forever. So there seems to be some kind of support for Wear but it's not really working. Is this something that is supposed to be working?

Posted about this in the forum a while ago and  judging by the comments I'm not alone with the problem: [https://community.openhab.org/t/app-in-android-wear-2-0/27024](url)

mueller-ma:
Duplicate of #73

I think we should remove the wear dummy folder to not confuse people

mueller-ma:
@DanielMalmgren Please close this issue

PR #382, opened 2017-09-26
Remove wear folder

mueller-ma:
The wear folder in its current state is just an empty wear app template.
It causes a lot of confusion on github and play store, because people
think they can use openhab on their android wear 2.

See #381 


digitaldan:
I agree, this should go!

PR #350, opened 2017-08-23
Menu rework

mueller-ma:
This PR will

- [x] Add several icons with different resolutions
- [x] Translate "Voice Recognizition"
- [x] Disable NFC Menu entry, if no NFC is available (will need help for this)
- [x] Reorder menu entries
- [x] Show NFC in the action bar
- [x] Apply material typing to strings

PR #384, opened 2017-09-29
Some improvements to the menu

mueller-ma:
This commit will
* Add several icons with different resolutions
* Translate "Voice Recognizition"
* Disable NFC Menu entry, if no NFC is available
* Reorder menu entries
* Apply material typing to strings

Same as #350 


mueller-ma:
The commit message is wrong: It never shows nfc in the action bar

mobile/src/main/res/values/strings.xml
```diff
@@ -102,6 +102,9 @@
     <string name="mtm_notification">Certificate verification</string>
     <string name="settings_openhab_none">none</string>
     <!-- theme -->
+    <string name="theme_value_light" translatable="false">light</string>
+    <string name="theme_value_dark" translatable="false">dark</string>
+    <string name="theme_value_black" translatable="false">black</string>
     <string name="theme_value_light">light</string>
```
digitaldan: Should this be removed?  They are listed here twice now.

digitaldan:
Hi, this LGTM except for the one comment and the merge conflict, I can merge when those are resolved.

digitaldan:
Also I noticed settings was moved from the first entry to the last, even though its probably the most used menu, is there a thought why?

mueller-ma:
I always expect that the settings is one of the last points

digitaldan:
@mueller-ma unfortunately the string files seem to be in constant change with every PR, can you fix the conflicts?

PR #388, opened 2017-10-01
Listen for gadgetbridge broadcast

mueller-ma:
This commit gives the user the ability to set up a custom broadcast receiver that changes an item on receive.

This was designed to send Mi Band 2 button presses to the openHAB server. This will only work if an app is installed, that send broadcasts on button presses. Since Gadgetbridge does so, it is the default broadcast.

I had to create a foreground service, that starts the receiver. Otherwise it would be killed by the os. Foreground services have to show a notification, but it shown as low priority, so it shouldnt disturb the user.

Before this will get merged, I want to test it on my phone if everything works fine.


mueller-ma:
Closing for now

PR #389, opened 2017-10-01
Add method that handles all messages shown to the user

mueller-ma:
I think it's bad to have multiple functions just for displaying messages in different froms. This PR creates one funtion that replaces all other "message" functions. Additionally we display some errors as dialogs and some as toasts. This seems very inconsistent.
Toasts are kind of outdated: IMO they don't look as good as snackbars and are displayed even when openhab is closed.
It also is a good compromise for #258. When you only have a local oh server, never display messages like "Connecting to local openhab". If you dont want this, you can enable all messages in the settings, kind of a debug mode.
HTTP error handling is improved and common errors can now be translated.


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -110,27 +112,37 @@
...
+                    showMessageToUser(String.format(getString(R.string.error_http_connection_failed), statusCode), "dialog", 5);
+                }
+            } else if (error instanceof java.net.UnknownHostException) {
                 Log.e(TAG, "Unable to resolve hostname");
-                showError(error.getMessage());
+                showMessageToUser(getString(R.string.error_unable_to_resolve_hostname), "dialog", 5);
             } else if (error instanceof SSLHandshakeException) {
-                showError(getString(R.string.error_connection_sslhandshake_failed));
+                // if ssl exception, check for some common problems
+                if (error.toString().contains("java.security.cert.CertPathValidatorException")) {
```
maniac103: Wouldn't `error.getCause() instanceof CertPathValidatorException` be a better check?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -551,8 +560,75 @@ public void onSuccess(Call call, int statusCode, Headers headers, String respons
...
+     *
+     * @param message message to show
+     * @param messageType can be dialog, snackbar or toast (deprecated)
+     * @param logLevel 0 when debug is enabled
+     *                 1 when remote url is configured
+     *                 2 when local url is configured
+     *                 4 when debug is disabled
+     *                 5 always
+     */
+    public void showMessageToUser(String message, String messageType, int logLevel) {
```
maniac103: mesageType should be an int or enum, shouldn't it?

mueller-ma: Or a constant?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
mueller-ma: Yes, it is. Thanks.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
mueller-ma: I created constant int for that. Thanks.

mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
```diff
@@ -111,9 +112,29 @@ public void onOpenHABTracked(String baseUrl, String message) {
         stopSelf(getLastStartId());
     }
 
+    /**
+     * @param message message to show
+     * @param messageType must be Constants.MESSAGES.DIALOG
+     * @param logLevel not implemented
+     */
+    public void showMessageToUser(String message, int messageType, int logLevel) {
+        switch (messageType) {
```
FlorianSW: if you only handle one possible value without any default, I would suggest using:
if (messageType != Constants.MESSAGES.DIALOG)
  return;

mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
```diff
@@ -111,9 +112,29 @@ public void onOpenHABTracked(String baseUrl, String message) {
         stopSelf(getLastStartId());
     }
 
+    /**
+     * @param message message to show
+     * @param messageType must be Constants.MESSAGES.DIALOG
```
FlorianSW: Should probably throw an exception if not DIALOG?

Another thing: If that must be MESSAGES.DIALOG and because logLevel is not implemented, you probably want to add an overloaded method with the message parameter only and call this one with the only accepted values. Or, even better, remove these parameters from this method, if you do not plan to use them in the very short future (they can easily be added later, if needed).

mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
```diff
@@ -203,7 +224,7 @@ private void showToast(final String message) {
         new Handler(Looper.getMainLooper()).post(new Runnable() {
             @Override
             public void run() {
-                Toast.makeText(getApplicationContext(), message, Toast.LENGTH_LONG).show();
+                showMessageToUser(message, Constants.MESSAGES.SNACKBAR, Constants.MESSAGES.LOGLEVEL.ALWAYS);
```
FlorianSW: That would never show anything, as MESSAGES.SNACKBAR is not handled, right?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -110,27 +112,37 @@
...
+                    showMessageToUser(String.format(getString(R.string.error_http_connection_failed), statusCode), Constants.MESSAGES.DIALOG, Constants.MESSAGES.LOGLEVEL.ALWAYS);
+                }
+            } else if (error instanceof java.net.UnknownHostException) {
                 Log.e(TAG, "Unable to resolve hostname");
-                showError(error.getMessage());
+                showMessageToUser(getString(R.string.error_unable_to_resolve_hostname), Constants.MESSAGES.DIALOG, Constants.MESSAGES.LOGLEVEL.ALWAYS);
             } else if (error instanceof SSLHandshakeException) {
-                showError(getString(R.string.error_connection_sslhandshake_failed));
+                // if ssl exception, check for some common problems
+                if (error.getCause() instanceof java.security.cert.CertPathValidatorException) {
```
FlorianSW: Why do you don't use imports here?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -551,8 +560,72 @@ public void onSuccess(Call call, int statusCode, Headers headers, String respons
...
+     * You might want to send two messages: One detailed one with
+     * logLevel Constants.MESSAGES.LOGLEVEL.DEBUG and one simple message with
+     * Constants.MESSAGES.LOGLEVEL.NO_DEBUG
+     *
+     * @param message message to show
+     * @param messageType can be one of Constants.MESSAGES.*
+     * @param logLevel can be on of Constants.MESSAGES.LOGLEVEL.*
+     */
+    public void showMessageToUser(String message, int messageType, int logLevel) {
+        if (message != null) {
```
FlorianSW: if (message == null)
  return;

That would make the code less complex :)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -551,8 +560,72 @@ public void onSuccess(Call call, int statusCode, Headers headers, String respons
...
+                            return;
+                        }
+                        break;
+                    case Constants.MESSAGES.LOGLEVEL.LOCAL:
+                        if (localUrl.length() > 1) {
+                            Log.d(TAG, "local URL set, show message: " + message);
+                        } else {
+                            Log.d(TAG, "no local URL set, dont show message: " + message);
+                            return;
+
```
FlorianSW: unneeded empty line.

mobile/src/main/java/org/openhab/habdroid/util/Constants.java
```diff
@@ -26,5 +26,19 @@
     public static final String PREFERENCE_TONE              = "default_openhab_alertringtone";
     public static final String PREFERENCE_SSLCLIENTCERT     = "default_openhab_sslclientcert";
     public static final String PREFERENCE_SSLCLIENTCERT_HOWTO = "default_openhab_sslclientcert_howto";
+    public static final String PREFERENCE_DEBUG_MESSAGES      = "default_openhab_debug_messages";
     public static final String DEFAULT_GCM_SENDER_ID        = "737820980945";
-}
+
+    public interface MESSAGES {
```
FlorianSW: I'm not sure if these values could be saved in a separate file? That would remove the requirement of using COnstants-MESSAGES.LOGLEVEL.NO_DEBUG, which, at least for me, looks pretty long :)

mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
mueller-ma: OpenHABVoiceService implements OpenHABTrackerReceiver, so I need a function with that parameters since they are used in MainActivity.

mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
mueller-ma: Right, I missed that. I change this to a toast since I want to be sure that it will be displayed

mobile/src/main/java/org/openhab/habdroid/util/Constants.java
mueller-ma: If you just type "ALWAYS" Android Studio asks to make a static import.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
FlorianSW: I would use imports :)

digitaldan:
@mueller-ma can you resolve the conflicts? Thanks!

mobile/src/main/res/values/strings.xml
```diff
@@ -66,19 +66,37 @@
...
     <string name="info_demo_mode">Running in demo mode. To connect to your openHAB server please go to settings menu.</string>
+    <string name="info_demo_mode_short">Running in demo mode</string>
     <string name="info_conn_url">Connecting to local URL</string>
     <string name="info_conn_rem_url">Connecting to remote URL</string>
     <string name="info_discovery">Discovering openHAB. Please wait...</string>
     <string name="error_no_url">Please check that openHAB is running or configure openHAB URL or remote URL</string>
-    <string name="error_authentication_failed">Authentication Failed. Please check Username/Password settings</string>
     <string name="error_empty_sitemap_list">openHAB returned empty sitemap list</string>
     <string name="error_network_not_available">Network is not available</string>
+    <string name="error_http_connection_failed">Got invalid answer from openHAB. Please check if openHAB is working correctly (HTTP response code %d)</string>
```
mueller-ma: Does it make clear, that oh server responed, but a http error occured?

digitaldan: I'm ok with it, but i would say "Received an invalid response from openHAB.  Please check if openHAB is working correctly (HTTP response code %d)"

Issue #359, opened 2017-09-05
Rendering issue with "mappings=[...]"


Adminius:
Hello,

i have some issues on Android app. Browser view is better, but also not perfect:
App:
![screenshot_20170905-092155](https://user-images.githubusercontent.com/17052262/30049858-7a648e64-921c-11e7-8d8e-8943d1f63c3f.png)
Browser:
![screenshot_20170905-092130](https://user-images.githubusercontent.com/17052262/30049864-8269ff9a-921c-11e7-99da-1064d2d92909.png)

Landscape mode is ok:
App:
![screenshot_20170905-092211](https://user-images.githubusercontent.com/17052262/30049886-8eb07b8a-921c-11e7-9ac4-88b41d3cfbca.png)
Browser:
![screenshot_20170905-092114](https://user-images.githubusercontent.com/17052262/30049893-95de788a-921c-11e7-8825-206a640138bf.png)

It well be better to have the same "..."  behavior in the app as in a browser .

mueller-ma:
Duplicate of #121

Adminius:
since 2015 not solved :/

mueller-ma:
No one is preventing you from fixing it ;)

mueller-ma:
I'm trying to fix it

mueller-ma:
@Adminius Since this is a duplicate can you close it?

Adminius:
Ok

mueller-ma:
Can you test my proposed fix in #360

PR #390, opened 2017-10-01
Second try to fix #121 and #359

### LINKED ISSUE #121: Switch label and item text overlaps if too many mappings defined
For a chart, I am using a switch item with 4 mappings:

```
Group item=Humidity label="Luftfeuchtigkeit" {
      Switch item=HumidityChartPeriod label="Zeitraum" mappings=[0="Stunde", 1="Tag", 2="Woche", 3="Monat"]
      Chart item=HumidityChart period=4h refresh=60000 visibility=[HumidityChartPeriod==0]
      Chart item=HumidityChart period=D refresh=60000 visibility=[HumidityChartPeriod==1, HumidityChartPeriod=="Uninitialized"]
      Chart item=HumidityChart period=W refresh=60000 visibility=[HumidityChartPeriod==2]
      Chart item=HumidityChart period=M refresh=60000 visibility=[HumidityChartPeriod==3]
    }
```

On my Nexus 4, the label of the switch item overlaps with the mappings:
switchitem

Would be nice if there was some solution where texts don't overlap
##

mueller-ma:
screenshot_2017-10-01-22-18-57

### DIFF
--- a/mobile/src/main/res/layout/openhabwidgetlist_icontext.xml
+++ b/mobile/src/main/res/layout/openhabwidgetlist_icontext.xml
@@ -23,6 +23,8 @@
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_gravity="center_vertical"
+        android:ellipsize="end"
+        android:maxLines="1"
         android:textAppearance="?android:attr/textAppearanceMedium" />
 
 </LinearLayout>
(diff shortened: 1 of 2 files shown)


FlorianSW:
Wouldn't it be possible to wrap the content of the widget name and fill out multiple lines?

mueller-ma:
Note that the screenshot was taken on a tablet. On my smartphone there isn't enough place for a second line (I think).

digitaldan:
So this is hard for me to test, and since it's since it's all layout logic, hard to read and know exactly what it's doing. So, my question is, are you comfortable with me merging this and it going out to users, or do you need time to better perfect it?  I understand how difficult GUI alignment issues can be.

mueller-ma:
I'm sure, that this solution is way better than the previous fix. I have some trouble connecting virtual devices to my own oh server, so I have only tested it on my low res tablet. Dan, can you show a screenshot like in the previous fix?

digitaldan:
Sounds good, I will test tomorrow and merge if all looks ok.

PR #374, opened 2017-09-16
Add app intro

mueller-ma:
I use a libary under Apache: https://github.com/apl-devs/AppIntro
How should we give credit to the dev? Under the about section?

Signed-off-by: mueller-ma <mueller-ma@users.noreply.github.com>

PR #391, opened 2017-10-02
Re-order connection settings and add separate user/pass for local ope…

### LINKED ISSUE #354: Different username for local and for remote access
(I hope it is not duplicate, couldn't find it by searching)

Expected Behavior
Having two set of fields for local and remote credentials.

Current Behavior
When accessing openhab that requires username and password from a local network a credentials are required. Same goes for when using remote access for openhab (via myopenhab.org f.eks)
However, currently there is only one set of fields for credential, requiring that local and remote credentials are the same. That complicates things even more with multiple users for single openhab instance.

Possible Solution
Separate login fields for local and remote access. F.eks. Having two extra login and password fields for remote access.

This will increase security in case someone gets on the local network, the username for openhab is required. Today almost only easy solution is to have local access is without authentication.

FlorianSW:
…nHAB

The user may want to connect to a local openHAB instance, which is also
protected by a username/password, which is most-likely a different set of
credentials as for the remote one. TO allow specifying different sets of
credentials for the remote and the local URL, a separate, second
username/password field is added for the local connection.

To have a cleaner and better separated preferences view, the connection
settings are also re-ordered and now grouped for the openHAB local
connection settings (URL, username, password) as well as the openHAB remote
connection settings (URL, username, password). Also, the encryption
settings were moved to it's own preferencescreen "Encryption settings".

Fixes #354


mobile/src/main/res/xml/preferences.xml
```diff
@@ -6,82 +6,103 @@
...
+            <Preference
+                android:clickable="true"
+                android:defaultValue="@string/settings_openhab_none"
+                android:dependency="default_openhab_demomode"
+                android:key="default_openhab_sslclientcert"
+                android:summary=""
+                android:title="@string/settings_openhab_sslclientcert"/>
+            <Preference
+                android:clickable="true"
+                android:dependency="default_openhab_demomode"
```
mueller-ma: I'm not allowed to view the howto when demo mode is activated?

FlorianSW: I haven't changed that, it was the same dependency before, too, so yes :/

digitaldan:
Thanks @FlorianSW , this is a simple solution to something that has perplexed people for sure.  Questions,  how will this work for people when they upgrade, will this break until they go into the settings?

FlorianSW:
@digitaldan The current setting will be used for remote connections. If the local URL also needs a username + password, this will be broken, until the user sets the username/password again in the settings :/

mueller-ma:
When we release this in the same version than #374 we could copy the username and password [here](https://github.com/openhab/openhab.android/pull/374/files?diff=unified#diff-07e4a6e281af02db331ff3f81d9f86d8R255).

digitaldan:
@FlorianSW can you a) resolve the merge conflicts and b) think about @mueller-ma 's suggestion about coping the credentials, I had a similar thought. It would be nice not to break this when people upgrade.

maniac103:
> When we release this in the same version than #374 we could copy the username and password at this point.

When doing that (which is a good idea) it's probably a good idea to add a 'version' pref that serves as a 'needs conversion' indicator.

mueller-ma:
> When doing that (which is a good idea) it's probably a good idea to add a 'version' pref that serves as a 'needs conversion' indicator.

Yes, I think thats a better solution, because we could reuse it at any new version.

mueller-ma:
> When doing that (which is a good idea) it's probably a good idea to add a 'version' pref that serves as a 'needs conversion' indicator.

Yes, I think thats a better solution, because we could reuse it at any new version.

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class OnUpdateBroadcastReceiver extends BroadcastReceiver {
+    private static final String TAG = OpenHABMainActivity.class.getSimpleName();
+
+    private static final String UPDATE_LOCAL_CREDENTIALS = "1.8.0.8";
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        if (getVersionNumber(context.getPackageManager(), context.getPackageName())
+                .equals(UPDATE_LOCAL_CREDENTIALS)) {
```
maniac103: This won't work if a user skips a version when updating. Preference versioning (storing the old version along with the data) likely will work better here.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -6,61 +6,81 @@
...
-            android:key="default_openhab_sslcert"
-            android:summary="@string/settings_openhab_sslcert_summary"
-            android:title="@string/settings_openhab_sslcert" />
-        <CheckBoxPreference
-            android:defaultValue="false"
-            android:key="default_openhab_sslhost"
-            android:summary="@string/settings_openhab_sslhost_summary"
-            android:title="@string/settings_openhab_sslhost" />
+            android:title="@string/settings_openhab_demomode"/>
+        <PreferenceScreen android:title="@string/settings_openhab_connection">
```
maniac103: Why not use a PreferenceCategory here? IIRC Google recommended to use them for smaller groups of 4-6 prefs, even though they seem to have dropped this from their latest guidelines.

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
FlorianSW: Hmm, you're right. Just to confirm that I understand correctly what you mean:
- There should be a preference, which saves the last used version (which needs to be updated with every onCreate of OpenHABMainActivity?)
- Here we check, if the last version is != the current version, and if so, update the username/password?

That would also require, that we remove this code snippet after we released the version, right?

mobile/src/main/res/xml/preferences.xml
FlorianSW: Personally, I think it's clearer to the user, why it's possible to add two sets of username and password, if they appear in a separate screen. However, if someone has good reasons why that should be a preference category, I'm fine with that, too :)

mobile/src/main/res/xml/preferences.xml
```diff
@@ -6,61 +6,81 @@
...
+                android:clickable="true"
+                android:defaultValue="@string/settings_openhab_none"
+                android:dependency="default_openhab_demomode"
+                android:key="default_openhab_sslclientcert"
+                android:summary=""
+                android:title="@string/settings_openhab_sslclientcert"/>
+            <Preference
+                android:clickable="true"
+                android:dependency="default_openhab_demomode"
+                android:key="default_openhab_sslclientcert_howto"
```
maniac103: You can't visit it when it's *not* enabled. This indeed works as intended, just tried it, even though I can't tell off-hand *why* it works like that.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -6,61 +6,81 @@
...
+                android:key="default_openhab_sslhost"
+                android:summary="@string/settings_openhab_sslhost_summary"
+                android:title="@string/settings_openhab_sslhost"/>
+        </PreferenceScreen>
     </PreferenceCategory>
     <PreferenceCategory android:title="@string/settings_display_title">
         <ListPreference
-            android:key="default_openhab_theme"
-            android:title="@string/settings_openhab_theme"
             android:defaultValue="@string/theme_value_light"
```
maniac103: This change seems off, doesn't it?

mobile/src/main/res/xml/preferences.xml
```diff
@@ -76,7 +96,7 @@
             android:defaultValue="false"
             android:key="default_openhab_screentimeroff"
             android:summary="@string/settings_openhab_screentimeroff_summary"
-            android:title="@string/settings_openhab_screentimeroff" />
+            android:title="@string/settings_openhab_screentimeroff"/>
```
maniac103: Useless whitespace change here and below.

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
maniac103: We'd assign a number as pref version, which is counted up with each pref change that needs conversion. Conversion code reads the number, compares it to the known latest version number and does conversion based on the difference. We wouldn't need to remove any code snippet, as the conversion just wouldn't do anything if the version matches.
And no need to do the conversion in the activity, doing it in this receiver is fine (props for discovering that intent action!)

mobile/src/main/res/xml/preferences.xml
maniac103: I find a mixture of categories and screens a bit weird. Not sure whether this + the Google guidelines count as good reason in your book ;-)

mobile/src/main/res/xml/preferences.xml
```diff
@@ -6,61 +6,81 @@
...
+                android:title="@string/settings_openhab_username"/>
+            <EditTextPreference
+                android:defaultValue="@string/empty_string"
+                android:dependency="default_openhab_demomode"
+                android:inputType="textPassword"
+                android:key="default_openhab_remote_password"
+                android:summary="@string/settings_openhab_password_summary"
+                android:title="@string/settings_openhab_password"/>
+        </PreferenceScreen>
+        <PreferenceScreen android:title="@string/settings_openhab_sslsettings">
```
maniac103: Why convert SSL settings to a pref screen? No potential confusion here AFAICT?

mobile/src/main/res/values-de/strings.xml
```diff
@@ -108,11 +108,6 @@
     <string name="mainmenu_openhab_voice_recognition">Sprachbefehl</string>
     <string name="theme_name_basic_ui">Basic UI</string>
     <string name="theme_name_basic_ui_dark">Basic UI dunkel</string>
-    <string name="minus">-</string>
```
mueller-ma: I will miss them

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
```diff
@@ -0,0 +1,49 @@
...
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+import android.preference.PreferenceManager;
+import android.util.Log;
+
+import org.openhab.habdroid.ui.OpenHABMainActivity;
+import org.openhab.habdroid.util.Constants;
+
+public class OnUpdateBroadcastReceiver extends BroadcastReceiver {
+    private static final String TAG = OpenHABMainActivity.class.getSimpleName();
```
mueller-ma: Wrong TAG

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private static final String UPDATE_LOCAL_CREDENTIALS = "1.8.0.8";
```
mueller-ma: Shouldn't be this the next version number? Will be 2.0 https://github.com/openhab/openhab.android/pull/341#issuecomment-322299485

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
FlorianSW: Ok, then probably I haven't understood the idea completely :) If we leave the code here over more versions, how we should decide, if a migration is really needed? Sorry, if the solution is pretty simple (which I assume), I really just don't get it :]

mobile/src/main/res/xml/preferences.xml
FlorianSW: oops.

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
FlorianSW: Fixed!

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
FlorianSW: Fixed!

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
FlorianSW: But what if the version of the app wasn't updated so far (but just installed the first time) and then updated? It will need some default value (let's say 0), which would trigger all conversions, which would be incorrect, wouldn't it?

mobile/src/main/res/xml/preferences.xml
FlorianSW: I don't think so. The SSL settings are common for both, the local and the remote settings, so they at least do not belong to one of them. We probably can create a separate PreferenceCategory for it, would that be ok?

maniac103: Yeah, using a PreferenceCategory sounds fine.

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
```diff
@@ -0,0 +1,48 @@
...
+import android.content.pm.PackageManager;
+import android.preference.PreferenceManager;
+import android.util.Log;
+
+import org.openhab.habdroid.util.Constants;
+
+public class OnUpdateBroadcastReceiver extends BroadcastReceiver {
+    private static final String TAG = OnUpdateBroadcastReceiver.class.getSimpleName();
+
+    private static final String UPDATE_LOCAL_CREDENTIALS = "2.0.0.0";
```
mueller-ma: Maybe its easier to check for version code?

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
```diff
@@ -0,0 +1,48 @@
...
+import android.content.Intent;
+import android.content.SharedPreferences;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+import android.preference.PreferenceManager;
+import android.util.Log;
+
+import org.openhab.habdroid.util.Constants;
+
+public class OnUpdateBroadcastReceiver extends BroadcastReceiver {
```
mueller-ma: Does openhab recognize this broadcast even when its not running?

FlorianSW: It should, the OS should create the Intent and delivers it to the app, which was "replaced" only, which then can respond to it as described in the manifest. In this case, the OnUpdateBroadcastReceiver will be called. That happens also, if the app is currently not running.

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
maniac103: Why? When installing an app for the first time, no conversion would be necessary as there's no existing data. When an installed app is updated, we'll get the broadcast.
And for deciding whether migration is needed, we'd read the current version number from prefs and decide based on the delta of the version numbers compared to the latest one supported by the new app version.
A challenge is indeed writing the version initially; we could probably make the settings activity write the latest app supported version unconditionally. On updates, we'll get the broadcast before the settings activity will run again.

digitaldan:
@FlorianSW so has this been changed to not break user's setting when upgrading from the current app?

digitaldan:
@FlorianSW ?

FlorianSW:
@digitaldan Not yet so far (sorry for the late feedback!). I'll now start trying to implement, what was suggested by @mueller-ma and @maniac103 :)

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
FlorianSW: I probably "found" a way without needing to save the version in a shared pref at all, please take a look at my next commit, I would appreciate your feedback :]

mueller-ma:
I think its better to get versionCode from the manifest and use this instead of versionName https://github.com/openhab/openhab.android/blob/master/mobile/src/main/AndroidManifest.xml#L4

maniac103:
And in addition, better use BuildConfig.VERSION_CODE instead of querying PackageManager.

digitaldan:
>@digitaldan Not yet so far (sorry for the late feedback!). I'll now start trying to implement, what was suggested by @mueller-ma and @maniac103 :)

Sounds good, let me know when you are finished.

FlorianSW:
Finished :P @mueller-ma and @maniac103 Can you take a look on the changes if there's something I missed or can do better? :)

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
```diff
@@ -0,0 +1,40 @@
...
+import static org.openhab.habdroid.util.Constants.PREFERENCE_COMPAREABLEVERSION;
+
+public class OnUpdateBroadcastReceiver extends BroadcastReceiver {
+    private static final String TAG = OnUpdateBroadcastReceiver.class.getSimpleName();
+
+    private static final int UPDATE_LOCAL_CREDENTIALS = 25;
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        if (!Intent.ACTION_MY_PACKAGE_REPLACED.equalsIgnoreCase(intent.getAction())) {
```
maniac103: equals() should be fine here, intent actions are case sensitive.

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
```diff
@@ -0,0 +1,40 @@
...
+            SharedPreferences.Editor edit = prefs.edit();
+            if (prefs.getString(Constants.PREFERENCE_LOCAL_USERNAME, null) == null) {
+                edit.putString(Constants.PREFERENCE_LOCAL_USERNAME, prefs.getString(Constants
+                        .PREFERENCE_USERNAME, null));
+            }
+            if (prefs.getString(Constants.PREFERENCE_LOCAL_PASSWORD, null) == null) {
+                edit.putString(Constants.PREFERENCE_LOCAL_PASSWORD, prefs.getString(Constants
+                        .PREFERENCE_PASSWORD, null));
+            }
+            edit.apply();
```
maniac103: Wouldn't you want to update PREFERENCE_COMPARABLEVERSION to the new version here? I realize the case of the user not starting the activity before the next update might be a bit academic, but fixing this doesn't hurt either I think.

mobile/src/main/res/xml/preferences.xml
maniac103: One compelling reason for not using PreferenceScreens, at least in the current version of the patch, is that the activity spawned for the new screen lacks an action/toolbar with back button and title. IMHO, this would need to be fixed.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -6,56 +6,74 @@
...
-        <CheckBoxPreference
-            android:defaultValue="false"
-            android:key="default_openhab_sslhost"
-            android:summary="@string/settings_openhab_sslhost_summary"
-            android:title="@string/settings_openhab_sslhost" />
+            android:title="@string/settings_openhab_demomode"/>
+        <PreferenceScreen android:title="@string/settings_openhab_connection">
+            <org.openhab.habdroid.ui.widget.URLInputPreference
+                android:defaultValue="@string/empty_string"
+                android:dependency="default_openhab_demomode"
```
maniac103: Those dependencies would need to be moved up to the PreferenceScreen.

mobile/src/main/res/xml/preferences.xml
maniac103: I think the point is that SSL settings make no sense in demo mode. But my comment above also applies here: the dependencies need to be moved out to the PreferenceScreen, if those are going to be kept, as they won't work otherwise.

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
FlorianSW: +1

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
FlorianSW: You're right, haven't thought about that :) I think it should always put the new version into the setting, even if no migration of data is needed :)

FlorianSW:
@mueller-ma Did you had already some time to take a look on this PR? :)

maniac103:
I've played around with attempts to make the sub preference screens have a toolbar, and none of the possible attempts really looks straightforward and/or elegant to me. The 'best' approach I could find is this one: https://stackoverflow.com/questions/27862299/toolbar-is-hidden-in-nested-preferencescreen ... and it still feels pretty kludgy, at least in our codebase.
I'd suggest one of the following approaches:
- Keep PreferenceScreen, use approach above, but split the current SettingsFragment into SettingsFragment + ConnectionFragment + SSLSettingsFragment. Make ConnectionFragment generic enough to work for both local + remote settings (e.g. pass URL summary string resource + keys into fragment arguments)
- Use PreferenceCategory for local, remote, SSL settings. Don't disable their prefs if demo mode is disabled, but instead pop up a dialog asking for demo mode to be disabled when clicking on them and demo mode is on. Move demo mode to misc settings, as it (IMHO) is currently presented in a more important place than justified by its actual usage frequency.

Thoughts?

maniac103:
> Use PreferenceCategory for local, remote, SSL settings. Don't disable their prefs if demo mode is disabled, but instead pop up a dialog asking for demo mode to be disabled when clicking on them and demo mode is on. Move demo mode to misc settings, as it (IMHO) is currently presented in a more important place than justified by its actual usage frequency.

Pushed a proof of concept for this, in case anyone wants to play around with it: https://github.com/maniac103/openhab.android/tree/settings-reorder
I'm not sure how I feel about it myself yet, and I guess I would be totally fine with implementing option 1 instead as well, but only with that work towards having toolbars in each sub-screen done.

mueller-ma:
@floriansw Soon (tm)

FlorianSW:
I tried the first approach already, however, the AppCompatActivity seems to require us to use the SupportFragmentManager (getSupportFragmentManager()) instead of the default FragmentManager (getFragmentManager). That's why replacing a fragment does not work for me in that way (for navigating through fragments). Does anyone of you has an idea how to do that instead? (Without using the support fragment manager, which would require us to use the v4 Fragments, too).

maniac103:
Please define 'does not work' ;-) Given adding our SettingsFragment works fine, I don't see why replacing it by some other fragment wouldn't work...

FlorianSW:
Hehe :P With "does not work" I mean, that the following code to replace is executed successfully, no error message or exception, however, the fragment is not being replaced. It simply happens nothing.
`getFragmentManager().beginTransaction().replace(R.id.prefs_container, new ConnectionSettingsFragment(), TAG_NESTED).addToBackStack(TAG_NESTED).commit();`

(Let's assume, that any required type is existing.

maniac103:
Seems to work fine for me. Can you push your code somewhere, so I can see whether I can repro with your code?

FlorianSW:
I've done it again from scratch based on the example you linked above (the StackOverflow post) and pushed it here: https://github.com/FlorianSW/openhab.android/commit/5c02f00a6f14e7725172ddc7729dc7a44d9503bd The same behavior, it simply does not replace the fragment, maybe I just overlook something :(

maniac103:
> The same behavior, it simply does not replace the fragment, maybe I just overlook something :(

Yes, you do :-P The fragment *is* replaced, it's just that your replacement fragment looks exactly identical because it inflates the very same prefs. Try inflating something else (not deriving from SettingsFragment would be a start) or changing a pref and you'll see :-)

mobile/src/main/java/org/openhab/habdroid/core/OnUpdateBroadcastReceiver.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.openhab.habdroid.BuildConfig;
+import org.openhab.habdroid.util.Constants;
+
+import static org.openhab.habdroid.util.Constants.PREFERENCE_COMPAREABLEVERSION;
+
+public class OnUpdateBroadcastReceiver extends BroadcastReceiver {
+    private static final String TAG = OnUpdateBroadcastReceiver.class.getSimpleName();
+
+    private static final int UPDATE_LOCAL_CREDENTIALS = 25;
```
mueller-ma: Current master is 26

FlorianSW:
DAMN IT! :D I was sure, I had changed that in one version when I tried various ways, and it seems I was that sure, that I haven't checked that again :P So, yes, it is working in that way. I'll will prepare a commit with the changes and will update the PR (it may take a bit more time, as I will start today in the evening or tomorrow at some time, and not yet :/). Thanks for your friendly pointer :P

FlorianSW:
The migration part does not work if you install the app from Android studio as the intent is not fired in this case :)

mueller-ma:
I build apks instead, but still same behavior.

Also the passwords are used wrong:
Local settings => User: user, password: mypassword, url: my local url
Remote settings => User: dummyUsername, password: dummyPW, url: https://myopenhab.org

Local Nginx log:
User "dummyUsername" was not found in ...

mobile/src/main/java/org/openhab/habdroid/util/Constants.java
```diff
@@ -10,23 +10,26 @@
...
-    public static final String PREFERENCE_URL               = "default_openhab_url";
-    public static final String PREFERENCE_THEME             = "default_openhab_theme";
-    public static final String PREFERENCE_ANIMATION         = "default_openhab_animation";
-    public static final String PREFERENCE_DEMOMODE          = "default_openhab_demomode";
-    public static final String PREFERENCE_FULLSCREEN        = "default_openhab_fullscreen";
-    public static final String PREFERENCE_TONE              = "default_openhab_alertringtone";
-    public static final String PREFERENCE_SSLCLIENTCERT     = "default_openhab_sslclientcert";
+    public static final String PREFERENCE_SCREENTIMEROFF      = "default_openhab_screentimeroff";
+    public static final String PREFERENCE_USERNAME            = "default_openhab_remote_username";
+    public static final String PREFERENCE_LOCAL_USERNAME      = "default_openhab_username";
```
mueller-ma: This seems to be confusing. I thought: "The user setting without local or remote in its name is remote", but "default_openhab_username" is local.

maniac103: Maybe the constant name should be changed to include _LOCAL_

FlorianSW: I change the constant name.

FlorianSW:
> I build apks instead, but still same behavior.
My fault, I fixed that! Thanks for pointing to it!

> Also the passwords are used wrong:
I assume, an image was requested in that case? The username/password was not set correctly to the http client, fixed that, too.

mueller-ma:
I think the settings for ssl certficates should be seperated, too. When someone uses openhab cloud, he should never disable cert warnings, but for local connections this might be ok. Please tell me if that's out of scope for this PR, than I open an issue for that.

FlorianSW:
@mueller-ma I agree, and I already thought about that case, too. However, this would, in my opinion, require a bit more things to be changed (probably, instead of having to use multiple different settings to handle, a separate class could handle all of this when needed). But the PR is already quite big, so I would like to do that in a separate PR, the issue mentioned in this one is already fulfilled (and widened) :)

mobile/src/main/res/values/strings.xml
```diff
@@ -48,6 +48,9 @@
     <string name="settings_openhab_icon_format_png">Bitmap</string>
     <string name="settings_openhab_icon_format_svg">Vector</string>
     <string name="settings_ringtone">Ring tone</string>
+    <string name="settings_openhab_sslsettings">Encryption settings</string>
+    <string name="settings_openhab_alt_connection">openHAB remote settings</string>
+    <string name="settings_openhab_connection">openHAB local settings</string>
```
maniac103: Maybe better as 'Remote server settings' / 'Local server settings'?

maniac103:
@FlorianSW Did you intentionally not apply my patch or just forgot about it?

FlorianSW:
> @FlorianSW Did you intentionally not apply my patch or just forgot about it?

Oh, I haven't seen that comment, sorry! :)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -77,23 +76,151 @@ public void finish() {
...
+        public void onStart() {
+            super.onStart();
+
+            getParentActivity().getSupportActionBar().setTitle(getTitleResId());
+        }
+
+        protected abstract void updateAndInitPreferences();
+
+        protected abstract @StringRes
+        int getTitleResId();
```
maniac103: This line break wasn't present in my patch ;-)

mueller-ma:
Also the local connection still fails. My setup:
1. Android 5 tablet
2. foss build (dont have up to date play services on that device)
3. nginx reverse proxy on port 80 over http (the openhab installation behind is broken, but that shouldnt brake the authentication)
4. no openhab cloud, since the oh is broken
5. Local settings: my local url, valid user and valid password.
6. Remote settings: dummy remote url, dummy user and password (user:dudxjd password: djdje)
7. Extended main activity L580 to output username and pw
8. Here is the log

````
/? I/art: Late-enabling -Xcheck:jni
/org.openhab.habdroid I/CrashlyticsCore: Initializing Crashlytics 2.4.1.19
/org.openhab.habdroid I/FA: App measurement is starting up, version: 11400
/org.openhab.habdroid I/FA: To enable debug logging run: adb shell setprop log.tag.FA VERBOSE
/org.openhab.habdroid V/FA: Collection enabled
/org.openhab.habdroid V/FA: App package, google app id: org.openhab.habdroid, 1:<some_id>:android:<some_id>
/org.openhab.habdroid I/FA: To enable faster debug mode event logging run:
                              adb shell setprop debug.firebase.analytics.app org.openhab.habdroid
/org.openhab.habdroid D/FA: Debug-level message logging enabled
/org.openhab.habdroid V/FA: Registered activity lifecycle callback
/org.openhab.habdroid I/CrashlyticsInitProvider: CrashlyticsInitProvider initialization successful
/org.openhab.habdroid D/FirebaseApp: com.google.firebase.auth.FirebaseAuth is not linked. Skipping initialization.
/org.openhab.habdroid D/FirebaseApp: com.google.firebase.crash.FirebaseCrash is not linked. Skipping initialization.
/org.openhab.habdroid I/FirebaseInitProvider: FirebaseApp initialization successful
/org.openhab.habdroid I/InstantRun: starting instant run server: is main process
/org.openhab.habdroid V/FA: Using measurement service
/org.openhab.habdroid D/OnUpdateBroadcastReceiver: Checking for putting local username/password to remote username/password.
/org.openhab.habdroid V/FA: Connecting to remote service
/org.openhab.habdroid W/GooglePlayServicesUtil: Google Play services out of date.  Requires 11400000 but found 9683236
/org.openhab.habdroid V/FA: Using measurement service
/org.openhab.habdroid V/FA: Connection attempt already in progress
/org.openhab.habdroid W/art: Suspending all threads took: 39.803ms
/org.openhab.habdroid I/art: Background sticky concurrent mark sweep GC freed 6831(791KB) AllocSpace objects, 9(144KB) LOS objects, 18% free, 3MB/4MB, paused 42.111ms total 91.354ms
/org.openhab.habdroid D/OpenHABMainActivity: onCreate()
/org.openhab.habdroid W/art: Before Android 4.1, method android.graphics.PorterDuffColorFilter android.support.graphics.drawable.VectorDrawableCompat.updateTintFilter(android.graphics.PorterDuffColorFilter, android.content.res.ColorStateList, android.graphics.PorterDuff$Mode) would have incorrectly overridden the package-private method in android.graphics.drawable.Drawable
/org.openhab.habdroid V/FA: onActivityCreated
/org.openhab.habdroid I/AppCompatViewInflater: app:theme is now deprecated. Please move to using android:theme instead.
/org.openhab.habdroid D/OpenHABMainActivity: Intent != null
/org.openhab.habdroid D/OpenHABMainActivity: Intent action = android.intent.action.MAIN
/org.openhab.habdroid D/OpenHABMainActivity: onResume()
/org.openhab.habdroid V/FA: Using measurement service
/org.openhab.habdroid V/FA: Connection attempt already in progress
/org.openhab.habdroid V/NFC: this device does not have NFC support
/org.openhab.habdroid V/FA: Activity resumed, time: 807199
/org.openhab.habdroid I/FA: Tag Manager is not found and thus will not be used
/org.openhab.habdroid D/FA: Logging event (FE): screen_view(_vs), Bundle[{firebase_event_origin(_o)=auto, firebase_screen_class(_sc)=OpenHABMainActivity, firebase_screen_id(_si)=<some_id>}]
/org.openhab.habdroid V/FA: Using measurement service
/org.openhab.habdroid V/FA: Connection attempt already in progress
/org.openhab.habdroid D/OpenHABTracker: Checking reachability of http://openhab2.local/
/org.openhab.habdroid D/OpenHABTracker: Socket connected
/org.openhab.habdroid D/OpenHABTracker: Connecting to local URL = http://openhab2.local/
/org.openhab.habdroid D/OpenHABMainActivity: Remote URL set, show message: Verbinde mit lokaler URL
/org.openhab.habdroid D/OpenGLRenderer: Use EGL_SWAP_BEHAVIOR_PRESERVED: true
/org.openhab.habdroid D/Atlas: Validating map...
/org.openhab.habdroid W/FA: Service connection failed: ConnectionResult{statusCode=SERVICE_VERSION_UPDATE_REQUIRED, resolution=null, message=null}
/org.openhab.habdroid V/FA: Processing queued up service tasks: 4
/org.openhab.habdroid E/FA: Discarding data. Failed to send app launch
/org.openhab.habdroid E/FA: Failed to get app instance id
/org.openhab.habdroid E/FA: Failed to send current screen to service
/org.openhab.habdroid E/FA: Discarding data. Failed to send event to service
/org.openhab.habdroid I/OpenGLRenderer: Initialized EGL, version 1.4
/org.openhab.habdroid D/OpenGLRenderer: Enabling debug mode 0
/org.openhab.habdroid I/MaliEGL: [Mali]window_type=1, is_framebuffer=0, errnum = 0
/org.openhab.habdroid I/MaliEGL: [Mali]surface->num_buffers=4, surface->num_frames=3, win_min_undequeued=1
/org.openhab.habdroid I/MaliEGL: [Mali]max_allowed_dequeued_buffers=3
/org.openhab.habdroid D/OpenHABDrawerAdapter: Not showing count 0
/org.openhab.habdroid D/OpenHABDrawerAdapter: Not showing count 0
/org.openhab.habdroid D/OpenHABDrawerAdapter: Not showing count 0
/org.openhab.habdroid D/OpenHABMainActivity: openHAB version 1 - got error java.io.IOException: Unauthorized accessing http://openhab2.local/rest/bindings user:dudxjd password: djdje
/org.openhab.habdroid D/OpenHABMainActivity: Loading sitemap list from http://openhab2.local/rest/sitemaps
/org.openhab.habdroid W/art: Before Android 4.1, method int android.support.v7.widget.ListViewCompat.lookForSelectablePosition(int, boolean) would have incorrectly overridden the package-private method in android.widget.ListView
/org.openhab.habdroid V/NFC: this device does not have NFC support
/org.openhab.habdroid D/OpenHABMainActivity: Hide nfc menu entry, because nfc is not available
/org.openhab.habdroid E/OpenHABMainActivity: Error: java.io.IOException: Unauthorized
/org.openhab.habdroid E/OpenHABMainActivity: HTTP status code: 401
/org.openhab.habdroid I/MaliEGL: [Mali]window_type=1, is_framebuffer=0, errnum = 0
/org.openhab.habdroid I/MaliEGL: [Mali]surface->num_buffers=4, surface->num_frames=3, win_min_undequeued=1
/org.openhab.habdroid I/MaliEGL: [Mali]max_allowed_dequeued_buffers=3
/org.openhab.habdroid W/InstanceID/Rpc: No response
/org.openhab.habdroid D/FirebaseInstanceId: background sync failed: TIMEOUT, retry in 10s
/org.openhab.habdroid W/InstanceID/Rpc: No response
/org.openhab.habdroid D/FirebaseInstanceId: background sync failed: TIMEOUT, retry in 20s
/org.openhab.habdroid W/InstanceID/Rpc: No response
/org.openhab.habdroid D/FirebaseInstanceId: background sync failed: TIMEOUT, retry in 40s
/org.openhab.habdroid V/FA: Recording user engagement, ms: 121409
/org.openhab.habdroid V/FA: Using measurement service
/org.openhab.habdroid V/FA: Connecting to remote service
/org.openhab.habdroid W/GooglePlayServicesUtil: Google Play services out of date.  Requires 11400000 but found 9683236
/org.openhab.habdroid V/FA: Activity paused, time: 928607
/org.openhab.habdroid W/FA: Service connection failed: ConnectionResult{statusCode=SERVICE_VERSION_UPDATE_REQUIRED, resolution=null, message=null}
/org.openhab.habdroid D/FA: Logging event (FE): user_engagement(_e), Bundle[{firebase_event_origin(_o)=auto, engagement_time_msec(_et)=121409, firebase_screen_class(_sc)=OpenHABMainActivity, firebase_screen_id(_si)=<some_id>}]
/org.openhab.habdroid D/OpenHABMainActivity: onSaveInstanceState
/org.openhab.habdroid D/OpenHABMainActivity: Saving 0 fragments
/org.openhab.habdroid D/OpenHABMainActivity: Saving current page = 0
/org.openhab.habdroid D/OpenHABMainActivity: onStop()
/org.openhab.habdroid V/FA: Using measurement service
/org.openhab.habdroid D/OpenHABTracker: Receiver not registered: org.openhab.habdroid.core.ConnectivityChangeReceiver@32d2aab8
/org.openhab.habdroid V/FA: Connecting to remote service
/org.openhab.habdroid W/GooglePlayServicesUtil: Google Play services out of date.  Requires 11400000 but found 9683236
/org.openhab.habdroid V/FA: Processing queued up service tasks: 2
/org.openhab.habdroid E/FA: Failed to send current screen to service
/org.openhab.habdroid E/FA: Discarding data. Failed to send event to service
/org.openhab.habdroid W/FA: Service connection failed: ConnectionResult{statusCode=SERVICE_VERSION_UPDATE_REQUIRED, resolution=null, message=null}
/org.openhab.habdroid V/FA: Processing queued up service tasks: 0
/org.openhab.habdroid D/OpenHABMainActivity: onResume()
/org.openhab.habdroid V/FA: Using measurement service
/org.openhab.habdroid V/FA: Connecting to remote service
/org.openhab.habdroid W/GooglePlayServicesUtil: Google Play services out of date.  Requires 11400000 but found 9683236
/org.openhab.habdroid V/FA: Activity resumed, time: 938190
/org.openhab.habdroid D/OpenHABTracker: Checking reachability of http://openhab2.local/
/org.openhab.habdroid D/OpenHABTracker: Socket connected
/org.openhab.habdroid D/OpenHABTracker: Connecting to local URL = http://openhab2.local/
/org.openhab.habdroid D/OpenHABMainActivity: Remote URL set, show message: Verbinde mit lokaler URL
/org.openhab.habdroid I/MaliEGL: [Mali]window_type=1, is_framebuffer=0, errnum = 0
/org.openhab.habdroid I/MaliEGL: [Mali]surface->num_buffers=4, surface->num_frames=3, win_min_undequeued=1
/org.openhab.habdroid I/MaliEGL: [Mali]max_allowed_dequeued_buffers=3
/org.openhab.habdroid I/MaliEGL: [Mali]window_type=1, is_framebuffer=0, errnum = 0
/org.openhab.habdroid I/MaliEGL: [Mali]surface->num_buffers=4, surface->num_frames=3, win_min_undequeued=1
/org.openhab.habdroid I/MaliEGL: [Mali]max_allowed_dequeued_buffers=3
/org.openhab.habdroid W/FA: Service connection failed: ConnectionResult{statusCode=SERVICE_VERSION_UPDATE_REQUIRED, resolution=null, message=null}
/org.openhab.habdroid V/FA: Processing queued up service tasks: 1
/org.openhab.habdroid E/FA: Failed to send current screen to service
/org.openhab.habdroid D/OpenHABMainActivity: openHAB version 1 - got error java.io.IOException: Unauthorized accessing http://openhab2.local/rest/bindings user:dudxjd password: djdje
/org.openhab.habdroid D/OpenHABMainActivity: Loading sitemap list from http://openhab2.local/rest/sitemaps
/org.openhab.habdroid E/OpenHABMainActivity: Error: java.io.IOException: Unauthorized
/org.openhab.habdroid E/OpenHABMainActivity: HTTP status code: 401
/org.openhab.habdroid I/MaliEGL: [Mali]window_type=1, is_framebuffer=0, errnum = 0
/org.openhab.habdroid I/MaliEGL: [Mali]surface->num_buffers=4, surface->num_frames=3, win_min_undequeued=1
/org.openhab.habdroid I/MaliEGL: [Mali]max_allowed_dequeued_buffers=3
````

**Comment by mueller-ma** [comment]
Fix for the latter issue: Change mainactivity L559 to 
```
if (baseUrl.equals(Util.normalizeUrl(mSettings.getString(Constants.PREFERENCE_URL, null)))) {
```

mueller-ma:
Got the oh working again. Local connection seems to work now (also charts), but from time to time the app still tries to connect with the remote user

digitaldan:
@FlorianSW  whats the current status here, should I test this or do you have changes based on @mueller-ma last comment?

FlorianSW:
Hmm, it works for me, however, I "just" used the default sidemap of a default openhab installation for the local part (and a productive one running openhab-cloud somewhere outside my network for the remote settings).

@mueller-ma What do you mean with "from time to time the app still tries to connect with the remote user". Does the app use the remote user when connecting to the local url or does it connect to the remote url completely? Do you know, what requests will be made with the remote user? Is it possible, that this are requests for the notifications? They are always made with the remote user (to the remote url, though)?

mueller-ma:
When I access the local server, it works. But every few seconds in the logfile of the local proxy I see the message "invalid user <remote user here>". Notification connections shouldnt show up on the local server.

FlorianSW:
@mueller-ma Doesn't you see what URL was requested? That would help to identify, where and why the false username/password is used :/

mueller-ma:
````
2017/12/09 11:08:01 [error] 13471#13471: *461 user "remote_user" was not found in "/etc/nginx/.htpasswd", client: <ip>, server: openhab2.local, request: "GET /icon/group HTTP/1.1", host: "openhab2.local"
````

FlorianSW:
Ah, ok, I think I found the place where the username/password is passed directly, too :/ Fixing now.

I'll probably create a separate PR to change the current behaviour of how changes to the url, username and password are tracked.

digitaldan:
@FlorianSW let me know when you think this is ready, or if you need me to test something specific

FlorianSW:
@digitaldan All things mentioned by @mueller-ma should work now (finally, thanks @mueller-ma for your time and patience :)), so this is ready from my side :)

digitaldan:
sounds good!

FlorianSW:
Great, thanks @digitaldan :)

PR #393, opened 2017-10-04
Selection: Add title and padding

mueller-ma:
Before:
screenshot_2017-10-04-20-39-30
After:
screenshot_2017-10-04-21-18-32

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -602,9 +602,7 @@ public class OpenHABWidgetAdapter extends ArrayAdapter<OpenHABWidget> {
                         OpenHABWidget openHABWidget = (OpenHABWidget) parent.getTag();
                         if (openHABWidget != null) {
                             Log.d(TAG, "Label selected = " + openHABWidget.getMapping(index).getLabel());
-                            Iterator<OpenHABWidgetMapping> mappingIterator = openHABWidget.getMappings().iterator();
-                            while (mappingIterator.hasNext()) {
-                                OpenHABWidgetMapping openHABWidgetMapping = mappingIterator.next();
+                            for (OpenHABWidgetMapping openHABWidgetMapping : openHABWidget.getMappings()) {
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -601,12 +599,13 @@ public void onItemClick(AdapterView<?> parent, View view, int index, long id) {
...
+					} catch (Exception ignored) {}
 				}
 			});
-    		spinnerAdapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
-    		selectionSpinner.setAdapter(spinnerAdapter);
-    		if (spinnerSelectedIndex >= 0) {
+			spinnerAdapter.setDropDownViewResource(R.layout.openhabwidgetlist_sectionswitchitem_spinner);
+			selectionSpinner.setAdapter(spinnerAdapter);
+			selectionSpinner.setPrompt(openHABWidget.getLabel());
+            if (spinnerSelectedIndex >= 0) {
```
FlorianSW: you've mixed whitespaces with tabs :)

mueller-ma:
I noticed that, but I couldnt find a way to use spaces in android studio, so I opened the file in vim and saw that the whole file is mixed.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
mueller-ma: I noticed that, but I couldnt find a way to use spaces in android studio, so I opened the file in vim and saw that the whole file is mixed.

mueller-ma: We should write down rules for text indetion. Four spaces?

digitaldan:
Due to the white space issue, the change set is not really readable :-) Can you point me to where in that file you changed/added?

mueller-ma:
Just look at the first commit

mueller-ma:
I have removed white space changes from this PR.

PR #394, opened 2017-10-06
Dim display when inactive

### LINKED ISSUE #39: Dim Display if there is no touchscreen input
It would be nice if there is an option to enable automatic display dimming if the display doesn't receive touch input for a configurable amount of time. If in dimmed state and the display is touched restore to the previous dimming value. This would be usefull if habdroid is used in conjunction with a wall mounted tablet and the 'disable display timer' option to reduce display wear and energy consumption. 
##

mueller-ma:
Sets the display brightness to 0, when the user is inactive for some
time and restore it on interaction or stop.
I moved the display settings related to the hardware display to a
new section.

Fix #39

That might be not enough, since the lowest brightness is still too bright on some devices. Maybe we could use some other methods to dim the display or change the theme to a dark theme.

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -10,6 +10,7 @@
     <uses-permission android:name="android.permission.READ_LOGS" />
     <uses-permission android:name="android.permission.GET_TASKS" />
     <uses-permission android:name="android.permission.NFC" />
+    <uses-permission android:name="android.permission.WRITE_SETTINGS" />
     <!-- Keeps the processor from sleeping when a message is received. -->
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="com.google.android.c2dm.permission.RECEIVE" />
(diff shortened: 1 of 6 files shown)


digitaldan:
Hi, so I think the intent is to dim the screen as much as possible, like say when this is used as a permanent touch screen?   One of the very first things i ever contributed to the Android client was full screen mode and Daydream support for exactly this.

 Any reason why Daydream does not work here?  Before I was using Habpanel for my wall mounts, I used this, which dims the screen after inactivity to the point where  it's almost off (but that's configurable), but the touchscreen is still active so you can wake it up by touching.  In fact, you can register your app as a Daydream module so you can display something when dimmed.  I just use the default clock or Timely as they look nice.  I remember looking to do what you are doing and decided it was best to use what's built into the OS then try to roll my own thing.

mueller-ma:
The name Daydream is now used for google's vr app, so the old daydream might not work anymore and apart from that I have never heard of it

maniac103:
Of course it still exists, it's just named 'Screensaver' now: https://developer.android.com/reference/android/service/dreams/DreamService.html

mueller-ma:
After clicking on the screensaver I have to unlock the tablet. Can this be changed?

digitaldan:
Yes, none of my tablets lock, I have daydream enabled when its docked or charging (in Daydream settings), then I use the night/dim mode setting for most of my tablets with the default clock.

mueller-ma:
@madon requested that the display should undim, when a item changes. And I think it is easier to implement when we dim the display ourself.

The app Red Moon https://github.com/LibreShift/red-moon can dim the screen way more than the official api, so I asked them if it is possible to toggle the filter from another app.

mueller-ma:
Closing for now

PR #337, opened 2017-07-21
Preference UX improvements

maniac103:
This started off as an improvement for the DE translation, but while working on the translation I found a number of improvement areas in the preference handling code, which I also fixed. I thought about adding them in a separate PR, but unfortunately they touch strings as well, so would conflict with the DE translation update.
If you prefer, I can split this PR to merge the translation updates first, then submit the preference code updates in a separate PR later.

PR #397, opened 2017-10-09
Fix up GCM status printing.

maniac103:
Use the string formatting built into getString(), and don't print the ID
if it's known to be null anyway.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABInfoFragment.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABInfoFragment.java
@@ -188,15 +188,11 @@ public class OpenHABInfoFragment extends DialogFragment {
 
 
     private void setGcmText() {
-        String infoString;
         if (OpenHABMainActivity.GCM_SENDER_ID == null) {
-            infoString = getString(R.string.info_openhab_gcm_not_connected);
+            mOpenHABNotificationText.setText(R.string.info_openhab_gcm_not_connected);
         } else {
-            infoString = getString(R.string.info_openhab_gcm_connected);
+            mOpenHABNotificationText.setText(
+                    getString(R.string.info_openhab_gcm_connected, OpenHABMainActivity.GCM_SENDER_ID));
         }
-
-        mOpenHABNotificationText.setText(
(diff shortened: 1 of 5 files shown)


maniac103:
Unfortunately I had to remove the respective translations for Greek and Russian as my understanding of those languages isn't good enough for adapting the string :-(

@edgiru @karafyllidis Can you please help me out there?

mueller-ma:
What about german?

maniac103:
> What about german?

values-de doesn't have a translation for this string yet; I'll add it in a follow-up PR as mentioned in https://github.com/openhab/openhab.android/pull/337#issuecomment-335075561

digitaldan:
>Unfortunately I had to remove the respective translations for Greek and Russian as my understanding of those languages isn't good enough for adapting the string :-(

Can you ask in the forums for translation help with that?

karafyllidis:
@digitaldan @maniac103 I can help with the Greek translation. What you need in Greek is the following
<string name="info_openhab_gcm_not_connected">Η εγγραφή συσκευής απέτυχε</string>

maniac103:
@karafyllidis Thanks! That leaves us with the Russian string.

karafyllidis:
@maniac103 Anytime. Sorry for the delay

jvitkauskas:
Please see updated Russian and Lithuanian translations here: https://github.com/jvitkauskas/openhab.android/commit/241fc89da95b0bd1a560b144922439e7a397b7b4

jvitkauskas:
Also, shouldn't "theme_value_light", "theme_value_dark" and "theme_value_black" be marked non-translatable?

maniac103:
@jvitkauskas Thanks! Will add your commit to this PR.

> BTW, shouldn't "theme_value_light", "theme_value_dark" and "theme_value_black" be marked non-translatable?

Yes, I'd think so.

maniac103:
The code side of this is merged via #337 now, but the strings are left out. @karafyllidis @jvitkauskas Should I submit a PR with your translation additions or will you do it?

karafyllidis:
@maniac103 I have just submitted PR #413 for the Greek translation

PR #398, opened 2017-10-09
Remove permission checks for permissions needed for discovery.

maniac103:
That check introduced hard to translate, technical strings, and its
outcome wasn't actually used anyway.


mueller-ma:
What will happen, when the permissions arr not granted?

maniac103:
> What will happen, when the permissions arr not granted?

Maybe discovery won't work? Not sure, but the result of the check wasn't actually used before either.

maniac103:
> What will happen, when the permissions arr not granted?

Maybe discovery won't work? Not sure, but the result of the check wasn't actually used before either.

Edit: From my understanding of the code, the only consequence this will have is removing the error message that discovery won't work for people who denied those permissions. I have to stress, though, that those permissions are *not* runtime permissions (and are *not* deemed dangerous either), so the only way to deny them is - AFAICT - using 'adb shell pm revoke', at which point I think we can safely assume we're dealing with expert users.

maniac103:
Merged in #337.

PR #400, opened 2017-10-09
396 chart crash

### LINKED ISSUE #396: Bad Chart URL request crashing app
We are seeing an increase in crashes do to a malformed URL, which looks like it's trying to load up the end of a chart URL.  Just in the last Month we have had more than 1000 crashes, all look identical.  This would be easy to catch and protect against a crash, but I also wonder why we have so many broken chart URL requests, sounds like there's an issue somewhere else?  @mueller-ma , @FlorianSW  do you have any ideas about the root cause? 
```

  | 0 | java.lang.IllegalArgumentException: unexpected url: &w=1080&h=540
-- | -- | --
  | 1 | at okhttp3.Request$Builder.url(Request.java:142)
  | 2 | at org.openhab.habdroid.util.MyAsyncHttpClient.method(MyAsyncHttpClient.java:36)
  | 3 | at org.openhab.habdroid.util.MyAsyncHttpClient.method(MyAsyncHttpClient.java:27)
  | 4 | at org.openhab.habdroid.util.MyHttpClient.get(MyHttpClient.java:131)
  | 5 | at org.openhab.habdroid.util.MyWebImage.getBitmapFromUrl(MyWebImage.java:93)
  | 6 | at org.openhab.habdroid.util.MyWebImage.getBitmap(MyWebImage.java:73)
  | 7 | at com.loopj.android.image.SmartImageTask.run(Unknown Source)
  | 8 | at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
  | 9 | at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
  | 10 | at java.lang.Thread.run(Thread.java:761)

```

mueller-ma:
I have tested it by setting chartItem to null before it is checked if it's null and the app didn't crash. This is just a quick fix and we should have a look why chartitem is null.

Fix #396

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -10,10 +10,8 @@
 package org.openhab.habdroid.ui;
 
 import android.content.Context;
-import android.content.SharedPreferences;
 import android.net.Uri;
 import android.os.Build;
-import android.preference.PreferenceManager;
 import android.support.v7.widget.SwitchCompat;
 import android.util.DisplayMetrics;
 import android.util.Log;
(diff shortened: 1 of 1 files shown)


FlorianSW:
I love to have the same code style on all files, however, could you add a PR with just the fix for #396 and one for the code-style correction? It's pretty hard to find out what was changed and what wasn't :(

mueller-ma:
https://github.com/openhab/openhab.android/pull/400/commits/85d29984ee7e9e50a86430ca3a9560d36500d2b8

I moved this bracket https://github.com/openhab/openhab.android/pull/400/commits/85d29984ee7e9e50a86430ca3a9560d36500d2b8#diff-7084e55d51f6579524d6f8ef39d55555L524 down so the if includes adding the w and h attributes

digitaldan:
this LGTM, thanks @mueller-ma  and thanks @FlorianSW

Issue #399, opened 2017-10-09
App crash in OpenHABFragmentPagerAdapter


digitaldan:
This has happened almost 600 times in last month
```

0 | java.lang.NullPointerException: Attempt to invoke virtual method 'android.os.Handler android.support.v4.app.FragmentHostCallback.getHandler()' on a null object reference
-- | --
  | 1 | at android.support.v4.app.FragmentManagerImpl.ensureExecReady(FragmentManager.java:1959)
  | 2 | at android.support.v4.app.FragmentManagerImpl.execSingleAction(FragmentManager.java:1980)
  | 3 | at android.support.v4.app.BackStackRecord.commitNowAllowingStateLoss(BackStackRecord.java:626)
  | 4 | at android.support.v4.app.FragmentStatePagerAdapter.finishUpdate(FragmentStatePagerAdapter.java:166)
  | 5 | at android.support.v4.view.ViewPager.dataSetChanged(ViewPager.java:1094)
  | 6 | at android.support.v4.view.ViewPager$PagerObserver.onChanged(ViewPager.java:3113)
  | 7 | at android.support.v4.view.PagerAdapter.notifyDataSetChanged(PagerAdapter.java:284)
  | 8 | at org.openhab.habdroid.ui.OpenHABFragmentPagerAdapter.clearFragmentList(OpenHABFragmentPagerAdapter.java:78)
  | 9 | at org.openhab.habdroid.ui.OpenHABMainActivity.openSitemap(OpenHABMainActivity.java:770)
  | 10 | at org.openhab.habdroid.ui.OpenHABMainActivity.access$600(OpenHABMainActivity.java:105)
  | 11 | at org.openhab.habdroid.ui.OpenHABMainActivity$6.onSuccess(OpenHABMainActivity.java:671)
  | 12 | at org.openhab.habdroid.util.MyAsyncHttpClient$1$2.run(MyAsyncHttpClient.java:71)
  | 13 | at android.os.Handler.handleCallback(Handler.java:751)
  | 14 | at android.os.Handler.dispatchMessage(Handler.java:95)
  | 15 | at android.os.Looper.loop(Looper.java:154)
  | 16 | at android.app.ActivityThread.main(ActivityThread.java:6123)
  | 17 | at java.lang.reflect.Method.invoke(Native Method)
  | 18 | at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:867)
[...]

PR #401, opened 2017-10-09
Catch null pointer exception

mueller-ma:
Related #399 
I'm not sure if thats really a fix for the problem, but I cannot reproduce it.


digitaldan:
I think we would want to catch the NPE here https://github.com/openhab/openhab.android/blob/69518bb553d85587ba0e084e1137344b7419800d/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABFragmentPagerAdapter.java#L78 

I have a guess that this is caused by the app going to the background, but trying to process the last REST request and update a view that is no longer there.  That or its updating a view that was destroyed for another view.   I would have to look more into how we are managing view fragments.

maniac103:
I'd agree to this assessment, but to me, catching an NPE seems like an ugly kludge. I would set an mResumed flag in OpenHABMainActivity (set in onResume/reset in onPause) and skip processing the REST sitemap callback completely if the flag is false.

mueller-ma:
I see, there are better solutions.

Issue #392, opened 2017-10-02
Remove "side menu" from the app


kaikreuzer:
In a rush of adding support for openHAB 2, Victor had introduced the side menu in the app as a first start, which adds "Discovery" and "Binding" menus - the implementation has been left very rudimentary though since then and I doubt that they are being used by anyone.

What is more, to not have such an empty menu with just 2 entries, the sitemaps were added there as well as an alternative way to setting them in the configuration. This is actually pretty counterproductive, because it leads people to believe that they should/could use sitemaps as a navigational element, to quickly switch between them. Instead, their intended use is to be set in the configuration and not to be changed by the user.

kubawolanin:
I believe same applies to [openhab.windows](https://github.com/openhab/openhab.windows). @NicoVermeir FYI.

mueller-ma:
I could move discovery and bindings to the action bar or we move the action bar menu to the side menu.

FlorianSW:
I kind of like the navigation drawer, because it's a more intuitive way for navigating _thorugh_ the app (e.g. the settings menu should be there, too, as well as the "About", I think). The toolbar menu, for me, is more something which should contain contextual menu items, which only applies and are only available for the current activity/fragment.

Also, just to mention: If we ever implement, in openhab-cloud, something to use different OH2 instances with one account, the sitemaps selection in the navigation drawer is the central point to switch between instances. However, this is not merged, yet.

mueller-ma:
Could we move "Clear icon cache" and "Write nfc tag" to settings and merge "About" and "openHAB info"?

kvoit:
I guess I was "lead to believe I could use sitemaps as navigational elements", though I also do that in the web UI. 
So what is the indented way?

mueller-ma:
Sitemaps are not intended for navigation. You can create mutliple sitemaps for different people, devices or for testing.

FlorianSW:
Even if it is not intended for navigation, I think, even for the "intended" usage of sitemaps, having a setting to select the sitemap is way too complicated. Having it in the navigation drawer is, by definition, a good practive, as it allows the user to navigate through different destinations of the app. However, selecting the default sitemap can maybe be moved to the settings menu, though, even if, or maybe exactly because, it already appears when a connection is opened to a new local/remote URL.

I would agree with moving the "Write NFC Tag" to the settings menu, however, I never used it so far, so probably I don't know, how much and when it is intended to be used.

[...]

PR #404, opened 2017-10-12
Update settings

FlorianSW:
This updates the settings menu and activity to be more in place with the style of the app.


FlorianSW:
When #389 is merged, this PR should be updated to not use a Toast anymore :)

maniac103:
I have implemented the PreferenceFragment conversion together with a lot of other preference related fixes in #337 already. Unfortunately that one is unmerged yet :-(

mueller-ma:
Can you both have a look at #392 ? Do we want to move everything to the action bar/menu or to the side menu?

maniac103:
Settings would make sense in the drawer, maybe also 'About', but the other stuff IMHO doesn't fit there: it's called *navigation* drawer for a reason, stuff like 'clear image cache' isn't a navigational item.

mueller-ma:
Most of the changes have been done my @maniac103. @FlorianSW, can you update the PR?

mobile/src/main/res/values-de/strings.xml
```diff
@@ -84,6 +84,8 @@
     <string name="action_settings">Einstellungen</string>
     <string name="nfc_dialog_title">Aktion des NFC-Tags</string>
     <string name="info_not_set">Nicht gesetzt</string>
+    <string name="info_settings_moved">Der Menüeintrag Einstellungen wurde in die linke
```
mueller-ma: Intentional two lines?

FlorianSW: Nope, thanks for the note, haven't seen that :)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -925,9 +928,12 @@ public boolean onOptionsItemSelected(MenuItem item) {
         //menu items
         switch (item.getItemId()) {
             case R.id.mainmenu_openhab_preferences:
-                Intent settingsIntent = new Intent(this.getApplicationContext(), OpenHABPreferencesActivity.class);
-                startActivityForResult(settingsIntent, SETTINGS_REQUEST_CODE);
-                Util.overridePendingTransition(this, false);
+                Toast.makeText(
```
mueller-ma: Replace with 
`showMessageToUser(getString(R.string.info_settings_moved), Constants.MESSAGES.TOAST, Constants.MESSAGES.LOGLEVEL.ALWAYS);`
Snackbar won't work here, because it is only visible in MainAcivity.

digitaldan:
@FlorianSW so this moves the settings to the left, which I agree is better, but now we have settings in two places?  I see we message the user to use the left one, but is there a plan to remove this in a latter release?  I would also ask if we really need to do this, or do we think most people would find this on their own?

And as I type, it now occurs to me that maybe the drop down menu needs to go away altogether? it seems more fragmented to have this in two different places?

mueller-ma:
I think the message is a good idea.

> And as I type, it now occurs to me that maybe the drop down menu needs to go away altogether? it seems more fragmented to have this in two different places?

See #392 
I think we should move some points of the drop down to a sub menu of the settings screen and create a new view for about and info.

FlorianSW:
@digitaldan Yes, this would need to be removed after the next release. However, I agree, that this is probably not the best idea. As you also think, that the users can figure it out themself, that the settings can be found in the navigation panel, I would also agree to remove it without keeping the entry in two places.

> And as I type, it now occurs to me that maybe the drop down menu needs to go away altogether? it seems more fragmented to have this in two different places?

That's kind of a difficult decision. If we look at the material design patterns and guidelines from Google, a menu entry "consists of a discrete option or action that can affect the app, the view, or selected elements within a view.", however, it "should not be used as a primary method for navigation within an app." (source: https://material.io/guidelines/components/menus.html#) that's for what the navigation drawer is used for. A navigation drawer should, on the other hand, contain elements to navigate to different destinations of the app (e.g. the settings, sitemap, ...). (source: https://material.io/guidelines/patterns/navigation-drawer.html).

So, in my opinion, the menu items should stay where they are, except the following items (which should, in my personal interpretation of the guidelines, should be moved to the navigation drawer):
* About
* openHAB-Info

All the other items sounds like options for me, which should be kept in the options menu. However, I would say, that this shouldn't be part of this PR, right? :)

digitaldan:
So I am just now following the conversation from #392, I'll wait to hear back from kai

digitaldan:
@FlorianSW so it sounds like we have a consensus  on #392 , we will keep sitemaps, remove bindings and discovery options.  So for this PR we want to move settings over, we also have #433 which moves over the about/info screen.  It also sounds like we need a PR to remove menus.

with #392 this does not leave much in our dropdown menu.  I would argue clearing the image cache could be in settings, and then we only have 2 entries here?  Is this acceptable? (I don't know actually)..  What I don't want to do is release something that is "half" done.  For better or worse, right now everything is one place which is better then spread all around the app in a state that will change again in another release.   Thoughts?

FlorianSW:
In fact, "select default sitemap" could (and should) also be in the settings, as it is not an option provided for the current context. After removing this one as well, keeping the settings only to point to the new location seems odd, so I would propose:
* remove the settings menu from the options menu completely in this PR
* Propose a new PR to remove the options menu completely and move the "select default sitemap" to the settings (or would you like to have it in the same PR?)

digitaldan:
I merged #470 which has causes conflict in this PR and in #433 , @FlorianSW  can you fix?

FlorianSW:
on it!

FlorianSW:
@digitaldan Did we agreed on removing the settings entry from the options menu completely in this PR, too?

digitaldan:
>did we agreed on removing the settings entry from the options menu completely in this PR, too?

I think we want to, but if we move "settings", "clear image cache" and "select default sitemap" (both now in settings),  that would leave just "Write NFC tag, and "about/info" in this menu.  Seems kinda out of place for just those two menus?  Do you have any ideas?

FlorianSW:
About/Info is taken care of in #433, so it's only writing NFC tags. I must say, I don't know exactly how it works, however, it seems to need the current context (when long pressing an item for truning on/off an item and so on, or when creating an NFC tag for opening a sitemap). While the long press thing is not a problem, we can simply keep it how it works currently, I agree, that the sitemap thing could be a bit complicated. It turns out, that this seems to be the only menu entry which should belong in such an option menu (as it is something, which is based on the currently viewed fragment as a context). So, it would be ok to show the options menu (with the NFC tag option only), whenever NFC is available, and not showing it otherwise.

Another possibility would be to move this option into each item long press menu (something like "Open this sitemap page").

digitaldan:
>Another possibility would be to move this option into each item long press menu (something like "Open this sitemap page").

Actually I had forgotten this is how NFC works  on items, if you long press one it gives you the option to trigger it via a NFC tag, butI I have not actually used the NFC feature. So the menu option is for tagging a sitemap page?  I wonder just how useful that is?

FlorianSW:
So, tagging for a specific action taken in the smart home network (like turning on of off something) seems reasonable for me. Opening a sitemap page based on a NFC tag is something where I've difficulties to see any advantage and reason why someone would like to do that :/

However, as we do not have any usage data of particular features in the app (?) I would like to not simply remove the feature from the app completely in one step, but moving the option somewhere else (where it isn't that prominent anymore). If someone complains about the new position, we're sure, that we've at least one user of this feature, otherwise we can still delete it after some versions?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -1332,7 +1332,6 @@ private void loadDrawerItems() {
         aboutDrawable.setColorFilter(
                 iconColor,
                 PorterDuff.Mode.SRC_IN);
-        mDrawerItemList.add(OpenHABDrawerItem.dividerItem());
```
mueller-ma: I opend #488, but I think this solution is better. I'll close my PR.

digitaldan:
I think we are good on this, @FlorianSW if you can fix the conflict I can merge

FlorianSW:
@digitaldan Done! :)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -904,6 +910,12 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
     }
 
+    private void openPreferences() {
+        Intent settingsIntent = new Intent(this.getApplicationContext(), OpenHABPreferencesActivity.class);
```
maniac103: Just using `this` is sufficient here.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -872,9 +875,12 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
             case R.id.mainmenu_openhab_preferences:
-                Intent settingsIntent = new Intent(this.getApplicationContext(), OpenHABPreferencesActivity.class);
-                startActivityForResult(settingsIntent, SETTINGS_REQUEST_CODE);
-                Util.overridePendingTransition(this, false);
+                showMessageToUser(
+                        getString(R.string.info_settings_moved),
+                        Constants.MESSAGES.TOAST,
+                        Constants.MESSAGES.LOGLEVEL.ALWAYS
+                );
+                openPreferences();
```
maniac103: I thought this menu item should be killed?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -27,20 +27,21 @@
...
+import org.openhab.habdroid.BuildConfig;
 import com.loopj.android.image.WebImageCache;
-
 import org.openhab.habdroid.R;
 import org.openhab.habdroid.util.Constants;
 import org.openhab.habdroid.util.MyWebImage;
 import org.openhab.habdroid.util.Util;
 
 import java.security.cert.X509Certificate;
+import java.text.DateFormat;
```
maniac103: The imports introduced here seem unused.

mobile/src/main/res/values-de/strings.xml
```diff
@@ -74,6 +74,7 @@
     <string name="action_settings">Einstellungen</string>
     <string name="nfc_dialog_title">Aktion des NFC-Tags</string>
     <string name="info_not_set">Nicht gesetzt</string>
+    <string name="info_settings_moved">Der Menüeintrag Einstellungen wurde in die linke Navigationsleiste verschoben. Bitte nutze diesen Eintrag zukünftig.</string>
```
maniac103: If we aren't going to remove it and keep the message for now:

`Das Einstellungsmenü hat einen neuen Platz: Es kann jetzt über die Navigationsleiste auf der linken Seite erreicht werden.`

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
FlorianSW: We probably haven't agreed on that, but I think it's ok to directly kill it :)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
FlorianSW: done!

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
FlorianSW: Done!

mobile/src/main/res/values-de/strings.xml
FlorianSW: Removed :)

Issue #30, opened 2014-11-10
HABDroid on f-droid


Zillode:
We should provide HABDroid on the F-Droid store (https://f-droid.org/)

I think this is an Android specific issue, maybe https://github.com/openhab/openhab/issues/394 needs to be moved here?
##

ejabbsrh:
As an F-Droid user, I am interested in any progress that has been made and if there are any difficulties in the process. I know the F-Droid userbase is comparitively small so your interest is much appreciated. If you have any questions there are much more knowledgable people than I on the [F-Droid forums](https://f-droid.org/forums).

belovictor:
From your old issue:
> I've been working on building HABDroid on the F-Droid platform and got it running. Unfortunately I
> observe dependencies to closed source code, such as GoogleCloudMessaging and Crittercism. For
> obvious reasons this is not allowed. Is it possible for HABDroid to provide a branch/Gradle flavour
> where these closed dependencies are stripped?
I'm afraid unfortunately not, cause this will break the user experience - my.openHAB features will not work and we will not be able to track and resolve crashes/bugs from F-Droid users.
From what I understood from https://f-droid.org/wiki/page/Inclusion_How-To there are ways to use Google. openHAB is published under Eclipse Public License. So probably you should ask those guys how to do it?

xylo:
The usual solution in this case is to split the app into two apps: One completely open source app and one that works as a plugin / connector to the proprietary stuff. Of course the second app needs to be optional.

However, I assume that most users would not install the second app, because it breaks the whole open source concept of the first app and opens a major security whole, since nobody knows what actually happens in closed source parts or proprietary services. The past has shown that one has to expect the worst.

If those closed source parts are just used for bug reports and for connecting to my.openHAB it should be possible to move this functionality into an optional app without breaking the user experience for users with privacy concern, because those users will anyway not use my.openHAB, because that would mean that they give a proprietary service access to their private data. The other users that don't care about privacy would just install the second app on demand. That way both groups should be satisfied.

Furthermore, there is another reason for splitting the app.  There are millions of users with Custom ROMs (e.g. over 12 million CyanogenMod users) and a lot of them did not install the Google Apps for obvious reasons. Since those users are missing the Google Play Services they won't be able to use the GoogleCloudMessaging and thus cannot even install the openHAB app even if they get it from another source. That means that they have to install the Google Play store just to be able to use the openHAB app. This is a pretty strong requirement for an "open source" app.

[...]

PR #409, opened 2017-10-15
Add foss build flavor

mueller-ma:
See #30 
I have moved non free components to mobile/src/full/java/org/openhab/habdroid and create dummy classes  and functions in mobile/src/foss/java/org/openhab/habdroid.

Of course this breaks cloud functionality for foss user.


digitaldan:
How would we support a foss build going forward, would we need to build and sign a f-droid version and deploy to that platform for every release?

mueller-ma:
No, fdroid builts are automated. I will some metadata about this app to the fdroid metadata repo and than a new version will be automaticly built, when you create a new tag. I have to look at the docs, maybe a version bump in the android manifest is enough.

digitaldan:
@mueller-ma can you resolve the conflict issues and I can merge?  Also I think we want to remove google analytics from the project as we really don't use it.  We also need to remove the critterism stuff and replace it with the Google supplied crash reporting.  Maybe something you want to try out.

mueller-ma:
I think I will first make a pr to remove that stuff, then try to integrate the google crash analytics and then make a foss flavor

mueller-ma:
On fdroid only specific maven repos are allowed. Since the only depency from the oh repo was smart-image-view I changed it to the github repo.
Now I get following error:
```
Error:Error converting bytecode to dex:
Cause: Dex cannot parse version 52 byte code.
This is caused by library dependencies that have been compiled using Java 8 or above.
If you are using the 'java' gradle plugin in a library submodule add 
targetCompatibility = '1.7'
sourceCompatibility = '1.7'
to that submodule's build.gradle file.
...while parsing com/loopj/android/image/BitmapImage.class
```

mueller-ma:
Works now. Can someone test if openhab cloud still works as expected in the full build?

mueller-ma:
The versions of the foss build are called 1.8.0.8-foss, so both versions can be installed.

digitaldan:
Hi @mueller-ma , in android studio it looks like it is running the foss version by default , do you know how to configure that so it runs the non foss version?

mueller-ma:
On the left sidebar there should be an entry with the green android logo. Click on it and a windows pops up where you can switch between foss and full

digitaldan:
Found it, thanks!

digitaldan:
I don't think device registration is working with myopenhab.org.  I have not looked to much into it, but my test device is not registering with the non foss build.

digitaldan:
wait!  This could be my setup, I had my credentials entered when testing the foss version, but forgot when switching to the non foss one, i needed to re-enter them.

digitaldan:
Nope, still not working.  Can you setup an account on myopenhab.org and test?  You don't actually need to hook up an openhab, there is a "devices" menu that shows registered mobile devices which also has the ability to send a test message to them.

mueller-ma:
Well, I have removed all gcm methods with dummy methods, so I think openhab cloud wont work.

mueller-ma:
Ah, you tested the full built. I will look into that

mueller-ma:
I connected my oh with myopenhab.org and tried with a build from current master branch. The device registered gcm as I can send messages to it.
After that I pruged the app data and tried with the full build from this PR to send messages => also worked.

mueller-ma:
Ok, initial registration is not working.

mueller-ma:
Reopend in #455

PR #414, opened 2017-10-19
Support for image item

### LINKED ISSUE #216: Add support for Image items
A similar request exists for Classic UI and Basic UI: https://github.com/eclipse/smarthome/issues/1800
##

lolodomo:
Fixes #216 

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -10,9 +10,12 @@
 package org.openhab.habdroid.ui;
 
 import android.content.Context;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
 import android.net.Uri;
 import android.os.Build;
 import android.support.v7.widget.SwitchCompat;
+import android.util.Base64;
 import android.util.DisplayMetrics;
 import android.util.Log;
 import android.util.TypedValue;
@@ -38,6 +41,8 @@ import android.widget.SeekBar.OnSeekBarChangeListener;
 import android.widget.Spinner;
 import android.widget.TextView;
 import android.widget.VideoView;
+
+import com.loopj.android.image.SmartImage;
 
 import org.openhab.habdroid.R;
 import org.openhab.habdroid.model.OpenHABItem;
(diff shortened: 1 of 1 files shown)


lolodomo:
Ready for review and test.
It is working but relatively slow to refresh from an Android emulator accessing my openHAB server through myopenhab. Would be interesting to test it on a real tablet and a local access to the server.

lolodomo:
I have now tested with my Nexus 10 tablet and the OH server on the local network and it works perfectly with a really correct refresh delay.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -472,16 +477,22 @@ public void onStopTrackingTouch(SeekBar seekBar) {
...
                 MySmartImageView imageImage = (MySmartImageView) widgetView.findViewById(R.id.imageimage);
-                imageImage.setImageUrl(ensureAbsoluteURL(openHABBaseUrl, openHABWidget.getUrl()), false,
+                OpenHABItem item = openHABWidget.getItem();
+                if (item != null && item.getType().equals("Image") && item.getState() != null
+                        && item.getState().startsWith("data:")) {
+                   imageImage.setImage(new MyImageFromItem(item.getState()));
+                } else {
+                    imageImage.setImageUrl(ensureAbsoluteURL(openHABBaseUrl, openHABWidget.getUrl()), false,
                         openHABUsername, openHABPassword);
+                    if (openHABWidget.getRefresh() > 0) {
```
mueller-ma: No refresh rate for the other type? Imagine following setup:
A selection that controls which webcam I want to see. The webcams make every few seconds a picture.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -906,4 +917,16 @@ public boolean isEnabled(int position) {
...
+    class MyImageFromItem implements SmartImage {
+        private String itemState;
+
+        public MyImageFromItem(String itemState) {
+            this.itemState = itemState;
+        }
+
+        public Bitmap getBitmap(Context context) {
+            byte[] data = Base64.decode(itemState.substring(itemState.indexOf(",") + 1), Base64.DEFAULT);
+            return BitmapFactory.decodeByteArray(data, 0, data.length);
```
mueller-ma: Can this also handle svg?

lolodomo: I don't find on the net what formats are supported by this Android method but I can imagine it supports BMP, GIF, JPEG and PNG. At my knowledge, SVG is vectorial and not bitmap and so probably not supported as a content for an item of type Image. It would be interesting to check how ESH defines a channel of type image.

lolodomo: Image item is defined in ESH as "Holds the binary data of an image". Not very precise on what formats have to be supported.
https://www.eclipse.org/smarthome/documentation/concepts/items.html

During my tests, I certainly succeeded in displaying JPEG and PNG formats in openHAB app

lolodomo: I could enhance the current code to try to decode a SVG in case BitmapFactory.decodeByteArray returns null. We already have a method that build a bitmap from a SVG stream, I could reuse it.

Do you know any binding that uses SVG format with channels of type image ???
Without any binding of this kind, it will be very difficult for me to test that.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
lolodomo: Each update of an item on the server side triggers a refresh of the page in openHAB app. That's why a specific refresh of an image element linked to an item of type image is not useful.
The refresh has a sense only when using an URL for the picture with a potential change of the picture behind the URL. It is the case for example if your item is of type String and contains an URL.

Sorry but I don't understand why your mean with selection of webcams. Can you provide items and sitemap example to explain your case ?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
mueller-ma: No, it was just an idea that someone could want to use it.

lolodomo: In this case, this could be implemented later when really required.

lolodomo:
@digitaldan : can this PR be merged ? It is a feature expected since so many months...

digitaldan:
LGTM, thanks

PR #415, opened 2017-10-19
Fix color switches and slider

### LINKED ISSUE #408: No Light state recognition for Color Items!
I have colored hue lights that all show the ON version of their icon at all times no matter the brightness or ON/OFF state. Sliders for Color Items also fail to represent the current brightness value as they do in Basic UI on loading of a sitemap with sliders for color items they always show a 0% value. Everything shows as expected through the web version of Basic UI on a pc so I know it is an android app issue!

mueller-ma:
Fix #408

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/OpenHAB2Widget.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/OpenHAB2Widget.java
@@ -1,5 +1,7 @@
 package org.openhab.habdroid.model;
 
+import android.util.Log;
+
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -7,7 +9,7 @@ import org.json.JSONObject;
 import java.util.ArrayList;
 
 public class OpenHAB2Widget extends OpenHABWidget {
-
+    private final static String TAG = OpenHAB2Widget.class.getSimpleName();
     private String iconFormat;
 
     public OpenHAB2Widget() {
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/model/OpenHAB2Widget.java
```diff
@@ -20,9 +22,27 @@ public String getIconPath() {
...
+                    try {
+                        itemState = String.valueOf(widgetItem.getStateAsBrightness());
+                        if(getType().equals("Switch")) {
+                            if (itemState.equals("0")) {
+                                itemState = "OFF";
+                            } else {
+                                itemState = "ON";
+                            }
+                        }
+                    } catch (Exception e) {
```
maniac103: Where should the exception in this try block come from?

mueller-ma:
Closing for now until we get some information here: https://github.com/eclipse/smarthome/issues/4445

lolodomo:
Why did you reopen it ?

mueller-ma:
Because it fixes an issue ;)
I only closed it because I was not sure how the icon of a color switch should look like.

lolodomo:
Isn't it something relative to the current discussion we have for Eclipse SmartHome ?

mueller-ma:
This pr fixes color slider and switches in general, but also about the fetched icon.

Dracrius:
Sorry it's my bad this got a bit befuddled. This pr does fix something it's just I was putting up a fuss about how it was fixed, as part of the fix was making the android app use the same method as Basic UI for setting dynamic icons. The fix is needed for color items, but if we do settle on updating how dynamic icons in general work in ESH a new pr will be needed. I had him close it originally as I thought they could be a single pr but getting responses on the ESH side is taking longer then I expected so I agree as well that color item switches and sliders should be fixed either way in the mean time.

PR #418, opened 2017-10-19
Video: fix #402

### LINKED ISSUE #402: Video element: add support for HTTP Live Streaming playlist
Please add support for HTTP Live Streaming playlist.
As it is working in Chrome on Android, I suppose it could work in the app too.
You can see how it was easy to update Basic UI and Classic UI: https://github.com/eclipse/smarthome/pull/4386
The general idea is to bypass the proxying and to use a particular media type.

lolodomo:
Fixes #402

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -539,7 +539,6 @@ public class OpenHABWidgetAdapter extends ArrayAdapter<OpenHABWidget> {
                 break;
             case TYPE_VIDEO:
                 VideoView videoVideo = (VideoView) widgetView.findViewById(R.id.videovideo);
-                Log.d(TAG, "Opening video at " + openHABWidget.getUrl());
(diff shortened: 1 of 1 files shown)


mueller-ma:
Doesn't this also implement the video item with the url as state, but only when encoding is hls?

lolodomo:
Nothing changed when encoding different from HLS. In this case, the URL is provided by the  proxy using the proxy URL returned by the REST API. The proxy uses the item value as URL if set or by default the value of the URL element in the video sitemap element.

mueller-ma:
When you remove the check for the encoding wouldn't it fix items like
`Video item=foo url=http://....`?

lolodomo:
Case with `Video item=foo url=http://....`remains unchanged and should work as before in app. Do you mean that there is a problem in such a case ? I don't see any issue reported. Note that I opened yesterday another issue relative to support of video URLs.
My current PR is clearly only to fix the behaviour when and only when encoding is set to HLS. Other cases remains unchaned.

lolodomo:
@digitaldan : can this PR be reviewed ? We need for Netatmo Welcome camera.

PR #421, opened 2017-10-21
Fix the label display for a selection item

### LINKED ISSUE #327: Habdroid always appends item state in label
Opening an issue based on the discussion [here](https://community.openhab.org/t/remove-item-state-from-a-label/31115/5).

It appears that habdroid does not respect the item or sitemap label and will always append item state.  If you put a [] in your sitemap or item label, habdroid will no longer append the state but you end up with [] in your label text.

Example:

Items:
`Number Thermostats_Downstairs_System_Mode "System Mode" { channel="omnilink:thermostat:home:downstairs:system_mode" }
`
Sitemap:
`Selection item=Thermostats_Downstairs_System_Mode mappings=[0="Off",1="Heat",2="Cool",3="Auto",4="Emergency Heat"]`

This should just render a label that says "System Mode", but habdroid appends the state:
image

lolodomo:
Fixes #327

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -579,8 +579,14 @@ public class OpenHABWidgetAdapter extends ArrayAdapter<OpenHABWidget> {
                 break;
             case TYPE_SELECTION:
                 int spinnerSelectedIndex = -1;
-                if (labelTextView != null)
-                    labelTextView.setText(openHABWidget.getLabel());
+                splitString = openHABWidget.getLabel().split("\\[|\\]");
+                if (labelTextView != null) {
+                    if (splitString.length > 0) {
+                        labelTextView.setText(splitString[0]);
+                    } else {
+                        labelTextView.setText(openHABWidget.getLabel());
+                    }
+                }
                 final Spinner selectionSpinner = (Spinner) widgetView.findViewById(R.id.selectionspinner);
                 ArrayList<String> spinnerArray = new ArrayList<String>();
                 Iterator<OpenHABWidgetMapping> mappingIterator = openHABWidget.getMappings().iterator();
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -579,8 +579,14 @@ public void onStopTrackingTouch(SeekBar seekBar) {
                 break;
             case TYPE_SELECTION:
                 int spinnerSelectedIndex = -1;
-                if (labelTextView != null)
-                    labelTextView.setText(openHABWidget.getLabel());
+                splitString = openHABWidget.getLabel().split("\\[|\\]");
```
mueller-ma: Shouldn't it be `String[] splitString = ...`

lolodomo: No because the variable is declared globally and used in several cases of the switch.

mueller-ma: ok, thats in line 115

mueller-ma:
Can I have square brackets inside the label?

lolodomo:
@mueller-ma : I don't understand your suggestion !
The object of this PR is to suppress the display of the state value surrounded by the brackets in the displayed label at left because we already have the value displayed in the selection list at right.
Please take a look to the issue 327 for the description.

lolodomo:
@mueller-ma : ok I see what you mean, you are talking about the title in the opened dialog box. I am fixing that.

lolodomo:
I will need your help because I don't find where this title is set in the sub-windows.

lolodomo:
One interesting point is that I don't have this title in the branch in which I implemented the support for image data. So there was a very very recent merge that added this title in the selection box.

lolodomo:
Ok, it was added by you in PR #393 :-)

lolodomo:
Ok, that is fixed now.

PR #427, opened 2017-10-27
Use a number picker for Setpoint, etc. 

sansei4:
image

Benefits:
- Native behaviour
- No delay - doesn't need to wait for server reply after each state change (sometimes it takes a very long time until it changes, )
- Sometimes the value is not updated at all, only after a screen refresh - i.e. I have to go to the next/previous page and back, to see the new value. Otherwise each pressure on the "plus" sends the same value.
- Reducing the data bus traffic.
- Elimination of an unexpected behaviour of the end device, through the obtaining an intermediate value.
 

I suppose, it depends on the core as well, but I don't know, what exactly is needed to implement this feature.


mueller-ma:
Hi @sansei4,
Thanks for your pull request. Since you are using Github for your first time, you might have mixed issues and pull request. Pull Request (PR) are for sending changed code back to the original repo where issues are used for bug tracking or requesting new features.  Please close this PR and open an issue with your request.

sansei4:
sorry, I'll close it.

PR #433, opened 2017-10-29
Merge About and openHAB-info screen to onw About activity

FlorianSW:
With this commit, both menu entries are also moved from the main menu to
the navigation drawer.

The version number of the app itself is now moved from the preferences
activity to the "about" activity, directly under the app name.

Issue #392 

I haven't changed much in the screens itself, as the commit is already larger as I expected :)


mobile/src/main/res/values/strings.xml
```diff
@@ -149,6 +147,7 @@
     <string name="about_links">Links</string>
     <string name="about_links_list"><![CDATA[\u00B7 <a href="https://github.com/openhab/openhab.android">openHAB App source code</a><br><br> \u00B7 <a href="https://github.com/openhab/openhab.android/issues">Report a problem</a><br><br> \u00B7 <a href="http://docs.openhab.org/">openHAB Documentation</a><br><br> \u00B7 <a href="https://community.openhab.org/">Community Forum</a>]]></string>
     <string name="about_copyright" translatable="false">© 2012 - %s openHAB Foundation</string>
+    <string name="about_version_string">%1$s - %2$s</string>
```
mueller-ma: This should be non-translatable

mobile/src/main/java/org/openhab/habdroid/ui/AboutFragment.java
```diff
@@ -10,36 +10,45 @@
...
         about.setText(Html.fromHtml(getString(R.string.about_license)));
         about.setMovementMethod(LinkMovementMethod.getInstance());
+        TextView appVersion = (TextView) view.findViewById(R.id.app_version);
+        appVersion.setText(getString(R.string.about_version_string,
+                BuildConfig.VERSION_NAME,
+                DateFormat.getDateTimeInstance().format(BuildConfig.buildTime)
+        ));
 
-        String year = new SimpleDateFormat("yyyy").format(Calendar.getInstance().getTime());
+        String year = new SimpleDateFormat("yyyy", Locale.US)
```
mueller-ma: Why locale US?

FlorianSW: Uh, I wasn't even sure that I've commited this change :) My IDE (IntelliJ) suggested to use an explicit Locale over SimpleDateFormat directly. As this is only a year, I though Locale.US fits in all cases.

mobile/src/main/res/values/strings.xml
FlorianSW: Will do that with the next commit! :)

mobile/src/main/java/org/openhab/habdroid/ui/AboutFragment.java
mueller-ma: Also in chinese? I think we can ignore China as long there are no chinese strings.

FlorianSW: Good argument, as this is nothing, which is required in this commit, I'll revert this particular change :)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -1401,6 +1392,16 @@ private void loadDrawerItems() {
             mDrawerItemList.add(OpenHABDrawerItem.menuItem("Discovery", getResources().getDrawable(R.drawable.ic_track_changes_grey600_36dp), DRAWER_INBOX));
             mDrawerItemList.add(OpenHABDrawerItem.menuItem("Bindings", getResources().getDrawable(R.drawable.ic_extension_grey600_36dp), DRAWER_BINDINGS));
         }
+        Drawable aboutDrawable = getResources().getDrawable(R.drawable.ic_info_outline);
+        aboutDrawable.setColorFilter(
+                ContextCompat.getColor(this, R.color.colorAccent_themeDark),
```
mueller-ma: That color is different from the other icons in the drawer.

FlorianSW: Yes, unfortunately. However, I would like to avoid introducing another png file or a separate grey fr the navigation drawer icons. To streamline the color I've proposed PR #434 :) Is that ok for you? :]

mueller-ma: Yes, thanks.
Do you think we should use the same color for all themes?

mueller-ma: I think the color is good for all themes. You save the color here https://github.com/openhab/openhab.android/pull/434/files#diff-07e4a6e281af02db331ff3f81d9f86d8R1395 and query it for the about drawable again. Could you rebase #434 on top of this pr?

mobile/src/main/res/menu/main_menu.xml
```diff
@@ -24,16 +24,6 @@
...
-        android:title="@string/mainmenu_openhab_info"
-        app:showAsAction="never" />
-    <item
-        android:id="@+id/mainmenu_about"
-        android:orderInCategory="105"
-        android:title="@string/about_title"
-        app:showAsAction="never" />
-    <item
         android:id="@+id/mainmenu_openhab_preferences"
         android:orderInCategory="106"
```
mueller-ma: Can you change this to 104?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
FlorianSW: Hmm, rebasing this PR based on 434 would include all changes from the one PR in this PR; which are unrelated :/ I think, that there will be merge conflicts in one PR, dependant on which one is merged first, and I would resolve the conflicts with keeping in mind, that the color is already saved, ok? :)

And to the theme thing: Yes, the color should be ok, I've switched all themes and any theme looks good with the grey color :)

mobile/src/main/res/menu/main_menu.xml
FlorianSW: yep :)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
mueller-ma: Ok.

lolodomo:
I just tested and it works well.
But note that it is not at all a fix for the issue 392. The main purpose of this issue was to suppress the side menu and in particular access to sitemaps.

FlorianSW:
@lolodomo I know, that's why I haven't used the "Fixes" word or one of the others which should close the issue automatically, so this PR shouldn't close the issue automatically, right? However, if you say, that it would be better to not even mention the issue, I'm ok with removing it from the desciption :)

mobile/src/main/java/org/openhab/habdroid/ui/AboutFragment.java
mueller-ma: I thought about this and now I think the change to Locale.US is usefull. Imagine some country with other years, than they would see "Copyright 2010 - 84512" or something like that.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
mueller-ma: The other pr is merged now.

PR #434, opened 2017-10-29
Replace png images for navigation drawer with Vector drawable equivalent

FlorianSW:
Vector drawables scales better on different sizes and looks lees fuzzy
when they are resized.

This also allows us to have different color for the icons in different
themes much more easily.


FlorianSW:
Some images for comparsion from my HTC 10:
Before:
screenshot_20171029-184414

After:
screenshot_20171029-183958

mueller-ma:
Can you also replace the microphone in the action bar with a svg? And does svg have any minimum api level?

FlorianSW:
Done! :)

mueller-ma:
Can you also remove the unused customMic attribute from the themes?

PR #436, opened 2017-10-30
Fix overlapping label and item text/buttons

### LINKED ISSUE #435: Label and item text overlap when label is too long
Same as #121, but for Selection, Switch without mappings, setpoint, text, slider, rollershutter

mueller-ma:
This is the same than #390, but for all other elements.
I changed the relative layouts to constraint layouts.

Fixes #435

After:
screenshot_2017-10-30-19-23-03
screenshot_2017-10-30-19-23-06

Sitemap:
```
        Selection item=number4 label="hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhdddddddddddddddddddddddddddddddddddddddddddddddddddddddddhhhhaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaahhhhh" mappings=[1="foo", 2="bar"]
        Switch item=number4 label="hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhdddddddddddddddddddddddddddddddddddddddddddddddddddddddddhhhhaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaahh" mappings=[1="foo", 2="bar"]
        Switch item=number4 label="hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhdddddddddddddddddddddddddddddddddddddddddddddddddddddddddhhhhaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaahh"
        Setpoint item=number4 label="hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhdddddddddddddddddddddddddddddddddddddddddddddddddddddddddhhhhaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaahhhh[%d]"
        Text item=number4 label="hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhdddddddddddddddddddddddddddddddddddddddddddddddddddddddddhhhhhhhhaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa[%d]"
        Slider item=number4 label="hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhdddddddddddddddddddddddddddddddddddddddddddddddddddddddddhhhhhhhhaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa[%d]"
        Default item=color1 label="hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhdddddddddddddddddddddddddddddddddddddddddddddddddddddddddhhhhhhhhaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa[%d]"
        Default item=roller1 label="hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhdddddddddddddddddddddddddddddddddddddddddddddddddddddddddhhhhhhhhaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa[%d]"
```
Would be great if someone could test this on another device.

### DIFF
--- a/mobile/src/main/res/layout/openhabwidgetlist_coloritem.xml
+++ b/mobile/src/main/res/layout/openhabwidgetlist_coloritem.xml
@@ -1,66 +1,83 @@
 <?xml version="1.0" encoding="utf-8"?>
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<android.support.constraint.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/relativeLayout"
     android:layout_width="fill_parent"
     android:layout_height="wrap_content"
+    android:background="?android:activatedBackgroundIndicator"
     android:descendantFocusability="blocksDescendants"
-    android:paddingLeft="@dimen/widgetlist_item_left_margin"
-    android:paddingRight="@dimen/widgetlist_item_right_margin"
-    android:background="?android:activatedBackgroundIndicator">
+    tools:layout_editor_absoluteY="25dp">
 
     <include
         android:id="@+id/colorleftlayout"
-        layout="@layout/openhabwidgetlist_icontext" />
-
-    <LinearLayout
-        android:layout_width="wrap_content"
+        layout="@layout/openhabwidgetlist_icontext"
+        android:layout_width="0dp"
         android:layout_height="wrap_content"
-        android:layout_alignParentEnd="true"
-        android:layout_alignParentRight="true"
-        android:layout_alignParentTop="true"
+        android:layout_marginEnd="8dp"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toStartOf="@+id/colorbutton_up"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        android:layout_marginRight="8dp" />
+
+
+    <ImageButton
+        android:id="@+id/colorbutton_up"
+        android:layout_width="46dip"
+        android:layout_height="wrap_content"
+        android:layout_marginEnd="0dp"
         android:focusable="false"
-        android:focusableInTouchMode="false">
+        android:focusableInTouchMode="false"
+        android:paddingLeft="3dip"
+        android:paddingRight="3dip"
+        android:src="@drawable/ic_action_up"
+        app:layout_constraintBottom_toBottomOf="parent"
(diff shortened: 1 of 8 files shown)


mueller-ma:
@wrosenauer Can you test following apk: [mobile-debug.zip](https://github.com/openhab/openhab.android/files/1428334/mobile-debug.zip)

ThomDietrich:
@mueller-ma thanks for solving the issue! Works as expected on my device.

mueller-ma:
@ThomDietrich Which device do you have?

ThomDietrich:
I've tested your attached app on a OnePlus 3 smartphone with android 7.
I have to add one bug report: If the value of an item (e.g. a string) is long, the label will be pushed out on the left. Can you do something about that?

ThomDietrich:
I thought about the whole issue some more, especially for Text sitemap elements. You are currently leaving the value unchanged and are shortening the label. I'm not sure if this is ideal. The sitemap was hand build by the user and the label is something meaningful. Without it (after being replaced by just "..." in the worst case) the meaning of the value text might get unclear. 
I wonder if it would be a better solution for the Text element to scroll the value string if an overlap would occur. Is that easy to implement?

mueller-ma:
I tought about implementing a max width for the text value, so it is always a few characters of the label shown.

mueller-ma:
The text now scrolls:
screenshot_2017-10-31-19-58-00
[mobile-debug.zip](https://github.com/openhab/openhab.android/files/1431966/mobile-debug.zip)

ThomDietrich:
Hey @mueller-ma,
doesn't seem to work as expected. See screenshot...

index

mueller-ma:
I dont know why this is happening. Now the icon and the value are always shown and (if space) the label.
[mobile-debug.zip](https://github.com/openhab/openhab.android/files/1441769/mobile-debug.zip)

ThomDietrich:
Same behavior. I'm also not so sure about the scrolling any more. It creates a lot of unwanted visual chaos.
Let's look at it differently: The sitemap is built by the end user and he/she has to make sure, that everything is in order. The case where a string is too long is hence really an exception. In this case it might be okay to just trim it down to a certain length, just as you do with the label for other element types currently.
At the end of the day the case where the app is forced to shorten strings will eventually be handled by the end user by manually shortening his/her strings to fit onto the screen.

mueller-ma:
Ok, I will undo my last commit. I think the actuall value is more important than the label. If the label is not visible, the user should split it up in two text elements.

mueller-ma:
Triggering travis

Issue #342, opened 2017-08-14
Use weblate for translations


mueller-ma:
Weblate offers free hosting for oss, has login via github and makes translations a lot easier.
https://weblate.org/en/hosting/

mueller-ma:
@digitaldan If I get a "yes" I can start working on this. It would make translations much easier.

jmvaz:
I can contribute with PT-PT language.

mueller-ma:
You can fork the repository, copy the file mobile/src/main/res/values/strings.xml to mobile/src/main/res/values-pt/strings.xml, remove all lines with "translatable=false" and translate the other lines. Than open a PR.

mueller-ma:
@ThomDietrich mentioned crowdin [here](https://github.com/openhab/openhab.android/pull/437#discussion_r148939226), however its not free.

ThomDietrich:
https://github.com/marketplace/category/localization
https://github.com/marketplace/crowdin

Crowdin is completely free for open source projects and has a strong integration with GitHub.

mueller-ma:
Ok, I didnt find that on their website. Is crowdin under a free license?

ThomDietrich:
Found it here: https://crowdin.com/page/open-source-project-setup-request
Not sure what you mean by "under a free license". They are a company looking for paying customers offering an free open source plan to lure in the potential customers of tomorrow. Very much the same strategy as of GitHub, Slack, Trello and co ;)

mueller-ma:
I mean with "under a free license": The weblate source code is under GPL: https://github.com/WeblateOrg/weblate

ThomDietrich:
Ah I see. Well no it is not. For me the offered functionality is a strong argument: https://support.crowdin.com

mueller-ma:
We can give it a try. I think @digitaldan must do the setup or can you do it?

ThomDietrich:
Let's give @digitaldan the honor of final decision 😃 I technically could but this is clearly not my call to make.

ThomDietrich:
@digitaldan if okay with you i can set up crowdin tomorrow. We are already using it in other repositories and it's quite nice. It would also be a good thing to have all openHAB projects translated there to attract more contributors.

@mueller-ma this will then also make your request for updated strings easier ;)

ThomDietrich:
[...]

PR #437, opened 2017-11-01
Selection: value not in list

### LINKED ISSUE #290: Display of selection elements

lolodomo:
Fixes #290

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -618,6 +618,10 @@ public class OpenHABWidgetAdapter extends ArrayAdapter<OpenHABWidget> {
                             spinnerSelectedIndex = spinnerArray.size() - 1;
                         }
                 }
+                if (spinnerSelectedIndex == -1) {
+                    spinnerArray.add("          ");
+                    spinnerSelectedIndex = spinnerArray.size() - 1;
+                }
                 ArrayAdapter<String> spinnerAdapter = new SpinnerClickAdapter<String>(this.getContext(),
                         android.R.layout.simple_spinner_item, spinnerArray, openHABWidget, new AdapterView.OnItemClickListener() {
                     @Override
@@ -626,7 +630,7 @@ public class OpenHABWidgetAdapter extends ArrayAdapter<OpenHABWidget> {
                         String selectedLabel = (String) parent.getAdapter().getItem(index);
                         Log.d(TAG, "Spinner onItemSelected selected label = " + selectedLabel);
                         OpenHABWidget openHABWidget = (OpenHABWidget) parent.getTag();
-                        if (openHABWidget != null) {
+                        if (openHABWidget != null && index < openHABWidget.getMappings().size()) {
                             Log.d(TAG, "Label selected = " + openHABWidget.getMapping(index).getLabel());
                             for (OpenHABWidgetMapping openHABWidgetMapping : openHABWidget.getMappings()) {
                                 if (openHABWidgetMapping.getLabel().equals(selectedLabel)) {


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -618,6 +618,10 @@ public void onStopTrackingTouch(SeekBar seekBar) {
                             spinnerSelectedIndex = spinnerArray.size() - 1;
                         }
                 }
+                if (spinnerSelectedIndex == -1) {
+                    spinnerArray.add("");
```
mueller-ma: Maybe we should add "Please select...", "Undefined", "Value not defined in selection" or just "       ". The empty spinner is pretty hard to hit.

ThomDietrich: Agreed!

lolodomo: I would like to avoid adding a new string to translate.
Is it ok if I use 3 spaces ?

mueller-ma: What about 10 spaces? 3 sounds a little bit too less.

ThomDietrich: > I would like to avoid adding a new string to translate.

If technicalities like this are hindering development we are doomed :) Did you consider using https://crowdin.com for the translation management?
Just spaces sounds unclear. Could you post a screenshot?

mueller-ma: See #342

lolodomo: Translation management is certainly not the purpose of this PR.
My idea with this PR was to propose a kind of workaround which is better than nothing. I hope it could be better handled later (with the blank line being hidden) by a developer having the correct skills. That is why I wanted to keep it as simple as possible.
Additionally Basic UI is displaying a blank in this case. I think it is a good idea to have a similar behaviour in the Android app.
If you really want a value, we will probably discuss during ages about this value. I could propose "Other".

mueller-ma: When basic ui uses whitespace, then we also should use it. You might want to use non breaking spaces.

lolodomo: "Value not defined in selection" is the right description but is largely too long.

ThomDietrich: > I think it is a good idea to have a similar behaviour in the Android app.

Agreed. Okay then 👍

lolodomo:
Change done.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -618,6 +618,10 @@ public void onStopTrackingTouch(SeekBar seekBar) {
                             spinnerSelectedIndex = spinnerArray.size() - 1;
                         }
                 }
+                if (spinnerSelectedIndex == -1) {
+                    spinnerArray.add("          ");
```
FlorianSW: Does this really need to be an arbitrary amount of whitespaces? :/

mueller-ma: Without whitespaces you need to hit that small arrow to open the dropdown.

FlorianSW: Hmm, at least for the default sitemap (Widget overview -> Scene selection), the area of the selection widget is big enough to open the selection window, so I don't need to tap on the arrow.

Also: Doesn't this adds an empty element to the selection list? This looks a bit weird :/

(I've changed the logic a bit, so that spinnerSelectedIndex is always -1).

mueller-ma: It is big enough, because it always displays an value. This PR is for the case that the selection item would be set to  - in this case - "sleeping".

Issue #429, opened 2017-10-28
Use a number picker for Setpoint, etc.


sansei4:
![image](https://user-images.githubusercontent.com/7754105/32126569-a57bc328-bb71-11e7-8f71-3c6c40cb613d.png)

Benefits:
- Native behaviour
- No delay - doesn't need to wait for server reply after each state change (sometimes it takes a very long time until it changes, )
- Sometimes the value is not updated at all, only after a screen refresh - i.e. I have to go to the next/previous page and back, to see the new value. Otherwise each pressure on the "plus" sends the same value.
- Reducing the data bus traffic.
- Elimination of an unexpected behaviour of the end device, through the obtaining an intermediate value.

I suppose, it depends on the core as well, but I don't know, what exactly is needed to implement this feature.

mueller-ma:
No need to add "#429" to the title.

sansei4:
How to set a label?

mueller-ma:
Labels can only be set by people with write access.

boc-tothefuture:
:+1:  On this idea.  Using the setpoint value right now for things like temperature and in some cases volume (I know there is a slider, but I often find I don't get enough control w/ the slider) can often not be the best user experience.  Often the presses are not registered as quickly as I want them to be.

mueller-ma:
I haven't found a way to implement this with the android dialog.

boc-tothefuture:
@mueller-ma  - Does it need to be a dialog?  I was playing with it and even if its not a dialog the user experience seems better than the button presses. (although I haven't wired it up)

PR #442, opened 2017-11-02
Number picker for setpoint

boc-tothefuture:
Tagged as a WIP because I am not sure volumeUp and volumeDown widgets were doing in the old model and how or if that would map into a number picker?

Additionally, I have not yet tested this actually make a change in OpenHab.

This is a pass at resolving #429


mueller-ma:
I'm working on step size.

mueller-ma:
@lolodomo Can we expect minValue, maxValue and step always be set in the REST API?

boc-tothefuture:
@mueller-ma - I am not an android expert, but from throwing this together it appears the number picker wont work without a min/max.  So we would have to default to one if not supplied - but no idea what a reasonable default is across the range of items that could be controlled.

lolodomo:
Min/max/step are present in the REST API for a setpoint widget but I think they can be null if the user has not defined them in his sitemap.

mueller-ma:
Please merge https://github.com/boc-tothefuture/openhab.android/pull/1/

boc-tothefuture:
@mueller-ma - I merged it.. But should it be stepSize instead of stepsize?

mueller-ma:
Yes, you can change it

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -707,19 +707,19 @@ public void onClick(DialogInterface dialog, int which) {
                             final NumberPicker numberPicker = (NumberPicker) dialogView.findViewById(R.id.numberpicker);
                             int minValue = (int) openHABWidget.getMinValue();
                             int maxValue = (int) openHABWidget.getMaxValue();
-                            int stepsize = (int) openHABWidget.getStep();
-                            final int stepsizeSave;
+                            int stepSize = (int) openHABWidget.getStep();
+                            final int stepSizeSave;
```
mueller-ma: Maybe there is a better way.

boc-tothefuture: What were you trying to accomplish?

mueller-ma: Changing stepsize even when its final.

boc-tothefuture:
```
                            int minValue = (int) openHABWidget.getMinValue();
                            int maxValue = (int) openHABWidget.getMaxValue();
                            final int stepSize;
                            if(minValue == maxValue) {
                                stepSize = 1;
                            } else {
                                stepSize = (int) openHABWidget.getStep();
                            }

                            numberPicker.setMaxValue(maxValue/stepSize);
                            numberPicker.setMinValue(minValue);
                            NumberPicker.Formatter formatter = new NumberPicker.Formatter() {
                                @Override
                                public String format(int value) {
                                    int temp = value * stepSize;
                                    return "" + temp;
                                }
                            };
```

?

mueller-ma: stepSize must be declared final, but I want to set it to 1 in case minValue=maxValue

boc-tothefuture: yeah, that is what I do above in that snippet.  declared final, if min==max its 1, otherwise it gets step from the widget.

mueller-ma: Thats better than mine. Can you commit it?

boc-tothefuture: done.

mueller-ma:
I've submitted a second PR that resolves some issue. One issue is still left: Last comment here https://stackoverflow.com/questions/12979643/change-the-step-size-of-a-numberpicker/30469826#30469826

boc-tothefuture:
@mueller-ma - Is the setDisplayedValues a better option for doing steps?  Would that resolve the initial display issue?

boc-tothefuture:
@mueller-ma  - let me know if you want me to look at the setDisplayedValues approach?

mueller-ma:
Yes, you can give it try. I think you can assume min and max value (if not ser by the user) to be 0 and 100.

boc-tothefuture:
@mueller-ma - It appears that in OpenHabWidget they default to 0 and 100, so no action should be necessary.  Correct?

boc-tothefuture:
@mueller-ma  - Before I started the approach with setDisplayedValues I attempted your approach and I can't replicate the issue you report.  Can share a screen shot?  It works fine for me unless the current value doesn't map to one of the step values.

boc-tothefuture:
@mueller-ma 
```
 try {
                                numberPicker.setValue(Integer.parseInt(valueTextView.getText().toString()) / stepSize);
                            } catch (NumberFormatException e) {
                                e.printStackTrace();
                                Log.e(TAG, "set 0");
                                numberPicker.setValue(0);
                            }
```
Should we be setting to zero or minValue in that case?

mueller-ma:
> It appears that in OpenHabWidget they default to 0 and 100, so no action should be necessary. Correct?

Not sure if thats in all oh versions.

> Should we be setting to zero or minValue in the error case?

I think minValue is better.

mueller-ma:
The item is "6", but 6 isnt displayed:

screenshot_2017-11-03-16-45-27

boc-tothefuture:
> It appears that in OpenHabWidget they default to 0 and 100, so no action should be necessary. Correct?
> Not sure if thats in all oh versions.

Its not in OH, its in the android code. 
```
private Boolean legend = null;
	private float minValue =0;
	private float maxValue = 100;
	private float step = 1;
```

boc-tothefuture:
@mueller-ma What version of android are you seeing that in and what is your step value?  I only see that when I have a step value set to something that doesn't properly align w/ the potential formatted values.

mueller-ma:
Android 5, step size 2, min 0 and max 10

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -664,47 +668,82 @@ public void onItemClick(AdapterView<?> parent, View view, int index, long id) {
...
+                                stepValues[i] = String.valueOf(minValue + (i*stepSize));
+                            }
+
+                            AlertDialog.Builder builder = new AlertDialog.Builder(context);
+
+                            if (labelTextView != null) {
+                                builder.setTitle(labelTextView.getText());
+                            }
+                            final LayoutInflater inflater = LayoutInflater.from(context);
+                            final View dialogView = inflater.inflate(R.layout.openhab_dialog_numberpicker, null);
```
mueller-ma: Android Studio suggest not to use null here

mueller-ma:
Item is set to 10. They all show the right value, but when I tab on one, they...
```
        Setpoint item=number4 label="aahhhh[%d]" shows 2
        Setpoint item=number4 label="aahhhh[%d]" maxValue=10 minValue=0 step=2 shows 2
        Setpoint item=number4 label="aahhhh[%d]" maxValue=9 minValue=0 step=2 shows 2
        Setpoint item=number4 label="aahhhh[%d]" maxValue=10 minValue=10 step=3 shows 10
        Setpoint item=number4 label="aahhhh[%d]" maxValue=10 minValue=10 shows 10
        Setpoint item=number4 label="aahhhh[%d]" minValue=0 step=2 shows 10
```

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
boc-tothefuture: Thanks!

If you read the link android studio send you to it says this at the end:

>  There are of course instances where you can truly justify a null parent during inflation, but they are few. One such instance occurs when you are inflating a custom layout to be attached to an AlertDialog.

That is what we are doing here. 

> The issue here is that AlertDialog.Builder supports a custom view, but does not provide an implementation of setView() that takes a layout resource; so you must inflate the XML manually. However, because the result will go into the dialog, which does not expose its root view (in fact, it doesn’t exist yet), we do not have access to the eventual parent of the layout, so we cannot use it for inflation. It turns out, this is irrelevant, because AlertDialog will erase any LayoutParams on the layout anyway and replace them with match_parent.

So I think using null there is the appropriate course of action.

boc-tothefuture:
So I understand:
```
     * Setpoint item=number4 label="aahhhh[%d]" shows 2
     * Setpoint item=number4 label="aahhhh[%d]" maxValue=10 minValue=0 step=2 shows 2
     * Setpoint item=number4 label="aahhhh[%d]" maxValue=9 minValue=0 step=2 shows 2
     Setpoint item=number4 label="aahhhh[%d]" maxValue=10 minValue=10 step=3 shows 10
     Setpoint item=number4 label="aahhhh[%d]" maxValue=10 minValue=10 shows 10
     Setpoint item=number4 label="aahhhh[%d]" minValue=0 step=2 shows 10
```
The ones with the * are the incorrect behavior?

mueller-ma:
Yes.

mueller-ma:
Did it work on your device?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -664,47 +668,82 @@ public void onItemClick(AdapterView<?> parent, View view, int index, long id) {
...
+                            numberPicker.setMinValue(0);
+                            numberPicker.setMaxValue(stepValues.length -1);
+                            numberPicker.setDisplayedValues(stepValues);
+
+                            // Find the closest value in the calculated step values.
+                            int stepIndex = Arrays.binarySearch(stepValues, valueTextView.getText());
+                            if ( stepIndex < 0 ){
+                                stepIndex = (-stepIndex) - 1; // Use the returned insertion point if value is not found and select the closest value.
+                             }
+                            numberPicker.setValue(stepIndex);
```
mueller-ma: Should this be `numberPicker.setValue(stepValues[stepIndex];`?

boc-tothefuture: I don't think so....   Based on SO at least.  You are now using the index once you start using setDisplayedValues.

https://stackoverflow.com/questions/39541318/android-numberpicker-setvalue-does-not-work-with-setdisplayedvalues

boc-tothefuture:
@mueller-ma can you provide your sitemap so I can test against it?

mueller-ma:
See https://github.com/openhab/openhab.android/pull/442#issuecomment-341810327. Its just a test server, so there is nothing more in the sitemap.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
mueller-ma: I see.

mueller-ma:
Found the issue: When you add values to stepValues they are e.g. 0 2 4 6 8 10.
From the java docs for binarySearch:

> The range must be sorted (as by the sort(int[]) method) prior to making this call. If it is not sorted, the results are undefined

Since the array is a string array it is sorted when it has the following order: 0 10 2 4 6 8

boc-tothefuture:
wow good find.  I will fix it.

boc-tothefuture:
@mueller-ma - should be fixed now.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -664,47 +669,87 @@ public void onItemClick(AdapterView<?> parent, View view, int index, long id) {
...
+
+                            // Find the closest value in the calculated step values.
+                            int stepIndex = Arrays.binarySearch(stepValues, valueTextView.getText(), new Comparator<CharSequence>() {
+                                @Override
+                                public int compare(CharSequence t1, CharSequence t2) {
+                                    return Integer.valueOf(t1.toString()).compareTo(Integer.valueOf(t2.toString()));
+                                }
+                            });
+                            if ( stepIndex < 0 ){
+                                stepIndex = (-stepIndex) - 1; // Use the returned insertion point if value is not found and select the closest value.
```
mueller-ma: I have an item set to 10, min 0, max 9 and step 2. The closest value is 8, but 0 is shown instead.

mueller-ma:
> @mueller-ma - should be fixed now.

@boc-tothefuture That works now.

mueller-ma:
Can you check in OpenHABWidget.java that stepsize > 0 and maxSize > minSize?
Otherwise the app may crash.

boc-tothefuture:
@mueller-ma Ok, it should display your latest test case correct.

Can I get a second set of eyes on this logic:
```
  if ( stepIndex < 0 ){
       stepIndex = (-(stepIndex+1)); // Use the returned insertion point if value is not found and select the closest value.
       stepIndex = Math.min(stepIndex, stepValues.length -1);  //handle case where insertion would be larger than the array
 }
```

That is based on reversing this logic here:
```
index of the search key, if it is contained in the array; otherwise, (-(insertion point) - 1). The insertion point is defined as the point at which the key would be inserted into the array: the index of the first element greater than the key, or a.length if all elements in the array are less than the specified key. Note that this guarantees that the return value will be >= 0 if and only if the key is found.
```

You see any potential corner cases?

boc-tothefuture:
> Can you check in OpenHABWidget.java that stepsize > 0 and maxSize > minSize?
Otherwise the app may crash.

In OpenHABWidget.java or the Adapter?  Also, what is my course of action if that doesn't hold?

mueller-ma:
`final String[] stepValues = new String[(maxValue - minValue)/stepSize +1];`
E.g. (0 - 10)/2+1=-4 so a array with negative size would be created, which causes an exception. But when I think about this, it should be catched by the server.

mueller-ma:
> Can I get a second set of eyes on this logic

Works now. Didnt find any corner cases, but I will think about it.

boc-tothefuture:
@mueller-ma -
  Any idea what to do about the volume widget stuff I have commented out?  I wasn't sure how they aligned with the setpoint or number picker implementation.

mueller-ma:
It seems that some items can be controlled by the volume buttons. That works for slider and setpoint. However it is always the first item on the current page controlled (or I just dont know, how to controll other items)

maniac103:
> It seems that some items can be controlled by the volume buttons. That works for slider and setpoint.

Commit https://github.com/openhab/openhab.android/commit/058c3f9ad8238eacd3a021b63a023ab424ce6f0f introduced that feature - IMHO, it's of questionable value and UX though as it
- is almost undiscoverable
- only works in a random subset of cases (only one setpoint/slider)

boc-tothefuture:
Hmmm.  
The number picker is now an alert dialog so could try and make it control that...
But, that is questionable user interface.  Having used Android for 7+ years if my volume button didn't control the volume because I had a number picker up I would be confused.

Can we remove that functionality from the code here or do we need to discuss?

mueller-ma:
I think its the best to remove it.

mueller-ma:
@digitaldan do you think we should preserve the possibility to controll items via volume buttons? It seems to me kind of buggy

boc-tothefuture:
Removed and also removed WIP from title.   This is ready to merge pending from my perspective.

maniac103:
> Removed volume control

That removal is incomplete though. When removing this, IMHO it should be removed completely, ideally by reverting the commit I linked to above.

boc-tothefuture:
@maniac103 -. 
  Shouldn't complete removal be tracked in a separate issue and PR?
  We could block this PR until that issue/PR is complete.

maniac103:
@boc-tothefuture Good point. That would make sense, yes.

boc-tothefuture:
@maniac103 I opened #446 for discussion of the volume issue.

boc-tothefuture:
Per discussion this should be merged after #447

digitaldan:
@boc-tothefuture can you look at the merge conflict?  I merged #447 first which is what I believe caused the conflict.

boc-tothefuture:
@digitaldan - resolved the merge conflict.

PR #444, opened 2017-11-03
Change icon type name

mueller-ma:
In Paper UI they are called "Bitmap" and "Vector", but I think "Vector graphic" is more common.

### DIFF
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -58,7 +58,7 @@
     <string name="settings_openhab_icon_format_value_png" translatable="false">PNG</string>
     <string name="settings_openhab_icon_format_value_svg" translatable="false">SVG</string>
     <string name="settings_openhab_icon_format_png">Bitmap</string>
-    <string name="settings_openhab_icon_format_svg">Vector graphics</string>
+    <string name="settings_openhab_icon_format_svg">Vector</string>
     <string name="settings_ringtone">Ring tone</string>
     <!-- App messages strings -->
     <string name="title_voice_widget">openHAB Voice Commands</string>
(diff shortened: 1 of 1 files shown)


lolodomo:
This would require a change in all translations. Is PNG and SVG not talking for everybody ?

mueller-ma:
> Is PNG and SVG not talking for everybody ?

Maybe. I opend this PR to get some feedback.

Dracrius:
PNG and SVG is clearer as it also tells a user the expected extension. Helpful when making custom icons seeing as there are countless threads of users having trouble making custom icon sets and many boiled down to users using the wrong format. I think a change like that would worsen these mistakes and make it easier for a user trying to diagnose the issue to over look having the setting set to the right format.

mueller-ma:
@maniac103 @FlorianSW Can I have your opinion?

maniac103:
I don't have a strong opinion on this one. On one hand I think being consistent with other UIs is a good idea/goal, on the other hand I assume the target audience of OpenHAB to have enough IT background to know the PNG and SVG abbreviations/file formats, so using them shouldn't be a problem either.
Dunno really ¯\_(ツ)_/¯

Dracrius:
Well I'd like to ask does openhab actually support multiple bitmap file types? Or not to be to stubrn up but any? PNG while storing a bitmap image isnt really a Bitmap file. Bitmap usually refers to the use of .bmp. I have very very rarely seen a png or png support refered to as bitmap for this reason. Especially because if using a bitmap for icons or something with transperency it'd require an alpha layer and the use of it so you either use .bmp's with alpha layers or png's not usually both.

mueller-ma:
I think only png works. I will close this PR since it doesn't seem to be an improvement.

mueller-ma:
I'll reopen since it is an improvement, when we use the same terms between all interfaces.

PR #449, opened 2017-11-08
Change selection widgets to use radio buttons

### LINKED ISSUE #423: Selection box: selected item not emphased
With the current master branch, selecting a selection opens a sub-window in which you can select your value. There is nothing that shows what is the current selection. I see there is some code for that so I assume there is a bug in the code.

FlorianSW:
Instead of just showing a list of all possible values where the current
value can not be easily identified anymore (without knowing it already
from the sitemap) is confusing when selecting the (probably new) item of
the selection.

Using radio buttons for the presentation of the values fixes this issue
and does not change the functionality at all.

Fixes #423


maniac103:
It might be an official layout from the Gingerbread era though ... maybe there's a similar more updated layout in the support library?

mueller-ma:
We could copy the [file](https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/res/res/layout/select_dialog_singlechoice.xml) to the project and replace textAppearanceLarge with textAppearanceMedium.

mueller-ma:
Triggering travis

PR #453, opened 2017-11-17
Add two espresso ui tests

mueller-ma:
This PR adds the espresso ui testing framework. The first test checks if everything in the intro is present and cancels it. The second test goes through all intro slides, selects the demo sitemap and checks if some entries are present (mic and dropdown menu buttons, sitemap groups, slider, colorpicker, rollershutter, etc.).

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -10,6 +10,7 @@ android {
         minSdkVersion 14
         targetSdkVersion 25
         multiDexEnabled = true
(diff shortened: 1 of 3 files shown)


digitaldan:
Nice!

mueller-ma:
FYI you need to purge the settings after the first test is run. Or set "firstStart" in settings to true, but I couldn't find a way to access the shared prefmanager from the ui tests

PR #455, opened 2017-11-19
Create foss build flavor

mueller-ma:
* Remove Google Analytics since it is unused
* Remove crittercism since they changed their payment plans
* Create dummy gcm libary that is used for the foss build
        => Notifications wont work in the foss build
* Remove cloudbees maven, use jitpack.io instead

See #409 


mueller-ma:
Device registration now works with the full build.

digitaldan:
tested both builds, LGTM!

PR #458, opened 2017-11-22
Create code of conduct

mueller-ma:
Github suggest to add a code of conduct and I also think its a good idea. In line 37 I had to insert a mail address. Is that one ok or does another one suit better?

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,46 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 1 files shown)


digitaldan:
@kaikreuzer this probably is something we would want to do across all projects or not as all?  Thoughts?

mueller-ma:
@kaikreuzer Poke

kaikreuzer:
Sorry @mueller-ma, I didn't see this before.

> this probably is something we would want to do across all projects or not as all? 

Correct and we actually have that in place over almost all projects as a part of the CONTRIBUTING.md see e.g. https://github.com/openhab/openhab-distro/blob/master/CONTRIBUTING.md#community-guidelines. Interestingly, this section has been left away in this repo here, don't ask me why 😟 

We also have this in the central developer documentation, see https://docs.openhab.org/developers/contributing/contributing#community-guidelines.

I'd therefore suggest to align the Android repo with the others (and if you feel that these guidelines are missing anything essential, let's first discuss this in the forum under the "code" section).

mueller-ma:
Ok, I will add the community guidelines to this repo.

PR #467, opened 2017-11-26
Update German translation

mueller-ma:
@maniac103 "Notification message" => "Benachrichtigung"?


mueller-ma:
@floriansw I think you introduced that string, maybe you can help.

maniac103:
@mueller-ma 'Benachrichtigung' sounds good to me; it's the title of a dialog that shows the contents of a GCM notification.

FlorianSW:
Sounds good to me, too :)

PR #469, opened 2017-11-27
Ensure step size has a minimum of 1

### LINKED ISSUE #468: Crash with PR: Number picker for setpoint
https://github.com/openhab/openhab.android/pull/442/files#diff-7084e55d51f6579524d6f8ef39d55555R705

@boc-tothefuture I did not test this when I merged, but now that I'm testing something else I'm getting a app crash on our demo site:

```
11-26 14:18:13.084 25000-25000/org.openhab.habdroid E/AndroidRuntime: FATAL EXCEPTION: main
                                                                      Process: org.openhab.habdroid, PID: 25000
                                                                      java.lang.ArithmeticException: divide by zero
                                                                          at org.openhab.habdroid.ui.OpenHABWidgetAdapter$12.onClick(OpenHABWidgetAdapter.java:705)
                                                                          at android.view.View.performClick(View.java:5204)
                                                                          at android.view.View$PerformClick.run(View.java:21153)
                                                                          at android.os.Looper.loop(Looper.java:148)
                                                                          at android.app.ActivityThread.main(ActivityThread.java:5417)
                                                                          at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                                                                          at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
11-26 14:19:12.534 25000-25062/org.openhab.habdroid I/MyWebImage: fetched bitmap for https://myopenhab.org/icon/rollershutter?state=ON&format=PNG
11-26 14:19:12.635 25000-25065/org.openhab.habdroid I/MyWebImage: fetched bitmap for https://myopenhab.org/icon/rollershutter?state=100&format=PNG
```

boc-tothefuture:
It appears of if the step size is set to zero we would have a divide by zero.  This should resolve #468

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -710,7 +710,16 @@ public class OpenHABWidgetAdapter extends ArrayAdapter<OpenHABWidget> {
 
                             final String[] stepValues = new String[((int)(Math.abs(maxValue - minValue)/stepSize)) +1];
                             for(int i = 0; i < stepValues.length; i++){
-                                stepValues[i] = String.valueOf(minValue + (i*stepSize));
+                                //Check if step size is a whole integer.
+                                if( stepSize == Math.ceil(stepSize)){
+                                    //Cast to int to prevent .0 being added to all values in picker
+                                    stepValues[i] = String.valueOf((int) (minValue + (i*stepSize)));
+                                }
+                                else{
+
+                                    stepValues[i] = String.valueOf(minValue + (i*stepSize));
+                                }
+
                             }
 
                             AlertDialog.Builder builder = new AlertDialog.Builder(context);
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -698,7 +698,8 @@ public class OpenHABWidgetAdapter extends ArrayAdapter<OpenHABWidget> {
                             if(minValue == maxValue) {
                                 stepSize = 1;
                             } else {
-                                stepSize = (int) openHABWidget.getStep();
+                                //Ensure min step size is 1
+                                stepSize = Math.max(1, (int) openHABWidget.getStep());
                             }
 
 
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -694,6 +694,11 @@ public class OpenHABWidgetAdapter extends ArrayAdapter<OpenHABWidget> {
 
                             int minValue = (int) openHABWidget.getMinValue();
                             int maxValue = (int) openHABWidget.getMaxValue();
+
+                            //This prevents an exception below. But could lead to user confusion if this case is ever encountered.
+                            if (minValue > maxValue){
+                                maxValue = minValue;
+                            }
                             final int stepSize;
                             if(minValue == maxValue) {
                                 stepSize = 1;
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -703,7 +703,7 @@ public class OpenHABWidgetAdapter extends ArrayAdapter<OpenHABWidget> {
                             }
 
 
-                            final String[] stepValues = new String[(maxValue - minValue)/stepSize +1];
+                            final String[] stepValues = new String[Math.abs(maxValue - minValue)/stepSize +1];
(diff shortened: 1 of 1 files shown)


mueller-ma:
Can you also check if maxSize > minSize?

boc-tothefuture:
What is the appropriate action if max < min?  Set max == min?

mueller-ma:
Yes, maybe show a message

boc-tothefuture:
I think openhab core should be logging/erroring when max < min. That wouldn't be limited to just the android platform. 
Is setting max = min the right thing to do here?  Or should I be swapping max and min? I am not sure what is consistent with the other platforms? 
Or is the appropriate method to just throw an exception?  
I think silently correcting this for one platform might be confusing to the user.  Essentially, if the user mixes up the min and max - they are going to get a setpoint that only has one value..
I am leaning towards throwing an exception so the user knows they have a configuration issue.

mueller-ma:
What do you mean with "Throw an exception"? I dont want the app to crash, but a dialog with some information are good.

boc-tothefuture:
@mueller-ma - Do you have an example of other items in openhab android that inform the users of configuration errors in the sitemap?  I would like to base it on that if we already have the alert mechanism in place.

If we accept that max < min is not a valid state..  If we don't throw an exception(crash) - what do we do?  Do we hide the widget? Grey it out?

mueller-ma:
I think the best idea is to just set max = min, because it should be catched server side. No other item validation is done in the app.

boc-tothefuture:
Not trying to be difficult here.  But if the server side should catch this should we not just crash the app if its not caught?  I am trying to understand why an 'undefined behavior' (from the user perspective), changing max to min, is better than a crash? 

I am happy to implement the max = min workaround - but trying to better understand the philosophy here.

mueller-ma:
> But if the server side should catch this should we not just crash the app if its not caught?

Yes, but currently the server side tests are not implemented and it would be a very bad user experience if the app crashes after the update.

boc-tothefuture:
@digitaldan Can you validate this fixes the crash you ran into?

digitaldan:
I'm getting another crash now:
```
11-27 16:52:32.516 23439-23439/org.openhab.habdroid E/AndroidRuntime: FATAL EXCEPTION: main
                                                                      Process: org.openhab.habdroid, PID: 23439
                                                                      java.lang.NumberFormatException: Invalid int: "21.5 °C"
                                                                          at java.lang.Integer.invalidInt(Integer.java:138)
                                                                          at java.lang.Integer.parse(Integer.java:410)
                                                                          at java.lang.Integer.parseInt(Integer.java:367)
                                                                          at java.lang.Integer.parseInt(Integer.java:334)
                                                                          at java.lang.Integer.valueOf(Integer.java:525)
                                                                          at org.openhab.habdroid.ui.OpenHABWidgetAdapter$12$3.compare(OpenHABWidgetAdapter.java:751)
                                                                          at org.openhab.habdroid.ui.OpenHABWidgetAdapter$12$3.compare(OpenHABWidgetAdapter.java:748)
                                                                          at java.util.Arrays.binarySearch(Arrays.java:589)
                                                                          at java.util.Arrays.binarySearch(Arrays.java:553)
                                                                          at org.openhab.habdroid.ui.OpenHABWidgetAdapter$12.onClick(OpenHABWidgetAdapter.java:748)
                                                                          at android.view.View.performClick(View.java:5204)
                                                                          at android.view.View$PerformClick.run(View.java:21153)
                                                                          at android.os.Looper.loop(Looper.java:148)
                                                                          at android.app.ActivityThread.main(ActivityThread.java:5417)
                                                                          at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                                                                          at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)

```

digitaldan:
Something else is not right, that widget is suppose to be a spinner I believe, but there is no + or - on it. @mueller-ma Is this also related to the other PR work on spinners?

boc-tothefuture:
@digitaldan That was part of the original PR was changing it from the +/- thing to an android number picker.
The original issue is #429 that contains the discussion of why to move to a number picker.

boc-tothefuture:
@digitaldan -
This PR was to turn it from a se

boc-tothefuture:
There is a lot that is going to need to be fixed here.  

The initial problem I can see was that the PR didn't support step values that were not integers.  So a step of 0.5 was causing it to crash.

I have added support for floats as steps, which makes everything a float.  So float values look ok, but now if you are using just integers you have an extra .0 on the end of all values which is not a good solution.

Additionally, the initial PR only supported pure numbers in the labels and additional text breaks the binary search method used to find the closest matching value.

boc-tothefuture:
@digitaldan and @mueller-ma 
The latest version I just pushed works with both float and int step sizes.  It also works with labels that have things besides numbers in them.  I tested it against my system and demo and doesn't crash.

digitaldan:
>That was part of the original PR was changing it from the +/- thing to an android number picker.

Gotcha, I don't think i fully grasped the UI change, and actually did not try it out (which is unusual for me).  I'll need to open a new issue as now there is not any feedback that the UI element is clickable, which is why it took me a bit to figure out what I had clicked which made the app crash.  

This does indeed fix the crash issue.

PR #470, opened 2017-11-27
Remove Bindings and Discovery funtionality from app

### LINKED ISSUE #392: Remove "side menu" from the app

FlorianSW:
Fixes #392


mueller-ma:
@FlorianSW Can you change the mail address in this commit or add it to your github account?

mueller-ma:
Did you check if other drawer entries still work as expected. I dont see why they shouldn't work, but just to be sure.

mueller-ma:
Can you remove all screenshots of that features?

digitaldan:
I can confirm sitemap selection works and the notification menu shows up if enabled.

Issue #257, opened 2016-12-25
Feature: Built-in Tasker integration


kubawolanin:
Currently openHAB Android client is a standalone app, with no connections with smartphone's sensors or environment. 
I think that combining the power of the app with Tasker plugin would revolutionize a way we interact with our home automation system. Or, at least, make it easier to integrate multiple Android apps with HABdroid (is it still accurate name?:))

Here's one tutorial on [how to create your own Tasker plugin](https://tasker.dinglisch.net/plugins.html) and a [thread on that matter](https://www.b4x.com/android/forum/threads/how-to-write-a-tasker-plugin.31136/) as well.

Best regards,
Kuba

[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=40386923)](https://www.bountysource.com/issues/40386923-feature-built-in-tasker-integration?utm_source=40386923&utm_medium=shield&utm_campaign=ISSUE_BADGE)

watou:
Agree! Along similar lines, I would like to be able to choose item names in the app which will be updated by the app based on location, motion, battery level, signal level, etc.

kubawolanin:
I created two mockups of basic plugin configuration steps.

In Tasker, after you choose the plugin and go to Configuration, you'd see a list of possible actions.
(e.g. Items Actions, Binding Actions, Things Actions... whatever our REST API allows us to use :))
I think we should focus on Items Actions at first. Baby steps.
![1](https://cloud.githubusercontent.com/assets/2270505/21472317/151cea20-cad4-11e6-8e5a-ae5014f9d31e.jpg)

And in the next screen you'd assign a command to the item you selected:
![2](https://cloud.githubusercontent.com/assets/2270505/21472318/1520c898-cad4-11e6-984d-6eafa084714c.jpg)

kubawolanin:
## openHAB Tasker integration

- [ ] Display an openHAB icon in Tasker Actions / Plugin
    - See [Action Plugins](https://tasker.dinglisch.net/plugins-intro.html) in Tasker Plugin introduction
- [ ] Show a OpenHABTaskerActionActivity window (title: "Action configuration")
    - A list view of two elements: openHAB Item and command
        - Clicking Item element should open a dialog 
            with radio list of all items (you can select only one item)
            with one button: Cancel
        - Clicking Command element should open a dialog 
            with Text field and two buttons: Cancel, Save
    - Two buttons: Cancel, Done
- [ ] Command element can receive Tasker variables (like %BATT)
[...]

PR #472, opened 2017-11-27
WIP: Listen for broadcasts

mueller-ma:
This PR gives the user the ability to set up a custom broadcast
receiver that changes an item on receive.

This was designed to send Mi Band 2 button presses to the openHAB
server. This will only work if an app is installed, that send broadcasts
on button presses. Since Gadgetbridge does so, it is the default
broadcast.

I had to create a foreground service, that starts the receiver.
Otherwise it would be killed by the os. Foreground services have to show
a notification, but it shown as low priority, so it shouldnt disturb the
user.

Reopen #388 
Related #257 

@kubawolanin I dont have tasker, so can you try if it is possible to send android broadcasts? If so, you can change one item.
[mobile-foss-debug.zip](https://github.com/openhab/openhab.android/files/1507074/mobile-foss-debug.zip)

Todo:
- [ ] Test if it works with tasker
- [ ] Start after boot
- [ ] Add better scalled icon for notification
- [ ] Add dropdown to select item
- [x] Change strings to let it sound more generall and less mi band

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/ForegroundService.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/ForegroundService.java
@@ -1,0 +1,80 @@
+package org.openhab.habdroid.core;
+
+import android.app.Notification;
+import android.app.PendingIntent;
+import android.app.Service;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.content.SharedPreferences;
+import android.os.IBinder;
+import android.preference.PreferenceManager;
+import android.support.v4.app.NotificationCompat;
+import android.support.v4.content.res.ResourcesCompat;
+import android.util.Log;
+
+import org.openhab.habdroid.R;
+import org.openhab.habdroid.ui.OpenHABMainActivity;
+import org.openhab.habdroid.util.Constants;
+
+public class ForegroundService extends Service {
+    private static final String TAG = "ForegroundService";
+
+    @Override
+    public void onCreate() {
+        Log.d(TAG, "onCreate()");
+        super.onCreate();
+    }
+
+    @Override
+    public int onStartCommand(Intent intent, int flags, int startId) {
+        CustomBroadcastReceiver cbr = new CustomBroadcastReceiver();
+        if (intent.getAction().equals(Constants.ACTION.STARTFOREGROUND_ACTION)) {
+            Log.i(TAG, "Received Start Foreground Intent ");
+            String broadcast = (String) intent.getExtras().get("broadcast");
+            IntentFilter cbr_intent = new IntentFilter();
+            cbr_intent.addAction(broadcast);
+            registerReceiver(cbr, cbr_intent);
+
+            Intent notificationIntent = new Intent(this, OpenHABMainActivity.class);
+            notificationIntent.setAction(Constants.ACTION.MAIN_ACTION);
+            notificationIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
+
+            PendingIntent pendingIntent =
+                    PendingIntent.getActivity(this, 0, notificationIntent, 0);
+
+            Notification notification = new NotificationCompat.Builder(this)
+                    .setContentTitle(getString(R.string.settings_custom_broadcast_listening))
+                    .setContentText(getString(R.string.notification_last_broadcast_none))
+                    .setSmallIcon(R.drawable.icon_blank)
+                    .setPriority(Notification.PRIORITY_LOW)
+                    .setColor(ResourcesCompat.getColor(getResources(), R.color.openhab_orange, null))
+                    .setContentIntent(pendingIntent)
+                    .setOngoing(true).build();
+
+            startForeground(Constants.NOTIFICATION_ID.FOREGROUND_SERVICE, notification);
+
+        } else if (intent.getAction().equals(Constants.ACTION.STOPFOREGROUND_ACTION)) {
+            // No API call available to check if a broadcast receiver is running: https://stackoverflow.com/questions/2682043/how-to-check-if-receiver-is-registered-in-android/3568906#3568906
+            try {
+                Log.e(TAG, "stop cbr");
+                unregisterReceiver(cbr);
+            } catch (IllegalArgumentException ignored) {}
+            Log.i(TAG, "Received Stop Foreground Intent");
+            stopForeground(true);
+            stopSelf();
+        }
+        return START_STICKY;
+    }
+
+    @Override
+    public void onDestroy() {
+        super.onDestroy();
+        Log.i(TAG, "onDestroy()");
+    }
+
+    @Override
+    public IBinder onBind(Intent intent) {
+        // Used only in case of bound services.
+        return null;
+    }
+}
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
@@ -11,7 +11,6 @@ package org.openhab.habdroid.ui;
 
 import android.app.AlertDialog;
 import android.app.Dialog;
-import android.app.Notification;
 import android.app.PendingIntent;
 import android.app.ProgressDialog;
 import android.content.ActivityNotFoundException;
@@ -47,7 +46,6 @@ import android.support.v7.widget.Toolbar;
 import android.text.TextUtils;
 import android.util.Log;
 import android.view.Gravity;
-import android.view.KeyEvent;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
@@ -75,7 +73,6 @@ import org.openhab.habdroid.core.notifications.GoogleCloudMessageConnector;
 import org.openhab.habdroid.core.notifications.NotificationSettings;
 import org.openhab.habdroid.model.OpenHABLinkedPage;
 import org.openhab.habdroid.model.OpenHABSitemap;
-import org.openhab.habdroid.model.thing.ThingType;
 import org.openhab.habdroid.ui.drawer.OpenHABDrawerAdapter;
 import org.openhab.habdroid.ui.drawer.OpenHABDrawerItem;
 import org.openhab.habdroid.util.Constants;
@@ -110,6 +107,8 @@ import de.duenndns.ssl.MemorizingResponder;
 import de.duenndns.ssl.MemorizingTrustManager;
 import okhttp3.Call;
 import okhttp3.Headers;
+
+import static org.openhab.habdroid.core.CustomBroadcastReceiver.CUSTOM_BROADCAST_RECEIVER_INTENT;
 
 public class OpenHABMainActivity extends AppCompatActivity implements OnWidgetSelectedListener,
         OpenHABTrackerReceiver, MemorizingResponder {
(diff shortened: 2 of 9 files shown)


kubawolanin:
Hey @mueller-ma!

Thank you so much for making it! I believe it will be a killer feature.

I couldn't figure out how to send broadcast from Tasker, though. I found some docs on [Intents](http://tasker.dinglisch.net/userguide/en/intents.html). Is the Android broadcast the same thing?
Sorry, my knowledge is limited here.

Here's some feedback on the overall experience with the feature:

1. It would be nice to have a list of all items (from `/rest/items`) to choose from. Typing it by hand is error prone and it would add a lot to the good UX :-)
1. Do you plan to extend the functionality to handle multiple broadcasts? It would be great, but then it would also require a separate view to store them.

Cheers

ThomDietrich:
Hey guys!

Just wanted to agree, great initiative @mueller-ma !!
I don't know about the Mi band button but a Tasker integration would be amazing! Sadly I can not bring any experience with tasker-related android development to the table.

rkoshak:
> I couldn't figure out how to send broadcast from Tasker, though. I found some docs on Intents. Is the Android broadcast the same thing?

I am about 90% certain that an Intent is what gets broadcast so to the Broadcast Listener. So if we have the proper format to form the Intent we should be able to broadcast to openHAB app. 

For example, to tell PocketCasts to start playing I send an intent with the Action

`au.com.shiftjelly.pocketcasts.action.PLAY`

and it is directed towards the Broadcast Receiver.

Similarly, to start Waze on my phone to automatically navigate to home I send an Intent with action

`android.intent.action.VIEW`

and Data

`waze://?ll=12.3456,-123.4567&navigate=yes`

and the Target is the Activity.

So we will need documentation for how to form the Intent to send to this receiver.

ThomDietrich:
@rkoshak or you add a [Plugin](http://tasker.wikidot.com/plug-ins-and-3rd-party) to tasker, teaching it specific interactions your app provides. I believe that is the goal of this Issue? http://tasker.dinglisch.net/developers.html

kubawolanin:
That's correct @ThomDietrich. 
@mueller-ma if you don't have Tasker let me show you real quick what the fuzz is all about :)

First you define some Tasks - they contain a sequence of Actions.
Tasks are then executed with Profiles (e.g. if you received a text message from certain number or something connected to phones Bluetooth etc)
screenshot_2017-11-29-19-55-41-901_net dinglisch android taskerm

Actions are the events (Intents?) like turning on the torch, airplane mode, launching an app etc. 
openHAB would be stored in 'Plugins' category.
screenshot_2017-11-29-19-56-01-673_net dinglisch android taskerm

If a certain app has a Tasker capability, it shows up in the Plugins category:
screenshot_2017-11-29-19-56-07-677_net dinglisch android taskerm

As you can see, I have the "Locale Toast Plug-In" from [this example code](http://www.twofortyfouram.com/developer/toast.zip) visible as well.

After you select a Plugin you'd like to work with, you'll be directed to plugin's configuration view which I roughly mocked-up in https://github.com/openhab/openhab.android/issues/257#issuecomment-269132020

Hope that this clears things up a bit :-)
Sorry if I'm off topic!

rkoshak:
A Tasker plugin would indeed be a major improvement. Tasker is like openHAB for your phone and I know I and others go to great lengths to integrate the two together.

However, I think this and #257 are slightly different. Having a broadcast listener is attractive because it will work with other automation programs that can send intents as well. But I don't see how it buys us much in Tasker. Intents are considerably more arcane than using the REST API with an HTTP task so I'm thinking that a plugin might be the way to go for Tasker.

So in short, I think we need both this to support other automation tools on Android and #257 to make Tasker integration easier than it currently is.

That's my two cents, spoken as someone who doesn't have the time right now to actually contribute to either, soon hopefully though, once I work off the rest of my backlog of todos.

mueller-ma:
Hey all,
Thanks for your feedback. I added some suggestions to the todo list.
I know tasker and I also know that is closed source, so dont expect too much support from me. I made this change very generall (you can configure broadcast and intent extra) so it may be used with gadgetbridge, tasker or any other app sending broadcasts. You could e.g. toggle a switch when battery runs low, you take a picture, unlock phone, etc.

FOSS tasker alternatives exists (https://github.com/biotinker/LibreTasks/issues), but development seems to be stuck.

Would be great to support multiple broadcasts, but I would need some help for that.

mueller-ma:
I cannot figure out, why it doesnt start after boot.

maniac103:
> I cannot figure out, why it doesnt start after boot.

You need to add the boot completed intent filter to a broadcast receiver, not the activity.

mueller-ma:
@maniac103 Now the mainactivity starts on boot. I think I have to move that setup function to a seperate activity?

maniac103:
Why don't you just start the service in the boot completed receiver?

maniac103:
You mean duplication between startup on boot and handling of a changed pref? You could solve that by sending a broadcast to the boot completed receiver whenever the pref changes, so it can reevaluate the service state.

(BTW, I'd unify CustomBroadcastReceiver and BootCompletedReceiver into a single one that handles all aspects of that custom broadcast handling. Might be a personal preference thing only though.)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -166,9 +167,9 @@ public void onFailure(Call call, int statusCode, Headers headers, byte[] respons
 
     // Loopj
 //    private static MyAsyncHttpClient mAsyncHttpClient;
-    private static MyAsyncHttpClient mAsyncHttpClient;
+    public static MyAsyncHttpClient mAsyncHttpClient;
     // Base URL of current openHAB connection
-    private String openHABBaseUrl = "http://demo.openhab.org:8080/";
+    public static String openHABBaseUrl = "http://demo.openhab.org:8080/";
```
maniac103: Is making everything static here really a good idea? If it was me to decide, mAsyncHttpClient shouldn't be static to begin with, let alone public static. I'd suggest giving the service its own HTTP client and OpenHABTracker instance (like OpenHABVoiceService already has), and making CustomBroadcastReceiver a private non-static member class of that service so it has access to the client.

mobile/src/main/java/org/openhab/habdroid/util/Constants.java
```diff
@@ -40,4 +40,18 @@
...
     }
-}
\ No newline at end of file
+    public static final String PREFERENCE_CUSTOM_BROADCAST    = "default_openhab_custom_broadcast_enabled";
+    public static final String PREFERENCE_CUSTOM_BROADCAST_BROADCAST = "default_openhab_custom_broadcast";
+    public static final String PREFERENCE_CUSTOM_BROADCAST_EXTRA = "default_openhab_custom_broadcast_extra";
+    public static final String PREFERENCE_CUSTOM_BROADCAST_ITEM = "default_openhab_custom_broadcast_item";
+
+    public interface ACTION {
+        public static String MAIN_ACTION = "org.openhab.habdroid.action.main";
```
maniac103: This doesn't seem to be actually used. It's set in the notification pending intent, but is never checked against.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -338,10 +341,29 @@ private void processIntent(Intent intent) {
...
+            Log.d(TAG, "start cbr");
+            Intent startIntent = new Intent(this, ForegroundService.class);
+            startIntent.setAction(Constants.ACTION.STARTFOREGROUND_ACTION);
+            startIntent.putExtra("broadcast", mSettings.getString(Constants.PREFERENCE_CUSTOM_BROADCAST_BROADCAST, ""));
+            startService(startIntent);
+        } else {
+            Log.d(TAG, "stop cbr");
+            Intent stopIntent = new Intent(this, ForegroundService.class);
+            stopIntent.setAction(Constants.ACTION.STOPFOREGROUND_ACTION);
+            startService(stopIntent);
```
maniac103: Just calling stopService() here (setting action not needed) should be sufficient. It's equivalent to calling stopSelf() in the service.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -338,10 +341,29 @@ private void processIntent(Intent intent) {
...
             }
         }
     }
 
+    private void setupCustomBroadcastReceiver() {
+        // Custom broadcast receiver
+        if (mSettings.getBoolean(Constants.PREFERENCE_CUSTOM_BROADCAST, false)) {
+            Log.d(TAG, "start cbr");
+            Intent startIntent = new Intent(this, ForegroundService.class);
+            startIntent.setAction(Constants.ACTION.STARTFOREGROUND_ACTION);
```
maniac103: Setting action isn't needed iMHO, the service can call startForeground() unconditionally in its onStartCommand().

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -338,10 +341,29 @@ private void processIntent(Intent intent) {
...
+            startIntent.putExtra("broadcast", mSettings.getString(Constants.PREFERENCE_CUSTOM_BROADCAST_BROADCAST, ""));
```
maniac103: Can't the service figure out the broadcast it should listen to by itself?

mobile/src/main/java/org/openhab/habdroid/core/ForegroundService.java
```diff
@@ -0,0 +1,80 @@
...
+                    .setSmallIcon(R.drawable.icon_blank)
+                    .setPriority(Notification.PRIORITY_LOW)
+                    .setColor(ResourcesCompat.getColor(getResources(), R.color.openhab_orange, null))
+                    .setContentIntent(pendingIntent)
+                    .setOngoing(true).build();
+
+            startForeground(Constants.NOTIFICATION_ID.FOREGROUND_SERVICE, notification);
+
+        } else if (intent.getAction().equals(Constants.ACTION.STOPFOREGROUND_ACTION)) {
+            // No API call available to check if a broadcast receiver is running: https://stackoverflow.com/questions/2682043/how-to-check-if-receiver-is-registered-in-android/3568906#3568906
```
maniac103: You can track a boolean whether it's registered though. Seems cleaner than catching the exception.

mobile/src/main/java/org/openhab/habdroid/core/ForegroundService.java
```diff
@@ -0,0 +1,80 @@
...
+import android.preference.PreferenceManager;
+import android.support.v4.app.NotificationCompat;
+import android.support.v4.content.res.ResourcesCompat;
+import android.util.Log;
+
+import org.openhab.habdroid.R;
+import org.openhab.habdroid.ui.OpenHABMainActivity;
+import org.openhab.habdroid.util.Constants;
+
+public class ForegroundService extends Service {
```
maniac103: Wouldn't this be better named `CustomBroadcastListenerService`?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
mueller-ma: My idea was to pass all information (broadcast, item to change, intent extra, openhab url) from mainactivity to foreground service and from there to CustomBroadcastListenerService. But I dont have an idea how I can get those infos in CustomBroadcastReceiver#onReceive()

mobile/src/main/java/org/openhab/habdroid/util/Constants.java
mueller-ma: Removed that

mobile/src/main/java/org/openhab/habdroid/core/ForegroundService.java
mueller-ma: I renamed it.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
mueller-ma: CustomBroadcastListenerService now reads all settings and writes them to CustomBroadcastReceiver.

mueller-ma:
Waiting for #562

mueller-ma:
Closing for now as listening for only one broadcast is a too specific use case

PR #480, opened 2017-11-29
Move the clear image cache from the options to the settings menu

### LINKED ISSUE #478: Move "clear image cache" to the settings menu
In https://github.com/openhab/openhab.android/pull/404#issuecomment-347045797 it seems we've a consensus to kill the options menu and move all the possible buttons into better places around in the app. One of the options is the "clear image cache", which should clearly be somewhere in the settings.

That's the issue for solving this.

FlorianSW:
The possibility to clear the image cache is not an option to the currently
viewed sitemap or section or whatever. It also shouldn't be necessary to
have this option in such a prominent component as the options menu.
Therefore, and for preparation to remove the options menu completely, the
menu is moved to the settings (next to the icon type selection).

Fixes #478

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
@@ -29,13 +29,11 @@ import android.view.MenuItem;
 
 import com.loopj.android.image.WebImageCache;
 
-import org.openhab.habdroid.BuildConfig;
 import org.openhab.habdroid.R;
 import org.openhab.habdroid.util.Constants;
 import org.openhab.habdroid.util.Util;
 
 import java.security.cert.X509Certificate;
-import java.text.DateFormat;
 
 /**
  * This is a class to provide preferences activity for application.
@@ -146,11 +144,8 @@ public class OpenHABPreferencesActivity extends AppCompatActivity {
             clearCachePreference.setOnPreferenceClickListener(new Preference.OnPreferenceClickListener() {
                 @Override
                 public boolean onPreferenceClick(Preference preference) {
-                    if (getActivity() == null) {
-                        return false;
-                    }
                     // Get launch intent for application
-                    Intent restartIntent = getActivity().getBaseContext().getPackageManager()
+                    Intent restartIntent = getActivity().getPackageManager()
                             .getLaunchIntentForPackage(getActivity().getBaseContext().getPackageName());
                     restartIntent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
                     // Finish current activity
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
@@ -27,7 +27,7 @@ import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.Toolbar;
 import android.view.MenuItem;
 
-import java.text.DateFormat;
+import com.loopj.android.image.WebImageCache;
 
 import org.openhab.habdroid.BuildConfig;
 import org.openhab.habdroid.R;
@@ -35,6 +35,7 @@ import org.openhab.habdroid.util.Constants;
 import org.openhab.habdroid.util.Util;
 
 import java.security.cert.X509Certificate;
+import java.text.DateFormat;
 
 /**
  * This is a class to provide preferences activity for application.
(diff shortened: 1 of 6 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -139,6 +142,27 @@ public boolean onPreferenceClick(Preference preference) {
                     return true;
                 }
             });
+
+            clearCachePreference.setOnPreferenceClickListener(new Preference.OnPreferenceClickListener() {
+                @Override
+                public boolean onPreferenceClick(Preference preference) {
+                    if (getActivity() == null) {
```
maniac103: I don't think activity being null is a condition that actually can happen here. How should a click on a detached fragment happen?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -139,6 +142,27 @@ public boolean onPreferenceClick(Preference preference) {
...
+                        return false;
+                    }
+                    // Get launch intent for application
+                    Intent restartIntent = getActivity().getBaseContext().getPackageManager()
+                            .getLaunchIntentForPackage(getActivity().getBaseContext().getPackageName());
```
maniac103: No need for getBaseContext() ... package manager stuff isn't theme dependent.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -139,6 +142,27 @@ public boolean onPreferenceClick(Preference preference) {
...
+                    restartIntent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+                    // Finish current activity
+                    getActivity().finish();
```
maniac103: Wouldn't you rather want to set an activity result here, start the settings with startActivityForResult and restart the main activity from its onActivityResult? And there also is Activity::recreate, I don't think that complicated stuff of really needed.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
FlorianSW: I'm a bit paranoid here, I'll remove it :D

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
FlorianSW: Good catch (haven't read the code, just copied it from main activity) :)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
FlorianSW: When I think about that, I'm not really sure if all this is really needed. Isn't the main activity restarted when I leave the settings menu? In this case, the cleared image cache should already be used and filled with new values? So probably just a message like "Image cache cleared" should be fine? What do you think?

mobile/src/main/java/org/openhab/habdroid/util/Constants.java
```diff
@@ -22,7 +22,7 @@
     public static final String PREFERENCE_ANIMATION         = "default_openhab_animation";
     public static final String PREFERENCE_DEMOMODE          = "default_openhab_demomode";
     public static final String PREFERENCE_FULLSCREEN        = "default_openhab_fullscreen";
-    public static final String PREFERENCE_TONE              = "default_openhab_alertringtone";
```
mueller-ma: I think you removed that by accident

FlorianSW: You're right, I already found that and fixed it in the next commit, however, I forgot to push it, sorry! :)

mobile/src/main/res/xml/preferences.xml
```diff
@@ -72,6 +72,10 @@
             android:summary="%s"
             android:entries="@array/iconTypeValues"
             android:entryValues="@array/iconTypeValues" />
+        <Preference
+            android:clickable="true"
+            android:key="default_openhab_cleacache"
+            android:title="@string/mainmenu_openhab_clearcache" />
```
mueller-ma: Maybe add a summary like "Click to clear image cache" so the user knows that it will be clear immediately. Thats not a critical action, so maybe we dont need this hint.

FlorianSW:
Yeah, I know why. The disk cache is cleared when you use clear and the in-memory cache of the current WebImageCache object. However, the method creates it's own new object of that class, the MyWebImage class still has it's own object with it's own separated in-memory cache, which is not cleared correctly. I'll fix that in this PR, too.

mobile/src/main/res/xml/preferences.xml
FlorianSW: Hmm, I think that the title is already pretty clear in what this method does :)

digitaldan:
LGTM.

PR #490, opened 2017-12-04
Add string for http code 503

mueller-ma:
Restart oh and go to http://oh/rest/items to see this code.

### DIFF
--- a/mobile/src/main/res/values-de/strings.xml
+++ b/mobile/src/main/res/values-de/strings.xml
@@ -123,7 +123,7 @@
     <string name="error_http_code_495">Ungültiges Client-Zertifikat (HTTP-Statuscode 495)</string>
     <string name="error_http_code_496">Client-Zertifikat wird gefordert (HTTP-Statuscode 496)</string>
     <string name="error_http_code_502">Ihr Reverse-Proxy bekam eine ungültige Antwort von openHAB (HTTP-Statuscode 502)</string>
-    <string name="error_http_code_503">openHAB startet gerade (HTTP-Statuscode 503)</string>
+    <string name="error_http_code_503">openHAB steht temporär nicht zur Verfügung (HTTP-Statuscode 503)</string>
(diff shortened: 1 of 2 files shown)


mobile/src/main/res/values-de/strings.xml
```diff
@@ -123,6 +123,7 @@
     <string name="error_http_code_495">Ungültiges Client-Zertifikat (HTTP-Statuscode 495)</string>
     <string name="error_http_code_496">Client-Zertifikat wird gefordert (HTTP-Statuscode 496)</string>
     <string name="error_http_code_502">Ihr Reverse-Proxy bekam eine ungültige Antwort von openHAB (HTTP-Statuscode 502)</string>
+    <string name="error_http_code_503">openHAB startet gerade (HTTP-Statuscode 503)</string>
```
FlorianSW: Hmm, that's not entirely correct. That may be one of the possible problems why the 503 http status code is sent, but it's not the only one. What happens if the user is connecting to a remote, which proxy is not available because of too high load and sents a 503 status code instead. The openHAB instance in the back may still be up and is not starting, so the string would be false.

mueller-ma: True, I'll change the strings.

mobile/src/main/res/values/strings.xml
```diff
@@ -79,6 +79,7 @@
     <string name="error_http_code_495">Invalid SSL client certificate (HTTP response code 495)</string>
     <string name="error_http_code_496">SSL client certificate is required (HTTP response code 496)</string>
     <string name="error_http_code_502">Your reverse proxy got an invalid answer from openHAB (HTTP response code 502)</string>
+    <string name="error_http_code_503">openHAB is currently unavailable (HTTP response code 503)</string>
```
mueller-ma: Does the user understand, that this is only temporary?

FlorianSW: I think so, but what do you think about:
openHAB is temporary unavailable (HTTP response code 503)
?

maniac103: *temporarily ;-)

Issue #248, opened 2016-11-11
More than one discoverable server leads to requests to other server


watou:
I have a 1.8.3 server on an RPi2 and a 2.0 nightly server on Mac.  When opening client to 1.8.3 server, I receive a few "HTTP Error: Unacceptable" popups and dismiss them.  When going to see charts on my 1.8.3 server, the 2.0 server's log reports "Illegal argument in chart: No Persistence service found."  So it appears that the client is addressing multiple servers, when obviously the user is trying to address just one server at a time.

mueller-ma:
Did you enter the URL manually in the settings?
I have entered my productive oh server, but also running a night oh and it never sends requests to the nightly.

PR #496, opened 2017-12-05
Improve discovery 

### LINKED ISSUE #8: How to setup the android client to my local demo server?
In the setting I got openHAB URL & openHAB Remote URL, which one should I put the "http://xxx:8080/openhab.app?sitemap=demo" in? What's different between these two?
##

### LINKED ISSUE #11: get openhab url via autodicovery
It would be great to get the URL of your openhab server via autodiscovery.

The app already does discover the server via mDNS, I think.
But you still have to type in the URL if you want to save this in the setting.

In case there is more than one openhab server, the app should display all discovered servers to let the user choose one.
##

### LINKED ISSUE #179: Disable demo mode if local instance is found
I heard many people being irritated by being connected to the demo server after having setup their openHAB instance and launching the app for the first time. Since the local demo and the demo server are more or less the same, they often did not directly notice that they are looking at the remote demo instead of their local setup.
Since the demo server as many regular outages as well, I think it might be a good idea to automatically disable the demo mode (at least on the very first start of the app), if a local openHAB instance is found.

@digitaldan what do you think?
##

mueller-ma:
- Demo mode is disabled by default and only enable it if no urls are
configured and bonjour fails. Fixes #179
- Save discovered url in settings. Fixes #11 and #248 
- Inform the user that the app tries to discover oh and if that failes
it informs the user that the demo mode will be enabled. Fixes #8
- Use https for demo mode
- Discover http oh, not https. I think most people just use the default
certificate. Certificate warnings have to be turned off to make this working, so this
adds no security at all.
- Only discover when no url is set.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHABTracker.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHABTracker.java
@@ -30,6 +30,9 @@ import java.util.concurrent.ExecutionException;
 
 import javax.jmdns.ServiceInfo;
 
+import static org.openhab.habdroid.util.Constants.PREFERENCE_DEMOMODE;
+import static org.openhab.habdroid.util.Constants.PREFERENCE_URL;
+
 /**
  * This class provides openHAB discovery and continuous network state tracking to
  * change openHAB connectivity URL during app use if needed
@@ -80,7 +83,7 @@ public class OpenHABTracker implements AsyncServiceResolverListener {
                 Context.CONNECTIVITY_SERVICE);
         NetworkInfo activeNetworkInfo = connectivityManager.getActiveNetworkInfo();
         if (activeNetworkInfo != null) {
-            if (settings.getBoolean(Constants.PREFERENCE_DEMOMODE, false)) {
+            if (settings.getBoolean(PREFERENCE_DEMOMODE, false)) {
                 mOpenHABUrl = mCtx.getString(R.string.openhab_demo_url);
                 Log.d(TAG, "Demo mode, url = " + mOpenHABUrl);
                 openHABTracked(mOpenHABUrl);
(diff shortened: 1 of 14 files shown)


mobile/src/main/res/values/strings.xml
```diff
@@ -151,4 +150,7 @@
     <string name="intro_whatis">A vendor and technology agnostic open source automation software for your home</string>
     <string name="intro_themes">Themes</string>
     <string name="intro_themes_description">Choose between several themes</string>
+    <string name="error_no_url_start_demo_mode">You need an openHAB server to use this app. Please go to settings to enter IP or DNS. Demo mode will be started now.</string>
+    <string name="intro_discovery">Discovery</string>
+    <string name="intro_discovery_summary">openHAB for Android tries to find the openHAB server automatically. This will not work, when authentication is enforced.</string>
```
maniac103: `Please note this will work only if authentication is not enforced by the server.`
?

mobile/src/main/res/values/strings.xml
```diff
@@ -151,4 +150,7 @@
     <string name="intro_whatis">A vendor and technology agnostic open source automation software for your home</string>
     <string name="intro_themes">Themes</string>
     <string name="intro_themes_description">Choose between several themes</string>
+    <string name="error_no_url_start_demo_mode">You need an openHAB server to use this app. Please go to settings to enter IP or DNS. Demo mode will be started now.</string>
```
maniac103: `Please configure the server IP address or host name in the server settings. For now, demo mode will be used.`
?

mueller-ma: Thanks. The strings I used were pretty bad.

mobile/src/main/java/org/openhab/habdroid/core/OpenHABTracker.java
```diff
@@ -153,24 +160,16 @@ public void onServiceResolved(ServiceInfo serviceInfo) {
...
-        mOpenHABUrl = "https://" + serviceInfo.getHostAddresses()[0] + ":" +
+        mOpenHABUrl = "http://" + serviceInfo.getHostAddresses()[0] + ":" +
                 String.valueOf(serviceInfo.getPort()) + "/";
+        PreferenceManager.getDefaultSharedPreferences(mCtx).edit().putString(PREFERENCE_URL, mOpenHABUrl).apply();
         openHABTracked(mOpenHABUrl);
     }
 
     public void onServiceResolveFailed() {
         bonjourDiscoveryFinished();
-        Log.i(TAG, "Service resolve failed, switching to remote URL");
```
mueller-ma: Not needed here, as it is only called of no remote url is set.

FlorianSW:
I was free and implemented most (all?) stuff of this PR in "my" bigger PR, which overhauls the whole connection management in this commit:
https://github.com/FlorianSW/openhab.android/commit/4e5a1827de0faeee69f570c63139093f760b3011

in PR #562. Do you think it's reasonable to close this PR and go forward with #562? :)

mueller-ma:
Closing it and I have a look at your PR.

PR #497, opened 2017-12-06
Scale images to be fully visible

lolodomo:


lolodomo:
This PR allows having images fully visible on the tablet/phone. Before this PR, for example the music album arts with their square resolution were scaled in such a way that you can't show them fully (height was to much important) when the tablet was in landscape mode.

lolodomo:
Why 95% ?

PS: it is very easy to adjust.

mueller-ma:
So you dont have to scroll at exactly that pixel where you can see the complete picture.

mueller-ma:
It doesnt have to be 95%, just an idea. When you change something, please do not force push, so everyone can see the changes after my review.

lolodomo:
Change done, the max height of an image is set to 90%. It allows seeing another row of data.

lolodomo:
I merged with the master branch to resolve the conflict introduced by the merge of my PR 481.

PR #498, opened 2017-12-07
Don't unnecessarily make variables static.

maniac103:
Having them static can cause unexpected side effects (e.g. when
recreating activities or fragments), and in most of the cases they
didn't need to be static to begin with.


digitaldan:
@maniac103 can you sign off on your commit? See https://github.com/openhab/openhab-android/blob/master/CONTRIBUTING.md , thanks!

maniac103:
@digitaldan Done. I always forget about it :-/

mueller-ma:
Can you rebase that, so the new tests will be run?

maniac103:
@mueller-ma Done.

PR #501, opened 2017-12-08
Avoid useless refresh every 30 seconds

### LINKED ISSUE #460: UI with single widget refreshes itself every 30 seconds
Hi all,

I am currently experimenting with openHAB for Android (or HABdroid, if you want) and a sitemap in which I have a Webview widget which displays a PHP page built by me hosted in one of my servers. The Webview is configured to take pretty much all of the space available in the phone screen because in this interface I am building that is the main component (the webpage serves as a MPD controller and file browser).

Because of that need, the user moves around inside the webpage, and maybe changes its address according to what he needs to do.

**Now the problem**: I have noticed HABdroid refreshes the entire UI every 30 seconds, hence "re-drawing" the Webview component which also starts over from the original address as specified in the sitemap file. This is annoying for me, because it means the user cannot "move freely" inside the Webview component and maybe change pages over minutes of browsing (he would be redirected to the original address after 30 seconds from first load).

**Please note: I don't know if this is an issue for you**, but I would like some insight as of why this behavior is implemented this way, as well as a way to disable it if possible.
I had a look at the application source code but, given that I am not very much skilled in Java programming I did not find anything useful about this.

Thank you all for your time and patience.
Have a good day.

Andrea

lolodomo:
Fixes #460

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetListFragment.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetListFragment.java
@@ -405,13 +405,13 @@ public class OpenHABWidgetListFragment extends ListFragment {
             return;
         }
 
-        // As we change the page we need to stop all videos on current page
-        // before going to the new page. This is quite dirty, but is the only
-        // way to do that...
-        openHABWidgetAdapter.stopVideoWidgets();
-        openHABWidgetAdapter.stopImageRefresh();
         // If openHAB verion = 1 get page from XML
         if (mActivity.getOpenHABVersion() == 1) {
+            // As we change the page we need to stop all videos on current page
+            // before going to the new page. This is quite dirty, but is the only
+            // way to do that...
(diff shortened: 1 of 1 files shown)


lolodomo:
To avoid any problem, please wait for the following fix to be merged in ESH and included in OH snapshots before merging it.

lolodomo:
The fix for the REST API is now merged but of course not yet in any openHAB snapshot. We still have to wait for a new ESH build before merging this PR.

lolodomo:
Kai is building a new server version including the fix.
So please include this PR in the release 2.2.

kaikreuzer:
FTR, the latest distro contains @lolodomo's runtime fixes.

lolodomo:
So please include this PR in the coming release 2.0.

mueller-ma:
How should this PR be labeled? Bug or enhancement or something else?

lolodomo:
Enhancement I think

PR #507, opened 2017-12-09
Fix discovery of number picker

### LINKED ISSUE #476: Number picker is not discoverable
After merging #442  the + and - buttons are now removed from the spinner widget, which now does not look like it can be selected.  There is no difference now between the number picker value and a non selectable value.  If you look at the official demo.openhab.org site, this is very apparent as under the widget overview we have a label of "Temperature" and then a picker to set the value.  If you drill down to any of the rooms we also have a "Temperature" label, but the value is NOT selectable, but looks identical to the clickable version under the widget overview.

boc-tothefuture:
This resolves #476 

Review request:  I am fairly new to android dev, specifically layouts.  I am not sure if the method I used to align the elements is optimal across all devices.  It lines up correctly with the selection dropdown on my device (Pixel).


maniac103:
For starters, the drawable should be a vector drawable ;-)

boc-tothefuture:
@maniac103  - I just pulled the image out of the android SDK.  I don't see any SVG version in the SDK.

maniac103:
Vector drawables are XMLs. If no such image can be found in the SDK, I'd check the android sources (android.googlesource.com under platform/frameworks/base/core/res/res/drawable)

boc-tothefuture:
@maniac103 -
  Might need your help down this rabbit hole.  Found the right file, its here:  https://android.googlesource.com/platform/frameworks/base/+/master/core/res/res/drawable/ic_btn_round_more.xml

When I take that and replace the png I had with it, I then get failures references the two drawables inside of that xml.  Those two objects are not in the same place as that one..

boc-tothefuture:
This SO post:  https://stackoverflow.com/questions/4738249/how-to-reference-the-system-select-button-image-for-custom-preference-in-android

Indicates that they are just PNGs in the platform...

maniac103:
Having then as PNGs is fine as well. Problem is just that you placed it in the resolution independent folder (drawable without suffix), while (appropriately sized) PNGs should live in drawable-[m|h|xh|xxh]dpi. Appropriately sized means their pixel size should be 'size in dp' times 1 (mdpi), 1.5 (hdpi), 2 (xhdpi), 3 (xxhdpi). When putting it in the resolution independent folder, the image will or at least might look blurry on some devices.

boc-tothefuture:
@maniac103 -
  Thanks.  I pushed additional resolutions for the images.  Looks better!

mobile/src/main/res/layout/openhabwidgetlist_setpointitem.xml
```diff
@@ -2,60 +2,45 @@
...
     android:layout_width="fill_parent"
     android:layout_height="wrap_content"
-    android:background="?android:activatedBackgroundIndicator"
-    android:paddingLeft="@dimen/widgetlist_item_left_margin"
-    android:paddingRight="@dimen/widgetlist_item_right_margin">
+    android:background="@drawable/list_selector">
 
     <include
-        android:id="@+id/setpointleftlayout"
+        android:id="@+id/textleftlayout"
```
mueller-ma: Please revert. It is still the left part of the setpoint.

mobile/src/main/res/layout/openhabwidgetlist_setpointitem.xml
```diff
@@ -2,60 +2,45 @@
...
-    <Button
-        android:id="@+id/setpointbutton_plus"
-        android:layout_width="46dip"
-        android:layout_height="wrap_content"
+    <ImageView
+        android:id="@+id/imageViewDownArrow"
+        android:layout_width="30dp"
+        android:layout_height="30dp"
+        android:scaleType="fitXY"
         android:layout_marginEnd="8dp"
```
mueller-ma: Maybe 0 here?

mobile/src/main/res/layout/openhabwidgetlist_setpointitem.xml
```diff
@@ -2,60 +2,45 @@
...
         android:layout_height="wrap_content"
-        android:layout_marginEnd="8dp"
+        android:layout_marginEnd="16dp"
         android:textAppearance="?android:attr/textAppearanceMedium"
         app:layout_constraintBottom_toBottomOf="parent"
-        app:layout_constraintEnd_toStartOf="@+id/setpointbutton_plus"
+        app:layout_constraintRight_toLeftOf="@id/imageViewDownArrow"
         app:layout_constraintTop_toTopOf="parent"
-        android:layout_marginRight="8dp" />
+        android:layout_marginRight="16dp" />
```
mueller-ma: Same here

mobile/src/main/res/layout/openhabwidgetlist_setpointitem.xml
```diff
@@ -68,5 +53,5 @@
         android:orientation="horizontal"
         app:layout_constraintLeft_toLeftOf="parent"
         app:layout_constraintRight_toRightOf="parent"
-        app:layout_constraintTop_toBottomOf="@+id/setpointleftlayout" />
+        app:layout_constraintTop_toBottomOf="@+id/textleftlayout" />
```
mueller-ma: See first comment

mobile/src/main/res/layout/openhabwidgetlist_setpointitem.xml
```diff
@@ -2,60 +2,45 @@
...
+        app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toTopOf="parent"
         android:layout_marginRight="8dp" />
 
     <TextView
         android:id="@+id/widgetvalue"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:layout_marginEnd="8dp"
+        android:layout_marginEnd="16dp"
```
mueller-ma: IMO 8 is enough

digitaldan:
I like the dropdown arrow!

boc-tothefuture:
@mueller-ma 
  I made all the changes except the marginEnd=0.  If you set marginEnd to zero then the arrows do not line up with the arrows for the selection item.  It doesn't look as good.

mueller-ma:
@boc-tothefuture Agreed. What about the hitbox?

boc-tothefuture:
Not sure how to make it bigger?  Before I had it where if you pressed anywhere on the line it would trigger. You requested I revert that to make it like the other widgets.  I have the onclicklistener on both the text element and the down arrow image.  Is there something else I can set the listener on besides the entire line?

boc-tothefuture:
@mueller-ma  - 
  I used a TouchDelegate to expand the hit area.  Let me know if this works better for you. In my resting it was pretty good.

digitaldan:
LGTM!

maniac103:
IMHO that TouchDelegate method is suboptimal. I'd rather make the ImageView larger and use fit type center instead of fitXY so the ImageView gets some padding. I think 48dp is the touch target size suggested in the guidelines.

boc-tothefuture:
@maniac103 - 
  Making the image view larger and going to center instead of fitXY makes the image the wrong size and wrong offset compared to the selection widgets:
screenshot_20171210-060414

Additionally, when I used this method rather than touch delegate there is a space between the image and the valueTextView that is dead for touches. 

Any other options?

maniac103:
If the image has the wrong size with center, the drawable size is off and needs to be adjusted. For touch target, making the number additionally clickable would help.
Can you post a screenshot with 'show layout borders' development option enabled?

boc-tothefuture:
@maniac103 - 
  I couldn't find show layout borders development option.  I have this from the layout designer:
image

That is how it looks when its fitXY and 30dp.  

I appreciate the help.

boc-tothefuture:
@maniac103 and @mueller-ma - 
  I played around for a bit and had no luck making the ImageView bigger but not scaling the corresponding image.  I just pushed a new build that still uses the touch delegate, but the bounds are not arbitrary now.  The right bound is the current bound +8dp which is right margin.  The left bound is now set to the left bound of the text area.  In theory this should then properly scale the touch area with the layout on any device.


Let me know what you think.

maniac103:
+8 pixels is not +8dp ;-)
I'll check it later today.

boc-tothefuture:
@maniac103 - Thanks for the patch!  Works great!

@mueller-ma - Back to you for merging.

digitaldan:
LGTM, thanks!

PR #514, opened 2017-12-10
Move "Select default sitemap" to Preferences

### LINKED ISSUE #479: Move "Select default sitemap" to the settings menu
In https://github.com/openhab/openhab.android/pull/404#issuecomment-347045797 it seems we've a consensus to kill the options menu and move all the possible buttons into better places around in the app. One of the options is the "select default sitemap", which should clearly be somewhere in the settings.

That's the issue for solving this.

FlorianSW:
Instead of a menu entry "Select default sitemap", there's now a setting in
the Preferences menu, which allows the user to "claer the default sitemap"
selection, if there's already a selection. Once cleared and going back to
the main activity, the user is prompted to select a new default sitemap.

Fixes #479

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/Constants.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Constants.java
@@ -10,24 +10,27 @@
 package org.openhab.habdroid.util;
 
 public class Constants {
-    public static final String PREFERENCE_SCREENTIMEROFF    = "default_openhab_screentimeroff";
-    public static final String PREFERENCE_USERNAME          = "default_openhab_username";
-    public static final String PREFERENCE_PASSWORD          = "default_openhab_password";
-    public static final String PREFERENCE_SITEMAP           = "default_openhab_sitemap";
-    public static final String PREFERENCE_SSLCERT           = "default_openhab_sslcert";
-    public static final String PREFERENCE_SSLHOST           = "default_openhab_sslhost";
-    public static final String PREFERENCE_ALTURL            = "default_openhab_alturl";
-    public static final String PREFERENCE_URL               = "default_openhab_url";
-    public static final String PREFERENCE_THEME             = "default_openhab_theme";
-    public static final String PREFERENCE_ANIMATION         = "default_openhab_animation";
-    public static final String PREFERENCE_DEMOMODE          = "default_openhab_demomode";
-    public static final String PREFERENCE_FULLSCREEN        = "default_openhab_fullscreen";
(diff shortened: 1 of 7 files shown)


mobile/src/main/res/values-de/strings.xml
```diff
@@ -136,4 +136,7 @@
     <string name="set">Setzen</string>
     <string name="cancel">Abbrechen</string>
     <string name="dlg_notification_title">Benachrichtigung</string>
+    <string name="settings_no_default_sitemap">Kein Standard-Sitemap festgelegt</string>
```
mueller-ma: Keine?

FlorianSW: Tatsache :)

Let me change that!

digitaldan:
@FlorianSW is this for 2.0 or 2.0.1?

FlorianSW:
2.0 please :)

mobile/src/main/java/org/openhab/habdroid/util/Constants.java
```diff
@@ -10,27 +10,30 @@
...
-    public static final String PREFERENCE_DEMOMODE            = "default_openhab_demomode";
-    public static final String PREFERENCE_FULLSCREEN          = "default_openhab_fullscreen";
-    public static final String PREFERENCE_TONE                = "default_openhab_alertringtone";
-    public static final String PREFERENCE_CLEAR_CACHE         = "default_openhab_cleacache";
-    public static final String PREFERENCE_SSLCLIENTCERT       = "default_openhab_sslclientcert";
-    public static final String PREFERENCE_SSLCLIENTCERT_HOWTO = "default_openhab_sslclientcert_howto";
-    public static final String PREFERENCE_DEBUG_MESSAGES      = "default_openhab_debug_messages";
-    public static final String DEFAULT_GCM_SENDER_ID          = "737820980945";
-    public static final String PREFERENCE_COMPAREABLEVERSION  = "versionAsInt";
+    public static final String PREFERENCE_SCREENTIMEROFF        = "default_openhab_screentimeroff";
```
FlorianSW: A general question into the round: Do we actually need these whitespaces to format this like a table (so that all values start at the same column)? It makes the git diffs unreadable in my opinion :P

mueller-ma: I think we dont need it. You have to increase it every time you introduce a long new variable.

digitaldan:
LGTM

digitaldan:
I'm going to merge this in just a bit after I get travis deploys working

PR #516, opened 2017-12-10
Remove write NFC tag for sitemap URLs only

### LINKED ISSUE #515: Decide what to do with NFC "Link to Sitemap page"
From what I read from the code, the NFC option at the options menu at the top right has the following functionality:
* Once the user taps this option, the NFC tag write activity is started
* the parameter sitemapPage is passed to that activity, which contains the url of the curretnly displayed sitemap page (I assume e.g. a sub-category or something like that)
* The user could write a NFC tag now, which, when recognized by the system, will open the app and the specific sitemap page

I don't use this feature, so probably I missed something when reading the code, or is this functionaliyt correct?

If so, this is the only menu entry left in the options menu and, to be honest, I really doubt, that someone really uses this feature. What would be the use case? What is the benefit over having an nfc tag, which directly controls an item (like you can create when long-pressing an item in the app)?

With this Issue I would like to ask into the round: Do we really need such a functionality or can we simply kill it?

FlorianSW:
The navigation with nfc tags to a speicific sitemap page only, without
taking any action, seems pointless and is most likely not really used
everywhere. There's also no valid use case, I can think of, which means,
we shouldn't keep this option.

Fixes #515


digitaldan:
What version do we want this in, I would vote 2.0 since we have done so much work with menus and it would seem odd to leave this in, by itself in a menu, for only a short time.

@FlorianSW can you fix the conflicts?

FlorianSW:
2.0 would be great! Resolving conflicts now :)

digitaldan:
Sorry, one more conflict to fix :-(

FlorianSW:
Done! :)

PR #517, opened 2017-12-11
Redesign write NFC tag activity

FlorianSW:
Instead of just a full-size png of the NFC logo and a little explaining
text, the activity now:

 * shares the same layout as the other activities, with a toolbar, e.g.
 * no big image is shown anymore (the image isn't visible at all)
 * The current tip for the user (turn on nfc or touch the nfc tag) is now
   at the top of the content fragment and has a little info icon, divided
   by a horizontal line from the rest of the content
 * If NFC is turned off, a button to the NFC settings (api >= 16) or to
   the wireless settings (api < 16) is shown
 * If NFC is enabled, a little progress bar and a text indicates, that the
   app is waiting for the tag to be touched


FlorianSW:
How it now looks like:

screenshot_20171211-012000
screenshot_20171211-011936

maniac103:
I like the current large image better :-/ Especially the divider line feels a bit out of place to me, due to the huge amount of white space below. The button or the progress hint could be shown below the large image as well (but better with the spinner to the left of the text).

mueller-ma:
I dont really like the large nfc picture. It feels like "We have some space here, why not display a picture?"

FlorianSW:
I don't like it, too, that's why I've proposed to change that :) It's something, which (sorry) feels like "we don't know how we can do it better, so let's put a picture there".

maniac103:
The problem is that a huge amount of white space isn't necessarily better that a huge picture either. From my POV, it looks unfinished/unpolished. Large pictures is a commonly used pattern for activities that don't need a lot of text. Try e.g. opening Google (or AOSP) dialer with an empty call log or after revoking permissions for it. The result will look like this:

screenshot_20171212-162531

IMHO such UI would also be suitable here.

FlorianSW:
Hmm, having this for the "NFC is not supported" or "NFC is not activated" would be better, yes :) However, a watermark (a light grey picture) would be better instead of the NFC image we currently have (which is simply "too much" I think :P). We could probably use the material icon for NFC (https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/material-design-icons/2.1.3/device/drawable-hdpi/ic_nfc_black_24dp.png) in a similar layout as the above "history" icon?

FlorianSW:
Ok, that's how it actually (after pushing my changes) would look like (inspired by the example of the contacts app above). I'm not sure, if the progressbar is really needed here :/

screenshot_20171213-001543
screenshot_20171213-001559

maniac103:
IMHO the progress bar is fine, but I wonder whether it would look better when making it larger and pushing it a bit further down (ideally center it between icon and bottom of screen).
The 'activate' button needs a proper style (ButtonBar style, I think).

mobile/src/main/res/drawable/ic_nfc_black_24dp.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+        android:width="24dp"
+        android:height="24dp"
```
maniac103: Why not make this 180dp (don't forget to rename the file to omit the size suffix ;-) ) and make the ImageView wrap_content? I think this would be better for Android < 5.0, which uses PNGs Android Studio generates from this image.

mueller-ma:
I think just a colored "activate" looks better than a button.

maniac103:
Yes, but the ButtonBar style doesn't make it look like a button - it's the style used in the dialer screenshot above.

mobile/src/main/res/drawable/ic_nfc_black_24dp.xml
FlorianSW: I haven't thought about that, changed! :)

FlorianSW:
> IMHO the progress bar is fine, but I wonder whether it would look better when making it larger and pushing it a bit further down (ideally center it between icon and bottom of screen).

I moved it more to the bottom. It's not centered directly, but should take space based on the screen size (used layout_weight to specify the size occupied by the layout).

For the progress bar: As it is "just" an indication for the user, that the app is waiting for some input, I used the small progressbar style, as it isn't the main widget of the fragment. The bigger (normal sized) progress bar looks too massive in my opinion :/

mobile/src/main/res/layout/fragment_writenfc.xml
```diff
@@ -0,0 +1,64 @@
...
+              android:gravity="center_horizontal|center_vertical"
+              android:orientation="vertical">
+
+    <LinearLayout
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:orientation="vertical"
+        android:gravity="center_horizontal|center_vertical"
+        android:layout_weight="6">
+        <ImageView
```
mueller-ma: This image is just decoration, so please add `tools:ignore="ContentDescription"` to avoid lint errors.

FlorianSW: Will change that in the next commit :)

FlorianSW:
Did I missed something to respond to that blocks this PR?

mueller-ma:
I just didnt had time for reviewing this pr. Some of your PRs arent signed, can you check that?

FlorianSW:
Do I really need to sign _all_ commits of a pull request (sorry, I don't see any reason to do that :()? That could be a lot for some PRs.

mueller-ma:
You also ask me to sign your commit, when I merge them.

FlorianSW:
Rebasing my 7 commits (where 2 are merges from upstream) is pretty difficult and would rewrite commits I haven't touched/changed. I'm not sure, how this will behave for the history especially when this PR will be merged. I amended the last commit and signed it off, is that ok in this case? :/

mueller-ma:
That's ok. I dont have a device with nfc, so I cannot really test it, but the "Your device has no nfc" screen looks good.

digitaldan:
@FlorianSW I have not tested but the changes lgtm.

mueller-ma:
Triggering travis

PR #525, opened 2017-12-12
Reverting automatic pushes for now

digitaldan:


mueller-ma:
@digitaldan Is there a reason why you disabled automatic pushes?

PR #542, opened 2017-12-19
Fix icon for rollershutter

mueller-ma:
See https://community.openhab.org/t/openhab-android-habdriod-december-2017-version-rollershutter-status-not-shown-correctly-anymore/37436


mueller-ma:
@FlorianSW Can you look at the commit I pushed. I think it was wrong before.

FlorianSW:
So, just to be clear: Whenever the widgetType or the groupType of the widgetItem is "Rollershutter", the else if condition should be skipped (given, that the type is Switch and the item has no mappings), correct? If so: +1!

mueller-ma:
Yes, thats what I want to achieve.

PR #547, opened 2017-12-20
Improve support for rtl languages

### LINKED ISSUE #533: Issues with right to left languages
This isnt an issue at all as long there is not translation in a rtl language.

Overlapping label and value:
screenshot_ - - - - -
Text goes in wrong direction (should be on the left side of the icons)
screenshot_ - - - - -
Not sure if the swipe gesture is in the correct direction:
screenshot_ - - - - -

mueller-ma:
* Copied all layout files to layout-v17 and change left/right to start/end
* Removed some useless v21 files
* Added "supports rtl" to manifest

Fix #533

The frame seperator is now only shown once on a page. I didnt find a way how to show it with a constraint layout. In the old relative layout the frame label was always shown on the left side.


maniac103:
You don't need separate v17 files. You can simply add the start/end tags to the existing files and they'll take precedence on API 17+.

mueller-ma:
Does api 17+ ignore things with left and right, when start and end is present?

maniac103:
Yes. That's what I meant by 'will take precedence': if both start and left are present, start will be used; same for end/right.

mobile/src/main/res/layout/openhabdrawer_menuwithcount.xml
```diff
@@ -19,22 +21,23 @@
...
     <LinearLayout
         android:layout_width="match_parent"
         android:layout_height="match_parent"
         android:layout_toRightOf="@+id/itemimage"
         android:layout_marginLeft="20dp"
         android:layout_centerVertical="true"
         android:orientation="horizontal"
-        android:gravity="left|center_vertical"
-        >
+        android:gravity="start|left|center_vertical"
```
mueller-ma: @maniac103 Does start and left combined also work here?

maniac103: Yes.

mueller-ma:
@gargamelonly @aviramr Can you test this apk?
[mobile-full-debug.zip](https://github.com/openhab/openhab.android/files/1586066/mobile-full-debug.zip)

gargamelonly:
Tested the apk on the S8. 
For me the icons and text appear in the correct direction. Although the text is in English it looks strange as it is aligned to the RTL view. Is there an option to set the app language as an optional setting?

mueller-ma:
@Aviramr finished the hebrew translation, so no english text is shown after we merge his PR.

Issue #538, opened 2017-12-19
Add NFC info screen


ThomDietrich:
The long-press action to write an action to an NFC tag comes as a surprise to new users.

I'd suggest to add a menu entry to the hamburger menu (e.g. "Write NFC tag") which simply described this function in a few words.

ThomDietrich:
Sounds like you are on the right track!

Feel free to close this Issue in favor of an issue with the proposal by yourself.

mueller-ma:
> but move the functionality there because

IMO it's better to have the long press action, because the sitemap is a structured and well-known representation of all items.
I have some improvements in mind: Add a new uri for item state changes (e.g. `openhab-icon://?i=item&s=state`). This way we can use two different receivers for nfc tags. The current one to open a sitemap and a worker to send the item updates. The current implementation for is opening the app, sending the update and closing the app, which is a quite bad ux IMO.
Mostly unrelated to nfc, but to the long press menu: Having the ability to share a chart or image. The share action should be located in the long press menu, so we have to add headers to this menu.

mueller-ma:
An info screen can be added to read tags, clear tags and inform the user how to write new tags.

mueller-ma:
> IMO it's better to have the long press action, because the sitemap is a structured and well-known representation of all items.

I didn't think at all the HABPanel-only users or that it might be useful to assign items that aren't used in a sitemap, e.g. assign an item that triggers a rule to do more advanced stuff.
I want to keep the possibility to write tags at the sitemap, but also add the possibility to get a full list of all items.

mueller-ma:
I'm working on this.

PR #556, opened 2017-12-25
Decouple sitemap load/selection from UI

FlorianSW:
In preparation for #538, it would be easier to have a central code class,
which handles the load of the possible sitemaps without any UI
interaction. This commit changes the current sitemap selection to use such
a class. The UI interaction (select sitemap dialog, e.g.) is still in the
UI class OpenHABMainActivity, the load of the data was moved to the
SitemapList class, which handles the load of the sitemap list.


mobile/src/main/java/org/openhab/habdroid/core/sitemap/SitemapList.java
```diff
@@ -0,0 +1,230 @@
...
+
+    public ArrayList<OpenHABSitemap> asList() {
+        return new ArrayList<>(sitemaps);
+    }
+
+    public void showSelectSitemapDialog(final Context context, final SitemapSelectCallback cb) {
+        Log.d(TAG, "Opening sitemap selection dialog");
+        final List<String> sitemapNameList = new ArrayList<>();
+        for (int i = 0; i < sitemaps.size(); i++) {
+            sitemapNameList.add(sitemaps.get(i).getName());
```
mueller-ma: getLabel() here

mueller-ma: Please look at #597

mueller-ma:
@FlorianSW I remember I reviewed it and some feature from a newer PR was missing, but I cannot see my comment here. I also don't find that missing feature anymore. Strange

mueller-ma:
Found it: #597 is missing here

maniac103:
This class doesn't exactly do what's written on the box, though: as it handles showing the selection dialog, it still includes UI pieces.

I'm not sure whether that decoupling really is that useful, given that fetching the sitemap list doesn't include much logic. IMHO it's more worthwhile cleaning the existing code, e.g. like in this commit: https://github.com/maniac103/openhab.android/commit/9b37f162d1f2fe71f621298ec089d6a067676c77

PR #557, opened 2017-12-25
Extend dialogs in debug mode

mueller-ma:
In debug mode url, username, password and stacktrace will be shown, when connections errors occure.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
@@ -105,6 +105,11 @@ import de.duenndns.ssl.MemorizingTrustManager;
 import okhttp3.Call;
 import okhttp3.Headers;
 
+import static org.openhab.habdroid.util.Constants.MESSAGES.DIALOG;
+import static org.openhab.habdroid.util.Constants.MESSAGES.LOGLEVEL.ALWAYS;
+import static org.openhab.habdroid.util.Constants.MESSAGES.LOGLEVEL.DEBUG;
+import static org.openhab.habdroid.util.Constants.MESSAGES.LOGLEVEL.NO_DEBUG;
+
 public class OpenHABMainActivity extends AppCompatActivity implements OnWidgetSelectedListener,
         OpenHABTrackerReceiver, MemorizingResponder {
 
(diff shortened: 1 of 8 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -115,36 +120,41 @@ public void onFailure(Call call, int statusCode, Headers headers, byte[] respons
...
+                message = getString(R.string.error_connection_failed);
             } else {
                 Log.e(TAG, error.getClass().toString());
-                showMessageToUser(error.getMessage(), Constants.MESSAGES.DIALOG, Constants.MESSAGES.LOGLEVEL.ALWAYS);
+                message = error.getMessage();
             }
+            showMessageToUser(message, DIALOG, NO_DEBUG);
+            message += "\nURL: " + openHABBaseUrl + "\nUsername: " + openHABUsername + "\nPassword: " + openHABPassword;
+            message += "\nStacktrace:\n" + Log.getStackTraceString(error);
+            showMessageToUser(message,DIALOG, DEBUG);
```
FlorianSW: missing whitespaces :)

mueller-ma: @FlorianSW Any other comments on that?

mueller-ma:
I found the issue with the two messages. Should be gone now.
I tested it with 401 and I can see a stacktrace there.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -629,10 +629,9 @@ public void showMessageToUser(String message, int messageType, int logLevel) {
         String localUrl = mSettings.getString(Constants.PREFERENCE_URL, "");
 
         // if debug mode is enabled, show all messages, except those with logLevel 4
-        if(debugEnabled) {
-            if (logLevel == Constants.MESSAGES.LOGLEVEL.NO_DEBUG) {
-                return;
-            }
+        if((debugEnabled && logLevel == Constants.MESSAGES.LOGLEVEL.NO_DEBUG) ||
+                (!debugEnabled && logLevel == Constants.MESSAGES.LOGLEVEL.DEBUG)) {
+            return;
```
FlorianSW: btw.: Having a separate else branch here is not neccessary, as the return will already end this method :)

PR #558, opened 2017-12-25
Extend OpenHAB2WidgetTest

mueller-ma:
I have a few questions about my PR:

1. Should simple getter and setter be tested?
2. Should I create a test case for each object I create or is it better to have a seperate test case for each assertEquals?
3. Why does the test testWithObject2 not fail? I set different colors for label, icon and value, but the 3 getter return all 0.
4. Should I test for the IndexOutOfBoundsException in testWithObject2?

### DIFF
--- a/mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
+++ b/mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
@@ -3,6 +3,7 @@ package org.openhab.habdroid.model;
 import net.bytebuddy.implementation.bytecode.Throw;
 
 import org.json.JSONObject;
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.runners.MockitoJUnitRunner;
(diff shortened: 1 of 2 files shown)


FlorianSW:
> Should simple getter and setter be tested?

-> Yes! What you "really" test here is, that the JSON property is returned when you use the getter :) Setters itself can be tested (should be pretty easy), and I would do it, but I can understand discussions how useful this is. However: If you expect a specific function/method to do something, why not have a test that proves that?

> Should I create a test case for each object I create or is it better to have a seperate test case for each assertEquals?

I would have a setup method (Before Annotation) that setups two member variables with the both available test objects (something like widgetSVGWithLinkedPage and widgetPNGWithoutLinkedPage) and write one test function for each get method you use. want to test (against the objects under test in the member variables). This would make it more clear, what the test function covers and do not cover :)

> Why does the test testWithObject2 not fail? I set different colors for label, icon and value, but the 3 getter return all 0.

Executing `Color.parseColor("blue")` on an Android device returns "-16776961" (as int), while doing the exact same call in a JVM out of Android Studio), where the unit test is running) returns 0 :/

> Should I test for the IndexOutOfBoundsException in testWithObject2?

See my above comment: If it is an expected behaviour that this should throw an indexoutofboundexception with this data set: Why shouldn't there a test that proves that functionality is wokring :)

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
```diff
@@ -1,67 +1,404 @@
...
+
 import static junit.framework.Assert.assertEquals;
 import static junit.framework.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 
 @RunWith(MockitoJUnitRunner.class)
 public class OpenHAB2WidgetTest {
-    OpenHABWidget rootWidget = mock(OpenHAB2Widget.class);
+    private OpenHABWidget rootWidget = mock(OpenHAB2Widget.class);
+    OpenHABWidget sut1;
```
FlorianSW: I like descriptive names, even for tests :P

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
```diff
@@ -1,67 +1,404 @@
...
     @Test
-    public void getIconPath_iconExists_returnIconUrlToIconServlet() throws Exception {
-        OpenHABWidget sut = OpenHAB2Widget.createOpenHABWidgetFromJson(rootWidget, createJSONObject(), "PNG");
-        assertEquals("icon/groupicon?state=OFF&format=PNG", sut.getIconPath());
+    @Before
+    public void testCreateOpenHABWidgetFromJson_createsOpenHAB2Widget() throws Exception {
+        sut1 = OpenHAB2Widget.createOpenHABWidgetFromJson(rootWidget, createJSONObject(1), "PNG");
+        sut2 = OpenHAB2Widget.createOpenHABWidgetFromJson(rootWidget, createJSONObject(2), "SVG");
+        sut3 = OpenHAB2Widget.createOpenHABWidgetFromJson(rootWidget, createJSONObject(3), "SVG");
+        assertTrue(sut1 instanceof OpenHAB2Widget);
```
FlorianSW: I personally dislike asserts in before(class)/after(class), as this is setup code, that will possibly run multiple times. I would create separate tests for them :)

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
```diff
@@ -1,67 +1,404 @@
...
-                "    },\n" +
-                "    \"widgets\": [{ \"widgetId\": \"demo11\" }]\n" +
-                "  }";
+    public void testGetIconPath() throws Exception {
+        assertEquals("icon/groupicon?state=OFF&format=PNG", sut1.getIconPath());
+        assertEquals("icon/groupicon?state=ON&format=SVG", sut2.getIconPath());
+        assertEquals("icon/slider?state=81&format=SVG", sut3.getChildren().get(0).getIconPath());
+
+        // Rollersutter icon should always be 0 to 100, not ON/OFF
+        //assertEquals("icon/rollershutter?state=0&format=SVG", sut3.getChildren().get(1).getIconPath());
```
FlorianSW: why is this line commented?

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
```diff
@@ -1,67 +1,404 @@
...
+        assertEquals("demo11", sut1.getChildren().get(0).getId());
+
+        // sut2 doesn't have children
+        IndexOutOfBoundsException e = null;
+        try {
+            sut2.getChildren().get(0).getId();
+        } catch (IndexOutOfBoundsException ex) {
+            e = ex;
+        }
+        assertTrue(e != null);
```
FlorianSW: assertNotNull
However, what do you think about having this as a separate test (testGetChildrenNoChildren or whatever) and use the test annotations expected:
@Test(expected = IndexOutOfBoundsException.class)

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
```diff
@@ -1,67 +1,404 @@
...
+    public void testHasChildren() throws Exception {
+        assertEquals(true, sut1.hasChildren());
+        assertEquals(false, sut2.hasChildren());
+        assertEquals(true, sut3.hasChildren());
+    }
+
+    @Test
+    public void testGetPeriode() throws Exception {
+        assertEquals("M", sut1.getPeriod());
+        assertEquals("D", sut2.getPeriod()); // Object has no period, should default to D
```
FlorianSW: comment to message?

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
```diff
@@ -1,67 +1,404 @@
...
-                "      \"leaf\": \"false\"\n" +
-                "    },\n" +
-                "    \"widgets\": [{ \"widgetId\": \"demo11\" }]\n" +
-                "  }";
+    public void testGetIconPath() throws Exception {
+        assertEquals("icon/groupicon?state=OFF&format=PNG", sut1.getIconPath());
+        assertEquals("icon/groupicon?state=ON&format=SVG", sut2.getIconPath());
+        assertEquals("icon/slider?state=81&format=SVG", sut3.getChildren().get(0).getIconPath());
+
+        // Rollersutter icon should always be 0 to 100, not ON/OFF
```
FlorianSW: If you've a comment for the assert, I would put this comment into the message, so when this assert fails, you get this message and probably know better why the test failed or what was expected instead :)

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
```diff
@@ -1,67 +1,404 @@
...
+    @Test
+    public void testGetStep() throws Exception {
+        assertEquals(1F, sut1.getStep());
+        assertEquals(0.1F, sut2.getStep());
+    }
+
+    @Test
+    public void testGetRefresh() throws Exception {
+        assertEquals(1000, sut1.getRefresh());
+        assertEquals(100, sut2.getRefresh()); // Min refresh is 100, object has set refresh to 10
```
FlorianSW: comment to message?

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
```diff
@@ -1,67 +1,404 @@
...
+    @Test
+    public void testChildrenHasNonlinkedPages() throws Exception {
+        assertEquals(true, sut1.childrenHasNonlinkedPages());
+        assertEquals(false, sut2.childrenHasNonlinkedPages());
+
+    }
+
+    @Test
+    public void testGetColors() throws Exception {
+        // Doesn't seem to work via Android Studio
```
FlorianSW: add:
```
import android.graphics.Color;
import static org.hamcrest.CoreMatchers.equalTo;
// ...
Assume.assumeThat(Color.parseColor("blue"), equalTo(Color.BLUE));
```

To the beginning of the task and the comment above. This will mark the test as skipped, so a build does not fail, but we know, that there's something not working :)

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
```diff
@@ -1,67 +1,404 @@
...
+        assertEquals("Dimmer [81 %]", sut3.getChildren().get(0).getLabel());
+    }
+
+    @Test
+    public void testGetMapping() throws Exception {
+        assertEquals("ON", sut1.getMapping(0).getCommand());
+        assertEquals("On", sut1.getMapping(0).getLabel());
+        assertEquals("abcäöüßẞèéóò\uD83D\uDE03", sut2.getMapping(0).getLabel());
+
+        IndexOutOfBoundsException e = null;
```
FlorianSW: See my comment above, where you used such kind of code. I would extract it as a separate test and use the expected parameter of the Test annotation.

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
mueller-ma: Descriptive names are better, but I cannot find one for these oh widgets. sutWithWidgetidDemoAndTypeGroupAnd... isn't good.

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
mueller-ma: Because this is currently broken and would lead to a failing travis build. When #542 I can rebase this PR and uncomment that line.

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
mueller-ma: Moved them to a new test

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
mueller-ma: How can I do that?

FlorianSW: For all assert* methods provided by JUnit (and most of ones provided by other frameworks) overload the assert* method to take a string as a first parameter, the expected result as the second and the current result as the third parameter (or only the current result, if the method itself contains the expectation, like assertTrue). That means, you can write:

````
assertEquals("Rollersutter icon should always be 0 to 100, not ON/OFF", "icon/slider?state=81&format=SVG", sut3.getChildren().get(0).getIconPath());
````
And, if the test fails, the output would be something like:
````
junit.framework.ComparisonFailure: Rollersutter icon should always be 0 to 100, not ON/OFF 
Expected :icon/slider?state=81&format=SVG
Actual   :
 <Click to see difference>


	at junit.framework.Assert.assertEquals(Assert.java:100)
	at org.openhab.habdroid.core.connection.ConnectionFactoryTest.testGetAnyConnectionWifiLocalRemote(ConnectionFactoryTest.java:157)
// ...

Process finished with exit code -1
````
So, the comment is still there in the code and also visible in the test output :)

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
FlorianSW: Ah, ok, we should not forget to update one of the PR's to change this :)

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
FlorianSW: Hmm, you're right, I'd some ideas when I read the code the first time, but forgot them :/ It's ok for now I think :)

mobile/src/test/java/org/openhab/habdroid/model/OpenHAB2WidgetTest.java
mueller-ma: Ok, I changed it. I was anyway looking for a way how to easily see which test failed (in travis)

mueller-ma:
Waiting for #542

PR #559, opened 2017-12-25
Change frame header style

mueller-ma:
The frame header was capitalized, which isnt the case in basic ui, so I
removed that property here.

The text overlaps with the shaddow above, when its longer than two
lines, so it will be cut now after two lines.

When you test it, you wont see any changes on devices running api 21+, since there is a v21 file. I dont see any reason for that v21 file, so I remove it on #547


mueller-ma:
After:
screenshot_2017-12-25-19-17-42
screenshot_2017-12-25-19-17-57
Before:
screenshot_2017-12-25-19-19-25
screenshot_2017-12-25-19-19-14

lolodomo:
Good change. I can test it if needed.

mueller-ma:
I'm not sure whether the visual change is good or not. Can I have some feedback on this?
@FlorianSW @maniac103 @digitaldan

FlorianSW:
As the headings are not capitalized in basic UI, I think this change is reasonable :)

PR #561, opened 2017-12-26
Make toast default

### LINKED ISSUE #548: crash on voiceCommand from widget
after a few years using OH, i finally decided to make a voiceControl item & some rules. the WAF on this one is high !! thank you for such a cool feature.

on my wife's phone it works flawlessly. on mine, when activated from the widget it crashes every time (after sending the command to OH). it does not crash when voice is launched from inside the app.

i turned on debug & grabbed [logs](https://www.dropbox.com/s/stvtmw3o7e0fjov/habDroid.debug-crash-log-171220.txt?dl=0)... but im not sure what it's telling me. help plz.

mueller-ma:
Fix #548 
Instead of crashing, because snackbar is not implemented here, fall back to toast. Snackbars are not available here, since they must be attached to a parent layout, which doesnt exist, when the user is on the home screen and uses the voice widget.


mueller-ma:
Fix #548

PR #568, opened 2017-12-28
Fix some lint errors

mueller-ma:
They are all small changes, so I put them in one PR instead of having 12.


mueller-ma:
@maniac103 Do you want to have this as part of #675?

maniac103:
@mueller-ma As long as it doesn't grow larger, it's probably fine to merge it now. If there's more, I'd prefer to see it postponed.

PR #570, opened 2017-12-29
Remove unused permission from the app

### LINKED ISSUE #244: Asking for too many permissions, looking scary
Love this app - tx for the great work! One small feedback:

On first install, I was reluctant due to what to appears to be too many system permissions requested:

Screenshot of openHAB Android App Permissions

I really don't want to share which apps I'm running, my browsing history, and bookmarks with you! ;-)

I guess this is a little more complicated than just a PR against https://github.com/openhab/openhab.android/blob/master/mobile/src/main/AndroidManifest.xml to remove <uses-permission android:name für READ_LOGS & GET_TASKS ? There is certainly a reason why you have them in there.. probably because of some Crash Detector you're using it? Perhaps that thing, whatever it is, could be made to use optional temporarily user granted permissions somehow, if there is a such a thing in Android now?

Re. READ_LOGS it says on https://developer.android.com/reference/android/Manifest.permission.html it's "Not for use by third-party applications, because Log entries can contain the user's private information." 

Re. GET_TASKS it says "This constant was deprecated in API level 21. No longer enforced." ..

Re. READ_HISTORY_BOOKMARKS interestingly that permission, which I assume is what would cause the screenshot above, doens't even appear in your current latest AndroidManifest.xml, so I'm not sure what's going on there.
##

FlorianSW:
It is currently not used anywhere, so we shouldn't ask for this
permission, which can provide access to private information.

Fixes #244

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -19,7 +19,6 @@
         android:protectionLevel="signature" />
 
     <uses-permission android:name="org.openhab.habdroid.gcm.permission.C2D_MESSAGE" />
-    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
 
     <uses-feature
         android:name="android.hardware.location.gps"
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -7,8 +7,6 @@
     <uses-permission android:name="android.permission.CHANGE_WIFI_MULTICAST_STATE" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
-    <uses-permission android:name="android.permission.READ_LOGS" />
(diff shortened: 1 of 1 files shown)


FlorianSW:
I would propose that for 2.0.1 (https://github.com/openhab/openhab.android/milestone/13)

mueller-ma:
Is there an easy way to check if they are used?

FlorianSW:
Just looking at the code and try to find common code that requires these permissions. E.g. the ACCESS_FINE_PERMISSION is most likely needed when using `LocationManager`, which is not used anywhere in our code. READ_LOGS and GET_TASKS where added in https://github.com/openhab/openhab-android/commit/d40354d8a28b83a1be953e95d40af6582c7aaf88 As we do not use crittercism.com anymore, this is an indication, that the permission is not needed/used anymore. GET_TASKS is already deprecated and does in fact nothing, and read_logs would only give access to running logcat (I'm not sure how the log reports in Crashlytics looks like, but they do not document, that the permission is needed at all).

> Is there an easy way to check if they are used?

So, in short: A clear: yes and no :D

PR #578, opened 2017-12-30
Remove things sitemap from side menu and sitemap selection

### LINKED ISSUE #577: Sitemap _default
When selecting a sitemap the dialog always shows a sitemap called "_default". It should be displayed with a better looking name like "Things".
Or we remove it entirely, since it doesnt show up in basic ui.

mueller-ma:
Fix #577

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/Util.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Util.java
@@ -99,7 +99,7 @@ public class Util {
             try {
                 JSONObject sitemapJson = jsonArray.getJSONObject(i);
                 OpenHABSitemap openHABSitemap = new OpenHAB2Sitemap(sitemapJson);
-                if(!openHABSitemap.getName().equals("_default")) {
+                if(! (openHABSitemap.getName().equals("_default") && jsonArray.length() != 1)) {
                     sitemapList.add(openHABSitemap);
                 }
             } catch (JSONException e) {
--- a/mobile/src/main/java/org/openhab/habdroid/util/Util.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Util.java
@@ -99,7 +99,9 @@ public class Util {
             try {
                 JSONObject sitemapJson = jsonArray.getJSONObject(i);
                 OpenHABSitemap openHABSitemap = new OpenHAB2Sitemap(sitemapJson);
-                sitemapList.add(openHABSitemap);
+                if(!openHABSitemap.getName().equals("_default")) {
+                    sitemapList.add(openHABSitemap);
+                }
             } catch (JSONException e) {
                 e.printStackTrace();
             }
--- a/mobile/src/test/java/org/openhab/habdroid/util/UtilTest.java
+++ b/mobile/src/test/java/org/openhab/habdroid/util/UtilTest.java
@@ -43,9 +43,8 @@ public class UtilTest {
     }
 
     @Test
-    public void parseSitemapList() throws Exception {
-        // openHAB 1
-        List<OpenHABSitemap> sitemapList = Util.parseSitemapList(createSitemapOH1Document());
+    public void parseOH1SitemapList() throws Exception {
+        List<OpenHABSitemap> sitemapList = Util.parseSitemapList(getSitemapOH1Document());
         assertFalse(sitemapList.isEmpty());
 
         // Should be sorted
@@ -59,9 +58,11 @@ public class UtilTest {
         assertEquals("outside", sitemapList.get(7).getLabel());
 
         assertEquals(8, sitemapList.size());
+    }
 
-        // openHAB 2
-        sitemapList = Util.parseSitemapList(createJsonArray(1));
+    @Test
+    public void parseOH2SitemapList() throws Exception {
+        List<OpenHABSitemap> sitemapList = Util.parseSitemapList(createJsonArray(1));
         assertFalse(sitemapList.isEmpty());
 
         assertEquals("Main Menu", sitemapList.get(0).getLabel());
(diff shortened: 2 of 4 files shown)


mueller-ma:
When you use basic ui and have no sitemap, it just displays a message like "No sitemap available".

mueller-ma:
@kaikreuzer Remove the _default sitemap completly from the app?

kaikreuzer:
I commented here: https://github.com/openhab/openhab-android/issues/577#issuecomment-355232682

mueller-ma:
@FlorianSW How should I check for the sitemap selection not beeing here?

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -99,7 +99,9 @@ public int compare(OpenHABSitemap sitemap1, OpenHABSitemap sitemap2) {
             try {
                 JSONObject sitemapJson = jsonArray.getJSONObject(i);
                 OpenHABSitemap openHABSitemap = new OpenHAB2Sitemap(sitemapJson);
-                sitemapList.add(openHABSitemap);
+                if(! (openHABSitemap.getName().equals("_default") && jsonArray.length() != 1)) {
```
FlorianSW: This will only work, if the _default sitemap is the last sitemap in the list. Is this always the case?

FlorianSW:
> @FlorianSW How should I check for the sitemap selection not beeing here?

Just remove the ViewAssertions from the tests, or, if you really want to assert, that the sitemap selection is not visible (I would prefer that), change the viewMatcher of the selection dialog ViewInteraction by removing the isDisplayed() matcher. In the View assertion (the check method), change the assertion to be `doesNotExist()`.

The assertion commented with // does it contain "demo"? needs to be removed, though (it's pointless to assert that it isn't there, as we already checked for the selection dialog). Also, the action commented with // click on demo needs to be removed to (nothing to click on.

And that, of course, for all tests where the sitemap selection should not be visible anymore :)

mobile/src/main/java/org/openhab/habdroid/util/Util.java
mueller-ma: Does it? jsonArray's size is always the same, doesnt matter at which position _default is.

FlorianSW: Damn, I haven't read jsonArray, I read sitemapList :( Sorry!

mueller-ma:
@FlorianSW Can you review this?

mobile/src/test/java/org/openhab/habdroid/util/UtilTest.java
```diff
@@ -42,16 +44,31 @@ public void normalizeUrl() throws Exception {
...
+        // openHAB 1
+        List<OpenHABSitemap> sitemapList = Util.parseSitemapList(createSitemapOH1Document());
         assertFalse(sitemapList.isEmpty());
 
         // Should be sorted, null first
         assertEquals(null, sitemapList.get(0).getLabel());
         assertEquals("Garden", sitemapList.get(1).getLabel());
         assertEquals(8, sitemapList.size());
+
+        // openHAB 2
```
FlorianSW: can you extract that as a separate unit tests? So, you can save the comment and it's easier to find the failing test, if you see what action was done. You also do not need to have multiple setup steps in that way :)

mobile/src/test/java/org/openhab/habdroid/util/UtilTest.java
```diff
@@ -42,16 +44,31 @@ public void normalizeUrl() throws Exception {
 
     @Test
     public void parseSitemapList() throws Exception {
-        List<OpenHABSitemap> sitemapList = Util.parseSitemapList(createSitemapDocument());
+        // openHAB 1
+        List<OpenHABSitemap> sitemapList = Util.parseSitemapList(createSitemapOH1Document());
         assertFalse(sitemapList.isEmpty());
 
         // Should be sorted, null first
```
FlorianSW: Please move this comment to a message of the assertEquals and explain why the first label should be empty, please :)

mobile/src/test/java/org/openhab/habdroid/util/UtilTest.java
```diff
@@ -42,16 +44,31 @@ public void normalizeUrl() throws Exception {
...
         assertEquals(8, sitemapList.size());
+
+        // openHAB 2
+        sitemapList = Util.parseSitemapList(createJsonArray(1));
+        assertFalse(sitemapList.isEmpty());
+
+        assertEquals("Main Menu", sitemapList.get(0).getLabel());
+        assertEquals(1, sitemapList.size());
+
+        sitemapList = Util.parseSitemapList(createJsonArray(2));
```
FlorianSW: should be a separate test, too :)

mobile/src/test/java/org/openhab/habdroid/util/UtilTest.java
```diff
@@ -42,16 +44,31 @@ public void normalizeUrl() throws Exception {
...
+        assertFalse(sitemapList.isEmpty());
+
+        assertEquals("Main Menu", sitemapList.get(0).getLabel());
+        assertEquals("HOME", sitemapList.get(1).getLabel());
+        assertEquals(2, sitemapList.size());
     }
 
-    private Document createSitemapDocument() throws ParserConfigurationException, IOException, SAXException {
+    private Document createSitemapOH1Document() throws ParserConfigurationException, IOException, SAXException {
```
FlorianSW: if you rename it, I would love to rename it to "getSitemapOH1Document", it's a getter as it returns something :)

mobile/src/test/java/org/openhab/habdroid/util/UtilTest.java
```diff
@@ -71,15 +88,48 @@ private Document createSitemapDocument() throws ParserConfigurationException, IO
     public void sitemapExists() throws Exception {
         assertTrue(Util.sitemapExists(sitemapList(), "garden"));
         assertFalse(Util.sitemapExists(sitemapList(), "monkies"));
+        assertTrue(Util.sitemapExists(Util.parseSitemapList(createJsonArray(1)), "demo"));
```
FlorianSW: These lines are very hard to interpret why you sometime expect true, and sometimes false. Could you add amessage to the assert functions and explain it a bit? :)

mobile/src/test/java/org/openhab/habdroid/util/UtilTest.java
mueller-ma: I wait for #623 as sitemap labels arent null anymore.

PR #589, opened 2018-01-01
Fix typo in string name

### LINKED ISSUE #587: Typo in strings.xml key erorr_invalid_url
https://github.com/openhab/openhab.android/blob/dcbee8baa3bb0a03b5e30ec8097f1c41ac8b1d8a/mobile/src/main/res/values/strings.xml#L69

mueller-ma:
Fix #587 

@ThomDietrich Is this the correct way to change a string name?

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/widget/URLInputPreference.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/widget/URLInputPreference.java
@@ -71,7 +71,7 @@ public class URLInputPreference extends EditTextPreference implements TextWatche
                 }
             }
         }
-        mEditor.setError(mUrlIsValid ? null : mEditor.getResources().getString(R.string.erorr_invalid_url));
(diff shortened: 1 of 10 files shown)


ThomDietrich:
Hey!

I am not sure how crowdin is best able to handle such a case. I guess it is not needed to delete all translations from the translated files as those are now effectively managed by crowdin and if a key doesn't exist in the original strings file, it will also not be part of the exported translated strings files.
Anyhow, besides that crowdin offers a function called "Translation Memory" which should recognize and auto-translate the "new" string. I'd say let's merge the PR as is and see what happens...

ThomDietrich:
Looks good!
As expected, the new string is now listed as untranslated BUT the previous translation is available in the TM:

grafik

ThomDietrich:
Turns out touching the translation files was NOT a good idea 😄 

grafik

mueller-ma:
Oh, well now we know it better for the next time :)

PR #562, opened 2017-12-26
Get rid of magically passed around connection data

FlorianSW:
This is an attempt to completely rewrite how the app handles connection
information to the openHAB server it is currently connected to. Instead of
passing around the openHABbaseUrl, username and password into multiple
places such as activities, drawers, adapters and fragment(list), this
commit tries to achieve the following:

 * A single point where you can get a connection from, no matter where you
   are and what connection you may need (local, remove, cloud, ...)
 * An easy way to re-use connections provided by the new
   ConnectionFactory: Instead of needed to save the conenction again in a
   local member variable (in an activity or somewhere), just call
   getConnection() with the same arguments again. If there's already a
   cached connection and if it is still valid, you'll get the exact same
   connection back
 * Abstraction of necessary preferences: The only point where the
   preferences are accessed is the COnnectionFactory, that makes it easier
   to change what preferences are needed to construct a connection and,
   e.g., later change when what preferences are used.
 * Decouple the connection data management from the openHAB discovery
   service: The discovery service should simply put the discovered OH into
   the preferences, if it was successful, the rest is managed by the
   ConnectionFactory, which, by itself, does not try to discovery
   anything.
 * Make it easier to construct the exact connection object you need: Just
   request a Connection with Connections.REMOTE, .LOCAL, .CLOUD if you
   require a specific type of connection, or use .ANY to get any working
   connection object back, no need to use if/elses to get what you want.
 * Error handling: If your Activity uses connections from the
   ConnectionFactory and currently implements AppCompatActivity, inherit
   from ConnectionAvailbilityAwareActivity now, which will handle most of
   the connection errors that may occur there (no network, no URL set),
   otherwise, simply catch these exceptions, if they could occur for your
   requested type (.ANY only currently, all others will return null if the
   connection could not be obtained.

This commit also takes over things from PR #496:
* Demo mode is disabled by default and only enable it if no urls are
configured and bonjour fails. Fixes #179
* Save discovered url in settings. Fixes #11 and #248
* Inform the user that the app tries to discover oh and if that failes
it informs the user that the demo mode will be enabled. Fixes #8
* Use https for demo mode
* Discover http oh, not https. I think most people just use the default
certificate. Certificate warnings have to be turned off to make this
working, so this
adds no security at all.
* Only discover when no url is set.

Co-authored-by: Danny Baumann <dannybaumann@web.de>

PR #526, opened 2017-12-14
Add codecov and run android tests via travis

mueller-ma:
See this PR for an example: https://github.com/mueller-ma-bot/openhab.android/pull/1
I also tested Sputnik there, but this is not included in this PR.

@digitaldan This is just the config for codecov. To add it, go to their website and login via github.

Hopefully this will help to increase the code coverage. Current is ~6%

PR #591, opened 2018-01-01
Make instrumentation tests more independant from each other

FlorianSW:
Some tests may require some additional or specific preferences to be set
to specific values in order to work correctly.

To achieve that, this commit adds a Before method to both instrumentation
tests, which will set these required preferences before starting the
activity. It also removes the "go through intro" from the BasicWigetTest2,
which is already tested in the CheckIntroAndCancel test, so there's no
need to test it again. It is also out of scope of the test case.

### DIFF
--- a/mobile/src/androidTest/java/org/openhab/habdroid/ui/BasicWidgetTest.java
+++ b/mobile/src/androidTest/java/org/openhab/habdroid/ui/BasicWidgetTest.java
@@ -1,7 +1,9 @@
 package org.openhab.habdroid.ui;
 
 
+import android.preference.PreferenceManager;
 import android.support.constraint.ConstraintLayout;
+import android.support.test.InstrumentationRegistry;
 import android.support.test.espresso.DataInteraction;
 import android.support.test.espresso.IdlingResource;
 import android.support.test.espresso.ViewInteraction;
@@ -18,11 +20,13 @@ import org.hamcrest.Matchers;
 import org.hamcrest.TypeSafeMatcher;
 import org.hamcrest.core.IsInstanceOf;
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.openhab.habdroid.OpenHABProgressbarIdlingResource;
 import org.openhab.habdroid.R;
+import org.openhab.habdroid.util.Constants;
 
 import static android.support.test.espresso.Espresso.onData;
 import static android.support.test.espresso.Espresso.onView;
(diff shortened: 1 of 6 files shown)


mobile/src/androidTest/java/org/openhab/habdroid/ui/BasicWidgetTest.java
```diff
@@ -45,48 +49,25 @@
...
-                                                withId(R.id.bottom),
-                                                1)),
-                                3),
-                        isDisplayed()));
-        appCompatImageButton2.perform(click());
+    @Before
+    public void setup() {
+        PreferenceManager
+                .getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext())
+                .edit()
```
FlorianSW: I tried first calling clear() here. However, even if it works perfectly fine for the first test run, during the second test run, the preferences are resetted to the default values and the ones I set next are applied, however, the app behaves weird when started:
* The theme is dark, not light, even though that light is selected in the preferences
* the discovery of the openHAB server is started, even though Demo mode is enabled in the preferences
* The app says, that I should check, if openHAB is running

Opening the settings and just going back will result in the expected behaviour (light theme, demo mode, sitemap selection). I've no idea why this happens, however it does not happen when not using clear().

This may not be important for the tests in travis, as uninstalling the app before running the test again also works (and travis should always install on a clean emulator), however, I don't think that it will help us as developers :)

mobile/src/androidTest/java/org/openhab/habdroid/ui/BasicWidgetTest.java
```diff
@@ -45,48 +49,25 @@
...
-                                                withId(R.id.bottom),
-                                                1)),
-                                4),
-                        isDisplayed()));
-        appCompatButton.perform(click());
+        mActivityTestRule.launchActivity(null);
+    }
 
+    @Test
+    public void openHABMainActivityTest2() throws InterruptedException {
```
mueller-ma: The "2" was introduced by me, when I capture it via Android Studio. Can you remove it?

mueller-ma:
CheckIntroAndCancel cancels the intro, BasicWigetTest2 goes throught the intro.

mobile/src/androidTest/java/org/openhab/habdroid/ui/BasicWidgetTest.java
```diff
@@ -45,48 +49,25 @@
...
-                                3),
-                        isDisplayed()));
-        appCompatImageButton2.perform(click());
+    @Before
+    public void setup() {
+        PreferenceManager
+                .getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext())
+                .edit()
+                .putString(Constants.PREFERENCE_SITEMAP, "")
+                .putBoolean(Constants.PREFERENCE_FIRST_START, false)
```
mueller-ma: You should also enable demo mode. Its disabled by default in #562

FlorianSW: I would do it in #562 if this PR is merge before it, or otherwise add it once #562 is merged, ok? Doing that keeps both PRs clean I think :]

mobile/src/androidTest/java/org/openhab/habdroid/ui/BasicWidgetTest.java
FlorianSW: +1 :)

FlorianSW:
> CheckIntroAndCancel cancels the intro, BasicWigetTest2 goes throught the intro.

I would move both tests for the intro into their own test class then, ok?

mueller-ma:
#526 still failing

PR #593, opened 2018-01-02
Add pullapprove config

### LINKED ISSUE #585: Use pullapprove to check for signed-off-by
As openhab requires the "Signed-off-by" in the commit message of each commited commit in a pull request, and because this is the only project where I need to do that, I constantly forget to commit with the signed off by line in my commit messages.

Instead of requiring the reviewers to take a look on that and forcing me to rebase an entire history (sometimes more than 5-10 commits) just to add the Signed-off line, it would be great, if github wouldn't tell me, that my PR is finished with a green checkmark (which is an indicator for me, that i do not need to do anything except responding to comments). It turns out pullapprove has such a feature, and the openhab2-addons repository already uses it:
https://github.com/openhab/openhab2-addons/blob/master/.pullapprove.yml#L5

So this issue is to request to have such a check for openhab.android, too, just that I get a reminder directly after creating the pull request.

mueller-ma:
Fixes #585

@kaikreuzer I think you have setup pullapprove on https://github.com/openhab/openhab2-addons. Can you add this repo on https://pullapprove.com ?

### DIFF
--- a/.pullapprove.yaml
+++ b/.pullapprove.yaml
@@ -1,0 +1,34 @@
+version: 2
+
+requirements:
+  signed_off_by:
+    required: true
+
+always_pending:
+  title_regex: '(WIP|wip)'
+  explanation: 'Work in progress...'
+
+group_defaults:
+  approve_by_comment:
+    enabled: true
+    approve_regex: '^(Approved|LGTM|:shipit:|:\+1:|Engage)'
+    reject_regex: '^(Rejected|:-1:|Borg)'
+  reset_on_push:
+    enabled: true
+  reset_on_reopened:
+    enabled: true
+  author_approval:
+    ignored: true
+
+groups:
+  maintainers:
+    required: 1
+    users:
+      - kaikreuzer
+      - digitaldan
+      - ThomDietrich
+      - mueller-ma
+    conditions:
+      branches:
+        - master
+    reject_value: -100


.pullapprove.yaml
```diff
@@ -0,0 +1,34 @@
...
+    required: true
+
+always_pending:
+  title_regex: '(WIP|wip)'
+  explanation: 'Work in progress...'
+
+group_defaults:
+  approve_by_comment:
+    enabled: true
+    approve_regex: '^(Approved|LGTM|:shipit:|:\+1:|Engage)'
```
mueller-ma: I used https://github.com/pi-hole/pi-hole/blob/master/.pullapprove.yml#L13 as template. Even when nobody uses "Engage", I would never vote for removing Star Trek eastereggs :)

.pullapprove.yaml
```diff
@@ -0,0 +1,34 @@
...
+    reject_regex: '^(Rejected|:-1:|Borg)'
+  reset_on_push:
+    enabled: true
+  reset_on_reopened:
+    enabled: true
+  author_approval:
+    ignored: true
```
mueller-ma: http://docs.pullapprove.com/groups/author_approval/
I'm not sure if we want to use ignored or required. Required is IMO the best option, but we would have to tell every contributer that he should approve his own PR. Which no repo I know does.

kaikreuzer:
I seem to be logged in as openhab-bot too often these days ;-)

PullApprove is now enabled on this repo!

mueller-ma:
> code-review/pullapprove — Approval required by openhab-bot 

On every PR, not sure why.

OT:  FF has a feature called Containers, which allows you to be logged in as kaikreuzer in one container and as openhab-bot in a different.

kaikreuzer:
>>code-review/pullapprove — Approval required by openhab-bot

> On every PR, not sure why.

It seems you have missed to add that part: https://github.com/openhab/openhab2-addons/blob/master/.pullapprove.yml#L20-L23

mueller-ma:
I have added a group called "maintainers". The name shouldnt matter

kaikreuzer:
It seems as if pullapprove hasn't yet correctly sync'ed the yaml - it still uses the default settings.
I tried to re-sync it manually, but it didn't do anything.
Maybe let's simply give it some time until tomorrow...

kaikreuzer:
Hm, waiting didn't help.
In any case, I would suggest not to maintain a specific list of individuals in the pullaprove yml, but to use the group @android-maintainers instead. This way, we only have to maintain this at a single place and it would be in line with openhab2-addons.

mueller-ma:
Fixed it: https://github.com/openhab/openhab-android/commit/f121e2c237e9c3cfcc9fc8b85925eabeb2416e94

I wanted to use the android-maintainers team first, but then I cannot merge PRs like #589, because @ThomDietrich or you are not in that group.

kaikreuzer:
> Fixed it: f121e2c

Ah, nice. I was looking hard at typos in the file name before, but this little 'a' escaped me indeed!

> but then I cannot merge PRs like #589, because @ThomDietrich or you are not in that group.

... which is absolutely ok as only the verdict of the official maintainers should count. So the question is rather: Should you be allowed to self-approve your own PRs (which I would answer with "yes" in order to not constrain ourselves too much).

mueller-ma:
> Should you be allowed to self-approve your own PRs (which I would answer with "yes" in order to not constrain ourselves too much).

Ok, I can change that.

In #559 I rebased via Github, which doesnt give you the possibility to sign the commit. Should those PRs be rejected until _every_ commit is signed?

kaikreuzer:
> In #559 I rebased via Github,

Best would be to avoid it as this introduces merge commits, which we usually don't like (a proper git rebase does not do that). 
I don't think we should have any hard automatic rejection though - it should be up to the maintainer to decide (and generated commits are ok without a sign-off) and when merging the PR it should anyhow happen as a "squash", so that the resulting commit will have the sign-off line.

PR #594, opened 2018-01-02
Use github teams instead of a list of users

mueller-ma:
See https://github.com/openhab/openhab-android/pull/593#issuecomment-354740254

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -23,12 +23,11 @@ group_defaults:
 groups:
   maintainers:
     required: 1
-    users:
-      - kaikreuzer
-      - digitaldan
-      - ThomDietrich
-      - mueller-ma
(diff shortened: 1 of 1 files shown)


mueller-ma:
Can you sync the config?

kaikreuzer:
No, it seems to have an error:
```
"{'ignored': True, 'auto': True}" is an invalid value for author_approval: only one of these settings can be true 
```

mueller-ma:
Should be fixed: https://github.com/openhab/openhab-android/commit/e1ccd182de1ab520bd76d1f887e7e46c01d67199
I tested it in my fork and it sync successful.

kaikreuzer:
Yes, that looks better - and it already sync'ed it automatically.

PR #597, opened 2018-01-02
Use sitemap labels instead of name in the sitemap selection dialog

### LINKED ISSUE #595: Use sitemap labels instead of name in the sitemap selection dialog

mueller-ma:
Fix #595

The sitemap name is also shown in the settings. @FlorianSW Whats the best way to show the label there?

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
@@ -773,7 +773,7 @@ public class OpenHABMainActivity extends AppCompatActivity implements OnWidgetSe
                 // Check if we have a sitemap configured to use
                 SharedPreferences settings =
                         PreferenceManager.getDefaultSharedPreferences(OpenHABMainActivity.this);
-                String configuredSitemap = settings.getString(Constants.PREFERENCE_SITEMAP, "");
+                String configuredSitemap = settings.getString(Constants.PREFERENCE_SITEMAP_NAME, "");
                 // If we have sitemap configured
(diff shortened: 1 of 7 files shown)


FlorianSW:
You would need to save the label of the sitemap in another preference and make the "real" sitemap preference (with the name of the sitemap) hidden. Then, when the user clicks the "Clear sitemap" preference, both preferences need to be cleared.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -821,13 +821,15 @@ public void onSuccess(Call call, int statusCode, Headers headers, byte[] respons
     private void showSitemapSelectionDialog(final List<OpenHABSitemap> sitemapList) {
         Log.d(TAG, "Opening sitemap selection dialog");
         final List<String> sitemapNameList = new ArrayList<String>();
+        final List<String> sitemapLabelList = new ArrayList<String>();
         for (int i = 0; i < sitemapList.size(); i++) {
             sitemapNameList.add(sitemapList.get(i).getName());
```
FlorianSW: is it really needed to have two lists here? Wouldn't it be possible to get rid of sitemapNameList? When accessing the sitemapNameList in the dialog, you could access sitemapList, get the element at the position of "item" and execute getName on that.

If we would've API Level 24 as the minimum supported, we could even use streams and wouldn't need any additional list here (at least not as a local variable).

FlorianSW:
LGTM, however, there're merge conflicts.

mueller-ma:
Going to wait for #526 before I merge, so I can be sure it doesnt brake #526

PR #602, opened 2018-01-03
Add checkstyle and stickler ci config

### LINKED ISSUE #598: Use Stickler CI to run Checkstyle on PRs
> Checkstyle is a development tool to help programmers write Java code that adheres to a coding standard. It automates the process of checking Java code to spare humans of this boring (but important) task. This makes it ideal for projects that want to enforce a coding standard. 

from http://checkstyle.sourceforge.net/

I have configured stickler on my fork: https://github.com/mueller-ma-bot/openhab.android/pull/4
The configuration is taken from https://github.com/checkstyle/checkstyle and based on Google's java style.

mueller-ma:
Fixes #598

Repo needs to be enabled on https://stickler-ci.com/

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -26,9 +26,9 @@
     <property name="fileExtensions" value="java, properties, xml"/>
     <!-- Checks for whitespace                               -->
     <!-- See http://checkstyle.sf.net/config_whitespace.html -->
-        <module name="FileTabCharacter">
-            <property name="eachLine" value="true"/>
-        </module>
+    <module name="FileTabCharacter">
+        <property name="eachLine" value="true"/>
+    </module>
 
     <module name="TreeWalker">
         <module name="OuterTypeFilename"/>
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -1,0 +1,241 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+          "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
+
+<!--
+    Checkstyle configuration that checks the Google coding conventions from Google Java Style
+    that can be found at https://google.github.io/styleguide/javaguide.html.
+
+    Checkstyle is very configurable. Be sure to read the documentation at
+    http://checkstyle.sf.net (or in your downloaded distribution).
+
+    To completely disable a check, just comment it out or delete it from the file.
+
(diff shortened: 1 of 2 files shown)


checkstyle.xml
```diff
@@ -0,0 +1,241 @@
...
+
+<module name = "Checker">
+    <property name="charset" value="UTF-8"/>
+
+    <property name="severity" value="warning"/>
+
+    <property name="fileExtensions" value="java, properties, xml"/>
+    <!-- Checks for whitespace                               -->
+    <!-- See http://checkstyle.sf.net/config_whitespace.html -->
+        <module name="FileTabCharacter">
```
FlorianSW: why is this indented?

mueller-ma:
@kaikreuzer Can you use @openhab-bot to add this repo on https://stickler-ci.com/ ?

checkstyle.xml
mueller-ma: No idea. See https://github.com/checkstyle/checkstyle/pull/5424

FlorianSW: Ah, ok, I thought you changed it :P Didn't know that this was also indented in this way in the original :)

mueller-ma: Changed it.

mueller-ma:
We will see if it usefull or not. Especially the indention rule might be a bit spammy and doesn't care that much, because Android Studio does a lot of auto indention.

mueller-ma:
See https://github.com/openhab/openhab-android/pull/651#discussion_r163613514
That indention was done by Android Studio (I think). Should we keep that rule? WDYT?

maniac103:
Of original indention, new indention and the Stickler comment, all seem off. Shouldn't line continuation indent be 8 spaces? Old was 4, new is 'more' (12? 16? Something like that...) and Stickler wants even more, which definitely seems wrong.

mueller-ma:
The config for stickler is from the checkstyle repo and _should_ be the Google code standard.
I could try to use auto format in Android Studio, but when I merge that, all other PRs needs to be rebased and git doesn't seem to be smart enough to notice a whitespace only change. Or I just dont know the git cli parameters.

maniac103:
I would suggest **not** to use auto format, because the outcome in most cases leaves quite a bit to be desired (I recently noticed that in the OH Eclipse setup). IMHO Stickler is the right way to go, we just need to find out what's off with that indention rule.
And what do you mean by 'smart enough to notice a whitespace only change'? Git doesn't differentiate between whitespace and other changes, because it doesn't know (or care) about the contents of the checked in files. Depending on the content of the files, whitespace might be important after all (e.g. think Python).

PR #604, opened 2018-01-03
Improve notification design

### LINKED ISSUE #603: Notifications have no accent color

mueller-ma:
* Use the real message as text, not "openHAB". See for reference: http://cdn.androidbeat.com/wp-content/uploads/2016/03/bundled-notification.png
* Add accent color

Fixes #603

I havent really tested it, maybe someone can create a test notification.

### DIFF
--- a/mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
+++ b/mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
@@ -19,6 +19,7 @@ import android.net.Uri;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.support.v4.app.NotificationCompat;
+import android.support.v4.content.ContextCompat;
 import android.util.Log;
 
 import com.google.android.gms.gcm.GoogleCloudMessaging;
@@ -102,9 +103,10 @@ public class GcmIntentService extends IntentService {
         NotificationCompat.Builder mBuilder =
                 new NotificationCompat.Builder(this)
                         .setSmallIcon(R.drawable.ic_openhab_appicon_24dp)
-                        .setContentTitle("openHAB")
+                        .setContentTitle(getBaseContext().getString(R.string.app_name))
                         .setStyle(new NotificationCompat.BigTextStyle()
                                 .bigText(msg))
+                        .setColor(ContextCompat.getColor(this, R.color.openhab_orange))
                         .setAutoCancel(true)
                         .setSound(alarmSound)
                         .setContentText(msg)


mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
```diff
@@ -102,12 +103,10 @@ private void sendNotification(String msg, int notificationId) {
         NotificationCompat.Builder mBuilder =
                 new NotificationCompat.Builder(this)
                         .setSmallIcon(R.drawable.ic_openhab_appicon_24dp)
-                        .setContentTitle("openHAB")
-                        .setStyle(new NotificationCompat.BigTextStyle()
-                                .bigText(msg))
+                        .setContentTitle(msg)
+                        .setColor(ResourcesCompat.getColor(getBaseContext().getResources(), R.color.openhab_orange, null))
```
maniac103: `ContextCompat.getColor(this, R.color.openhab_orange)` should do the same thing in a shorter way.

mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
```diff
@@ -102,12 +103,10 @@ private void sendNotification(String msg, int notificationId) {
         NotificationCompat.Builder mBuilder =
                 new NotificationCompat.Builder(this)
                         .setSmallIcon(R.drawable.ic_openhab_appicon_24dp)
-                        .setContentTitle("openHAB")
-                        .setStyle(new NotificationCompat.BigTextStyle()
-                                .bigText(msg))
```
maniac103: Removing the BigTextStyle has the consequence of no longer being able to expand notifications with longer texts ... I would keep that one.

mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
```diff
@@ -19,6 +19,7 @@
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.support.v4.app.NotificationCompat;
+import android.support.v4.content.res.ResourcesCompat;
```
FlorianSW: That looks like the false import?

mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
```diff
@@ -102,12 +103,12 @@ private void sendNotification(String msg, int notificationId) {
         NotificationCompat.Builder mBuilder =
                 new NotificationCompat.Builder(this)
                         .setSmallIcon(R.drawable.ic_openhab_appicon_24dp)
-                        .setContentTitle("openHAB")
+                        .setContentTitle(msg)
```
FlorianSW: The message title will now be the same as the message content. I'm not sure, if I like that :(

mueller-ma:
When I look at some notifications from messengers, the contact's name is set as title and the message as content.
I can revert my changes and display "openHAB" as title, but I dont think thats a good idea.

maniac103:
> When I look at some notifications from messengers, the contact's name is set as title and the message as content.

Yes, but that display is generated by the messengers using MessagingStyle, which is not suitable for our use case.

FlorianSW:
I would set the title to openHAB (or even better the string `app_name`, so if that ever changes you do not need to update it at two places) and set the message as the content title.

Just some theoretical thinking: If we once have support for multiple servers in the app, we could show the common name of the server, which issued this notification. However, this is highly theoretical :P

mueller-ma:
@FlorianSW I also thought about this

maniac103:
Sorry I'm late for review comments, but: why are you calling getBaseContext()?

mueller-ma:
I haven't seen any context var in this class and wanted to avoid to pass a context.

maniac103:
`Service`, which `IntentService` derives from, already implements `Context`, so you could simply have used `getString()` without any prefix.

Issue #10, opened 2014-07-19
support multiple openhab server


CWempe:
In case you got more than one openHAB server this app is not very usefull.

I am using a separate openhab server for developing.
It is not nice to change the URL in the settings each time I want to use the other server.
##

kidalabama:
[](https://play.google.com/store/apps/details?id=treehou.se.habit)

Alpoy:
I have 2 different OH servers (one for my house and one for the cabin) would be very nice to see this feature implemented!

mueller-ma:
When every oh server is in the same network you can assign it to the same IP or dns name

FlorianSW:
I alredy thought about implementing such a functionality. However, it would probably a bit easier, once #562 is merged (I would start working on this after #562). The UI I had in mind so far was pretty similar how GMail does it with different e-mail accounts (providing a switch in the flyout menu).

However, some things we need to consider:
* The connection settings would need to be moved into a database, instead of preferences itself (this could be done before, as a preparation step, which is already quite big I think)
* We would need to try to register the device using GCM for each "remote" connection, even if it is not selected in order for notifications to work
* I would say, that discovering the OH instance to connect to is pretty difficult with more than two servers (and probably takes a long time, as timeouts would be applied to each try), so it's possible that, during development, this feature needs to go (?)

That are my thoughts so far :)

mueller-ma:
> I would say, that discovering the OH instance to connect to is pretty difficult with more than two servers

Currently if you have more one server, a random server is discorvery. Probably that one, which response faster.

FlorianSW:
Uh, I haven't thought about that, my point with "discovering" was: When you start the app, it automatically selects if it connects to the local or remote server, which is based on the current network. If you've multiple servers in your configuration, the app probably selects a matching server randomly (if multiple remote and/or local servers are configured). So, I misused the term "discovery" here, sorry :)

mueller-ma:
As a workaround you can use the stable version for one server and the beta for another one.

curlyel:
> Autoconnect to the last used instance.

Agree 👍 
If this is not the right instance, the user should choose from the options menu to connect to another openHAB instance in case she needs to switch over to some other instance. 

So we need to maintain a number of openHAB connection peers (local+remote-Url settings + the needed credentials).

I'd suggest to **assign an `instance id`** or **a `configuration slot id`** to each configuration "slot". This id clearly identifies the meant openHAB instance and can/should be used in the 

[...]

PR #608, opened 2018-01-04
Improve english strings

mueller-ma:
* Change "phone" to "device"
* Remove redundant "setting" from settings menu strings

@ThomDietrich @maniac103

### DIFF
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -72,7 +72,7 @@
     <string name="error_network_not_available">Network is not available</string>
     <string name="error_http_connection_failed">Received an invalid response from openHAB. Please check if openHAB is working correctly (HTTP response code %d)</string>
     <string name="error_no_uuid">Unable to get openHAB UUID, connection failed</string>
-    <string name="error_invalid_url">Please enter a valid URL in a \'protocol://host:port/\' form</string>
+    <string name="error_invalid_url">Please enter a valid URL in a \‘http(s)://host(:port)/\’ form</string>
     <string name="error_connection_failed">Connection to host failed</string>
     <string name="error_unable_to_resolve_hostname">Unable to resolve hostname</string>
     <string name="error_connection_sslhandshake_failed">SSL Handshake failed - maybe you need a valid client certificate</string>
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -72,7 +72,7 @@
     <string name="error_no_url">Please check that openHAB is running or configure openHAB URL or remote URL</string>
     <string name="error_empty_sitemap_list">openHAB returned empty sitemap list</string>
     <string name="error_network_not_available">Network is not available</string>
-    <string name="error_http_connection_failed">Connection unsuccessful. An unexpected response was received while trying to connect to the configured openHAB server. Please check the provided connection details, ensure the accessibility and operation of the server or contact your server administrator. (Received HTTP response code: %d)</string>
+    <string name="error_http_connection_failed">Connection unsuccessful. An unexpected response was received while trying to connect to the configured openHAB server (Received HTTP response code: %d)</string>
     <string name="error_no_uuid">Unable to get openHAB UUID, connection failed</string>
     <string name="error_invalid_url">Please enter a valid URL in a \'http(s)://host(:port)/\' form</string>
     <string name="error_connection_failed">Connection to host failed</string>
@@ -90,10 +90,10 @@
     <!-- Nginx specific codes, but since it is the recommended reverse proxy, we should check for them: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#nginx -->
     <string name="error_http_code_495">Authentication failed. Please check the provided SSL client certificate (HTTP response code 495)</string>
     <string name="error_http_code_496">SSL client certificate is required (HTTP response code 496)</string>
-    <string name="error_http_code_502">Your reverse proxy got an invalid answer from the openHAB server. Please check the operation of the server or contact your server administrator (HTTP response code 502)</string>
-    <string name="error_http_code_503">openHAB is temporarily unavailable. Please check the operation of the server or contact your server administrator(HTTP response code 503)</string>
-    <string name="error_http_code_504">Gateway Time-out. Seems that your reverse proxy is running, but couldn\'t get an answer from openHAB. Please check the operation of the server or contact your server administrator (HTTP response code 504)</string>
-    <string name="error_http_code_507">The openHAB server hasn\'t enough free space. Please check the operation of the server or contact your server administrator (HTTP response code 507)</string>
+    <string name="error_http_code_502">Your reverse proxy got an invalid answer from the openHAB server (HTTP response code 502)</string>
+    <string name="error_http_code_503">openHAB is temporarily unavailable (HTTP response code 503)</string>
+    <string name="error_http_code_504">Gateway Time-out. Seems that your reverse proxy is running, but couldn\'t get an answer from openHAB (HTTP response code 504)</string>
+    <string name="error_http_code_507">The openHAB server hasn\'t enough free space (HTTP response code 507)</string>
     <string name="error_http_code_511">Network authentication is required (HTTP response code 511)</string>
     <string name="title_activity_openhabwritetag">Write NFC tag</string>
     <string name="title_activity_openhabinfo">openHAB Information</string>
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -23,7 +23,7 @@
     <!-- App settings strings -->
     <string name="settings_connection_title">Connection</string>
     <string name="settings_display_title">Display</string>
-    <string name="settings_misc_title">Misc</string>
+    <string name="settings_misc_title">Miscellaneous</string>
     <string name="settings_openhab_url">Local server URL</string>
     <string name="settings_openhab_url_summary">The URL of the openHAB dashboard (based on hostname or IP). If configured, autodiscovery is disabled. Current setting: %s</string>
     <string name="settings_openhab_alturl">Remote server URL</string>
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -77,7 +77,7 @@
     <string name="error_invalid_url">Please enter a valid URL in a \'http(s)://host(:port)/\' form</string>
     <string name="error_connection_failed">Connection to host failed</string>
     <string name="error_unable_to_resolve_hostname">Unable to resolve hostname</string>
-    <string name="error_connection_sslhandshake_failed">SSL Handshake failed - maybe you need a valid client certificate</string>
+    <string name="error_connection_sslhandshake_failed">SSL handshake failed</string>
     <string name="error_certificate_not_trusted">SSL certificate is not trusted. Please import the root certificate or disable certificate validation during SSL handshake</string>
     <string name="error_certificate_expired">SSL certificate has been expired. Please get a new certificate or disable certificate validation during SSL handshake</string>
     <string name="error_certificate_not_valid_yet">SSL certificate is not valid yet. Please get a correct certificate or disable certificate validation during SSL handshake. Also check if the clock on your device is set correctly</string>
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -67,10 +67,12 @@
     <string name="info_conn_url">Connecting to local URL</string>
     <string name="info_conn_rem_url">Connecting to remote URL</string>
     <string name="info_discovery">Discovering openHAB. Please wait\u2026</string>
+
+    <!-- Error messages -->
     <string name="error_no_url">Please check that openHAB is running or configure openHAB URL or remote URL</string>
     <string name="error_empty_sitemap_list">openHAB returned empty sitemap list</string>
     <string name="error_network_not_available">Network is not available</string>
-    <string name="error_http_connection_failed">Received an invalid response from openHAB. Please check if openHAB is working correctly (HTTP response code %d)</string>
+    <string name="error_http_connection_failed">Connection unsuccessful. An unexpected response was received while trying to connect to the configured openHAB server. Please check the provided connection details, ensure the accessibility and operation of the server or contact your server administrator. (Received HTTP response code: %d)</string>
     <string name="error_no_uuid">Unable to get openHAB UUID, connection failed</string>
     <string name="error_invalid_url">Please enter a valid URL in a \'http(s)://host(:port)/\' form</string>
     <string name="error_connection_failed">Connection to host failed</string>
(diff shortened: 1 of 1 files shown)


mobile/src/main/res/values/strings.xml
```diff
@@ -54,9 +54,9 @@
     <string name="settings_clear_default_sitemap">Clear default sitemap</string>
     <string name="settings_no_default_sitemap">No default sitemap selected</string>
     <string name="settings_current_default_sitemap">Current default sitemap: %1$s</string>
-    <string name="settings_openhab_sslsettings">Encryption settings</string>
-    <string name="settings_openhab_alt_connection">Remote server settings</string>
-    <string name="settings_openhab_connection">Local server settings</string>
+    <string name="settings_openhab_sslsettings">Encryption</string>
+    <string name="settings_openhab_alt_connection">Remote server</string>
+    <string name="settings_openhab_connection">Local server</string>
```
ThomDietrich: "... server" doesn't seem like the best fit. After all this is not primarily about the server but about the way the server is reached. How about something similar to "Local connection" or "Local address"... Not sure what's the best solution here.

mobile/src/main/res/values/strings.xml
```diff
@@ -72,13 +72,13 @@
     <string name="error_network_not_available">Network is not available</string>
     <string name="error_http_connection_failed">Received an invalid response from openHAB. Please check if openHAB is working correctly (HTTP response code %d)</string>
```
ThomDietrich: Any idea how to improve this string? I stumbled over it a couple of times during working with translations.
The second sentence doesn't add value, also I don't like the mentioning of "openHAB" without any specifier like "server".

mobile/src/main/res/values/strings.xml
```diff
@@ -72,13 +72,13 @@
...
     <string name="error_no_uuid">Unable to get openHAB UUID, connection failed</string>
-    <string name="error_invalid_url">Please enter a valid URL in a \'protocol://host:port/\' form</string>
+    <string name="error_invalid_url">Please enter a valid URL in a \‘http(s)://host(:port)/\’ form</string>
```
ThomDietrich: Attention: You've inserted non-neutral single-quotes

ThomDietrich:
Hey @mueller-ma, thanks for taking these on ;)
There is one other issue I'd like to mention but which would probably need it's own PR. The string `about_links_list` is not a good fit to be translated, simply counted on the times I had to correct HTML related errors in translations the last couple of days...

mobile/src/main/res/values/strings.xml
mueller-ma: The quote change was suggested by Android Studio. Not sure if it's really better

mobile/src/main/res/values/strings.xml
mueller-ma: Or maybe just "local"?

mueller-ma:
The about page needs a redesign, then we can get rid of this string.

mobile/src/main/res/values/strings.xml
FlorianSW: So, "connection" is already implied by the title of the prefernces category. Duplicating it would result in something similar as "Connection Settings", where we remove the "Settings", as it's already clear based on the context.

So, I think simply saying "Local", "Remote" is something which is short and still reasonable; and it would fit into the "Encryption" category name.

ThomDietrich: Agreed 👍

mobile/src/main/res/values/strings.xml
ThomDietrich: Okay... Could you find out the reasoning for that? What are these quotes normally used for? Do they need to be escaped? I'd vote to go back to the old quotes or to unicode-escape them.

mobile/src/main/res/values/strings.xml
mueller-ma: I can remove the second sentence and add "server" to openhab.

mobile/src/main/res/values/strings.xml
mueller-ma: I reverted the change. Does double quotes `"` suit better here?

mobile/src/main/res/values/strings.xml
ThomDietrich: Sounds good. We could also hold on to the idea of the second sentence but describe possible causes better:

> An invalid response was received from the openHAB server (HTTP response code %d). Please check *this or that*.

What could actually lead to this situation?

mobile/src/main/res/values/strings.xml
ThomDietrich: Single quotes seem normal for these kinds of technical strings. Would be okay with me.

mobile/src/main/res/values/strings.xml
FlorianSW: That heavily relies on the returned http status code. There're special messages, iirc, for common status codes, like 500, 403, 401 and so on, where we most likely know the cause, or can point to somewhere, so the user can find out easily where the problem is. However, this message is used, when we do not have a separate message for the status code. That itself already means, that we most likely do not know the cause of the problem. Suggesting things here is pretty complicated and I would be against anything here, as we can not make sure, that our suggestion is even a good idea in all possible http status codes :(

mueller-ma: "HTTP error 42 occured, when connecting to you openHAB server"?

ThomDietrich: I'd not move the error code into more focus. That's a technical detail the average user will not be able to understand....

Actually this is exactly why I started this discussion. Put yourself into the position of an average no-technical end user of the app. Upon reading the error message, what did you learn? Something seems to be off. The message doesn't offer more insight.

I agree with both of you that (1) it's not easy to interpret the error message and (2) suggest next steps.

To that end, how about a more detailed message?

> Connection unsuccessful. An unexpected response was received while trying to connect to the configured openHAB server. Please ensure a working network connection, check the accessibility and operation of the server or contact your sever administrator. (Received HTTP response code: %d)

mobile/src/main/res/values/strings.xml
```diff
@@ -72,13 +72,13 @@
     <string name="error_network_not_available">Network is not available</string>
     <string name="error_http_connection_failed">Received an invalid response from openHAB. Please check if openHAB is working correctly (HTTP response code %d)</string>
     <string name="error_no_uuid">Unable to get openHAB UUID, connection failed</string>
-    <string name="error_invalid_url">Please enter a valid URL in a \'protocol://host:port/\' form</string>
+    <string name="error_invalid_url">Please enter a valid URL in a \'http(s)://host(:port)/\' form</string>
     <string name="error_connection_failed">Connection to host failed</string>
     <string name="error_unable_to_resolve_hostname">Unable to resolve hostname</string>
     <string name="error_connection_sslhandshake_failed">SSL Handshake failed - maybe you need a valid client certificate</string>
```
mueller-ma: Does it really show this error, when client cert auth fails? Doesn't this cause a http error (401, 495, 496) instead of failing the ssl handshake?

ThomDietrich: Funny you ask. I'm currently seeing it as I disabled demo mode but didn't configure my server address. Now that I think about it: Does auto discovery look for openHAB on https and if so, how can we expect it to own a valid certificate?

mobile/src/main/res/values/strings.xml
mueller-ma: That one sounds good. Should other http error strings also be extended with the second sentence?

mobile/src/main/res/values/strings.xml
```diff
@@ -72,13 +72,13 @@
...
     <string name="error_certificate_not_trusted">SSL certificate is not trusted. Please import the root certificate or disable certificate validation during SSL handshake</string>
     <string name="error_certificate_expired">SSL certificate has been expired. Please get a new certificate or disable certificate validation during SSL handshake</string>
-    <string name="error_certificate_not_valid_yet">SSL certificate is not valid yet. Please get a correct certificate or disable certificate validation during SSL handshake. Also check if the clock on your phone is set correctly</string>
+    <string name="error_certificate_not_valid_yet">SSL certificate is not valid yet. Please get a correct certificate or disable certificate validation during SSL handshake. Also check if the clock on your device is set correctly</string>
     <string name="error_certificate_revoked">SSL certificate has been revoked. Please get a new certificate or disable certificate validation during SSL handshake</string>
     <string name="error_http_code_401">User authentication has failed. Please check username, password and SSL client certificate (HTTP response code 401)</string>
```
mueller-ma: > username, password and SSL client certificate

User can think he needs all three of this. `username/password or SSL client certificate`?

mobile/src/main/res/values/strings.xml
mueller-ma: > Does auto discovery look for openHAB on https and if so, how can we expect it to own a valid certificate

Yes, and it will be changed in #562 to http. Ignoring ssl certs by default isnt good and the oh server has probably an invalid cert.

mobile/src/main/res/values/strings.xml
ThomDietrich: As you wish. Where it makes sense, why not. ;)

mobile/src/main/res/values/strings.xml
ThomDietrich: Also get rid of the first "User" and the technical detail?

> Authentication failed. Please check the configured username and password or the provided SSL client certificate.

??

mueller-ma: Yes, thats better.

mobile/src/main/res/values/strings.xml
```diff
@@ -80,19 +82,18 @@
...
     <string name="error_certificate_revoked">SSL certificate has been revoked. Please get a new certificate or disable certificate validation during SSL handshake</string>
-    <string name="error_http_code_401">User authentication has failed. Please check username, password and SSL client certificate (HTTP response code 401)</string>
-    <string name="error_http_code_403">User authentication has failed. Please check username, password and SSL client certificate (HTTP response code 403)</string>
-    <string name="error_http_code_407">User authentication has failed. Please check username, password and SSL client certificate (HTTP response code 407)</string>
-    <string name="error_http_code_414">openHAB URL is too long (HTTP response code 414)</string>
+    <string name="error_http_code_401">Authentication failed. Please check the configured username and password respectively the provided SSL client certificate (HTTP response code 401)</string>
```
mueller-ma: Should I add a full stop at the end of the message?

mobile/src/main/res/values/strings.xml
maniac103: I'm not sure whether that 5 line message, even though it's way more helpful for the user, is well suited here: I simply think it's too long for the amount of time it's shown. For Snackbars we could solve it by using DURATION_INFINITE instead of DURATION_LONG for certain messages, but that isn't an option for toasts. I'm really not sure a nontechnical user fully understands such verbose messages shown in a toast...

ThomDietrich: Well... doesn't sound like a message that should be shown in a toast, should it? Toast messages are for unimportant confirmations and stuff.
I'm not the android development expert here but toasts are imho seldom used for important stuff.

mobile/src/main/res/values/strings.xml
ThomDietrich: Whoever decides that also has to go through many of the other strings...

mobile/src/main/res/values/strings.xml
maniac103: That's right, but spawning a foreground dialog from a background service (e.g. voice recognition from widget) isn't good UI either.

mobile/src/main/res/values/strings.xml
FlorianSW: I think the error message mueller-ma is referring to is "error_connection_sslhandshake_failed", isn't it? There're multiple different scenarios where this may happen, however, most of them has their own messages already (see https://github.com/openhab/openhab-android/blob/c31e282b764b33a8657f38c589ecb4798c1bda98/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java#L135-L146), so this is most likely be really related to the client certificate (or, but that should be very rare, any other misconfiguration or mismatch of the server to the client (supported TLS version, support cipher suites,...). If it's related to the client cert, it depends on how the server is configured:

1) If the webserver requests a client certificate, but proceeds with the tls handshake, even if the client sends an empty certificate message (means: no client certificate), then the webserver will normally reply with a http status code (401, 200, but with an error page, ...)
2) If the webserver requests a client certificate and fails, if no valid certificate is provided by the client, a fatal handshake_failure will be send by the server, while immediately aborting the handshake. This message will be shown in this case by the app.

mobile/src/main/res/values/strings.xml
mueller-ma: @maniac103 Is right, I forgot that these messages are displayed in toasts, when using the voice widget. I shortend the messages.

mobile/src/main/res/values/strings.xml
mueller-ma: There seems to be a bug.
Log: `E/OpenHABMainActivity: Error: javax.net.ssl.SSLHandshakeException: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.`
But shows the `error_connection_sslhandshake_failed` string.

mueller-ma: Found the issue, fix is on its way.

mueller-ma: See #616

mueller-ma:
@ThomDietrich @maniac103 @FlorianSW Can I merge this PR?

mobile/src/main/res/values/strings.xml
```diff
@@ -21,9 +21,9 @@
     <string name="mainmenu_openhab_clearcache">Clear images cache</string>
 
     <!-- App settings strings -->
-    <string name="settings_connection_title">Connection Settings</string>
-    <string name="settings_display_title">Display Settings</string>
-    <string name="settings_misc_title">Misc Settings</string>
+    <string name="settings_connection_title">Connection</string>
+    <string name="settings_display_title">Display</string>
+    <string name="settings_misc_title">Misc</string>
```
maniac103: Better use the unabbreviated version here: `Miscellaneous`

mobile/src/main/res/values/strings.xml
```diff
@@ -54,9 +54,9 @@
     <string name="settings_clear_default_sitemap">Clear default sitemap</string>
     <string name="settings_no_default_sitemap">No default sitemap selected</string>
     <string name="settings_current_default_sitemap">Current default sitemap: %1$s</string>
-    <string name="settings_openhab_sslsettings">Encryption settings</string>
-    <string name="settings_openhab_alt_connection">Remote server settings</string>
-    <string name="settings_openhab_connection">Local server settings</string>
+    <string name="settings_openhab_sslsettings">Encryption</string>
+    <string name="settings_openhab_alt_connection">Remote</string>
+    <string name="settings_openhab_connection">Local</string>
```
maniac103: `Local server` / `Remote server`? Only local/remote seens a bit short. I saw the previous discussion, but I think omitting both 'connection' and 'settings' oversimplifies things and makes the string less clear for no real gain.

ThomDietrich: Your suggestion is a problem on its own. The menu entry is not necessarily related to a "server". Imagine your wife using the app or a user who's landlord provides an openHAB system for the central heating system. Those users do not know the server, they want to connect to a service using address and credentials.

All: I've been thinking about this for a while: Are you considering support of multiple accounts? Part of the redesign needed for such a function could also be the redesign of this local/remote part. 

My proposal would be to step away from local and remote completely. Instead use something like:

grafik

https://ninjamock.com/s/6DQ7MWx

FlorianSW: > Are you considering support of multiple accounts?

In a very very very long term: I think so, yes. See #10 for the discussion :)

I'm not sure, if everyone can easily know when a "primary connection" and when the "alternative connection" will be used. One could think, that the primary connection is always used, and only, if it is not reachable, the alternative connection is used. This is IHMO false, as the primary would not be used when using mobile data (it would not even be tried to connect to this service) :/

However, I don't have any better idea, so :(

mueller-ma: We should also look at #10 and how the GUI needs to be changed for that.
The heading of "Local" and "Remote" is "Connection", so we would have it twice.

maniac103: IMHO there is nothing wrong with a tiny bit of redundancy, as long as it isn't obviously repetitive.

ThomDietrich: I don't want to continue this discussion here but just one more thought: How likely is it that a user needs both URLs? If a system is accessible via a public address, why not use it all the time. There is no technical reason why anyone would want to provide the local address in this case.

PR #609, opened 2018-01-04
Use SpeechRecognizer.isRecognitionAvailable() to check if available

FlorianSW:
Instead of quering for an intent, which is used for implementation, we'll
use official API to check, if voice recognition is available on the
device. This will change the intent checked from ACTION_RECOGNIZE_SPEECH
to SERVICE_INTERFACE.

Also: Extract the Voice recognition test from the BasicWidget test and
execute it only, if voice recognition is actually available on the
instrumented device.


mueller-ma:
LGTM!

mueller-ma:
@kaikreuzer Did you or pullapprove add the badge?

PR #611, opened 2018-01-05
Travis: Run with Google Play Services for Voice Recognition

FlorianSW:
Getting the Voice recognition running is only possible, if at least one
handler handles the RecognitionService.SERVICE_INTERFACE intent. The
easiest way to achive this is to run the instrumented test on an emulator
which actually has the Google Play services installed.

This commit changes the travis-ci configuration so that the emulator
started has the Play services installed. It also adds some extras, which
are already pre-installed, but accoridng to the travis documentation[1]
should be explicitly mentioned to ensure build stability. However, for the
current build they don't change anything. The extras are namely:

extra-android-support
extra-google-google_play_services
extra-google-m2repository
extra-android-m2repository

[1]
https://docs.travis-ci.com/user/languages/android/#Pre-installed-components

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ jdk: oraclejdk8
 env:
   global:
   - secure: fMBUkSsCBTw/U907w7Xm/yNzw4lq9yt7zFy0JyEZrA543LI7zQRIclIOtsDTUs0bwY+4KI08MNoYPLJ4TQhSJaH4wQ8b9C5TRsqXNfc0V718TcuYqSXqU6VPRVtX46/fCEWv/HAs0ksUKrHIlpWOwLPQpYDcKxwwlUndaakYKJ8=
+  - ADB_INSTALL_TIMEOUT=8
 
 before_cache:
 - cd ${TRAVIS_BUILD_DIR}/gradle/caches/
@@ -24,8 +25,17 @@ android:
   - platform-tools
   - tools
   - android-22
-  - sys-img-armeabi-v7a-android-22
-  
+  - extra-google-google_play_services
+  - extra-google-m2repository
+  - extra-android-m2repository
+  - extra-android-support
+  - addon-google_apis-google-22 # Google play services
+  - sys-img-armeabi-v7a-google_apis-22
+
+licenses:
+  - android-sdk-license-.+
+  - google-gdk-license-.+
+
 install:
 - echo yes | sdkmanager "extras;m2repository;com;android;support;constraint;constraint-layout;1.0.2"
 - echo yes | sdkmanager "extras;m2repository;com;android;support;constraint;constraint-layout-solver;1.0.2"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -24,8 +25,17 @@ android:
...
-  - sys-img-armeabi-v7a-android-22
-  
+  - extra-google-google_play_services
+  - extra-google-m2repository
+  - extra-android-m2repository
+  - extra-android-support
+  - addon-google_apis-google-22 # Google play services
+  - sys-img-armeabi-v7a-google_apis-22
+
+licenses:
```
mueller-ma: "By default, Travis CI will accept all the requested licenses, but it is also possible to define a white list of licenses to be accepted, as shown in the following example:"
from https://docs.travis-ci.com/user/languages/android/#Pre-installed-components

FlorianSW: You're right, we most likely do not need the license list. I'll remove it, once I'm sure, that the build succeeds :)

FlorianSW:
Ok, I've no idea why the build is failing. On my fork I tested a variety of different solutions and suggestions I found, which may or may not fix this issue, however, the build is always failing. Does someone has an idea?

If not, I'll probably open an issue at https://github.com/travis-ci/travis-ci to ask, why the build on a non-google-apis including image works fine, however, an installtion on an image including google-apis fails.

mueller-ma:
I have read (couldnt find a source) that google doesnt release google-free abi of recent android version anymore, so they should always do the voice check.
Do you want to give it a try to use matrix to run tests on api 22 and 27?

.travis.yml
```diff
@@ -3,6 +3,10 @@ jdk: oraclejdk8
 env:
   global:
   - secure: fMBUkSsCBTw/U907w7Xm/yNzw4lq9yt7zFy0JyEZrA543LI7zQRIclIOtsDTUs0bwY+4KI08MNoYPLJ4TQhSJaH4wQ8b9C5TRsqXNfc0V718TcuYqSXqU6VPRVtX46/fCEWv/HAs0ksUKrHIlpWOwLPQpYDcKxwwlUndaakYKJ8=
+  - ADB_INSTALL_TIMEOUT=8
+  matrix:
+    - ANDROID_TARGET=android-22 ANDROID_ABI=armeabi-v7a
+    - ANDROID_TARGET=android-27 ANDROID_ABI=armeabi-v7a
```
FlorianSW: Ah, should be x86_64, as armeabi-v7a is not available for this api :/

mueller-ma:
When I understand it correctly, travis uses "cookbooks" as template for ci builds. Android has the min api 25: https://github.com/travis-ci/travis-cookbooks/blob/master/community-cookbooks/android-sdk/CHANGELOG.md

FlorianSW:
Hmm, ok, just a question (I'm not sure if my answer "yes", is correct): Can I simply install the sys-image with sdkmanager? Then no sys-img would be installed with android-components.

mueller-ma:
Yes, should work.

mueller-ma:
I opened an issue for Api 25+
https://github.com/travis-ci/travis-cookbooks/issues/965

mueller-ma:
Closing due to conflicts introduced by #801

PR #612, opened 2018-01-05
Remove unneeded getBaseContext

mueller-ma:
@maniac103 
Suggested change in #604


maniac103:
Did you check whether we have more of those in the code base? Also, we could check for possible calls to getApplicationContext() that can be removed - those are also only rarely needed.

kaikreuzer:
Closing and re-opening in order to test codecov check.

kaikreuzer:
Hm, might only work for new PRs then...?

kaikreuzer:
Codecov integration works now!

PR #616, opened 2018-01-07
Display correct ssl error message

mueller-ma:
The cause of the error a CertificateException and its cause is
CertPathValidatorException, so we need to call .getCause twice.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
@@ -109,6 +109,7 @@ import static org.openhab.habdroid.util.Constants.MESSAGES.DIALOG;
 import static org.openhab.habdroid.util.Constants.MESSAGES.LOGLEVEL.ALWAYS;
 import static org.openhab.habdroid.util.Constants.MESSAGES.LOGLEVEL.DEBUG;
 import static org.openhab.habdroid.util.Constants.MESSAGES.LOGLEVEL.NO_DEBUG;
+import static org.openhab.habdroid.util.Util.exceptionHasCause;
 
 public class OpenHABMainActivity extends AppCompatActivity implements OnWidgetSelectedListener,
         OpenHABTrackerReceiver, MemorizingResponder {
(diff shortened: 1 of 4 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -133,13 +133,13 @@ public void onFailure(Call call, int statusCode, Headers headers, byte[] respons
                 message = getString(R.string.error_unable_to_resolve_hostname);
             } else if (error instanceof SSLHandshakeException) {
                 // if ssl exception, check for some common problems
-                if (error.getCause() instanceof CertPathValidatorException) {
+                if (error.getCause().getCause() instanceof CertPathValidatorException) {
                     message = getString(R.string.error_certificate_not_trusted);
-                } else if (error.getCause() instanceof CertificateExpiredException) {
+                } else if (error.getCause().getCause() instanceof CertificateExpiredException) {
```
FlorianSW: When testing against https://expired.badssl.com the CertificateExpiredException is even one level under this:
e.getCause().getCause().getCause()

However, don't we've a better way of doing it? E.g. having a loop which calls getCause() until the returned object is not the same as the current one and then using this one for these checks, here?

mueller-ma: The first implementation had .toString.equals(CertificateExpiredException). I'm not sure if thats better.

mueller-ma:
Untrusted root and expired exceptions work now, however https://revoked.badssl.com/ gives a 404.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -146,4 +146,14 @@ public static void setActivityTheme(@NonNull final Activity activity, String the
         }
         activity.setTheme(themeRes);
     }
+
```
maniac103: Why not pass cause as `Class<?> causeType` and compare `error.getCause()` to that? Saves the instance creations...

mueller-ma: Yes, thats exactly what I'm currently trying. How can I call the method then?

mueller-ma: Nevermind. got it

FlorianSW:
Sorry for not repsonding anymore :( However, the change looks sane now! :)

Issue #270, opened 2017-02-24
Order sitemaps alphabetically


kvoit:
Especially in the app, sitemaps suggest themselves as the primary unit of order. Unfortunately, the order of sitemaps in the sidebar seems to be absolutely random with no possibility to influence it. Would it be possible to order the sitemaps alphabetically, at least?

ldet:
Good idea
Sorting alphabetically schould be possible.

Currently the order is determined by the rest api.
We could sort them by their name (the filename) or the label defined inside the file. I personally think filename would work best, because we could do something like this to force a specific order:
`0Lights.sitemap`
`1Heating.sitemap`

On the long-term, we might want to consider adding an index attribute to sitemaps we can sort by:
`sitemap name label="label" index="0"{`
But this would need to be done elsewhere.
And even then alphabetically is a good fallback solution I think.

lolodomo:
To be closed due to #392.

mueller-ma:
I would go for alphabetic order with the sitemap label, since it is shown in the ui. Sitemaps are not meant for navigation, so IMO some kind of index is not needed.

mueller-ma:
Will be fixed in #638

PR #620, opened 2018-01-09
Remove default sitemap and sort sitemap list

MAndretti:
- remove _default sitemap
- sort sitemap by name (openhab 1 and openhab 2)

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/Util.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Util.java
@@ -85,7 +85,7 @@ public class Util {
             for (int i = 0; i < sitemapNodes.getLength(); i++) {
                 Node sitemapNode = sitemapNodes.item(i);
                 OpenHABSitemap openhabSitemap = new OpenHAB1Sitemap(sitemapNode);
-                if (openhabSitemap.getName().equals("_default")) {
+                if (openhabSitemap.getName().equals("default")) {
                     continue;
                 }
                 sitemapList.add(openhabSitemap);
--- a/mobile/src/main/java/org/openhab/habdroid/util/Util.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Util.java
@@ -66,6 +66,18 @@ public class Util {
         return normalizedUrl;
     }
 
+    private static List<OpenHABSitemap> sortSitemapList(List<OpenHABSitemap> sitemapList) {
+        // Sort by sitename name
+        Collections.sort(sitemapList, new Comparator<OpenHABSitemap>() {
+            @Override
+            public int compare(OpenHABSitemap sitemap1, OpenHABSitemap sitemap2) {
+                return  sitemap1.getName().compareTo(sitemap2.getName());
+            }
+        });
+
+        return sitemapList;
+    }
+
     public static List<OpenHABSitemap> parseSitemapList(Document document) {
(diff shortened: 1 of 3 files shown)


MAndretti:
Hey Folks,

I don't understand the error on the ui test. 
Did I cause the error?

I remove _default or default sitemap. But only if there is more than one sitemap. So always a sitemap is given. After that I sort the sitemaps by name. 

If I run the App all looks fine. Only the build fails.

As newbie I need some help.

Thnaks,
Mario

mueller-ma:
Hey @MAndretti,

I have some suggestions for your PR:

1. Use a descripive name for the PR, not "Fix #1234"
2. #577 is covered by  #578, so please remove that part from your PR. Generell its always good to make small PRs, so one for #577 and one for #270 
3. Currently we have two checks for PRs:
3.1. Pullapprove to check whether the commits are signed. See https://github.com/openhab/openhab-android/blob/master/CONTRIBUTING.md#sign-your-work
3.2. Travis to build and run tests. Travis is [failing](https://travis-ci.org/openhab/openhab-android/builds/327212228#L3297), because we have UI tests, that expect the sitemap selection to come up. It doesnt come up, because there is only one sitemap now.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -66,6 +66,38 @@ public static String normalizeUrl(String sourceUrl) {
...
+
+        return sitemapListResult;
+    }
+
+    private static List<OpenHABSitemap> sortSitemapList(List<OpenHABSitemap> sitemapList) {
+        // Sort by sitename name
+        Collections.sort(sitemapList, new Comparator<OpenHABSitemap>() {
+            @Override
+            public int compare(OpenHABSitemap sitemap1, OpenHABSitemap sitemap2) {
+                return  sitemap1.getName().compareTo(sitemap2.getName());
```
mueller-ma: Sitemaps have the attribute "Label", so please use that to compare

MAndretti: Hey @mueller-ma 

Ok, next time I use a descripive name :) sorry. And one fix per pull.

1) I think that my change is a little more elegant then the fix implemented in #578. I remove the default sitemap only if one sitemap is configured on the server. And I use only one method (no copy and paste).

2) Yes sitemap has a attribute label. But some times a user want to sort the sitemap list by e.g. "priority". So the user can use
     Filename/Name -> Label 
     1Livinroom -> Livingroom
     2Bathroom -> Bathroom
     ...

If name=label (happens frequently) -> works too

Should I really go back to Label?

3) I run the App without a sitemap, one sitemap and ten sitemaps. No error. Which class performs the test?

Thanks,
Mario

mueller-ma: > next time I use a descripive name 

I prefere this time :) You can edit PR names.

> Yes sitemap has a attribute label. But some times a user want to sort the sitemap list by e.g. "priority"

We should check how basic ui sorts sitemaps.

> I run the App without a sitemap, one sitemap and ten sitemaps. No error. Which class performs the test?

When you run the app with one ui, no selection popup is shown, but the test expects a popup.

mueller-ma:
Nevermind, there is a open PR for sitemap sorting now.

PR #623, opened 2018-01-10
Display sitemap name when no label is present and add sitemap tests

### LINKED ISSUE #622: Crash, when sitemap has no label

mueller-ma:
Fixes #622

I found out, that the field "leaf" isnt set in OpenHAB2Sitemap. Is this intended?

### DIFF
--- a/mobile/src/test/java/org/openhab/habdroid/model/OpenHABSitemapTest.java
+++ b/mobile/src/test/java/org/openhab/habdroid/model/OpenHABSitemapTest.java
@@ -1,0 +1,69 @@
+package org.openhab.habdroid.model;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import static org.junit.Assert.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class OpenHABSitemapTest {
+    OpenHABSitemap sitemap1;
+    OpenHABSitemap sitemap2;
+
+    @Before
+    public void initSitemaps() throws JSONException {
+        String jsonString = "{\"name\":\"demo\",\"label\":\"Main Menu\",\"link\":\"http://demo.openhab.org:8080/rest/sitemaps/demo\",\"homepage\":{\"link\":\"http://demo.openhab.org:8080/rest/sitemaps/demo/demo\",\"leaf\":false,\"timeout\":false,\"widgets\":[]}}";
+        JSONObject jsonObject = new JSONObject(jsonString);
+        sitemap1 = new OpenHAB2Sitemap(jsonObject);
+
+        jsonString = "{\"name\":\"home\",\"icon\":\"home\",\"link\":\"http://demo.openhab.org:8080/rest/sitemaps/home\",\"homepage\":{\"link\":\"http://demo.openhab.org:8080/rest/sitemaps/home/home\",\"leaf\":true,\"timeout\":false,\"widgets\":[]}}";
+        jsonObject = new JSONObject(jsonString);
+        sitemap2 = new OpenHAB2Sitemap(jsonObject);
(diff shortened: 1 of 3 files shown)


mobile/src/test/java/org/openhab/habdroid/model/OpenHABSitemapTest.java
```diff
@@ -0,0 +1,69 @@
...
+        JSONObject jsonObject = new JSONObject(jsonString);
+        sitemap1 = new OpenHAB2Sitemap(jsonObject);
+
+        jsonString = "{\"name\":\"home\",\"icon\":\"home\",\"link\":\"http://demo.openhab.org:8080/rest/sitemaps/home\",\"homepage\":{\"link\":\"http://demo.openhab.org:8080/rest/sitemaps/home/home\",\"leaf\":true,\"timeout\":false,\"widgets\":[]}}";
+        jsonObject = new JSONObject(jsonString);
+        sitemap2 = new OpenHAB2Sitemap(jsonObject);
+    }
+
+    @Test
+    public void sitemapInstanceOf() {
```
FlorianSW: Hmm, this check seems pretty pointless, as the before method at both times invokes the constructor of OpenHAB2Sitemap, so the instanceof check is guaranteed to return true when used against OpenHAB2Sitemap?

This would probably make sense, when we would use factory methds to create the test objects, but we don't do that :)

mobile/src/test/java/org/openhab/habdroid/model/OpenHABSitemapTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import static org.junit.Assert.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class OpenHABSitemapTest {
+    OpenHABSitemap sitemap1;
```
FlorianSW: demoSitemapWithLabel

mobile/src/test/java/org/openhab/habdroid/model/OpenHABSitemapTest.java
```diff
@@ -0,0 +1,69 @@
...
+    OpenHABSitemap sitemap2;
```
FlorianSW: homeSitemapWithoutLabel

mobile/src/test/java/org/openhab/habdroid/model/OpenHABSitemapTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Test
+    public void sitemapInstanceOf() {
+        assertTrue(sitemap1 instanceof OpenHAB2Sitemap);
+        assertTrue(sitemap2 instanceof OpenHAB2Sitemap);
+    }
+
+    @Test
+    public void sitemapLabelNonNull() {
+        assertEquals("Main Menu", sitemap1.getLabel());
+        assertEquals("home", sitemap2.getLabel());
```
FlorianSW: Message for the assert:
"Sitemap without explicit label should return name for getLabel"

mobile/src/test/java/org/openhab/habdroid/model/OpenHABSitemapTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Test
+    public void testGetIcon() {
+        assertNull(sitemap1.getIcon());
+        assertEquals("home", sitemap2.getIcon());
+    }
+
+    @Test
+    public void testGetIsLeaf() {
+        assertFalse(sitemap1.isLeaf());
+        assertFalse("Leaf isn't set in OpenHAB2Sitemap", sitemap2.isLeaf());
```
FlorianSW: "Leaf should not be set in OpenHAB2Sitemap"

The message is shown when the test fails, the current message would be a bit misleading, wouldn't it? :)

Issue #581, opened 2017-12-30
Widget has no preview picture


mueller-ma:
https://developer.android.com/guide/topics/appwidgets/index.html#preview

cribskip:
I'm adding this in my open PR #545 today

Issue #549, opened 2017-12-20
Next Release


digitaldan:
This issue is used to coordinate new releases.

lolodomo:
The translations have to be emhanced too as some texts were added without translation just before the release 2.0.

FlorianSW:
#561 should be included, too.

digitaldan:
Hi guys, was away for the holidays and other engagements, but I'm back now.

cribskip:
Just a little follow-up: I've deployed a build of the [current master](https://github.com/openhab/openhab-android/tree/e6466c5d7debe947558db4608e9788e665ddb265) to my phone and tested the enhancements to this scope:

- #628 Voice widget
- #517 NFC

btw, these are the only ones merged from the list above

mueller-ma:
@digitaldan I would suggest to release a new version soon, because 2.0.0 crashes on all kitkat devices, which is about ~15% of all users.

mueller-ma:
2.0.24-beta (rename to .apk before): [mobile-full-debug.pdf](https://github.com/openhab/openhab-android/files/1799296/mobile-full-debug.pdf)

mueller-ma:
For now I have found following changes that should be in the play store changelog:

```
1. Allow the widget list to be refreshed when swiping (#639)
2. Fix flickering icons on sitemap reload. (#669)
3. Improve look'n'feel of section switch button (#687)
4. Improved auto discovery (#562)
5. Changed visual look of color and rollershutter buttons (#736)
6. New error fragment (#688) 
```

I think we can merge some of those:

```
1. Allow the widget list to be refreshed when swiping
3+5. Improve look'n'feel of some items
6. <removed>
2+4. Stability and performance improvements
```

maniac103:
6. + the navigation drawer overhaul can be summarized as something like 'UI improvements'

mueller-ma:
And can be merged with 3+5:
```
1. Allow the widget list to be refreshed when swiping
3+5+6. UI improvements
2+4. Stability and performance improvements
```

mueller-ma:
Updated changelog:
```
Allow the widget list to be refreshed when swiping
Support for Mapview widgets
Adapt voice service to use new voice command endpoint
UI improvements
Stability and performance improvements
```

mueller-ma:
#830, #817 and maybe #816 and than it should be fine for 2.1.0

maniac103:
Should add #828 as well.

[...]

PR #545, opened 2017-12-20
Improve Voice Widget

cribskip:
* make resizable
* fix too big logo
* improve icon

Issue #531, opened 2017-12-18
Widget broken - Wrong iconsize or microphone graphic is missing


Confectrician:
Hi,

After updating i recognized that the app widget is displayed wrong.

![image](https://user-images.githubusercontent.com/7161177/34114976-8492d522-e414-11e7-95df-0bb6fb5ed72b.png)
![image](https://user-images.githubusercontent.com/7161177/34114997-944ff044-e414-11e7-9a9c-11425138dea1.png)

Seems that something with the icons is wrong.
The voice button is usable if you have small fingers, but i don't think that this is the intended behavior.

BR 
Jerome

Edit:

Phone: Motorola Moto G Play
Android: 6.0.1

mueller-ma:
![screenshot_20171218-172145](https://user-images.githubusercontent.com/22525368/34116223-36db40b8-e418-11e7-8526-f0d508e1aa9d.png)
I created a new widget (didnt had one before the update) and it looks like intended.

Android 7.1.2
LineageOS 14.1
Oneplus X

mueller-ma:
Whats your grid size?

Confectrician:
4x4

mueller-ma:
I have 4x5 at 5 inch display. Did you try to remove and recreate the widget?

Confectrician:
Did it right now, but that doesn't change anything.

cribskip:
I'm having the same problem as @Confectrician.
I'll see if I can get my dev env ready today and take a look at this.

While at it, I'll make the widget resizable.

cribskip:
Ok, finally got github to do what I wanted ;-) Please see #545

robbyb67:
Same problem here - waiting for #545

mueller-ma:
Fixed in #628

PR #628, opened 2018-01-10
Improve Voice Widget

cribskip:
- Removed unused icons
- Added preview picture of widget
- Improved widget (scaling, readability (now with white microphone))
2018-01-10 22_11_42-android emulator - nexus_5x_api_23_5554
2018-01-10 22_11_11-android emulator - nexus_5x_api_23_5554

Hopefully this time I got it right ;-)
Thanks for your patience.

This is a follow-up on PR #545 
This should fix Issues #581 #531 and just to mention #549


mueller-ma:
Just one thing: In the menu where I can select which widget I want to create, I still see the old preview, but this could be caused by the launcher.

cribskip:
Ok, thanks for testing. I deleted my emulation device, reinstalled the build and looked at the widget selection. Works just fine (just like the screenshot above).

Even the Widget Preview App works fine:
2018-01-12 20_15_10-android emulator - nexus_5x_api_23_5554

PR #631, opened 2018-01-11
Remove default sidemap

MAndretti:
- Remove default or _default sitemap if more than one sitemap is given
- Change unit tests
- Change UI tests

Fix: sitemap _default #577

Signed-off-by: Mario Adamo Andretti <mario.adamo.andretti@gmail.com>


mueller-ma:
Hey @MAndretti ,

I think there is a small misunderstanding. When I said you should open a PR for both issues, this was meant as a generall advice. In this case #577 is covered by  #578.

Thanks for your understanding.

PR #639, opened 2018-01-13
Allow the widget list to be refreshed when swiping

### LINKED ISSUE #539: Refresh on pulldown
When inside a sitemap view pulling down the view beyond the length of the sitemap should trigger a refresh action as known from many apps. Currently there is no easy way to trigger a refresh of the sitemap view.

Rationale: Even if that might technically not be needed, a user might want to do that and one of the google app design guidelines is that every user input should result in appropriate action...

FlorianSW:
Fixes #539


mueller-ma:
@FlorianSW Can you adjust the test?

mueller-ma:
Hm, log from last travis is "null"

FlorianSW:
This failing test works for me on multiple devices and emulators just fine :/

mueller-ma:
Looks good so far, just two things:
1. In landscape mode a pull only refreshes one column, but I'm not sure if that's correct or if both should be refreshed.
2. The arrow in the reload indicator is black. I really like those indicator, when they switch the color: https://stackoverflow.com/a/25059466/9053008 Or at least it should be in primary_color

mueller-ma:
Travis works now

mueller-ma:
When we include this, people might think the app has no auto fresh and start refreshing over and over. We could show a snackbar when the user reloads that the feature is usefull for e.g. images. It should also contain "got it" to mute this snackbar.

FlorianSW:
Nice idea! Implemented it in the commits above :)

We can discuss if the message content needs to be changed, though :)

mueller-ma:
The snackbar looks good so far.

What about my other two points?

mobile/src/main/res/values/strings.xml
```diff
@@ -183,4 +183,6 @@
     <string name="content_description_open_number_picker">Open number picker</string>
     <string name="content_description_video">Video</string>
     <string name="content_description_open_openhab">Open openHAB</string>
+    <string name="swipe_to_refresh_description">The app refreshes widgets regularly, you mostly do not need to do it yourself.</string>
```
mueller-ma: Maybe describe Kai's usecase from https://github.com/openhab/openhab-android/issues/539#issuecomment-355902286

FlorianSW: The problem is, that the snackbar message is very limited, or it will be truncated (this can maybe "turned off"?), so it needs to be somehow short. This is the one I came up with, which is short enough but still descriptive :]

FlorianSW:
> In landscape mode a pull only refreshes one column, but I'm not sure if that's correct or if both should be refreshed.

I always forget landscape. Can you refresh both sides separatedly? This would be what I would expect :/

> The arrow in the reload indicator is black. I really like those indicator, when they switch the color: https://stackoverflow.com/a/25059466/9053008 Or at least it should be in primary_color

Good idea, will do that! :)

mueller-ma:
> I always forget landscape. Can you refresh both sides separatedly? This would be what I would expect :/

Yes, i can refresh both sides seperatly.

mobile/src/main/res/values/strings.xml
mueller-ma: I think its better to remove the full stop in this sentence, otherwise lgtm

maniac103: I'd make it say `The app also refreshes automatically, you normally do not need to do it manually`

mueller-ma:
Engage! :)

PR #640, opened 2018-01-13
Simplify layout and viewpager hierarchy

FlorianSW:
Instead of implementing our own ViewPager to implement features we do not
currently use (swiping pages) or where (imho) the benefit is at least
questionable, use the default ViewPager delivered with the support
library.

Also: Do not maintain a main activity layout for API level 21, where the
only difference to the one used for all other api levels is, that one
property is added. This property can be added to the main layout file and
will simply be ignored in api levels, which do not support it.

Also: Remove the main activity twopane layout file, which is unused in the
whole project.


maniac103:
I'm wondering whether we use **any** feature of ViewPager - do we actually page something? If not, we could/should kill it off altogether...

FlorianSW:
Uhm, I don't think so. Do you suggest removing the pager adapter and the viewpager and replacing it by methods that simply add the fragment to the app on top of the current fragment (or replacing it)?

maniac103:
Yes, I'm suggesting exactly that (unless I'm missing something). Personally, I'd do it in this PR.

maniac103:
Actually the ViewPager **is** used: for paging the fragments in multi-pane layout. Having two containers to insert the respective fragments into would likely be a much cleaner solution, though.

FlorianSW:
Hmm, I would be more comfortable with doing that in a separate PR, I do not want to have a big PR again, which changes a lot of things at once :(

mueller-ma:
LGTM! @FlorianSW rtm?

PR #641, opened 2018-01-13
Do not access static member via instance reference

FlorianSW:


FlorianSW:
Uhm, is it correct, that codecov is "filing" when the coverage is decreased?

maniac103:
Please see #498.

FlorianSW:
Ok, that would fix the second access, however, the first one still applies :)

maniac103:
Yes, of course :-) I just wanted to point out the second potential conflict.

mueller-ma:
The mentioned PR is merged, can you remove the second fix?

mueller-ma:
> Uhm, is it correct, that codecov is "failing" when the coverage is decreased?

It works as intended, but the question is, if it should fail when someone does only small changes to the code.

FlorianSW:
Done!

PR #645, opened 2018-01-14
Add Travis matrix for multiple emulators

mueller-ma:
@FlorianSW I just want to see if api 14 works without messing up your PR.

If it works I can either merge this one and you make a new one, or you include this change in #611


FlorianSW:
@mueller-ma feel free to add it to my existing PR. However, your changes still skip the voice test, so this problem is not resolved with that.

mueller-ma:
I wonder why the tests pass for api 16. I run kitkat on an avd and that crashed instantly due to a multidex issue (#619)

mueller-ma:
Ok, the test fails with api 19, but not with api 16. I have checked my fix #619 on an avd, so I'm going to merge that one.

@FlorianSW I'm going to add api 19 to your PR.

Issue #319, opened 2017-06-19
When entering a subpage and turning the tablet in landscape, no entry is marked


mueller-ma:
First: I selected "First Floor"
![screenshot_2017-06-19-20-21-56](https://user-images.githubusercontent.com/22525368/27299547-fa270b54-552c-11e7-9315-d2f50a225e4e.png)
Second: Turn tablet into landscape. How it looks:
![screenshot_2017-06-19-20-22-00](https://user-images.githubusercontent.com/22525368/27299548-fa2902e2-552c-11e7-9131-089e71d14561.png)
And how it should look:
![screenshot_2017-06-19-20-22-05](https://user-images.githubusercontent.com/22525368/27299549-fa293ca8-552c-11e7-9e91-ea775f49f18b.png)

lolodomo:
In fact, it looks like when you remains in landscapoe mode and select in the left column, the selection appears a short time and then disappears when the right column is updated.
So there is something wrong even when remaining in landscape mode.

mueller-ma:
The blue selection is now completly broken.

PR #625, opened 2018-01-10
Refresh only upgraded widgets, not all the fragments (Replace ListVie…

9a4gl:
…w with RecyclerView)

Signed-off-by: Tihomir Heidelberg <tihomir.heidelberg@lite.hr>

PR #650, opened 2018-01-17
Convert widget list fragment to use RecyclerView

maniac103:
This allows more fine-grained change handling later, and the ViewHolder pattern built into RecyclerView allows for great refactoring of the adapter code: Instead of having one gigantic getView() method, the binding of the different view types is now contained in a separate ViewHolder subclass for each type.


maniac103:
If any Espresso wizard around could help with updating the tests, that'd be great.

maniac103:
Currently two-pane selection highlight is broken; I'm looking into it.

mobile/src/main/res/drawable/list_arrow_activated_tinted.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<nine-patch xmlns:android="http://schemas.android.com/apk/res/android"
+    android:src="@drawable/list_arrow_activated"
+    android:tint="?attr/colorControlHighlight"
```
maniac103: Maybe could also use `?attr/colorControlActivated` here? Thoughts?

maniac103:
Two-pane layout fixed; only open issue is the Espresso test (which I'm not familiar with :-( )
@mueller-ma @FlorianSW Any help regarding the test would be appreciated.

lolodomo:
Is it a concurrent to PR #625 ?

FlorianSW:
> only open issue is the Espresso test (which I'm not familiar with :-( )

Most of the tests heavily rely on the current structure of the view, which can change easily (even in different API levels), like we see here. I ran into the same issue and simplified some of the tests in PR #639, I would take some time to do the same here, too, as most of the viewmatchers aren't really needed to still fulfill the testing goal. However, doing the same work as in 639 again here, sounds like waste, I would say we wait until the one is merged, update this PR and see, what tests still needs to be adjusted. WDYT?

maniac103:
@FlorianSW That's fine with me.

@lolodomo Yes and no. Yes, #625 also does ListView -> RecyclerView conversion, but no, it doesn't do the massive cleanup that this PR does. It also does additional things (optimize how updates are processed), which I'd prefer to keep in a separate PR and/or leave it to the SSE rewrite.

lolodomo:
Ok, that means I should wait for this PR to be merged before restarting and finishing my SSE work ?

maniac103:
> Ok, that means I should wait for this PR to be merged before restarting and finishing my SSE work ?

IMHO it's not directly related. All this PR allows is fine granular updates of single changed widgets in the adapter instead of forcing the adapter to go over the whole list and checking what updated. Replacing long polling by SSE should be mostly unaffected by it.

mueller-ma:
> on the current structure of the view, which can change easily (even in different API levels)

I made the test via the espresso recorder. When I run the test afterwards, the test failed... So the layout seems to change sometimes.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetListFragment.java
```diff
@@ -404,13 +396,14 @@ public void processContent(String responseString, boolean longPolling) {
...
+        List<OpenHABWidget> widgetList = new ArrayList<>();
+
         // If openHAB verion = 1 get page from XML
         if (mActivity.getOpenHABVersion() == 1) {
             // As we change the page we need to stop all videos on current page
             // before going to the new page. This is quite dirty, but is the only
             // way to do that...
-            openHABWidgetAdapter.stopVideoWidgets();
-            openHABWidgetAdapter.stopImageRefresh();
+            stopVisibleViewHolders();
```
maniac103: I wonder whether this is still needed now. Need to investigate.

FlorianSW:
> I made the test via the espresso recorder. When I run the test afterwards, the test failed... So the layout seems to change sometimes.

I already tried doing that for #639, too (as it was frustating selecting the correct spinner and thought it would be easier with the Espresso recorder), but after a simple test failed after recording and replaying it, I ended up doing it by my own again :)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -59,918 +55,1054 @@
...
+                MyAsyncHttpClient httpClient, ConnectionInfo connection) {
+            super(inflater, parent, R.layout.openhabwidgetlist_groupitem, httpClient, connection);
+            mLabelView = itemView.findViewById(R.id.widgetlabel);
+            mValueView = itemView.findViewById(R.id.widgetvalue);
+            mIconView = itemView.findViewById(R.id.widgetimage);
+        }
+
+        @Override
+        public void bind(OpenHABWidget widget) {
+            String[] splitString = widget.getLabel().split("\\[|\\]");
```
mueller-ma: Consider moving `"\\[|\\]"` to a global variable. That's a simple regex, but a variable name could describe it.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -59,918 +55,1054 @@
...
+        private int mRefreshRate;
+
+        ImageViewHolder(LayoutInflater inflater, ViewGroup parent,
+                MyAsyncHttpClient httpClient, ConnectionInfo connection) {
+            super(inflater, parent, R.layout.openhabwidgetlist_imageitem, httpClient, connection);
+            mImageView = itemView.findViewById(R.id.imageimage);
+
+            WindowManager wm = (WindowManager) parent.getContext().getSystemService(Context.WINDOW_SERVICE);
+            wm.getDefaultDisplay().getSize(mScreenSize);
+
```
mueller-ma: Unneeded new line

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -59,918 +55,1054 @@
...
+        }
+
+        @Override
+        public void bind(OpenHABWidget widget) {
+            View parent = (View) itemView.getParent();
+            // We scale the image at max 90% of the available height
+            int parentWidth = parent != null ? parent.getWidth() : 0;
+            int parentHeight = parent != null ? parent.getHeight() : 0;
+            mImageView.setMaxSize(parentWidth > 0 ? pare
```
mueller-ma: Isn't the item type always Image here?

maniac103: Yes, it is. It was taken straight from the old code, but as it's redundant, I'll remove it.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
maniac103: I actually thought about moving the split-label-to-label-and-value parsing to a common base class, but the handling is subtly different for each different type (fallback handling if there's less than 2 strings in the resulting array differs). I wonder whether that's intended or a slight bug which could be solved by using a base class?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
maniac103: Fixed

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
maniac103: Removed; also replaced the startsWith() with a more meaningful check.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -59,918 +55,1053 @@
...
+        mChartTheme = a.getString(1);
+
         a.recycle();
     }
 
-    @SuppressWarnings("deprecation")
+    public void update(List<OpenHABWidget> widgets) {
+        mItems.clear();
+        mItems.addAll(widgets);
+        notifyDataSetChanged();
```
9a4gl: This call is problem. notifyDataSetChanged(); does refresh all. In my PR there is
void reload(ArrayList<OpenHABWidget> oldItems, ArrayList<OpenHABWidget> newItems)
I had to implement requireRefresh on OpenHABWidget to support that.
I tested your branch and it does refresh my WebViews on page if something else refreshes.

maniac103: From the initial description of this PR:

> This allows more fine-grained change handling **later**

(emphasis added by me)

Having both RecyclerView conversion and update logic change is what bothered me in your PR, as (IMHO) it makes following the update logic changes harder, thus I believe the right thing to do is splitting those two items into separate PRs.

9a4gl: Ok, starting position for me was to fix unnecessary refreshes, not refactoring.. Refactoring was required to achieve it.. IMO, this is blocking (at least for me) to use application. So I suggest to proceed. Logic can go in next PR, that is fine. But is required.

maniac103:
Something is definitely off with Stickler.

mueller-ma:
Testing new stickler configs here: https://github.com/mueller-ma/openhab.android/pull/6

mueller-ma:
Do you want to fix #319 in this PR, too?

mobile/src/main/res/drawable/list_arrow_activated_tinted.xml
mueller-ma: I would go for colorControlHighlight

maniac103:
You mean the selection highlight? That one already should be fixed in this PR.

mueller-ma:
No, I was able to reproduce that bug.

mueller-ma:
Can you test landscape mode? It seems to me very buggy. I'll come up with some detailed information what's bugged

maniac103:
Yes, it certainly is buggy, which probably is related to the #319 - and which is what triggered #677. Honestly, the whole fragment management code is a complete and utter mess currently; I'm pretty sure the original author didn't understand how fragment state save/restore is meant to properly work. This is also what causes #319: On each rotation, one widget list fragment instance is leaked - initially, you have 2 list fragments, but when rotating say 5 times, you have 7 of them. One of them has the selection set correctly, but it isn't shown anymore.

lolodomo:
I can confirm the last point, see my issue #659.

mueller-ma:
Assume landscape and demo mode for both bugs:
1. Tap on "First floor" => First floor opens on the right side
2. Tap on "First floor" => Nothing happens, thats intended
3. Tap on "First floor" => The part from the right goes to the left, not sure if intended or not


Another bug:
1. Tap in "First floor"
2. Tap on Astro data => "First floor" is still marked as active

maniac103:
Let's leave the landscape mode issues for another PR. See #677 and the commit linked from there; it should fix all landscape mode issues.

maniac103:
Fixed the test & rebased. Should be good to merge now.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -55,921 +51,1075 @@
...
-                    //RollerShutterItem changed to RollerShutter in later builds of OH2
-                    if ("RollershutterItem".equals(openHABWidget.getItem().getType()) ||
-                            "Rollershutter".equals(openHABWidget.getItem().getType()) ||
-                            "Rollershutter".equals(openHABWidget.getItem().getGroupType()))
-                        return TYPE_ROLLERSHUTTER;
-                    else
-                        return TYPE_SWITCH;
-                } else
-                    return TYPE_SWITCH;
+
```
mueller-ma: Is there a reason why you have chosen OnTouchListener over OnClickListener? From what I read (https://stackoverflow.com/a/12811038/9053008) onTouch does not work with mouse and keyboard.

maniac103: Who is that 'you'? ;-) I just took what was already there, the OnTouchListener was part of that. FWIW, I wondered about the very same question myself though. In some cases (e.g. switch) there are even more specific change listeners we could use.

mueller-ma: Oh, I haven't checked what was before your change...

PR #661, opened 2018-01-22
Use pipe instead of slash in sed command

mueller-ma:
* Use https for all links 
* Prohibit pipe and semicolon in descriptions 

@kubawolanin @Confectrician You will have to resubmit your translation after this is merged.


Confectrician:
Hey @mueller-ma i think you have rsubmitted the german version already.
Something left to do for me or everything fine already?

mueller-ma:
No, German just hit 100%, so everything is done. Thanks for your translations.

PR #666, opened 2018-01-25
Remove obsolete code

cribskip:

While looking around, I noticed this huge code block which is never compiled. Should be cleaned up.
https://github.com/openhab/openhab-android/blob/a393c6faa24c10e8b488314ffac05c3e79c0a042/mobile/src/main/java/org/openhab/habdroid/core/GcmBroadcastReceiver.java#L35-L79


mueller-ma:
LGTM! Thanks

PR #667, opened 2018-01-25
Use time for notification time

### LINKED ISSUE #663: Notification time should be event time
Currently, the time that is shown in the notifications is the time when the notification was received by the device not the actual event time.

How to reproduce:
Set device to flight mode, send notification, wait a minute, go back online.

It should be be enough to add a `.setWhen(time_in_millis)` in GcmIntentService. Event time should be part of the message already.

cribskip:

Fixes #663 

**WARNING**
I could not get GCM working in the emulator, so this is untested but should work just fine.

### DIFF
--- a/mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
+++ b/mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
@@ -64,7 +64,11 @@ public class GcmIntentService extends IntentService {
                     notificationId = Integer.parseInt(intent.getExtras().getString(EXTRA_NOTIFICATION_ID));
                 }
                 if ("notification".equals(intent.getExtras().getString("type"))) {
-                    sendNotification(intent.getExtras().getString(EXTRA_MSG), notificationId);
+                    long time = System.currentTimeMillis();
+                    if (intent.getExtras().containsKey("time")) {
+                        time = Long.parseLong(intent.getExtras().getString("time")) * 1000;
+                    }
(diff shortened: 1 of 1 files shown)


mueller-ma:
@cribskip Can you rebase your PR to get rid of all the falsy reported @stickler-ci errors?

redeye86:
I tried to find a solution myself (c.f https://github.com/openhab/openhab-android/commit/7104cce9a8a5b4b2ae67e05389205a38308d81e6) and confirmed that it is working on my device. It seems that somewhere in the cloud service some information (time, severity, ...) is lost and not passed to android. In my implementation i used "google.sent_time", which is the time the gcm servers received the message from openhab cloud servers. This is not the actual event time, but i think the notifications are forwarded immediately anyway. I guess it would be best to use "time" if present (in case it gets implemented in the cloud service), otherwise "google.sent_time" and finally System.currentMillis() as a fallback.

cribskip:
@redeye86 Thanks, your implementation is sleeker than mine, this should be used.

@mueller-ma Please go for #673

maniac103:
> and finally System.currentMillis() as a fallback.

That's the default the Android framework chooses anyway if setWhen() isn't called. I'd call the timestamp extra rather `timestamp` instead of `time`, but I'd agree it would make sense to add it if there's a chance of implementing it in the cloud service (is this information available there?) ... currently it isn't sent: https://github.com/openhab/openhab-cloud/blob/master/app.js#L411

PR #668, opened 2018-01-25
Fix checkstyle Indentation module

mueller-ma:
This change was a great improvement to a local checkstyle run on MainActivity. I tried it in two repos, but stickler is still throwing tons of indent rule violations.
This has two possible reasons:

1. The checkstyle config is cached
2. It checkstyle config from the PR branch is taken

Anyway I would like to merge this one and @maniac103 please rebase #650 after this is merged.


maniac103:
@mueller-ma Rebase just to re-trigger Stickler you mean? If yes, I'll do that.

PR #673, opened 2018-01-26
Time in notification is now is sending time

### LINKED ISSUE #663: Notification time should be event time

redeye86:
Ensures that notifications show the actual event/sending time instead of the reception time on the device when you have a weak signal and receive the messages late.

Fixes #663
AFAIK the the google.sent_time is set when the cloud service sends the message to google, so technically it is not the actual event time, but


mueller-ma:
Is google.sent_time always provided or do we need any fallback?

mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
```diff
@@ -65,8 +64,20 @@ protected void onHandleIntent(Intent intent) {
                     notificationId = Integer.parseInt(intent.getExtras().getString(EXTRA_NOTIFICATION_ID));
                 }
                 if ("notification".equals(intent.getExtras().getString("type"))) {
-                    sendNotification(intent.getExtras().getString(EXTRA_MSG), intent.getExtras().getLong(EXTRA_SENT_TIME), notificationId);
-                // If this is hideNotification, cancel existing notification with it's id
+                    //we use local time as fallback
+                    long timestamp = System.currentTimeMillis();
+                    if(intent.getExtras().containsKey("timestamp")){
```
mueller-ma: Is there an open issue on openhab-cloud for timestamp?

mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
```diff
@@ -33,7 +33,6 @@
     private static final String TAG = GcmIntentService.class.getSimpleName();
 
     public static final String EXTRA_MSG = "message";
-    public static final String EXTRA_SENT_TIME = "google.sent_time";
```
mueller-ma: Can you use constants for both keys?

redeye86:
Ok, 

seems that i have messed it up. I accidentally merged my old attempt to implement home screen widgets. I will try to fix it....

redeye86:
ok, i made a new fork so that we have a clean PR without all the unrelated commits. Please ignore this pr and use #685 instead!

PR #678, opened 2018-01-27
Set the correct default value of icon format ListPreference

### LINKED ISSUE #670: After first start wrong icon format is fetched
While testing a PR I saw this log messages on a device set to German.
In the setting no value is selected for "Icon format". After selecting "Pixelgrafik" (Bitmap) the correct icon format is fetched.

````
01-26 16:29:21.943 10051-10151/org.openhab.habdroid I/MyWebImage: fetched bitmap for http://demo.openhab.org:8080/icon/office?state=null&format=Pixelgrafik
01-26 16:29:21.949 10051-10153/org.openhab.habdroid I/MyWebImage: fetched bitmap for http://demo.openhab.org:8080/icon/boy_1?state=null&format=Pixelgrafik
01-26 16:29:21.959 10051-10156/org.openhab.habdroid I/MyWebImage: fetched bitmap for http://demo.openhab.org:8080/icon/bedroom?state=null&format=Pixelgrafik
````

FlorianSW:
The default value does not need to be one of the elements in entries, but
one of the elements entryValues, otherwise the default value returned by
SharedPreferences#getString() will be the display value, not the correct
value itself.

Fixes #670

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetListFragment.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetListFragment.java
@@ -17,6 +17,7 @@ import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.os.Handler;
 import android.preference.PreferenceManager;
+import android.support.annotation.NonNull;
 import android.support.v4.app.ListFragment;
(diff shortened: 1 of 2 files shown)


mueller-ma:
LGTM!

PR #685, opened 2018-01-31
Time in notification is now is sending time / 2nd try

redeye86:
Successor of #673, now with clean code from a fresh fork (without all the other commits)

Ensures that notifications show the actual event/sending time instead of the reception time on the device when you have a weak signal and receive the messages late.

### DIFF
--- a/mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
+++ b/mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
@@ -33,6 +33,8 @@ public class GcmIntentService extends IntentService {
     private static final String TAG = GcmIntentService.class.getSimpleName();
 
     public static final String EXTRA_MSG = "message";
+    public static final String EXTRA_TIMESTAMP = "timestamp";
+    public static final String EXTRA_GCM_TIMESTAMP = "google.sent_time";
     public static final String EXTRA_NOTIFICATION_ID = "notificationId";
     public static final String ACTION_NOTIFICATION_SELECTED = "org.openhab.notification.selected";
     public static final String ACTION_NOTIFICATION_DELETED = "org.openhab.notification.deleted";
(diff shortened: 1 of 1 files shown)


mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
```diff
@@ -64,7 +66,19 @@ protected void onHandleIntent(Intent intent) {
...
+                    long timestamp = System.currentTimeMillis();
+                    if(intent.getExtras().containsKey(EXTRA_TIMESTAMP)){
+                        //in case timestamp is passed by cloud service in later versions
+                        timestamp = intent.getExtras().getLong(EXTRA_TIMESTAMP);
+                    }else if(intent.getExtras().containsKey(EXTRA_GCM_TIMESTAMP)){
+                        //use google.sent_time for now as it is the only source
+                        //of the event time available right now
+                        //(or rather the time cloud service forwareded it to gcm)
+                        timestamp = intent.getExtras().getLong(EXTRA_GCM_TIMESTAMP);
+                    }
```
maniac103: Could also simplify this logic:

```
long timestamp = intent.getLongExtra(EXTRA_TIMESTAMP, 0);
if (timestamp == 0) {
    timestamp = intent.getLongExtra(EXTRA_GCM_TIMESTAMP, 0);
}
if (timestamp == 0) {
   timestamp = System.currentTimeMillis();
}
```

or even

```
long timestamp = intent.getLongExtra(EXTRA_TIMESTAMP,
        intent.getLongExtra(EXTRA_GCM_TIMESTAMP, System.currentTimeMillis());
```

The latter certainly deserves a comment, though ;-)

mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
```diff
@@ -64,7 +66,19 @@ protected void onHandleIntent(Intent intent) {
                     notificationId = Integer.parseInt(intent.getExtras().getString(EXTRA_NOTIFICATION_ID));
                 }
                 if ("notification".equals(intent.getExtras().getString("type"))) {
-                    sendNotification(intent.getExtras().getString(EXTRA_MSG), notificationId);
+                    //we use local time as fallback
+                    long timestamp = System.currentTimeMillis();
+                    if(intent.getExtras().containsKey(EXTRA_TIMESTAMP)){
+                        //in case timestamp is passed by cloud service in later versions
+                        timestamp = intent.getExtras().getLong(EXTRA_TIMESTAMP);
+                    }else if(intent.getExtras().containsKey(EXTRA_GCM_TIMESTAMP)){
```
maniac103: If you don't want to simplify as per comment below, please keep spaces before/after curly braces and after keywords such as `if`.

mobile/src/full/java/org/openhab/habdroid/core/GcmIntentService.java
```diff
@@ -64,7 +66,19 @@ protected void onHandleIntent(Intent intent) {
                     notificationId = Integer.parseInt(intent.getExtras().getString(EXTRA_NOTIFICATION_ID));
                 }
                 if ("notification".equals(intent.getExtras().getString("type"))) {
-                    sendNotification(intent.getExtras().getString(EXTRA_MSG), notificationId);
+                    //we use local time as fallback
+                    long timestamp = System.currentTimeMillis();
+                    if(intent.getExtras().containsKey(EXTRA_TIMESTAMP)){
```
mueller-ma: Are there any plans on openhab-cloud to support this?

redeye86: I created an issue in the cloud repository at least: https://github.com/openhab/openhab-cloud/issues/160

mueller-ma: Should we merge this change before the feature exists server side? I dont think so.  We can wait until its in openhab-cloud or just merge the time from gcm for now and timestamp later.

redeye86:
I removed the timestamp stuff now. 

Once openhab-cloud#160 is resolved, it should be considered to support both, "gcm.sent_time" and "timestamp" for a while as the changes for both, cloud service and app, are not be rolled out immediately and at the same time. On the other hand we still have the fallback.

redeye86:
As far as i remember i commited using the -S option. I also have this verified badge behind my comment log. Or am I missing something?

commit

mueller-ma:
I'm talking about the option `-s` which adds a line at the commit message "Signed-of-by..."

mueller-ma:
Thanks!

PR #687, opened 2018-02-01
Improve look'n'feel of section switch button

### LINKED ISSUE #156: Switch item [mappings] are not sized correct (OH2)
using [mapping] eg for scenes do overlay in text and / or icon instead of resizing correct
on ios this works nice

this happens only with OH2
##

maniac103:
Should fix #156


maniac103:
See https://github.com/openhab/openhab-android/issues/156#issuecomment-360726376 for comparison screenshot.

mueller-ma:
Looks good so far, but I found these three things:
1. On a tablet in landscape mode the buttons could still have the old size without overlapping the label
2. When I tap on a button, this one and the old state are selected until the command is sent. Is this intended?
3. As you can see on your screen, the text color for unselected buttons and e.g. setpoint are different. This was the case before, too, but maybe you want to change it in this PR.

maniac103:
> On a tablet in landscape mode the buttons could still have the old size without overlapping the label

I don't really like the idea of using different font sizes for different screen layouts. We don't do this for selections and setpoints either, and honestly, I found the font size much too large from the beginning of my usage of the app.

> When I tap on a button, this one and the old state are selected until the command is sent. Is this intended?

Nope, that's fixed now. I thought of introducing another 'half-active' state which is used between releasing the button and the state update appearing, to give feedback to the user. One complication this would introduce is that we'd need to feed the result of the REST API call to the view to clear the 'half-active' state if the API call fails. I'm not sure whether the UX improvement is worth that effort - thoughts?

> As you can see on your screen, the text color for unselected buttons and e.g. setpoint are different. This was the case before, too, but maybe you want to change it in this PR.

Actually there's 3 font colors here:
- Selection widget -> `?attr/textColorPrimary`
- Inactive buttons -> `@color/normal`
- Text and setpoint widgets -> `?attr/textColorSecondary`

IMHO the light grey is fine for the inactive buttons because it highlights their inactive state. I'd rather make selections be consistent to text widgets...

mueller-ma:
> IMHO the light grey is fine for the inactive buttons because it highlights their inactive state.

I think the orange background makes it clear enough to distinguish between active and inactive.

maniac103:
I shall restore the value view. Thought it's not needed, but my rollershutters are an example of it being necessary ;-)

maniac103:
Value view restored. Only open question is the font color, but that also could be deferred to a later PR.

maniac103:
Nope, should be fine.

Issue #677, opened 2018-01-27
Remove usage of ViewPager for Widgetlist fragments


FlorianSW:
In #640 @maniac103 suggested:

> Actually the ViewPager is used: for paging the fragments in multi-pane layout. Having two containers to insert the respective fragments into would likely be a much cleaner solution, though.

mueller-ma:
Please wait for #675

maniac103:
Too late: https://github.com/maniac103/openhab.android/tree/cleanup-fragment-management ;-)

(based on #650)

maniac103:
Notification fragment is fine now; just need someone to test going to sitemap item via NFC

PR #688, opened 2018-02-01
Move from ViewPager to explicitly assigning fragments to container.

### LINKED ISSUE #659: Changing tablet orientation: fragment duplication
Testing my SSE code, I discovered a big and annoying bug.
First I start the app. I can see, as expected, 1 onResume for OpenHABWidgetListFragment which finally leads to 1 async long polling HTTP call for the updates.
Then I change the orientation of my tablet, I can now see 2 onResume for OpenHABWidgetListFragment (same page) which finally lead to 2 async long polling HTTP call for the updates.
Then I change the orientation again, I can now see 3 onResume for OpenHABWidgetListFragment (same page) which finally lead to 3 async long polling HTTP call for the updates.
And so on.

It looks like fragments are not really cleared.

I have not yet found the solution but I can see that in one side onResume on the activity is updating the list of fragments using the state fragment and trigger a refresh. This refresh of the data source is triggering the call of onResume for each fragment which  loads the page and initialize the async long polling call.
I can see in parallel that each fragment is going through all its lifecycle: onAttach onCreate onCreateView onViewCreated onActivityCreated and finally onResume which again loads the page and initialize the async long polling call.

It looks like the state fragment is ok as I have the right number of fragments.
But apparently, onResume is still run for normally suppressed fragments.

maniac103:
Fixes #659, #677, #399


maniac103:
While this works fine (and fixes all of the landscape bugs), there's one (important?) behavioral change: unlike before, fragments are now destroyed when moving to the next one, which means when coming back to the original page its data needs to be fetched again as the long polling request is shut down when the fragment is torn down.
One could fix this by moving the polling control (or polling itself) to a state retaining fragment controlled by ActivityController (as it knows what fragments are on the stack anyway), but I wonder whether that problem is actually worth the complication?

lolodomo:
Super.
I will wait for this new refactoring to be merged and then plug again my code to update the widgets from SSE messages. I will have to understand all the recent and massive changes.

lolodomo:
Fetching again the data when moving to a new fragment is not a problem. By the way, it was already the case before !
If you destroy fragments that are not visible to the user, I think that it is perfect.
It could just make the back button a little slower than before ?

maniac103:
> Fetching again the data when moving to a new fragment is not a problem. By the way, it was already the case before !

Yes, but only when going from A to B, not when going back to A: A's data was still up-to-date, wasn't it?

> If you destroy fragments that are not visible to the user, I think that it is perfect.

That's the case currently with this PR.

> It could just make the back button a little slower than before ?

And that's the potential disadvantage ;-) It's solvable by having a global connection holder, prototype here: https://github.com/maniac103/openhab.android/tree/global-connection-holder ... I'm just not sure whether the added complexity is worth it.

maniac103:
> I will wait for this new refactoring to be merged and then plug again my code to update the widgets from SSE messages. I will have to understand all the recent and massive changes.

SSE handling is actually easier for you with RecyclerView:
- get list of items from adapter
- find position of updated widget by comparing ID sent via SSE to IDs of items in widget list
- call adapter.notifyItemChanged(position) for every changed widget

Without SSE, it's a bit more complicated though, as we need to (deeply) compare old and new items.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -288,24 +280,47 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
         mViewPool = new RecyclerView.RecycledViewPool();
         MemorizingTrustManager.setResponder(this);
 
         // Check if we have openHAB page url in saved instance state?
         if (savedInstanceState != null) {
             openHABBaseUrl = savedInstanceState.getString("openHABBaseUrl");
-            sitemapRootUrl = savedInstanceState.getString("sitemapRootUrl");
-            mStartedWithNetworkConnectivityInfo = savedInstanceState.getParcelable("startedWithNetworkConnectivityInfo");
+            mLastKnownNetworkInfo = savedInstanceState.getParcelable("lastNetworkInfo");
```
mueller-ma: Constants for "lastNetworkInfo" and others?

mueller-ma:
I really like the new animations, which look a lot better and smoother than the old ones.
Two things:
- The bug with rotating to landscape and have nothing selected is fixed, but it would be nice not to have to scroll the left side down until the selected one is visible.
- The animation for going to a subpage is kind of laggy on my tablet: The view moves ~0.5cm to the left, waits and then it resumes the motion until the screen is occupied by the subpage.

lolodomo:
@maniac103 : thank you for the explanation on how I should trigger the update, that looks excellent if only updated widgets will be really updated.

Yes of course long polling will be kept and be used for any servers that were built before 2018.

maniac103:
> The bug with rotating to landscape and have nothing selected is fixed, but it would be nice not to have to scroll the left side down until the selected one is visible.

That one should be easy to fix, will do that.

> The animation for going to a subpage is kind of laggy on my tablet: The view moves ~0.5cm to the left, waits and then it resumes the motion until the screen is occupied by the subpage

I've noticed that one myself. I suspect it happens because the page data response happens in the middle of the animation, which leads to UI thread being busy with view inflation and binding. Not sure what the best way to fix this is; will need to profile to see whether the above suspicion is correct.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
maniac103: Yeah, sure, but maybe leave it to the separate cleanup work to do it consistently everywhere?

maniac103:
> I've noticed that one myself. I suspect it happens because the page data response happens in the middle of the animation, which leads to UI thread being busy with view inflation and binding. Not sure what the best way to fix this is; will need to profile to see whether the above suspicion is correct.

Turns out the suspicion actually is correct: the stutter happens because of the view creation and binding in the RecyclerView. One can optimize that somewhat (see e.g. https://github.com/openhab/openhab-android/pull/687/commits/cbb4c7ae2baa7b6d5f551104fcb7ce267c351772 - there's more potential [1]), and it'll improve when going back and forth due to the shared recycled view pool, but initial view creation will stay a problem: profiling shows that when entering e.g. the widget overview page on the demo server, main thread is completely busy with around 35% view creation and 35% view binding (30% are other things).
Not sure what to do there; all options have issues:
- Delay fragment transition until page data is loaded? This will move initial view creation to fragment view creation (and thus to the start of the animation), but requires complicated setup
- Delay content processing (and thus view creation) until end of fragment transition ... but animating an empty fragment with views suddenly popping into place doesn't seem nice either...

[1] When opening the 'Cellar' page in the demo server, 4% of CPU time is spent in generating the exceptions in OpenHABItem.getStateAsBrightness(), called via OpenHABItem.getStateAsBoolean() ... stuff like this seems like easy optimization potential, e.g. by evaluating the states to int/float/HSV on initialization, which happens in a background thread, and using the cached values afterwards. Same applies to OpenHAB2Widget.getIconPath() - could also cache the result here and save another 4% of CPU time.

mueller-ma:
> Please give [2] a try and let me know about your findings.

At first I thought this will make the app kind of unresponsive when going to a subpage, but it turns out the "lag" is minimal and you dont even notice it when not looking for it. The widget page in demo mode needs a bit longer, but the load icon appears, so the user knows, that the app is loading.
I would feel comfortable with adding this option.

mueller-ma:
Do you want to add cbb4c7ae2baa7b6d5f551104fcb7ce267c351772 and https://github.com/maniac103/openhab.android/tree/global-connection-holder to this PR?

maniac103:
The section switch binding improvement is part of #687 already, so if you don't have strong feelings where to put it I'd prefer to leave it there.
About the two commits in that mentioned branch, I can add them here, sure. Just need to add some Javadoc comments explaining how everything is supposed to work.

maniac103:
Also need to stop connections in onStop/restart in onStart.

maniac103:
Connections are stopped in background now, and fixed fragment restoration after screen rotation. Still need to add Javadoc.

maniac103:
Started rebasing this (and adding some follow-up fixes that'll likely land in a follow-up PR) on top of #562 here. In case somebody wants to try: https://github.com/maniac103/openhab.android/tree/cleanup-fragment-management-after-pr562

maniac103:
Rebased to master. Worth a try now :-)

maniac103:
> Crash at start:

Oops, looks like I failed at rebasing. The mSitemapList assignment in onCreate needs to happen before the setupDrawer call.

lolodomo:
I will test next weekend.
Once merged, I hope I can have a frame without massive code refactoring so that I can try reintroducing my SSE code.

maniac103:
> Once merged, I hope I can have a frame without massive code refactoring so that I can try reintroducing my SSE code.

Bad news: I have more refactoring coming (https://github.com/maniac103/openhab.android/tree/navdrawer-cleanup, https://github.com/maniac103/openhab.android/tree/autovalue and https://github.com/maniac103/openhab.android/tree/gcm-refactor) ... good news: none of this should interfere with the SSE changes, which now should be in the connection holder fragment only with a tiny bit of plumbing into the widget list fragment. Thinking about it, I can already add the plumbing in this PR, so the SSE changes are isolated to the connection holder fragment.

maniac103:
@lolodomo With the above commit, the SSE changes should be isolated to `PageConnectionHolderFragment`. Impact by further refactoring should be minimal (the only potentially related thing is that the autovalue branch changes the signature of the `OpenHABWidget` factory method, but adapting that should be trivial).

mueller-ma:
On change from landscape to portrait or via verca:
````
03-01 17:12:51.553 10216-10216/org.openhab.habdroid E/AndroidRuntime: FATAL EXCEPTION: main
                                                                      Process: org.openhab.habdroid, PID: 10216
                                                                      java.lang.RuntimeException: Unable to start activity ComponentInfo{org.openhab.habdroid/org.openhab.habdroid.ui.OpenHABMainActivity}: java.lang.IllegalStateException: Can't change container ID of fragment OpenHABWidgetListFragment{acf65a3 #17 id=0x7f09004f}: was 2131296335 now 2131296332
                                                                          at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2325)
                                                                          at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2387)
                                                                          at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:3947)
                                                                          at android.app.ActivityThread.access$900(ActivityThread.java:151)
                                                                          at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1309)
                                                                          at android.os.Handler.dispatchMessage(Handler.java:102)
                                                                          at android.os.Looper.loop(Looper.java:135)
                                                                          at android.app.ActivityThread.main(ActivityThread.java:5254)
                                                                          at java.lang.reflect.Method.invoke(Method.java:372)
                                                                          at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
                                                                          at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
                                                                       Caused by: java.lang.IllegalStateException: Can't change container ID of fragment OpenHABWidgetListFragment{acf65a3 #17 id=0x7f09004f}: was 2131296335 now 2131296332
                                                                          at android.support.v4.app.BackStackRecord.doAddOp(BackStackRecord.java:420)
                                                                          at android.support.v4.app.BackStackRecord.replace(BackStackRecord.java:441)
                                                                          at android.support.v4.app.BackStackRecord.replace(BackStackRecord.java:432)
                                                                          at org.openhab.habdroid.ui.activity.FragmentControllerOnePane.updateFragmentState(FragmentControllerOnePane.java:40)
                                                                          at org.openhab.habdroid.ui.activity.FragmentController.updateFragmentState(FragmentController.java:239)
                                                                          at org.openhab.habdroid.ui.OpenHABMainActivity.onCreate(OpenHABMainActivity.java:336)
                                                                          at android.app.Activity.performCreate(Activity.java:5990)
                                                                          at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1106)
                                                                          at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2278)
                                                                          at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2387) 
                                                                          at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:3947) 
                                                                          at android.app.ActivityThread.access$900(ActivityThread.java:151) 
                                                                          at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1309) 
                                                                          at android.os.Handler.dispatchMessage(Handler.java:102) 
                                                                          at android.os.Looper.loop(Looper.java:135) 
                                                                          at android.app.ActivityThread.main(ActivityThread.java:5254) 
                                                                          at java.lang.reflect.Method.invoke(Method.java:372) 
                                                                          at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903) 
                                                                          at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698) 
````

**Comment by mueller-ma** [comment]
I configured a wrong user name and pw. The hamburger icon is frozen during its animation:
screenshot_2018-03-01-17-17-37

**Comment by mueller-ma** [comment]
In debug mode the stacktrace font is pretty small and the view is not scrollable. Maybe hide the icon is debug mode:
screenshot_2018-03-01-17-21-02

**Comment by maniac103** [comment]
> On change from landscape to portrait or via verca:

Huh. While the exception kinda makes sense, I wonder why I don't see it in the emulator. I'll check it.

> I configured a wrong user name and pw. The hamburger icon is frozen during its animation:

And it stays like that? And touching it makes it work again?

> In debug mode the stacktrace font is pretty small and the view is not scrollable. Maybe hide the icon is debug mode

I'd rather make the stacktrace expandable with a 'show details' button. Thinking about it, what's the purpose of showing a stack trace to the user though?

**Comment by maniac103** [comment]
Exception fixed.

**Comment by mueller-ma** [comment]
> And it stays like that? And touching it makes it work again?

Cannot reproduce :(


I had the stacktrace shown and entered the settings, when this stacktrace occured. I cannot reproduce this, too, but maybe the stacktrace is helpful for you:

````
03-01 18:19:12.484 14438-14438/org.openhab.habdroid E/AndroidRuntime: FATAL EXCEPTION: main
                                                                      Process: org.openhab.habdroid, PID: 14438
                                                                          at android.support.v4.app.FragmentManagerImpl.checkStateLoss(FragmentManager.java:2044)
                                                                          at android.support.v4.app.FragmentManagerImpl.ensureExecReady(FragmentManager.java:2175)
                                                                          at android.support.v4.app.FragmentManagerImpl.execSingleAction(FragmentManager.java:2195)
                                                                          at android.support.v4.app.BackStackRecord.commitNow(BackStackRecord.java:645)
                                                                          at org.openhab.habdroid.ui.activity.FragmentControllerTwoPane.updateFragmentState(FragmentControllerTwoPane.java:73)
                                                                          at org.openhab.habdroid.ui.activity.FragmentController.updateFragmentState(FragmentController.java:239)
                                                                          at org.openhab.habdroid.ui.activity.FragmentController.indicateServerCommunicationFailure(FragmentController.java:173)
                                                                          at org.openhab.habdroid.ui.OpenHABMainActivity$DefaultHttpResponseHandler.onFailure(OpenHABMainActivity.java:200)
                                                                          at org.openhab.habdroid.util.MyAsyncHttpClient$1$2.run(MyAsyncHttpClient.java:73)
                                                                          at android.os.Looper.loop(Looper.java:135)
                                                                          at android.app.ActivityThread.main(ActivityThread.java:5254)
                                                                          at java.lang.reflect.Method.invoke(Method.java:372)
                                                                          at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
                                                                          at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
````

> I'd rather make the stacktrace expandable with a 'show details' button. Thinking about it, what's the purpose of showing a stack trace to the user though?

I think you're right, the stacktrace doesn't make a log of sence here. But I would keep the exception message.

File: mobile/src/main/res/values/strings.xml [hunk 2]
```diff
@@ -173,8 +173,9 @@
     <string name="try_again_button">Try again</string>
     <string name="network_error">Network error</string>
     <string name="go_to_settings_button">Go to settings</string>
+    <string name="configuration_missing">Your OpenHAB server could not be discovered automatically. Please configure its IP address or host name in the server settings.</string>
```
mueller-ma: "openHAB"

mobile/src/main/res/values/strings.xml
```diff
@@ -173,8 +173,9 @@
...
+    <string name="resolving_openhab">Trying to discover OpenHAB server instance\u2026</string>
```
mueller-ma: "openHAB"

mueller-ma:
One more non-reproducable bugs. After playing a bit around (entering sub pages, chaning orientation), I could not go to a different sitemap. When I clicked on it in the drawer nothing happend (tried it several times):
````
03-01 18:37:21.368 15579-15579/org.openhab.habdroid D/OpenHABDrawerAdapter: No count label
03-01 18:37:21.377 15579-15579/org.openhab.habdroid D/OpenHABDrawerAdapter: Not showing count 0
03-01 18:37:21.378 15579-15579/org.openhab.habdroid D/OpenHABDrawerAdapter: Not showing count 0
03-01 18:37:22.037 15579-15579/org.openhab.habdroid D/OpenHABMainActivity: Drawer selected item 1
03-01 18:37:22.037 15579-15579/org.openhab.habdroid D/OpenHABMainActivity: This is sitemap http://openhab:8080/rest/sitemaps/sitemapname
03-01 18:37:22.037 15579-15579/org.openhab.habdroid I/OpenHABMainActivity: Opening sitemap at http://openhab:8080/rest/sitemaps/sitemapname/sitemapname
03-01 18:37:24.560 15579-15579/org.openhab.habdroid D/OpenHABMainActivity: Loading sitemap list from /rest/sitemaps
03-01 18:37:25.354 15579-15579/org.openhab.habdroid D/OpenHABMainActivity: <sitemap json here>
03-01 18:37:25.367 15579-15579/org.openhab.habdroid D/OpenHABDrawerAdapter: No count label
03-01 18:37:25.378 15579-15579/org.openhab.habdroid D/OpenHABDrawerAdapter: Not showing count 0
03-01 18:37:25.379 15579-15579/org.openhab.habdroid D/OpenHABDrawerAdapter: Not showing count 0
03-01 18:37:27.212 15579-15579/org.openhab.habdroid D/OpenHABMainActivity: Drawer selected item 1
03-01 18:37:27.212 15579-15579/org.openhab.habdroid D/OpenHABMainActivity: This is sitemap http://openhab:8080/rest/sitemaps/sitemapname
03-01 18:37:27.212 15579-15579/org.openhab.habdroid I/OpenHABMainActivity: Opening sitemap at http://openhab:8080/rest/sitemaps/sitemapname/sitemapname
03-01 18:37:30.608 15579-15579/org.openhab.habdroid D/OpenHABMainActivity: Loading sitemap list from /rest/sitemaps
03-01 18:37:30.659 15579-15579/org.openhab.habdroid D/OpenHABMainActivity: <sitemap json here>
03-01 18:37:30.672 15579-15579/org.openhab.habdroid D/OpenHABDrawerAdapter: No count label
03-01 18:37:30.681 15579-15579/org.openhab.habdroid D/OpenHABDrawerAdapter: Not showing count 0
03-01 18:37:30.683 15579-15579/org.openhab.habdroid D/OpenHABDrawerAdapter: Not showing count 0
03-01 18:37:31.123 15579-15579/org.openhab.habdroid D/OpenHABMainActivity: Drawer selected item 1
03-01 18:37:31.123 15579-15579/org.openhab.habdroid D/OpenHABMainActivity: This is sitemap http://openhab:8080/rest/sitemaps/sitemapname
03-01 18:37:31.123 15579-15579/org.openhab.habdroid I/OpenHABMainActivity: Opening sitemap at http://openhab:8080/rest/sitemaps/sitemapname/sitemapname
````

mueller-ma:
Do you think a vertical line between the two sides in landscape will look good?

Two "issues" that might be not worth to fix as users are not changing orientation every few seconds::

When the first frame has a title, I select the first subpage in this frame and change the orientation to landscape, I need to scroll up to see the frame title.

Images seems to be refetched when changing orientation.

maniac103:
Drawer toggle and strings fixed.

> I had the stacktrace shown and entered the settings, when this stacktrace occured. I cannot reproduce this, too, but maybe the stacktrace is helpful for you:

Yes, it is. We need to kill off pending HTTP requests when the activity is going to the background, as we only can save fragment updates when the acitivity is in the foreground. I'll look into it.

> One more non-reproducable bugs. After playing a bit around (entering sub pages, chaning orientation), I could not go to a different sitemap. When I clicked on it in the drawer nothing happend (tried it several times):

Only thing I can imagine is that both sitemaps have the same link? Can you check this? (See the check at the beginning of `openSitemap`)

> When I go back from settings and there was an error before, the error is still shown instead of a progress icon.

Did you change anything in settings at this point?

maniac103:
> Do you think a vertical line between the two sides in landscape will look good?

Yeah, a thin grey line probably will look better.

> When the first frame has a title, I select the first subpage in this frame and change the orientation to landscape, I need to scroll up to see the frame title.

I'll check whether there's a method that moves the list as less as possible, but it's probably low priority.

> Images seems to be refetched when changing orientation.

Yes, as all views are recreated. The image cache should take care of it though.

maniac103:
> Sometimes after going to settings, clearing the url and going back, I get the "Your OpenHAB server could not be discovered automatically. Please configure its IP address or host name in the server settings" error instantly. Log doesn't contain much:

Is there anything in the log before that?

mueller-ma:
> Did you change anything in settings at this point?

You can enable or disable demo mode. Than you see, that the stacktrace doesn't appear or disappear until the error occurs (see in log)

> Is there anything in the log before that? But anyway, onResume should trigger onConnectionChange, which triggers the resolving.
Edit: if it's instantly ... did you clear both URLs? If that's the case, that message is to be expected.

No, that's all from the time I clicked on back.
The remote url was empty before and I cleared the local url. Why is it expected, that discovery doesnt work, if no url is set?

maniac103:
> The remote url was empty before and I cleared the local url. Why is it expected, that discovery doesnt work, if no url is set?

I guess discovery ran before already. The way it's set up right now discovery is only attempted once.

maniac103:
So for discovery, under what circumstances do we want to run it?
- Only in 'local' environments, I guess?
- Should it run if a local connection is configured, but not reachable?
- Should it run more than once (that is, re-run if settings or network change after it failed before)?

About the error not disappearing when going to demo mode, please try now.

mueller-ma:
>  Only in 'local' environments, I guess?

Well, we can only discorver local server (Wifi, wimax, vpn). As we don't know, if a local server is available, we should always discover if no local url is configured.
This leads to the case where you're on a different wifi and want to connect to oh cloud, but discovery is started, fails and switches the app to demo mode. So if we have a remote url, discovery should be disabled.

>   Should it run if a local connection is configured, but not reachable?

No, see above.

>   Should it run more than once (that is, re-run if settings or network change after it failed before)?

I would say yes, it should.

maniac103:
Adapted service logic as per above.

mueller-ma:
I already tried it and the discovery seems to be fixed now.

I looked through all comments and think these issues are still open:

> One more non-reproducable bugs. After playing a bit around (entering sub pages, chaning orientation), I could not go to a different sitemap. When I clicked on it in the drawer nothing happend (tried it several times)

Can you add some debug logging? If it occurs next time, we have more information.

> I think you're right, the stacktrace doesn't make a log of sense here. But I would keep the exception message.

Still shown.

> When I go back from settings and there was an error before, the error is still shown instead of a progress icon.

Better explanation:
1. Debug mode enabled => debug information are shown
2. Go to settings and disable debug mode
3. Go back
4. Debug information are still shown for a short time

maniac103:
The items in your last comment should be fixed now.
(BTW, I think we should use task lists in those todo lists ;-) )

mueller-ma:
Can you have look at the failing tests?

mueller-ma:
Portrait => select a subpage => change to landscape => select something on the left side:
````
03-07 15:55:27.148 3738-3738/org.openhab.habdroid E/AndroidRuntime: FATAL EXCEPTION: main
                                                                    Process: org.openhab.habdroid, PID: 3738
                                                                    java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String android.content.Context.getPackageName()' on a null object reference
                                                                        at android.preference.PreferenceManager.getDefaultSharedPreferencesName(PreferenceManager.java:374)
                                                                        at android.preference.PreferenceManager.getDefaultSharedPreferences(PreferenceManager.java:369)
                                                                        at org.openhab.habdroid.ui.activity.PageConnectionHolderFragment.updateActiveConnections(PageConnectionHolderFragment.java:110)
                                                                        at org.openhab.habdroid.ui.activity.FragmentController.updateConnectionState(FragmentController.java:412)
                                                                        at org.openhab.habdroid.ui.activity.FragmentController.openPage(FragmentController.java:165)
                                                                        at org.openhab.habdroid.ui.activity.FragmentControllerTwoPane.openPage(FragmentControllerTwoPane.java:102)
                                                                        at org.openhab.habdroid.ui.OpenHABMainActivity.onWidgetSelected(OpenHABMainActivity.java:985)
                                                                        at org.openhab.habdroid.ui.OpenHABWidgetListFragment.onItemClicked(OpenHABWidgetListFragment.java:123)
                                                                        at org.openhab.habdroid.ui.OpenHABWidgetAdapter.onClick(OpenHABWidgetAdapter.java:299)
                                                                        at android.view.View.performClick(View.java:4780)
                                                                        at android.view.View$PerformClick.run(View.java:19866)
                                                                        at android.os.Looper.loop(Looper.java:135)
                                                                        at android.app.ActivityThread.main(ActivityThread.java:5254)
                                                                        at java.lang.reflect.Method.invoke(Method.java:372)
                                                                        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
                                                                        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
````

Activity title doesn't change (in demo mode):
1. Ground floor (Activity title is Main Menu)
2. Kitchen (Activity title is Main Menu)
3. Back  (Activity title is Ground floor)

maniac103:
Tests fixed, as well as both of the issues you found (both were a regression of https://github.com/openhab/openhab-android/pull/688/commits/eaba22ef41a36c7932322f9f1218d10f9d9e267d).

maniac103:
Thinking about it, I can probably simplify the test fix and remove the test only method from `PageConnectionHolderFragment`, as the progress spinner is shown during that time anyway.

Besides that, I still want to rename `FragmentController` to `ContentController` and add some Javadoc to `PageConnectionHolderFragment`. When this is done, I consider this PR to be finished if no more bugs pop up. The mentioned changes at least shouldn't introduce new ones :-)

mueller-ma:
Two more when leaving the about screen:

````
03-07 18:28:02.126 6564-6564/org.openhab.habdroid E/AndroidRuntime: FATAL EXCEPTION: main
                                                                    Process: org.openhab.habdroid, PID: 6564
                                                                    java.lang.RuntimeException: Unable to resume activity {org.openhab.habdroid/org.openhab.habdroid.ui.OpenHABMainActivity}: java.lang.IllegalStateException: FragmentManager is already executing transactions
                                                                        at android.app.ActivityThread.performResumeActivity(ActivityThread.java:2986)
                                                                        at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3017)
                                                                        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1347)
                                                                        at android.os.Handler.dispatchMessage(Handler.java:102)
                                                                        at android.os.Looper.loop(Looper.java:135)
                                                                        at android.app.ActivityThread.main(ActivityThread.java:5254)
                                                                        at java.lang.reflect.Method.invoke(Method.java:372)
                                                                        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
                                                                        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
                                                                     Caused by: java.lang.IllegalStateException: FragmentManager is already executing transactions
                                                                        at android.support.v4.app.FragmentManagerImpl.ensureExecReady(FragmentManager.java:2167)
                                                                        at android.support.v4.app.FragmentManagerImpl.execSingleAction(FragmentManager.java:2195)
                                                                        at android.support.v4.app.BackStackRecord.commitNow(BackStackRecord.java:645)
                                                                        at org.openhab.habdroid.ui.activity.FragmentControllerTwoPane.updateFragmentState(FragmentControllerTwoPane.java:73)
                                                                        at org.openhab.habdroid.ui.activity.FragmentController.onPageUpdated(FragmentController.java:377)
                                                                        at org.openhab.habdroid.ui.activity.PageConnectionHolderFragment$ConnectionHandler.triggerUpdate(PageConnectionHolderFragment.java:188)
                                                                        at org.openhab.habdroid.ui.activity.PageConnectionHolderFragment.triggerUpdate(PageConnectionHolderFragment.java:143)
                                                                        at org.openhab.habdroid.ui.activity.FragmentController.triggerPageUpdate(FragmentController.java:298)
                                                                        at org.openhab.habdroid.ui.OpenHABMainActivity.triggerPageUpdate(OpenHABMainActivity.java:610)
                                                                        at org.openhab.habdroid.ui.OpenHABWidgetListFragment.onStart(OpenHABWidgetListFragment.java:211)
                                                                        at android.support.v4.app.Fragment.performStart(Fragment.java:2380)
                                                                        at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1458)
                                                                        at android.support.v4.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManager.java:1740)
                                                                        at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1809)
                                                                        at android.support.v4.app.FragmentManagerImpl.dispatchStateChange(FragmentManager.java:3217)
                                                                        at android.support.v4.app.FragmentManagerImpl.dispatchStart(FragmentManager.java:3176)
                                                                        at android.support.v4.app.FragmentController.dispatchStart(FragmentController.java:203)
                                                                        at android.support.v4.app.FragmentActivity.onStart(FragmentActivity.java:570)
                                                                        at android.support.v7.app.AppCompatActivity.onStart(AppCompatActivity.java:177)
                                                                        at org.openhab.habdroid.ui.OpenHABMainActivity.onStart(OpenHABMainActivity.java:582)
                                                                        at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1236)
                                                                        at android.app.Activity.performStart(Activity.java:6006)
                                                                        at android.app.Activity.performRestart(Activity.java:6063)
                                                                        at android.app.Activity.performResume(Activity.java:6068)
                                                                        at android.app.ActivityThread.performResumeActivity(ActivityThread.java:2975)
                                                                        at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3017) 
                                                                        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1347) 
                                                                        at android.os.Handler.dispatchMessage(Handler.java:102) 
                                                                        at android.os.Looper.loop(Looper.java:135) 
                                                                        at android.app.ActivityThread.main(ActivityThread.java:5254) 
                                                                        at java.lang.reflect.Method.invoke(Method.java:372) 
                                                                        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903) 
                                                                        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698) 
````

About => About libs => While loading, quickly go back twice. I think this is related to #742
````
03-07 18:28:38.345 6811-6853/org.openhab.habdroid E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #3
                                                                    Process: org.openhab.habdroid, PID: 6811
                                                                    java.lang.RuntimeException: An error occured while executing doInBackground()
                                                                        at android.os.AsyncTask$3.done(AsyncTask.java:304)
                                                                        at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:355)
                                                                        at java.util.concurrent.FutureTask.setException(FutureTask.java:222)
                                                                        at java.util.concurrent.FutureTask.run(FutureTask.java:242)
                                                                        at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231)
                                                                        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
                                                                        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
                                                                        at java.lang.Thread.run(Thread.java:818)
                                                                     Caused by: java.lang.IllegalStateException: Fragment AboutMainFragment{27633981} not attached to Activity
                                                                        at android.support.v4.app.Fragment.getResources(Fragment.java:715)
                                                                        at android.support.v4.app.Fragment.getString(Fragment.java:737)
                                                                        at org.openhab.habdroid.ui.AboutActivity$AboutMainFragment.getGcmText(AboutActivity.java:273)
                                                                        at org.openhab.habdroid.ui.AboutActivity$AboutMainFragment.getMaterialAboutList(AboutActivity.java:221)
                                                                        at com.danielstone.materialaboutlibrary.MaterialAboutFragment$ListTask.doInBackground(MaterialAboutFragment.java:87)
                                                                        at com.danielstone.materialaboutlibrary.MaterialAboutFragment$ListTask.doInBackground(MaterialAboutFragment.java:77)
                                                                        at android.os.AsyncTask$2.call(AsyncTask.java:292)
                                                                        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
                                                                        at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231) 
                                                                        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112) 
                                                                        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587) 
                                                                        at java.lang.Thread.run(Thread.java:818) 
````

maniac103:
The latter is unrelated to this PR. Looks like the load task should be cancelled after stopping; maybe we should save the context at the start of that method and use it throughout loading the about list.

I'll look into the former crash, is it 100% reproducible for you?

mueller-ma:
> is it 100% reproducible for you?

No

maniac103:
Did the mentioned refactoring, and added some debug output in an attempt to track down the crash above. Can you try to repro it and gather logcat? From the stack trace it looks like the activity is resumed, data is immediately delivered to the controller (which suggests it already was loaded before), but the controller updates the fragments because it thought that the load operation for the given page had never finished so far, which does't really make sense if the data is already available.

mueller-ma:
Test fails again:
````
org.openhab.habdroid.ui.BasicWidgetTest > openHABMainActivityTest[test(AVD) - 5.1.1] FAILED 
	java.lang.NullPointerException: Attempt to read from field 'android.view.View android.support.v7.widget.RecyclerView$ViewHolder.itemView' on a null object reference
	at org.openhab.habdroid.ui.BasicWidgetTest$1.matchesSafely(BasicWidgetTest.java:114)
org.openhab.habdroid.ui.IntroActivityTest > appShowsIntro[test(AVD) - 5.1.1] FAILED 
	android.support.test.espresso.NoMatchingViewException: No views in hierarchy found matching: (with id: org.openhab.habdroid:id/widgetlabel and with text: is "First Floor" and is displayed on the screen to the user)
	If the target view is not part of the view hierarchy, you may need to use Espresso.onData to load it from one of the following AdapterViews:android.widget.ListView{3ae57601 IFED.VC. ......ID -320,0-0,719 #7f090084 app:id/left_drawer}
org.openhab.habdroid.ui.IntroActivityTest > goThroughIntro[test(AVD) - 5.1.1] FAILED 
	android.support.test.espresso.NoMatchingViewException: No views in hierarchy found matching: (with id: org.openhab.habdroid:id/widgetlabel and with text: is "First Floor" and is displayed on the screen to the user)
	If the target view is not part of the view hierarchy, you may need to use Espresso.onData to load it from one of the following AdapterViews:android.widget.ListView{268fccb0 IFED.VC. ......ID -320,0-0,719 #7f090084 app:id/left_drawer}
org.openhab.habdroid.ui.VoiceRecognitionTest > checkVoiceAvailbility[test(AVD) - 5.1.1] SKIPPED 
	org.junit.AssumptionViolatedException: Voice recognition not available, skipping tests for it.
	at org.junit.Assume.assumeTrue(Assume.java:59)
:mobile:connectedFossDebugAndroidTest FAILED
````

I'm planing to tag a new beta version after I merged this PR. What should I write in the changelog for this PR? The PR title is too technical.

maniac103:
Fixed the tests (I hope).

> I'm planing to tag a new beta version after I merged this PR. What should I write in the changelog for this PR? The PR title is too technical.

Yes, it indeed is. How about a generic 'Code structure improvements' plus the 3 fixed issues?

mueller-ma:
Restarted the test
````
org.openhab.habdroid.ui.IntroActivityTest > goThroughIntro[test(AVD) - 5.1.1] FAILED 
	android.support.test.espresso.PerformException: Error performing 'single click - At Coordinates: 421, 768 and precision: 16, 16' on view '(with id: org.openhab.habdroid:id/next and Child at position 3 in parent (with id: org.openhab.habdroid:id/bottomContainer and Child at position 1 in parent with id: org.openhab.habdroid:id/bottom) and is displayed on the screen to the user)'.
	at android.support.test.espresso.PerformException$Builder.build(PerformException.java:84)
````

maniac103:
Yeah, that one must be a glitch; my attempt to run the tests locally succeeded and I didn't touch the intro code.

PR #691, opened 2018-02-02
Fix resetting widget TextViews back to default color.

### LINKED ISSUE #690: Value colors are not changed
Following sitemap:
````
Text item=foo valuecolor=[OFF="red", ON="default"]
````

when chaning from OFF to ON, the text is still red. After reopening that subpage, the text has the default color again.

maniac103:
TextView.setColor() overwrites the default ColorStateList, so we have to
store it in order to restore it later.

Fixes #690


maniac103:
@mueller-ma Can you please verify?

mueller-ma:
Yes, fixed it.

Issue #445, opened 2017-11-04
Different chart resolution in landcape and portrait modes


lolodomo:
The chart resolution should be the same when the tablet/phone is in landscape or portrait mode.
The current implementation leads to lower texts in landscape mode than in portrait mode.
The reason is known: the chart is requested in the two cases with the full tablet width while finally only displayed (resized) in a part of the tablet width when in landscape mode.

lolodomo:
@mueller-ma : have you an idea how this could be fixed ? Can we get the width of the final display ?

lolodomo:
@mueller-ma : any idea how I can get the width of the right frame in case of a display in portrait mode on a tablet ?

mueller-ma:
I have an idea, but need some help with the implementation. Code coming soon...

mueller-ma:
https://github.com/mueller-ma/openhab.android/tree/445-size-chart
OpenHABFragmentPagerAdapter getPageWidth() seems to return wrong values. This only works for sub-sub-panes.

lolodomo:
I got it, we can use `parent.getWidth()`. I will propose a fix.

mueller-ma:
This bug was introduced, I think by #650

maniac103:
@mueller-ma A bug from 2017 was introduced by my yesterday's pull request? I am confused.

mueller-ma:
*reintroduced. I'm not sure if it was your PR, but it could be as you made a lot of changes to the internal layout.

PR #693, opened 2018-02-02
Improve & fix widget sizing code

maniac103:
After #650, #445 resurfaced; this PR should fix it.

Also remove some weird manual fiddling with widget sizing.


maniac103:
@mueller-ma @lolodomo Can you please give this a try?

mobile/src/main/res/values/dimens.xml
```diff
@@ -10,4 +10,5 @@
     <dimen name="widgetlist_item_left_margin">0dp</dimen>
     <dimen name="widgetlist_item_right_margin">0dp</dimen>
     <dimen name="widget_margin">0dp</dimen>
+    <dimen name="webview_row_height">36dp</dimen>
```
maniac103: I'm open for suggestions regarding this value. BasicUI uses 36 pixels per row, which is what I based this on. Old code used 80 pixels per row, which amounts to either 80dp (mdpi), ~53dp (hdpi), 40dp (xhdpi) and ~27dp (xxhdpi). Most devices should be xxhdpi nowadays.

mueller-ma:
I tested it with the multimedia page in demo mode.
Before we had in landscape

1. Radio/Volume
2. Picture with multiple smaller pictures
3. A video which was just a black box on my device
4. Heise via Webview, took up the whole reserved space
5. A pic with "320x240 openHAB", with width equals subpage width (is this a placeholder for anything?). Reintroduced #497

Now we have
1. Radio/Volume
2. Picture with multiple smaller pictures
3. Video is missing
4. Heise via Webview, is only on the half of space of the reserved space
5. A pic with "320x240 openHAB", fixed #497

For the video I think it looks better, but the webview bug should fill up all the reserved space.

maniac103:
The WebView has a fixed height set in the sitemap though. Compare e.g. to its rendering in BasicUI: http://demo.openhab.org:8080/basicui/app?w=0203&sitemap=demo ... it didn't take the full height before either, but a fixed amount of pixels. For the demo sitemap one, height was set to 640px; now it's set to 228dp, which e.g. on xxhdpi amounts to 864px. BasicUI renders it at 228px.

In general: previously, with decreasing display density the WebView was rendered larger. Now it's rendered equally large for all devices, which - if you e.g. have an mdpi device -means it might be rendered smaller for you.
@kaikreuzer [The docs](https://docs.openhab.org/configuration/sitemaps.html#element-type-webview) say the height parameter is supposed to be `the number of element rows to fill`, but AFAICT there's no specification how large one row is supposed to be ... is that correct and the interpretation is up to the client?

I'd be interested in seeing the sitemap + item definition for items 3 and 5 as well; I guess something got broken in there.

mueller-ma:
Before
screenshot_2018-02-02-15-52-52
After
screenshot_2018-02-02-15-54-06

maniac103:
7 inch tablet, I assume? If yes, that's mdpi, thus 640px before and 228px afterwards. Compare to BasicUI on the same phone ... or to a somewhat recent phone :-)

mueller-ma:
I dont really care about the height. You see the space after the webview? There should be the 320*240 pic directly after the webview and not free space.

maniac103:
Ah, now I understand. Will check it.

mueller-ma:
LGTM! Can you resolve the merge conflict?

maniac103:
Fixed.

PR #700, opened 2018-02-05
Simplify intro code and change font

### LINKED ISSUE #697: Remove custom font "Helvetica" from intro
The font isn't found on two devices where I tested it and it's not under a free license.

mueller-ma:
Get rid of the stactraces, when the font is not found.

Fixes #697

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/IntroActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/IntroActivity.java
@@ -30,32 +30,36 @@ public class IntroActivity extends AppIntro {
         }
     }
 
+    /**
+     * Must be overridden to ensure that the intro will be closed when clicking on "SKIP"
+     * @param currentFragment
+     */
     @Override
     public void onSkipPressed(Fragment currentFragment) {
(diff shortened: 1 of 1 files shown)


mueller-ma:
The introActivity code could be simplified. Should I do this here or just change the font here and everything else (Simplification, comments, removal of unneeded methods) in a new PR?

maniac103:
You're the team member, you can set up the rules ;-) Personally I don't care.

Are by simplification you're referring to the API level distinction and the code duplication because of that being unnecessary? If no, that certainly is simplification potential, just use `ContextCompat.getColor()`

maniac103:
And BTW, is it just me or do the codecov reports look kinda useless? This PR only touches IntroActivity, but codecov reports changes in random other files...

mueller-ma:
No, its not just you. Codecov seems to me really useless, spammy and wrong. I think I'm going to disable it

The api level distinction is needed, as getRessources is depricated (works, but may not work in future). I reffer to the multiple getColor; get them once and save them.

maniac103:
The mentioned `ContextCompat.getColor()` provides the abstraction for exactly that deprecation :-)

PR #703, opened 2018-02-05
Use Android Log instead of System.out.print

mueller-ma:


maniac103:
:+1: ... Could also treat the various Exception.printStackTrace() occurrences similarly.

mueller-ma:
How? `Log.e(TAG, e.getMessage(), e)`?

maniac103:
Yeah, or a more meaningful (own) message instead of `e.getMessage()`

mobile/src/main/java/org/openhab/habdroid/model/OpenHABItem.java
```diff
@@ -151,7 +152,7 @@ public Float getStateAsFloat() {
 			try {
 				result = Float.parseFloat(state);
 			} catch (NumberFormatException e) {
-				e.printStackTrace();
+				Log.e(TAG, e.getMessage(), e);
```
mueller-ma: Not sure if that's really an error we should log as error.

maniac103: Probably not. In general I wouldn't log exceptions that occur only due to broken user/server input.

mueller-ma: I change their log level to d as it might be good to see, when those exceptions occur.

PR #706, opened 2018-02-06
Disallow # and \ in store descriptions

mueller-ma:
The intention behind this commit is to disallow new line or similar
chars in store descriptions.

Since crowdin inserts two lines comments, we need to check if sharp count > 2

@ThomDietrich

### DIFF
--- a/assets/store_descriptions/generate_and_validate.sh
+++ b/assets/store_descriptions/generate_and_validate.sh
@@ -24,7 +24,7 @@ do
     if [ -f "${folder}/strings.sh" ]
     then
         lang=${folder#${string_base}/}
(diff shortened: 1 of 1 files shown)


mueller-ma:
Travis will fail now, since two new line chars are in the current descriptions. I removed them both in crowdin and they are already in #680

ThomDietrich:
This is also great because otherwise something like a line break can be missed easily in crowdin.

I'm not complete sure about the # solution. It works now but will need to be changed if a new comment is added e.g. between entries. If that case becomes interesting maybe count `#` that are not the first character of a line?

mueller-ma:
In that case we can comment all lines in one commit and manually change the count in the script

PR #712, opened 2018-02-09
Change title bar color in "last apps" to themed color

### LINKED ISSUE #710: Color in "last apps" view is not always primary color
I have the app set to basic ui theme. In the last apps view the title bar is orange. Android 7.1.2

On Android 5.1.1 the title color in last apps is blue.

mueller-ma:
Fixes #710

I changed all activities to inherit from baseAcitivity instead of AppCompatActivity, except IntroActivity. Doesn't care, since the theme cannot be changed before going throught the intro.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/Util.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Util.java
@@ -43,7 +43,8 @@ import java.util.List;
 import okhttp3.Call;
 import okhttp3.Headers;
 
-import static android.os.Build.*;
+import android.os.Build.VERSION;
+import android.os.Build.VERSION_CODES;
 
 public class Util {
 
@@ -167,11 +168,10 @@ public class Util {
         if (VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP) {
             TypedValue typedValue = new TypedValue();
             activity.getTheme().resolveAttribute(R.attr.colorPrimary, typedValue, true);
-            activity.setTaskDescription(
-                    new ActivityManager.TaskDescription(activity.getString(R.string.app_name),
-                            BitmapFactory.decodeResource(activity.getResources(),
-                                    R.mipmap.icon_round),
-                            typedValue.data));
+            activity.setTaskDescription(new ActivityManager.TaskDescription(
+                    activity.getString(R.string.app_name),
+                    BitmapFactory.decodeResource(activity.getResources(), R.mipmap.icon_round),
+                    typedValue.data));
         }
     }
 
--- a/mobile/src/main/java/org/openhab/habdroid/util/Util.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Util.java
@@ -10,10 +10,16 @@
 package org.openhab.habdroid.util;
 
 import android.app.Activity;
+import android.app.ActivityManager;
+import android.content.res.Resources;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
 import android.net.Uri;
 import android.preference.PreferenceManager;
+import android.support.annotation.ColorInt;
 import android.support.annotation.NonNull;
 import android.util.Log;
+import android.util.TypedValue;
 
 import org.json.JSONArray;
 import org.json.JSONException;
@@ -140,7 +146,7 @@ public class Util {
 
     public static void setActivityTheme(@NonNull final Activity activity, String theme) {
         if (theme == null) {
-            theme = PreferenceManager.getDefaultSharedPreferences(activity).getString(Constants.PREFERENCE_THEME, activity.getString(R.string.theme_value_dark));
+            theme = PreferenceManager.getDefaultSharedPreferences(activity).getString(Constants.PREFERENCE_THEME, activity.getString(R.string.theme_value_light));
         }
         int themeRes;
         if (theme.equals(activity.getString(R.string.theme_value_dark))) {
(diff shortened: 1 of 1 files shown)


maniac103:
Can't you do the same thing from the util method that applies the theme?

mueller-ma:
Yes, that solution is much simpler.
The base for my implementation was https://github.com/T-Rex96/Easy_xkcd/blob/master/app/src/main/java/de/tap/easy_xkcd/Activities/BaseActivity.java#L32. Some lines are still the same, so should/do I need to credit them?

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -155,6 +161,17 @@ public static void setActivityTheme(@NonNull final Activity activity, String the
             themeRes = R.style.HABDroid_Light;
         }
         activity.setTheme(themeRes);
+
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
+            Bitmap icon = BitmapFactory.decodeResource(activity.getResources(), R.mipmap.icon_round);
+            TypedValue typedValue = new TypedValue();
+            Resources.Theme themeActivity = activity.getTheme();
```
mueller-ma: The theme is set in L163, but as res int, not as Theme. Can I convert res id to theme or should I use the getter here?

maniac103: Using the getter is fine, I think.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -155,6 +161,17 @@ public static void setActivityTheme(@NonNull final Activity activity, String the
...
+            themeActivity.resolveAttribute(R.attr.colorPrimary, typedValue, true);
+            @ColorInt int color = typedValue.data;
+            String name = activity.getString(R.string.app_name);
+            ActivityManager.TaskDescription description = new ActivityManager.TaskDescription(name, icon, color);
```
maniac103: Can probably omit the intermediate variable here, can't you?

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -155,6 +161,17 @@ public static void setActivityTheme(@NonNull final Activity activity, String the
             themeRes = R.style.HABDroid_Light;
         }
         activity.setTheme(themeRes);
+
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
+            Bitmap icon = BitmapFactory.decodeResource(activity.getResources(), R.mipmap.icon_round);
+            TypedValue typedValue = new TypedValue();
+            Resources.Theme themeActivity = activity.getTheme();
+            themeActivity.resolveAttribute(R.attr.colorPrimary, typedValue, true);
+            @ColorInt int color = typedValue.data;
```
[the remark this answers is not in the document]
mueller-ma: https://developer.android.com/reference/android/content/res/Resources.Theme.html#resolveAttribute(int,%20android.util.TypedValue,%20boolean)

resolveRefs is true, so the actuall color value is returned.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
mueller-ma: Done!

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -155,6 +161,17 @@ public static void setActivityTheme(@NonNull final Activity activity, String the
             themeRes = R.style.HABDroid_Light;
         }
         activity.setTheme(themeRes);
+
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
```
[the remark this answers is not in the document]
mueller-ma: Done!

mobile/src/main/java/org/openhab/habdroid/util/Util.java
mueller-ma: No, changed it.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -140,7 +148,7 @@ public static void setActivityTheme(@NonNull final Activity activity) {
 
     public static void setActivityTheme(@NonNull final Activity activity, String theme) {
         if (theme == null) {
-            theme = PreferenceManager.getDefaultSharedPreferences(activity).getString(Constants.PREFERENCE_THEME, activity.getString(R.string.theme_value_dark));
+            theme = PreferenceManager.getDefaultSharedPreferences(activity).getString(Constants.PREFERENCE_THEME, activity.getString(R.string.theme_value_light));
```
mueller-ma: Just in case the pref read fails, fall back to default.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -37,6 +43,8 @@
 import okhttp3.Call;
 import okhttp3.Headers;
 
+import static android.os.Build.*;
```
maniac103: I think generic imports are against Google code style: https://source.android.com/setup/code-style#fully-qualify-imports

I'd only import android.os.Build non-statically.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -155,6 +163,16 @@ public static void setActivityTheme(@NonNull final Activity activity, String the
...
         activity.setTheme(themeRes);
+
+        if (VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP) {
+            TypedValue typedValue = new TypedValue();
+            activity.getTheme().resolveAttribute(R.attr.colorPrimary, typedValue, true);
+            activity.setTaskDescription(
+                    new ActivityManager.TaskDescription(activity.getString(R.string.app_name),
+                            BitmapFactory.decodeResource(activity.getResources(),
+                                    R.mipmap.icon_round),
+                            typedValue.data));
```
maniac103: I realize that's total nitpicking, but wouldn't make this more sense?

```
activity.setTaskDescription(new ActivityManager.TaskDescription(
        activity.getString(...),
        BitmapFactory.decodeResource(...),
        typedValue.data));
```

mobile/src/main/java/org/openhab/habdroid/util/Util.java
mueller-ma: Ok, they are against. Changed it.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
mueller-ma: You're right, that looks better.

> that's total nitpicking

Doesn't care as long code gets better :)

PR #719, opened 2018-02-10
Disallow "HTTP error code" in translations

mueller-ma:


assets/store_descriptions/generate_and_validate.sh
```diff
@@ -64,21 +65,38 @@ do
         if [ "$chars" -gt 4000 ]
         then
             echo "Full descriptions exceeds 4000 char limit" 1>&2
-            let error++
+            let error++ || true
```
mueller-ma: If I don't add `|| true` the script stops here, because `let` returns 1 when incrementing from 0 to 1.

PR #726, opened 2018-02-27
Redesign about screen

### LINKED ISSUE #709: Redesign about screen
I will come up with some ideas soon.

mueller-ma:
Fixes #709 

I'm a bit unsure about the oh secret. Should we really show it instantly? Maybe it's better to "click to show" it or remove it completely.


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
```diff
@@ -43,9 +61,21 @@ public void onCreate(Bundle savedInstanceState) {
                     .commit();
         }
 
+        try {
+            mOpenHABVersion = Integer.valueOf(getIntent().getExtras().get("openHABVersion").toString());
```
mueller-ma: Can I trust the MainActivity to give me always the correct version?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
```diff
@@ -61,81 +91,234 @@ public void finish() {
...
+                return getString(R.string.info_openhab_gcm_connected, OpenHABMainActivity.GCM_SENDER_ID);
             }
         }
 
-        @Override
-        public int getCount() {
-            return 3;
+        private String getServerSecret(Connection conn) {
+            final String[] secret = {null};
+            conn.getSyncHttpClient().get("/static/secret", new MyHttpClient.TextResponseHandler() {
```
maniac103: I don't like the usage of the sync HTTP client here, as - depending on the server speed - it'll inevitably lead to ANR being reported.
IMHO we should show a progress indicator as long as it's loading.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
maniac103: As long as it can reach the server, yes. Why the complicated setup, though?

```
mOpenHABVersion = getIntent().getIntExtra("openHABVersion", 0);
```

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
```diff
@@ -1,27 +1,45 @@
...
+import java.util.Calendar;
+import java.util.Locale;
+
+import okhttp3.Call;
+import okhttp3.Headers;
+
 public class OpenHABAboutActivity extends AppCompatActivity {
+    private final static String TAG = OpenHABAboutActivity.class.getSimpleName();
+    private static Connection conn = null;
+    private static int mOpenHABVersion = 0;
```
maniac103: Why make those variables static?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
maniac103: Never mind. I just saw that this method is running in an AsyncTask. Too bad, I would love to get rid of the sync HTTP client ;-)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
mueller-ma: AboutMainFragment must be static, so those vars, too.

maniac103: A static Fragment instance is a no-go, as it'll leak the activity. Why must it be static? There must be a different solution...

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
mueller-ma: Yes, that's better

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
mueller-ma: The oh version is in the intent, so it will be passed as fragment argument. But how can I add the connection? I tried to add it to the getIntent().getExtras() bundle and pass that here https://github.com/mueller-ma/openhab.android/blob/430b7eb382a02fea46161e6f0abf47ed2949ed02/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java#L57 , but I cannot find the function to add the connection to the bundle. Their are only add functions for int, strings, etc.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -222,4 +222,14 @@ public void onSuccess(Call call, int statusCode, Headers headers, String respons
             }
         });
     }
+
+    public static String makeStringIrrecognizable(String string) {
+        return makeStringIrrecognizable(string, 3);
+    }
+
+    public static String makeStringIrrecognizable(String string, int showClearTextChars) {
```
mueller-ma: obfuscateString is probably a better name!?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
maniac103: You can re-fetch it from ConnectionFactory.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
maniac103: Yes, indeed. `showClearTextChars` also probably should be renamed to `clearTextCharCount`.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
```diff
@@ -61,81 +80,246 @@ public void finish() {
...
+                        .text(R.string.info_openhab_uuid_label)
+                        .subText(uuid)
+                        .icon(R.drawable.ic_info_outline)
+                        .build());
 
-            final TabLayout tabLayout = (TabLayout) view.findViewById(R.id.tab_layout);
-            tabLayout.setupWithViewPager(viewPager);
-            tabLayout.setTabsFromPagerAdapter(mPagerAdapter);
-            tabLayout.setTabMode(TabLayout.MODE_FIXED);
+                String secret = getServerSecret(conn);
```
mueller-ma: Is the secret only available via http in oh 1?

mobile/src/main/res/values/strings.xml
```diff
@@ -153,11 +154,19 @@
     <!-- About -->
     <string name="about_title">About</string>
     <string name="about_license_title">License</string>
-    <string name="about_license" translatable="false"><![CDATA[<a href="https://raw.githubusercontent.com/openhab/openhab.android/master/LICENSE">Eclipse Public License - v 1.0</a>]]></string>
-    <string name="about_links">Links</string>
-    <string name="about_links_list"><![CDATA[\u00B7 <a href="https://github.com/openhab/openhab.android">openHAB App source code</a><br><br> \u00B7 <a href="https://github.com/openhab/openhab.android/issues">Report a problem</a><br><br> \u00B7 <a href="http://docs.openhab.org/">openHAB Documentation</a><br><br> \u00B7 <a href="https://community.openhab.org/">Community Forum</a>]]></string>
```
mueller-ma: @ThomDietrich FYI this PR will remove this horrible to translate string.

mobile/src/main/res/values/strings.xml
```diff
@@ -153,11 +154,19 @@
...
     <string name="about_version_string" translatable="false">%1$s - %2$s</string>
+    <string name="about_changelog">Changelog</string>
+    <string name="about_source_code">Fork us on GitHub!</string>
+    <string name="about_issues">Report an issue</string>
+    <string name="about_docs">Documentation</string>
+    <string name="about_community_forum">Community forum</string>
+    <string name="about_foundation">openHAB Foundation</string>
+    <string name="about_community">openHAB community</string>
+    <string name="about_server">openHAB server</string>
+    <string name="about_translation">Help us to translate!</string>
```
mueller-ma: "Help us to translate openHAB products"?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
```diff
@@ -248,9 +248,7 @@ private String getServerSecret(Connection conn) {
             conn.getSyncHttpClient().get("/static/secret", new MyHttpClient.TextResponseHandler() {
                 @Override
                 public void onFailure(Call call, int statusCode, Headers headers, String responseString, Throwable error) {
-                    if (error.getMessage() != null) {
-                        Log.d(TAG, error.getMessage());
-                    }
+                    Log.d(TAG, "Could not fetch server secret", error);
```
mueller-ma: @maniac103 Do you think we need the stacktrace here?

mobile/src/main/res/values/strings.xml
maniac103: 'Products'? I thought the link only points to the app translation? If it points to all OH translations, how about `Help us translating openHAB` or - a bit more direct - `Translate openHAB to your language`?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
maniac103: No, probably not. We know the call path here anyway.

mobile/src/main/res/values/strings.xml
mueller-ma: It points to the profile of the openhab-bot on crowdin, which has a list with all oh projects.

maniac103: OK. Then my suggestions above are valid :-)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
mueller-ma: For oh2 instances the secret won't be fetched (which fails anyway).

mueller-ma:
@maniac103 Can you check if everything works fine on a oh1 instance?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABAboutActivity.java
```diff
@@ -61,81 +80,243 @@ public void finish() {
...
+                        .icon(R.drawable.ic_info_outline)
+                        .build());
 
-            final TabLayout tabLayout = (TabLayout) view.findViewById(R.id.tab_layout);
-            tabLayout.setupWithViewPager(viewPager);
-            tabLayout.setTabsFromPagerAdapter(mPagerAdapter);
-            tabLayout.setTabMode(TabLayout.MODE_FIXED);
+                if (mOpenHABVersion == 1) {
+                    String secret = getServerSecret(conn);
+                    if (!TextUtils.isEmpty(secret)) {
```
maniac103: That negation is wrong.

Maybe even hide the card item completely in that case, given secret is only populated if the cloud connector is installed?

maniac103:
@mueller-ma Besides the review comment everything else looks good.

mobile/src/main/res/values/themes.xml
```diff
@@ -17,6 +17,10 @@
         <item name="themedDrawerBackground">?android:colorBackground</item>
         <item name="themedHeaderTextColor">@color/colorAccent_themeDark</item>
 
+        <item name="mal_card_background">@color/about_libraries_window_background</item>
```
mueller-ma: Going to make a oh theme with dark and one with bright background and use them as parent theme for all oh themes.

maniac103:
@mueller-ma There's one thing I missed before: the arrow in the library list activity action bar is black instead of white.

mueller-ma:
Thanks for spotting this.

PR #727, opened 2018-03-02
Fix rollershutter height

mueller-ma:
Before:
screenshot_2018-03-02-15-31-15
After:
screenshot_2018-03-02-15-29-59


maniac103:
I can't really see a difference?

mueller-ma:
Look at the line under "Toilet" on the left side and "Livingroom" on the right.

PR #732, opened 2018-03-03
Improve themes

mueller-ma:
* Move common attributes to new themes and use these as parents for all openHAB themes
* Fix back arrow in about libaries
* Add new themes (red and teal, bright and dark)


maniac103:
Is there any precedent for those red and teal themes in other openHAB UIs? If not, why add red and teal, but not yellow, pink, purple, indigo or whatever?

mueller-ma:
Hm, it's probably better to add all themes at once.

mueller-ma:
Removed the themes for now as I wasn't sure if I want to merge them.

mueller-ma:
I justed noticed, that the arrow color cannot be changed so easily without breaking something else.
When I override colorControlNormal with white, the arrow is white, but also unchecked checkboxes.

maniac103:
I would suspect it's rather the action bar theme that is off (or not set at all).

maniac103:
Judging from the lib code, it should be sufficient to add this:
```
    <item name="drawerArrowStyle">@style/AboutLibrariesTheme.DrawerArrowStyle</item>
```

or

```
    <item name="drawerArrowStyle">@style/AboutLibrariesTheme.Light.DrawerArrowStyle</item>
```

as appropriate.

mueller-ma:
When I add `<item name="drawerArrowStyle">@style/AboutLibrariesTheme.Light.DrawerArrowStyle</item>` to the light theme, the back arrow in the main activity is black, in the about white and in used libaries dark grey.
Also the animation how the arrow is built changes.

maniac103:
Created #742 for the arrow fix.

PR #736, opened 2018-03-04
Replace PNG with SVG

mueller-ma:
and get rid of unused pictures.

This slithly changes the look of the rollershutter and color items:
screenshot_2018-03-04-14-32-43


mueller-ma:
@maniac103 Can you remove list_selector.xml and list_arrow* in #688?

maniac103:
Sure.

maniac103:
BTW, if you're changing the drawables anyway, I'd consider removing their background, replacing it by `?attr/selectableItemBackgroundBorderless` and making the foreground black or white.

mueller-ma:
I wanted to tint them with colorAccent, but now they are always pink, indepenet from the theme.

maniac103:
> I wanted to tint them with colorAccent, but now they are always pink, indepenet from the theme.

Tint how exactly? Using android:tint in the XML won't work for us anyway, as it won't work on API < 21.

mueller-ma:
> Using android:tint in the XML won't work for us anyway, as it won't work on API < 21.

I didn't know that, normally Android Studio marks everything that isn't available on a supported API level, but not this time.
I added some code that colors the buttons.

maniac103:
I think simply adding two different drawables (same path, but different color) and referencing them by attr ID in the XML would be better (at least it would be the 'do it as Google would do it' way), as it keeps design details in XML and out of code.

mueller-ma:
The icons in the navigation drawer and the about screen are in all themes "#9E9E9E", so I could use the color here, too.

maniac103:
> Can you remove list_selector.xml and list_arrow* in #688?

It turns out those actually are used (for the sub-page selection highlight in two-pane mode) ... what should I replace them with?

mueller-ma:
Ok, didn't see that it's tint mode src_in for list_arrow_activated.9.png, so I cannot see the blue.

> what should I replace them with?

If it's used, we can leave it as it is.

mobile/src/main/res/layout/openhabwidgetlist_rollershutteritem.xml
```diff
@@ -23,41 +23,44 @@
     <ImageButton
         android:id="@+id/rollershutterbutton_up"
         android:layout_width="46dip"
-        android:layout_height="46dip"
+        android:layout_height="0dp"
```
maniac103: Are those height changes intended? Keeping the explicit height of 46dp would probably make the sizing intent a bit clearer, even though the end result should be the same.

mueller-ma:
It's probably better to wait for #732. I can change the colors of the color and rollershutter button here, but the other elements should be in a new PR.

mobile/src/main/res/layout/openhabwidgetlist_rollershutteritem.xml
mueller-ma: Having the height set to match_constraint (0dp) makes them a bit larger (left one 46, right on 0 dp height):
buttons

PR #738, opened 2018-03-04
Add support for Slider option "switchSupport"

### LINKED ISSUE #735: Support Slider "switchSupport"
It's pretty common for home automation software to support having both a switch and a slider-like element combined into one for dimmable lights – Z-Way, ImperiHome, Fibaro Home Center, Vera etc. Slider elements in openHab already have the "switchSupport" parameter, but this is only used in Classic UI, and there only to modify the function of the already existing up and down buttons. In the Android UI one is left with having to create two elements for the same thing, one Switch and one Slider, with basically the same label and controlling the same thing, taking up twice the real estate and making the control slightly more awkward than it needs to be.

I would love to see switchSupport implemented in the Android app as well, but rendered as a single element with both a switch and a slider bar, much like this (although this image is from manipulating a Basic UI element, you should get the idea):

switchSupport rendered

The usefulness of this is e.g. for Z-Wave items that have support for ON and OFF commands as well as for precise levels. Usually ON has a value of 255, and will often turn the device back on to its previous dimmer level (or a specific preset value, or something else entirely, all dependent on the physical device and its configuration). So this must not be confused with ON just being a shortcut for setting the slider to 100% – something on which the documentation is a bit unclear with regards to Classic UI functionality.

mueller-ma:
Fixes #735

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -499,7 +499,7 @@ public class OpenHABWidgetAdapter extends RecyclerView.Adapter<OpenHABWidgetAdap
             mSeekBar.setProgress(progress);
             mBoundItem = widget.getItem();
             if(widget.getSwitchSupport()) {
-                mSliderSwitch.setChecked(progress != 0 ? true : false);
+                mSliderSwitch.setChecked(progress != 0);
                 mSliderSwitch.setVisibility(View.VISIBLE);
             }
         }
@@ -524,7 +524,7 @@ public class OpenHABWidgetAdapter extends RecyclerView.Adapter<OpenHABWidgetAdap
         public boolean onTouch(View v, MotionEvent motionEvent) {
             if (motionEvent.getActionMasked() == MotionEvent.ACTION_UP) {
                 Util.sendItemCommand(mConnection.getAsyncHttpClient(), mBoundItem,
-                        mSliderSwitch.isChecked() ? "0" : "100");
+                        mSliderSwitch.isChecked() ? "OFF" : "ON");
             }
             return false;
         }
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -501,6 +501,8 @@ public class OpenHABWidgetAdapter extends RecyclerView.Adapter<OpenHABWidgetAdap
             if(widget.getSwitchSupport()) {
                 mSliderSwitch.setChecked(progress != 0);
                 mSliderSwitch.setVisibility(View.VISIBLE);
+            } else {
+                mSliderSwitch.setVisibility(View.GONE);
             }
         }
 
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -459,9 +459,10 @@ public class OpenHABWidgetAdapter extends RecyclerView.Adapter<OpenHABWidgetAdap
         }
     }
 
-    public static class SliderViewHolder extends ViewHolder implements SeekBar.OnSeekBarChangeListener {
+    public static class SliderViewHolder extends ViewHolder implements SeekBar.OnSeekBarChangeListener, View.OnTouchListener {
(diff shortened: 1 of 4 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -490,11 +493,15 @@ public void bind(OpenHABWidget widget) {
...
             } else {
-                mSeekBar.setProgress(0);
+                progress = 0;
             }
+            mSeekBar.setProgress(progress);
             mBoundItem = widget.getItem();
+            if(widget.getSwitchSupport()) {
+                mSliderSwitch.setChecked(progress != 0 ? true : false);
+                mSliderSwitch.setVisibility(View.VISIBLE);
+            }
```
maniac103: ... else set visibility GONE.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -490,11 +493,15 @@ public void bind(OpenHABWidget widget) {
...
                 }
-                mSeekBar.setProgress(progress);
             } else {
-                mSeekBar.setProgress(0);
+                progress = 0;
             }
+            mSeekBar.setProgress(progress);
             mBoundItem = widget.getItem();
+            if(widget.getSwitchSupport()) {
+                mSliderSwitch.setChecked(progress != 0 ? true : false);
```
maniac103: `? true : false` is completely redundant.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
```diff
@@ -512,6 +519,15 @@ public void onStopTrackingTouch(SeekBar seekBar) {
             Util.sendItemCommand(mConnection.getAsyncHttpClient(), mBoundItem,
                     String.valueOf(seekBar.getProgress()));
         }
+
+        @Override
+        public boolean onTouch(View v, MotionEvent motionEvent) {
+            if (motionEvent.getActionMasked() == MotionEvent.ACTION_UP) {
+                Util.sendItemCommand(mConnection.getAsyncHttpClient(), mBoundItem,
+                        mSliderSwitch.isChecked() ? "0" : "100");
```
maniac103: The feature request explicitly asked for this *not* being a shortcut to 0/100. Sending ON/OFF would be appropriate here IMHO.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
mueller-ma: I set the visibility to gone in the xml. Do I need to change it here?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
mueller-ma: Removed.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
mueller-ma: Right, changed it to ON/OFF

mueller-ma:
@kvolden Can you test this apk?
[mobile-foss-debug.zip](https://github.com/openhab/openhab-android/files/1778811/mobile-foss-debug.zip)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
maniac103: Yes, as the holder may be re-used for a widget not supporting the switch when scrolling the list.

mueller-ma: Done

kvolden:
@mueller-ma That was quick. :smile: Just tested it, and the app seems to be a bit crashy, but I don't know if that's related to this change. The switchSupport feature itself works and looks exactly like I hoped.

The switch is present with or without the switchSupport parameter in the version I downloaded, but 8b5b4fc should take care of that I guess.

maniac103:
@kvolden Do you have stack traces of the crashes? No, we don't knowingly put or leave crashing code in ;-)

kaikreuzer:
Please see my https://github.com/openhab/openhab-android/issues/735#issuecomment-370342895 - I'm very much against implementing it as suggested in the issue.

PR #739, opened 2018-03-04
Fix flickering icon when state changes

### LINKED ISSUE #684: Prefetch icons for other states
Currently the icon for a specific state is fetched, when the state appears the first time. Maybe we should load icons for other states on the first app start, e.g. on and off for switches, 0 to 100 for slider.
This would avoid flickering icons, not related to #669

mueller-ma:
...and the new icon isn't already fetched.

Fixes #684

It seems that the old icon is now displayed until the new one is fetched.


maniac103:
I'm not entirely sure this is a good idea. It's correct for the intended use case, but also means that when a view is re-used on scrolling, the icon of the view you just scrolled out suddenly appears in the newly displayed view for a split second.

mueller-ma:
I tried to reproduce that, but even on my slow tablet, I couldn't see reused icons. Also cleared the image cache.

maniac103:
You need to delay the server response to reproduce (think slow network), the client device speed doesn't matter as much.

mueller-ma:
I was able to reproduce what you said, closing for now.

PR #740, opened 2018-03-05
Use gradle witness for dependency verification

mueller-ma:
The plugin compairs the hashsums of the downloaded dependencies with
those in the verify list.
Because we have some libaries (e.g. gcm) not in all flavors, witness
throws an exception, when the dependency is not found.
The two options I see are:
1. Don't verify libs like gcm. However an attacker could manipulate
those libaries.
2. Remove the exception: https://github.com/mueller-ma/gradle-witness/commit/1d6be152dbaca688f3f6755a2710272e95c0bcec

The jar file is compiled with 1d6be152dbaca688f3f6755a2710272e95c0bcec
from the repo above.


mueller-ma:
Since it's all trust on first use, another one should check if all hash sums are the same. The directory ~/.gradle must be removed/renamed in order to run the gradle hash sum check.

PR #742, opened 2018-03-05
Use about library fragment instead of activity.

maniac103:
Doing so works around the about library activity not setting an action
bar overlay appropriate for dark action bars when using it together with
a custom activity theme.


maniac103:
@mueller-ma This (indirectly) fixes the black action bar arrow by not using the libs activity anymore.

mueller-ma:
I think it would look better, when the activity title changes to "Used libraries" (title_activity_libraries)

maniac103:
Done.

mueller-ma:
The title is set to "About", when the device is rotated.

maniac103:
Props to your testing efforts :-) Fixed.

Issue #724, opened 2018-02-24
Issues with split screen


mueller-ma:
### Actual behaviour
When using split screen, the reload icon should be centered.

### Expected behaviour
It is horizontal moved to the right.

 
### Steps to reproduce
1. Open the app
2. Use split screen feature
3. Reload by swiping down

### Can you reproduce the issue in demo mode?

Yes

#### Client
* Android version: Android 7.2.1
* Device model: Oneplus X
* Stock or customized ROM: LineageOS 14.1
* App version : 2.0.14-beta
* Installation source : Github
* Device language: EN

The manifest entry for "app supports split screen" was added, but no optimization was done to ensure the app really supports it (or at least I dont see one): https://github.com/openhab/openhab-android/commit/839a8a918d4b313bd4289dc35aa570e5df2f4f8f

mueller-ma:
![screenshot_20180305-092602](https://user-images.githubusercontent.com/22525368/37033565-c59bde5a-2146-11e8-8175-b54a83ff87b5.png)
While resizing, the width is correct. Otherwise the switches and values are not visible.
![screenshot_20180305-092708](https://user-images.githubusercontent.com/22525368/37033566-c5b6eeb6-2146-11e8-978b-3e4aa92f054a.png)

PR #744, opened 2018-03-06
Fix split screen mode

mueller-ma:
Both entries were added to the manifest, but no code was added to
support these features.

See #724

As this PR doesn't really fix the issue, I'll wait for a real fix and only merge this PR, if the bug isn't fixed until the next release.

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -30,7 +30,7 @@
         android:fullBackupContent="true"
         android:icon="@mipmap/icon"
         android:label="@string/app_name"
-        android:resizeableActivity="true"
+        android:resizeableActivity="false"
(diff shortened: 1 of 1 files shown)


maniac103:
@fat-tire As you introduced those flags, any comments?

mueller-ma:
By adjusting minHeight and minWidth the values are now visible for me. However the app sometimes reloads when resizing and leaves the subpage.

maniac103:
> However the app sometimes reloads when resizing and leaves the subpage.

This will probably get better with #688.

fat-tire:
Hey sorry to come in late-- I'm not sure if there's a way to enable resizable windows (kind of a necessity for Chromebooks) but disable split screen.  i think they're both using the same setting in the manifest...  But it sounds like you guys are on it :)

PR #746, opened 2018-03-06
Fix SVG image items

mueller-ma:
The old check only worked for icons, but not for image items, since they
don't include "format=SVG" in their URL.

I think I have seen an issue for that, but I cannot find it anymore :(

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/MyWebImage.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/MyWebImage.java
@@ -90,11 +90,7 @@ public class MyWebImage implements SmartImage {
             	bitmap = getWebImageCache(context).get(url);
             if(bitmap == null) {
             	Log.i("MyWebImage", "Cache for " + url + " is empty, getting image");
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/util/MyWebImage.java
```diff
@@ -128,8 +124,11 @@ public void onFailure(Call call, int statusCode, Headers headers, byte[] respons
             @Override
             public void onSuccess(Call call, int statusCode, Headers headers, byte[] responseBody) {
                 InputStream is = new ByteArrayInputStream(responseBody);
+                Map headersMap = headers.toMultimap();
+                boolean isSVG = headersMap.containsKey("content-type") &&
+                        headersMap.get("content-type").toString().contains("svg");
```
maniac103: This can be optimized by passing `response.body()` instead of `response.body().bytes()` to the callback and then using `responseBody.contentType()`. It's too invasive to do in this PR though; I'll prepare a follow-up one.

PR #748, opened 2018-03-08
Use given context to populate about fragment strings.

maniac103:
As getMaterialAboutList() is running in a background thread, the
fragment might become detached while the thread is running, so better
use the passed-in context.


maniac103:
This should fix the second crash shown in https://github.com/openhab/openhab-android/pull/688#issuecomment-371217334

PR #749, opened 2018-03-08
Adapt voice service to use new voice command endpoint.

### LINKED ISSUE #242: Voice command support for OH2
So far, voice commands are sent to an hard-coded String item "VoiceCommand".

OH2 now has a dedicated voice handling infrastructure and offers a rest endpoint for it (starting with OH2 distro 519).
HABDroid should be adapted to use this new endpoint, if it is available; otherwise it can use the current logic as a fallback.

Here is the documentation of the new endpoint:

screen shot 2016-10-08 at 11 28 15

maniac103:
Closes #242


mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
```diff
@@ -112,22 +115,53 @@ private String extractVoiceCommand(Intent data) {
...
+                    sendRawVoiceCommand(conn, command, startId);
+                } else {
+                    Log.e(TAG, "Sending voice command to new endpoint failed: " + statusCode, error);
+                    stopSelf(startId);
+                }
+            }
+
+            @Override
+            public void onSuccess(Call call, int statusCode, Headers headers, byte[] body) {
+                Log.d(TAG, "Command was sent successfully");
```
mueller-ma: Add the actual command here?

maniac103: We can do that, but the recognized command is logged in `extractVoiceCommand()` already.

mueller-ma:
@maniac103 Can you resolve the merge conflict?

maniac103:
@mueller-ma Done.

PR #751, opened 2018-03-08
Replace custom drawer layout by NavigationView.

maniac103:


maniac103:
Before:
screenshot_1520511808

After:
screenshot_1520511871

mueller-ma:
When I'm forcing RTL layout for all languages (in dev options) "Sitemaps" is still on the left side.

maniac103:
> When I'm forcing RTL layout for all languages (in dev options) "Sitemaps" is still on the left side.

The same happens e.g. in Gmail, which I'm pretty sure is tested successfully in RTL layouts. I think we can ignore that as a limitation of the developer option.

PR #753, opened 2018-03-08
Improve connection retry UX.

maniac103:
Show a spinning wheel during the retry, and don't allow multiple server
queries to be active at the same time.


mueller-ma:
LGTM!
Something else related to the error fragment: The debug message still says "Exception stack", but there is no stack anymore. Also the exception message doesn't have to be so small.

maniac103:
It shows a stack of exceptions, if the root exception's cause is != the exception itself.

PR #754, opened 2018-03-08
Simplify image view refresh logic.

maniac103:
Instead of having a fixed timer scheduling a handler and a boolean
suppressing timer events during reload, do all scheduling using the
handler.


maniac103:
@lolodomo As you introduced the mentioned boolean, do you want to check this? :-)

mueller-ma:
RTM?

maniac103:
I think so, yes.

Issue #757, opened 2018-03-08
AboutLibaries adjustments


mueller-ma:
I just checked if all libaries are actually shown:

## Shown, but not used diretly

- [ ] Android-Iconics
- [ ] Fastadapter
- [ ] Okio

## Not shown, but used

- [x]  jmdns https://github.com/mikepenz/AboutLibraries/pull/380
- [x]  shell-software:fab Removed
- [ ]  androidsvg
- [x]  material-about-library https://github.com/mikepenz/AboutLibraries/pull/377

Do we actually need to show all Android libaries?

I asked myself where we use the fab, so I tried to remove it and the app still runs. I think it was used in the discovery screen.
Why can I remove support libaries and the app still complies? https://github.com/mueller-ma/openhab.android/commit/578b61f91e91b8bf995eb97d0c44d3d999c0c6f8

maniac103:
Support libraries are probably pulled in by some library we use, but I'd keep it in nevertheless so we have control over the used version.
The fab library and gson can die though, instead of the former we could use the FAB in the design support library if needed.

daniel-stoneuk:
@mueller-ma Hi man, thanks for bringing up the issue with referencing my library. I'm a bit bogged down with work currently and don't have time to fix the issue in my PR in AboutLibraries repo. I was wondering if you could fix and submit it for me. Thank you!

mueller-ma:
Unfortunately I cannot help you with the PR, I had to ask for help as well in https://github.com/mikepenz/AboutLibraries/pull/380

mueller-ma:
With AboutLibraries 8.0 only Toasty is missing. I opened https://github.com/GrenderG/Toasty/issues/137 for that.

PR #759, opened 2018-03-08
Remove unused dependencies

mueller-ma:
See #757


maniac103:
Both seem to have been introduced in https://github.com/openhab/openhab-android/commit/01dfbbb7e29927151c02055e9e0c73eefbff5978 - most of which has been removed later.

Issue #637, opened 2018-01-13
Colors in light theme too bright


cribskip:
When using `valuecolor=[ON="green"]`, basic UI shows a darker green than the app. Even when chaning the basic-ui theme, an intensiv light green (seems `#00FF00') is shown.

This is too bright for a white background.

We should change the colors at least in the basic-ui theme to match the basic ui and provide better accessibility through higher contrast.

Here's a quick comparison between basic UI (first) and app (second)
![2018-01-13 08_36_35-technik](https://user-images.githubusercontent.com/1429498/34903996-e2488d24-f83c-11e7-8ac8-06e365c6dc81.png)

![img_20180113_084202](https://user-images.githubusercontent.com/1429498/34904048-c384de3c-f83d-11e7-8c84-1646f0f7a5e1.jpg)

mueller-ma:
Can you do some research which colors are used in basic ui?

FlorianSW:
This green is the default green when you parse a color string (`green`) through Color.parseColor. I'm not quite sure, if we should change this behaviour :/

@mueller-ma  This are the color codes, I think:
https://docs.openhab.org/configuration/sitemaps.html#label-and-value-colors

However, in my opinion, this documentation is quite inaccurate. The color code is just implemented in core products of openHAB (eclipse smarthome), however, there's no guarantee, that these color codes are used by other clients, too. If the color codes are guaranteed, then the server should send us the color code and not the name of the color (where the assigned color code for this color name van vary from client to client). IMHO: When using the color name, instead of the code, the user should currently accept, that the exact color used can be different on different clients.

TL;DR:
I wouldn't change that in the app, I would suggest changing the behaviour of the server (to send the exact color code, which should be used, instead of the name).

What do you think?

mueller-ma:
When we have names, we can change it easier when having a dark color on a dark theme.

I opened an issue in esh since we should discuss it there: https://github.com/eclipse/smarthome/issues/4916

maniac103:
@FlorianSW From my understanding, the color names are guaranteed, not the color codes:

> Generally, you can expected that valid HTML colors will be accepted (e.g. “green”, “lightgrey”, “#334455”), but note that a UI may only accept internally defined colors, or work with a special theme. The color **names** above are agreed on between all openHAB UIs and are therefor your safest choice

(Emphasis mine)

So AFAICT we could have a lookup table of known colors to (per-theme) color resources and fall back to Color.parseColor if that lookup fails.

[...]

PR #762, opened 2018-03-08
Convert all model objects to be immutable

maniac103:
Use the AutoValue library to reduce boilerplate code, and make construction logic of the objects easier to understand.


maniac103:
Included also are some perfomance optimizations (generate derived item state at construction time instead of repeatedly at runtime). Good thing we had unit tests for those things :-)

mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java
```diff
@@ -10,293 +10,329 @@
...
+        public abstract Builder period(String period);
+        public abstract Builder service(String service);
+        public abstract Builder legend(@Nullable Boolean legend);
+        public abstract Builder height(int height);
+
+        public OpenHABWidget build() {
+            // Consider a minimal refresh rate of 100 ms
+            refresh(Math.max(refresh(), 100));
+            // Default period to 'D'
+            if (period() == null || period().isEmpty()) {
```
mueller-ma: TextUtils.isEmpty() ?

mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java
```diff
@@ -10,293 +10,329 @@
...
+                // set the state to "OFF" instead of 0 or to "ON" to fetch the correct icon
+                try {
+                    int itemStateNumber = Integer.valueOf(itemState);
+                    if (itemStateNumber == 0) {
+                        itemState = "OFF";
+                    } else {
+                        itemState = "ON";
+                    }
+                } catch (java.lang.NumberFormatException e) {
+                    // Item state is not a number, not sure if that can happen, but good to catch
```
mueller-ma: Should itemState set to null here?

mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java
```diff
@@ -10,293 +10,329 @@
...
 
-    public void setEncoding(String encoding) {
-        this.encoding = encoding;
+    private static Integer parseColor(String color) {
+        if (color != null && !color.isEmpty()) {
+            if (color.equals("orange")) {
+                color = "#FFA500";
+            }
+            try {
+                return new Integer(Color.parseColor(color));
```
mueller-ma: Android Studio suggests to change that to `Integer.valueOf(Color.parseColor(color))`.

mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java
```diff
@@ -10,293 +10,329 @@
...
+                // fall through
+            }
+        }
+        return Type.Unknown;
     }
 
-    public void setEncoding(String encoding) {
-        this.encoding = encoding;
+    private static Integer parseColor(String color) {
+        if (color != null && !color.isEmpty()) {
```
mueller-ma: TextUtils.isEmpty() ?

mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java
```diff
@@ -0,0 +1,418 @@
...
+    public void getIconPath_iconExists_returnIconUrlfromImages() throws Exception {
+        assertEquals("images/groupicon.png", sutXml.get(0).iconPath());
+    }
+
+    @Test
+    public void testGetIconPath() throws Exception {
+        assertEquals("icon/groupicon?state=OFF&format=PNG", sut1.get(0).iconPath());
+        assertEquals("icon/groupicon?state=ON&format=SVG", sut2.get(0).iconPath());
+        assertEquals("icon/slider?state=81&format=SVG", sut3.get(1).iconPath());
+        assertEquals("Rollersutter icon should always be 0 to 100, not ON/OFF", "icon/rollershutter?state=0&format=SVG", sut3.get(2).iconPath());
```
mueller-ma: s/should/must/ ?

mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java
```diff
@@ -0,0 +1,418 @@
...
+    @Test
+    public void testGetChildren() throws Exception {
+        assertEquals("demo11", sut1.get(1).id());
+        assertEquals("0202_0_0_1", sut3.get(2).id());
+    }
+
+    @Test
+    public void testGetPeriod() throws Exception {
+        assertEquals("M", sut1.get(0).period());
+        assertEquals("Object has no period, should default to D", "D", sut2.get(0).period());
```
mueller-ma: ...default to 'D' ?

mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java
```diff
@@ -0,0 +1,418 @@
...
+    @Test
+    public void testGetStep() throws Exception {
+        assertEquals(1F, sut1.get(0).step());
+        assertEquals(0.1F, sut2.get(0).step());
+    }
+
+    @Test
+    public void testGetRefresh() throws Exception {
+        assertEquals(1000, sut1.get(0).refresh());
+        assertEquals("Min refresh is 100, object has set refresh to 10",100, sut2.get(0).refresh());
```
mueller-ma: Whitespace after `10",`

mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java
```diff
@@ -0,0 +1,418 @@
...
+    @Test
+    public void testHasMappings() throws Exception {
+        assertEquals(true, sut1.get(0).hasMappings());
+        assertEquals(true, sut2.get(0).hasMappings());
+    }
+
+    @Test
+    public void testGetColors() throws Exception {
+        Assume.assumeThat(Color.parseColor("blue"), equalTo(Color.BLUE));
+        // Doesn't seem to work via Android Studio
```
mueller-ma: I haven't tested it, when I added this test, but doesn't it work, when you move it to mobile/src/androidTest/java/org/openhab/habdroid/model/OpenHABWidgetTest.java ?

mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java
```diff
@@ -0,0 +1,418 @@
...
+        Document document = builder.parse(new InputSource(new StringReader(xml)));
+        Node rootNode = document.getFirstChild();
+        return rootNode;
+    }
+
+    /**
+     * @param id get different json objects depending on the id
+     *           1: All values are set
+     *           2: Different colors, no periode, max step size < 1, chart refresh < 100
+     *           3: Frame with Slider, Rollershutter switch and Rollershutter group
```
mueller-ma: Can you add and test
* setpoint with minValue > maxValue
* setpoint with stepSize < 0 ?

mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java
maniac103: I think @FlorianSW previously asked (in a comment in #562 IIRC) for TextUtils not to be used in base code because it doesn't work for the tests.

mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java
maniac103: I have no idea, I just copied what was previously present in OpenHAB2Widget.java.

mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java
mueller-ma: Right, it's an Android method. However we could write your own method in Utils.

mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java
maniac103: Couldn't we also use the constant (#ff0000ff that is)? Moving this just for the constant definition seems a bit too much.

mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java
mueller-ma: Ah, that's my code. The `else if (type == Type.Switch && !hasMappings` branch can be simplified.
````java
if (itemState.equals("0")) {
itemState = "OFF";
} else {
itemState = "ON";
}
````

Comment (maniac103) [reply mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java]: Indeed, or even further: `itemState = itemState.equals("0") ? "OFF" : "ON"; `

File: mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java (hunk 4 above)
Comment (maniac103) [reply mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java]: See above ;-)

File: mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java (hunk 1 above)
Comment (maniac103) [reply mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java]: I'd leave this up for later and replace all such occurrences in our code. Alternatively, we could mock this method accordingly in our tests.

File: mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java (hunk 2 above)
Comment (maniac103) [reply mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java]: Done.

File: mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java (hunk 3 above)
Comment (maniac103) [reply mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java]: Done.

File: mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java (hunk 5 above)
Comment (maniac103) [reply mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java]: Done.

File: mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java (hunk 6 above)
Comment (maniac103) [reply mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java]: Done.

File: mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java (hunk 7 above)
Comment (maniac103) [reply mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java]: Done.

File: mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java (hunk 8 above)
Comment (maniac103) [reply mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java]: Removed the assumption, it was unneeded anyway (blue wasn't used in test JSON).

File: mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java (hunk 9 above)
Comment (maniac103) [reply mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java]: Done.

File: mobile/src/test/java/org/openhab/habdroid/model/OpenHABWidgetTest.java [hunk 10]
```diff
@@ -0,0 +1,412 @@
...
+        assertEquals(true, sut2.get(0).hasMappings());
+    }
+
+    @Test
+    public void testGetColors() throws Exception {
+        Integer i = new Integer(0);
+        assertEquals(i, sut1.get(0).labelColor());
+        assertEquals(i, sut1.get(0).iconColor());
+        assertEquals(i, sut1.get(0).valueColor());
+        assertEquals(i, sut2.get(0).labelColor());
```
mueller-ma: This test is still wrong. It passes, but has the wrong assertions. The colors if sut2 are orange, blue and red.

maniac103: That's right, but it always has been like that. We either can fix this now by copying the implementation of Color.parseColor into a mocked implementation, or we could wait for #637 after which we could assert against the color names. I'd prefer the latter ;-)

mueller-ma: Yes, the latter one sounds better.

mueller-ma:
Is this PR RTM? The app works like intended and unit tests pass.

maniac103:
From my side it is.

mobile/src/main/java/org/openhab/habdroid/model/OpenHABNFCActionList.java
```diff
@@ -1,87 +0,0 @@
...
-                        || "Switch".equals(openHABWidget.getItem().getType())
-                        || "Switch".equals(openHABWidget.getItem().getGroupType())) {
-                    actionNames.add(ctx.getString(R.string.nfc_action_on));
-                    actionCommands.add("ON");
-                    actionNames.add(ctx.getString(R.string.nfc_action_off));
-                    actionCommands.add("OFF");
-                    actionNames.add(ctx.getString(R.string.nfc_action_toggle));
-                    actionCommands.add("TOGGLE");
-                    //RollerShutterItem changed to RollerShutter in later builds of OH2
-                } else if ("RollershutterItem".equals(openHABWidget.getItem().getType())
```
mueller-ma: Something I missed during review: Is `RollerShutterItem` still supported?

maniac103: Yep: https://github.com/openhab/openhab-android/pull/762/files/ef87edaf0bc580c1e75e0a088d17f6292cf6fb3c#diff-0d55b6c1316f1f8bdeeaa49111054025R169

Issue #649, opened 2018-01-16
Google Cloud Messaging connection state


ThomDietrich:
Hello,
the "About" section in my installation states that GCM registration was not successful. A user (but honestly also I) doesn't know what that means and if he has to do something, and what. Any ideas how to cope with that? Maybe we just need to enhance the error message.

mueller-ma:
See #560

ThomDietrich:
ups... I see, everything is under control 😃

mueller-ma:
@ThomDietrich The string "GCM status" was changed to "Push notification status". It should be clear now, what it is does.

ThomDietrich:
Nice ;)

Issue #560, opened 2017-12-25
Inform user why GCM registration failed


mueller-ma:
Currently it just shows that gcm registration failed, but not why. Possible reason:

1. No play services
2. Outdated play services
3. Foss build
4. No remote url configured

maniac103:
FOSS build probably shouldn't show this line at all.

mueller-ma:
We should definitly show it and that it is not available in the foss build.
"You're running the foss build, gcm is not available in this version. Click here to download the full version from play store" or similar.

maniac103:
It's obviously not my call to make, but I disagree there. Play Store is the 'normal' way of getting the app, as in: used by the majority of people. If someone went to F-Droid to get the app, he already made an informed decision to not want non-OSS software. Google parts being missing is a natural consequence of that, and it's also the case among all other apps on F-Droid. Therefore IMHO there is no point directing those users to Play Store in the about dialog, as those users made an informed decision to not use Play Store when they installed the app.

mueller-ma:
Ok, I agree to not display a link. But think of devs building accidentally the wrong flavor and trying to fix gcm for hours, until they realise that foss flavor is default.

FlorianSW:
I think, that will be useful. After fixing #572 this would also look much snoozer I think :)

PR #775, opened 2018-03-11
Refactor GCM implementation

maniac103:
Keep notion of 'Google Cloud Messaging' in full flavor code. Also generate cloud connection object only if settings endpoint of server API can be reached.


maniac103:
The only problem I see in this PR is that the connection unit test now fails when checking the returned cloud connection object, as it's created both asynchronously and dependent on a HTTP call. @FlorianSW Any thoughts on how to solve this? Or should we simply drop the corresponding unit test?

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -0,0 +1,103 @@
...
+    public GcmRegistrationService() {
+        super("GcmRegistrationService");
+    }
+
+    @Override
+    protected void onHandleIntent(@Nullable Intent intent) {
+        CloudConnection connection =
+                (CloudConnection) ConnectionFactory.getConnection(Connection.TYPE_CLOUD);
+        if (connection == null) {
+            return;
```
maniac103: We may need to handle this differently (register listener, wait for connection to appear or for timeout).

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.java
```diff
@@ -296,19 +307,59 @@ private static Connection determineAvailableConnection(Context context,
...
+                        } catch (JSONException e) {
+                            Log.d(TAG, "Unable to parse notification settings JSON", e);
+                        }
+                        mCloudConnectionCheckHandle = null;
+                        CloudMessagingHelper.onConnectionUpdated(ctx, mCloudConnection);
+                    }
+                });
+            } else {
+                CloudMessagingHelper.onConnectionUpdated(ctx, null);
+            }
```
maniac103: We could also move all of this to the update handler thread and do the HTTP call synchronously there, so that available and cloud connection finally are updated together. Thoughts?

maniac103: Thinking about it I like this solution. Additionally we can add a blockingWaitForInitialization() call that the GCM service then can call; we then can also convert the voice service back to an IntentService and make it use that method.

maniac103:
Last two comments are untested so far; will test them tomorrow. The problem of the failing unit test remains, though...

FlorianSW:
Hi,

I've only seen this comment in my inbox so far and haven't looked at it deeply, yet. I just want to inform you, that I'll most likely take a look at this tomorrow :) sorry for this delay :]

Best

mueller-ma:
Do you also want to take care of #560 in this PR?

maniac103:
> Do you also want to take care of #560 in this PR?

I'd prefer not to. Especially the Play Services version check takes a bit of effort.

mobile/src/full/java/org/openhab/habdroid/core/GcmListenerService.java
```diff
@@ -0,0 +1,84 @@
...
+        int notificationId = notificationIdString != null
+                ? Integer.parseInt(notificationIdString) : 1;
+
+        if ("notification".equals(data.getString("type"))) {
+            //for now we use google.sent_time as a time reference for our notifications as the gcm
+            //message does not contain the actual event time from the openhab instance at the moment,
+            //in case gcm time is also missing, we fall back to the reception time which may be delayed (old behaviour)
+            long timestamp = data.getLong("google.sent_time", System.currentTimeMillis());
+            sendNotification(data.getString("message"), timestamp, notificationId);
+            // If this is hideNotification, cancel existing notification with it's id
```
mueller-ma: `its`

mobile/src/full/java/org/openhab/habdroid/core/GcmListenerService.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    public void onMessageReceived(String from, Bundle data) {
+        String notificationIdString = data.getString(EXTRA_NOTIFICATION_ID);
+        int notificationId = notificationIdString != null
+                ? Integer.parseInt(notificationIdString) : 1;
+
+        if ("notification".equals(data.getString("type"))) {
+            //for now we use google.sent_time as a time reference for our notifications as the gcm
+            //message does not contain the actual event time from the openhab instance at the moment,
+            //in case gcm time is also missing, we fall back to the reception time which may be delayed (old behaviour)
```
mueller-ma: I would remove `(old behaviour)` here as it is unclear what "old" is.

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.java
```diff
@@ -100,14 +116,26 @@ public static void shutdown() {
         sInstance.mUpdateThread.quit();
     }
 
+    public static void blockingWaitForInitialization() {
+        synchronized (sInstance.mInitializationLock) {
+            while (!sInstance.mIsInitialized) {
+                try {
+                    sInstance.mInitializationLock.wait();
+                } catch (InterruptedException e) {
```
mueller-ma: `} catch (InterruptedException ignored) {}` ?

mobile/src/full/java/org/openhab/habdroid/core/GcmListenerService.java
maniac103: You're aware that I just moved this and it came in via another PR, right? :-P

But yeah, we surely can fix this here :-)

mueller-ma: Yes, I'm aware of that. Seems like I missed it in an earlier review.

mueller-ma:
I've tested it and it looks good so far. Nice to get rid of gcmdummy and the "two permissions with the same name" error.

mobile/src/full/java/org/openhab/habdroid/core/GcmListenerService.java
```diff
@@ -0,0 +1,84 @@
...
+import android.os.Bundle;
+import android.preference.PreferenceManager;
+import android.support.v4.app.NotificationCompat;
+import android.support.v4.content.ContextCompat;
+
+import org.openhab.habdroid.R;
+import org.openhab.habdroid.ui.OpenHABMainActivity;
+import org.openhab.habdroid.util.Constants;
+
+public class GcmListenerService extends com.google.android.gms.gcm.GcmListenerService {
```
FlorianSW: Does this need to be the same class name? I kind of dislike that, as it doesn't make clear from the class name, why we extend it on our side. I also do not really like to prefix a class with the app name, but here it would maybe make sense: OpenHABGcmListenerService?

mobile/src/full/java/org/openhab/habdroid/core/GcmListenerService.java
```diff
@@ -0,0 +1,84 @@
...
+        int notificationId = notificationIdString != null
+                ? Integer.parseInt(notificationIdString) : 1;
+
+        if ("notification".equals(data.getString("type"))) {
+            //for now we use google.sent_time as a time reference for our notifications as the gcm
+            //message does not contain the actual event time from the openhab instance at the moment,
+            //in case gcm time is also missing, we fall back to the reception time which may be delayed
+            long timestamp = data.getLong("google.sent_time", System.currentTimeMillis());
+            sendNotification(data.getString("message"), timestamp, notificationId);
+            // If this is hideNotification, cancel existing notification with its id
```
FlorianSW: I think that's obvious from the code, I would remove this comment. Also, the else-if block is not needed, you can do that in a separate if.

You can probably also extract the body of the first if branch into it's own method with a better name (maybe the comment is than not needed anymore, or at least not 3 lines). The same for the second if body, which could make the code easier to understand :)

mobile/src/full/java/org/openhab/habdroid/core/GcmListenerService.java
```diff
@@ -0,0 +1,84 @@
...
+        } else if ("hideNotification".equals(data.getString("type"))) {
+            NotificationManager nm = (NotificationManager) getSystemService(NOTIFICATION_SERVICE);
+            nm.cancel(notificationId);
+        }
+    }
+
+    private void sendNotification(String msg, long timestamp, int notificationId) {
+        NotificationManager nm = (NotificationManager) getSystemService(NOTIFICATION_SERVICE);
```
FlorianSW: The part "(NotificationManager) getSystemService(NOTIFICATION_SERVICE)" can be extracted to it's own method, so you do not need to copy it from line 43 :)

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -0,0 +1,104 @@
...
+
+import okhttp3.Call;
+import okhttp3.Headers;
+
+public class GcmRegistrationService extends IntentService {
+    private static final String TAG = GcmRegistrationService.class.getSimpleName();
+
+    static final String ACTION_REGISTER = "org.openhab.habdroid.action.REGISTER_GCM";
+    static final String ACTION_HIDE_NOTIFICATION = "org.openhab.habdroid.action.HIDE_NOTIFICATION";
+    static final String EXTRA_NOTIFICATION_ID = "notificationId";
```
FlorianSW: Shouldn't the the static from GcmListenerService be used instead of this one? Or the other way around, at least I would recommend not to declare the same constant twice :)

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public GcmRegistrationService() {
+        super("GcmRegistrationService");
+    }
+
+    @Override
+    protected void onHandleIntent(@Nullable Intent intent) {
+        ConnectionFactory.blockingWaitForInitialization();
+        CloudConnection connection =
```
FlorianSW: I wouldn't cast this to a CloudConnection object, more on that later (at the CloudConnection part :))

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -0,0 +1,104 @@
...
+    static final String ACTION_HIDE_NOTIFICATION = "org.openhab.habdroid.action.HIDE_NOTIFICATION";
+    static final String EXTRA_NOTIFICATION_ID = "notificationId";
+
+    public GcmRegistrationService() {
+        super("GcmRegistrationService");
+    }
+
+    @Override
+    protected void onHandleIntent(@Nullable Intent intent) {
+        ConnectionFactory.blockingWaitForInitialization();
```
FlorianSW: hmm, isn't "waitForInitialization" already pretty obvious that it's blocking? I would remove the "blocking" part from the method name. However, I'll most likely repeat this comment later on, too :D

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -0,0 +1,104 @@
...
+            return;
+        }
+
+        if (ACTION_REGISTER.equals(intent.getAction())) {
+            try {
+                registerGcm(connection, connection.getMessagingSenderId());
+            } catch (IOException e) {
+                Log.e(TAG, "GCM registration failed", e);
+            }
+        } else if (ACTION_HIDE_NOTIFICATION.equals(intent.getAction())) {
```
FlorianSW: no need to put it into an else if, a second if is good, too :)

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -0,0 +1,104 @@
...
+            int notificationId = intent.getIntExtra(EXTRA_NOTIFICATION_ID, -1);
```
FlorianSW: This seems like a lot of code for the if body, what do you think about extracting it to it's own method, which makes the code a bit easier to read, if the name if good, like "hideNotification" or something?

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -0,0 +1,104 @@
...
+            if (notificationId >= 0) {
+                GoogleCloudMessaging gcm = GoogleCloudMessaging.getInstance(this);
+                Bundle sendBundle = new Bundle();
+                sendBundle.putString("type", "hideNotification");
+                sendBundle.putString("notificationId", String.valueOf(notificationId));
```
FlorianSW: "notificationId" should be replaced by it's constant value, shouldn't it?

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private void registerGcm(Connection connection, String senderId) throws IOException {
+        InstanceID instanceID = InstanceID.getInstance(this);
+        String token = instanceID.getToken(senderId,
+                GoogleCloudMessaging.INSTANCE_ID_SCOPE, null);
+        String deviceModel = URLEncoder.encode(Build.MODEL, "UTF-8");
+        String deviceId = Settings.Secure.getString(getContentResolver(), Settings.Secure.ANDROID_ID);
+
+        String regUrl = String.format(Locale.US,
+                "/addAndroidRegistration?deviceId=%s&deviceModel=%s&regId=%s",
```
FlorianSW: I would extract this template into it's own constant with a better name, so you do not need to think about what's that when reading the code. Something like "ADD_ANDROID_REG_URL_TEMPLATE"?

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -0,0 +1,104 @@
...
+                deviceId, deviceModel, token);
+
+        Log.d(TAG, "Register device at openHAB-cloud with URL: " + regUrl);
+        connection.getSyncHttpClient().get(regUrl, new MyHttpClient.ResponseHandler() {
+            @Override
+            public void onFailure(Call call, int statusCode, Headers headers, byte[] responseBody, Throwable error) {
+                Log.e(TAG, "GCM reg id error: " + error.getMessage());
+                if (responseBody != null)
```
FlorianSW: Don't we have a codestyle to not use shorthand if/else? If so, here're the brackets missing :)

mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
```diff
@@ -34,68 +31,35 @@
  * This service handles voice commands and sends them to OpenHAB.
  * It will use the openHAB base URL if passed in the intent's extra.
  */
-public class OpenHABVoiceService extends Service implements ConnectionFactory.UpdateListener {
+public class OpenHABVoiceService extends IntentService {
```
FlorianSW: Wasn't it your argument to not use IntentService here? I can't really remember anymore, why, but if so, would you be so kind to explain to me, what've changed now? Just for my interest :)

mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
```diff
@@ -34,68 +31,35 @@
...
-        super.onDestroy();
-        ConnectionFactory.removeListener(this);
-    }
 
-    @Nullable
-    @Override
-    public IBinder onBind(Intent intent) {
-        return null;
-    }
+        ConnectionFactory.blockingWaitForInitialization();
```
FlorianSW: ...probably that one? As it now does not handle async things anymore?

mobile/src/main/java/org/openhab/habdroid/core/connection/CloudConnection.java
```diff
@@ -0,0 +1,19 @@
...
+
+    private String mSenderId;
+
+    public CloudConnection(@NonNull AbstractConnection baseConnection, @NonNull String senderId) {
+        super(baseConnection, TYPE_CLOUD);
+        mSenderId = senderId;
+    }
+
+    @NonNull
+    public String getMessagingSenderId() {
```
FlorianSW: I don't feel that this belongs into this class :/ It shouldn't be needed to cast a Connection object to a concrete implementation, just because so you get access to a specific method of this class. Can't we put this setting back somewhere else?

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.java
```diff
@@ -100,14 +116,26 @@ public static void shutdown() {
         sInstance.mUpdateThread.quit();
     }
 
+    public static void blockingWaitForInitialization() {
```
FlorianSW: in my opinion, waitForInitialization would be enough, as this (from my point of view) already implies a blocking behaviour :)

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.java
```diff
@@ -100,14 +116,26 @@ public static void shutdown() {
...
+        synchronized (sInstance.mInitializationLock) {
+            while (!sInstance.mIsInitialized) {
+                try {
+                    sInstance.mInitializationLock.wait();
+                } catch (InterruptedException ignored) {
+                    // ignored
```
FlorianSW: do we've a checkstyle rule for that (a catch block needs a body)? If not, I would remove the comment, it's already clear from the name of the variable the exception is saved in :)

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.java
```diff
@@ -100,14 +116,26 @@ public static void shutdown() {
...
+
     public static void addListener(UpdateListener l) {
         sInstance.addListenerInternal(l);
     }
 
     private void addListenerInternal(UpdateListener l) {
         if (mListeners.add(l)) {
             if (mNeedsUpdate) {
-                mUpdateHandler.sendEmptyMessage(MSG_TRIGGER_UPDATE);
+                triggerConnectionUpdateIfNeeded();
```
FlorianSW: +1 for removing duplicated code! :)

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.java
```diff
@@ -178,67 +206,95 @@ private Connection getConnectionInternal(int connectionType) {
...
+                        @Override
+                        public void onFailure(Call call, int statusCode, Headers headers, String responseBody, Throwable error) {
+                            Log.e(TAG, "Error loading notification settings: " + error.getMessage());
+                        }
+
+                        @Override
+                        public void onSuccess(Call call, int statusCode, Headers headers, String responseBody) {
+                            try {
+                                JSONObject json = new JSONObject(responseBody);
+                                String senderId = json.getJSONObject("gcm").getString("senderId");
```
FlorianSW: As already stated somewhere else: I don't think, that this belongs into the Connection object. And if it does, I would argue, that it should be part of the Connection interface, so that no cast is needed.

Or, when I think about it: If a connection to specific server has special features/parameters, it should probably be it's own type (class or interface). However, it feeld strange to shift the need to cast the connection object to the caller of the COnnectionFactory, though :(

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -463,7 +449,8 @@ private void processIntent(Intent intent) {
         Log.d(TAG, "Got intent: " + intent);
         if (NfcAdapter.ACTION_NDEF_DISCOVERED.equals(intent.getAction())) {
             onNfcTag(intent.getData());
-        } else if (GcmIntentService.ACTION_NOTIFICATION_SELECTED.equals(intent.getAction())) {
+        } else if (ACTION_NOTIFICATION_SELECTED.equals(intent.getAction())) {
```
FlorianSW: no else-if needed, which is, however, maybe a task for somewhere else, as there're more branches, too. But, if you've time :D

mobile/src/main/java/org/openhab/habdroid/util/Constants.java
```diff
@@ -33,7 +33,6 @@
     public static final String PREFERENCE_DEBUG_MESSAGES        = "default_openhab_debug_messages";
     public static final String PREFERENCE_CLEAR_DEFAULT_SITEMAP =
             "default_openhab_clear_default_sitemap";
-    public static final String DEFAULT_GCM_SENDER_ID            = "737820980945";
```
FlorianSW: *yay* I totally missed the update of myopenhab.org, it's cool that we can now remove this specific value to make the app more generic :)

mobile/src/full/java/org/openhab/habdroid/core/GcmListenerService.java
maniac103: `GcmMessageListenerService`?

mobile/src/full/java/org/openhab/habdroid/core/GcmListenerService.java
maniac103: The second branch already *is* extracted to a separate method. It's only the comment that makes it more than 2 lines.

mobile/src/full/java/org/openhab/habdroid/core/GcmListenerService.java
maniac103: Sorry, but I think one can also overdo 'methodization'. Putting a one-line variable assignment into a separate method doesn't feel meaningful to me.

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
maniac103: I purposefully did that, because I want to decouple the GCM logic (which is only present in full flavor) from the generic activity logic. That's why I defined one constant for each intent 'target'.

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
maniac103: Dunno; don't have strong feelings here. I just wanted to make it 100% clear that this method must not be called from the main thread.

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
maniac103: Huh? Why check twice if the first if already matched and thus the second one can't?

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
maniac103: No, as that one is unrelated to the (intent extra) constant: the value here is defined by the cloud app API.

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
maniac103: Yeah, I can do that.

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
maniac103: I'm not a fan of introducing constants for stuff that is only used once under very specific circumstances. The reason for that is that it IMHO is unneeded indirection, which actually makes it less obvious what's going on.

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
maniac103: Right. Copied that mistake over.

mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
maniac103: My main problem was the approach of ContinuingIntentService, that is, letting the the service run indefinitely due to the async HTTP requests, which made the IntentService useless anyway.
Now I'm needing it for the blocking wait, thus I converted the HTTP to be sync requests and everything is in order :-)

mobile/src/full/java/org/openhab/habdroid/core/GcmListenerService.java
FlorianSW: Good for me :)

mobile/src/full/java/org/openhab/habdroid/core/GcmListenerService.java
FlorianSW: hmm, you're right. Ok, it's not ideal, but I think it's ok :)

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
FlorianSW: You can use return; in the first if, or just keep it. It's just a string comparsion. An else if feels more complicated from my point of view.

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
FlorianSW: Got it! :)

mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
FlorianSW: Ah, ok, thanks for the explanation! :)

mobile/src/main/java/org/openhab/habdroid/core/connection/CloudConnection.java
maniac103: This is a point where I *do* have strong feelings: IMHO it totally makes sense to put the sender ID into this class: it's a property of this specific server connection (as it was fetched from there) and it's only present if the  cloud connection is present (as it's fetched from the cloud API). Why would we want to introduce another random data holder here?
What's your proposed alternative solution?

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.java
maniac103: I'm used to working with AOSP code, where such a check style rule exists. Not sure about our linter configuration.

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.java
maniac103: > And if it does, I would argue, that it should be part of the Connection interface, so that no cast is needed.

I really have to disagree here: why move a property *specific to one particular type of connection* to the base class? This really only leads to method bloat.
If you're unhappy with the cast, this can be solved by splitting `ConnectionFactory.getConnection()` into 3 separate methods to fetch local/remote/cloud connection. Personally, I think the cast is fine.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
maniac103: Maybe not strictly needed, but perfectly fine here as it's an either-or condition. One could even switch() on the action.

maniac103:
Decided to partly implement #560 but left out Play Services version check for now. GCM registration status is now hidden on FOSS build, and more detailed on full build.

mueller-ma:
Can you change "gcm status" to "Notificatiom support" or similar? And can you show in the foss build a message that gcm is not supported?

maniac103:
> Can you change "gcm status" to "Notificatiom support" or similar?

I could, but why? I'd rather move the strings to the full flavor, I think.

> And can you show in the foss build a message that gcm is not supported?

Again, why? I thought in #560 we reached the consensus to not hint towards the full build in that case.

mueller-ma:
I read some comments of people not knowing what "gcm status" means (I think it was @ThomDietrich but I dont know for sure).

In my comment in #560 I meant not to display a sentence that gcm is not supported, but no further hints towards the play store.

maniac103:
> I read some comments of people not knowing what "gcm status" means. See #649

OK, so 'Push notification channel' or something among those lines?

> In my comment in #560 I meant not to display a sentence that gcm is not supported, but no further hints towards the play store.

It still doesn't make sense to me to show the lack of support of a given feature in a given build in the UI. We aren't going to display the lack of e.g. Mapview support in the foss build in the UI either, are we?

As mentioned in https://github.com/openhab/openhab-android/issues/560#issuecomment-353929458 the user already made an informed decision when he installed the foss build via F-Droid (and is unlikely to switch to the Play Store build for this). If he built himself, he should be able to either read the source or the README file.

mueller-ma:
> But shouldn't we also remove the 'sender ID' from the UI then? That value doesn't seem intuitively understandable either.

Is it usefull to see it? If yes, maybe show a toast with the id after clicking on "GCM Status"

> It still doesn't make sense to me to show the lack of support of a given feature in a given build in the UI.

Give me some time to think about it.

maniac103:
> Is it usefull to see it?

Not sure. The value shown equals the GCM account ID the user has to configure in the openhab-cloud instance config.

mueller-ma:
> Not sure. The value shown equals the GCM account ID the user has to configure in the openhab-cloud instance config.

I think showing it as toast or - maybe even better - as dialog should do it.

From the docs:
````java
                .setOnClickListener(new MaterialAboutActionItem.OnClickListener() {
                    @Override
                    public void onClick() {
                        Toast.makeText(ExampleMaterialAboutActivity.this, "Version Tapped", Toast.LENGTH_SHORT).show();
                    }
                })
````

maniac103:
Nah, that's a bit too obscure. IMHO either we think the value is important and keep showing it, or we think it's not important and don't show it.

mueller-ma:
I think it's better to show the gcm status on foss builds.
The ID should be shown, otherwise people cannot connect to their oh cloud, right?

maniac103:
> I think it's better to show the gcm status on foss builds.

And what text would you like to show? `Not available in this build` might be a bit unclear ('build' is a technical term, and it doesn't help in determining what version to install to get it), `Not available in this version` might imply that it's a matter of time until it's supported.

> The ID should be shown, otherwise people cannot connect to their oh cloud, right?

No. The ID generally is set up only on the openhab-cloud installation. The app fetches it from there and then uses it for registration. For people setting up their own cloud interface instance, showing it might be useful as it allows double checking whether the shown sender ID matches the one they configured; for people using myopenhab.org showing it doesn't serve any meaningful purpose, AFAICT.

mueller-ma:
"Not available in the foss version"?

I don't want that people, who install their own oh cloud, have to install adb to get the id. Maybe the snackbar isn't that bad? It's obscure, but we can tell those people how to see the id, if they open an issue.

maniac103:
> I don't want that people, who install their own oh cloud, have to install adb to get the id

They don't need to. Again: they need to configure it in OH cloud *first*, the app displaying it is a *confirmation only*, nothing more. As the app fetches that ID from the cloud instance, there's no way the cloud instance configuration can depend on the app.

mueller-ma:
> I don't want that people, who install their own oh cloud, have to install adb to get the id **EDIT: for confirmation**

If someone misses the id and opens a ticket, you can just answer "Click on the gcm status and see it" without having to add it and build an apk for him.

maniac103:
> If someone misses the id and opens a ticket

What should he want to use the ID value for?

(Don't get me wrong, I have no issues keeping the ID in the string. I just want to understand what exactly you're heading at.)

mueller-ma:
> What should he want to use the ID value for?

@FlorianSW What was your motivation for displaying the gcm id?

maniac103:
@mueller-ma I'd guess being able to differentiate whether the sender ID provided by openhab-cloud was used or the fallback ID that previously was hardcoded into the app. This PR removes the fallback ID though.

mueller-ma:
Ok, in this case I can live without the id.

mueller-ma:
@floriansw Does the pr look good for you?

maniac103:
Thinking about it, I wonder whether I should revert the last commit. As it stands, it's not possible to load the notification list for FOSS builds. I wonder whether making that possible would be useful even without being able to receive notifications?

mueller-ma:
> I wonder whether making that possible would be useful even without being able to receive notifications?

Right, it seems usefull to me.

maniac103:
OK, in that case I'll revert that commit.

mueller-ma:
Two things:
1. Does the "Push notification status" auto update from "Registering..." to "Registered"?
2. It shows me "Gerät mit ID %s registriert"

maniac103:
> 1. Does the "Push notification status" auto update from "Registering..." to "Registered"?

No. Is this worth the effort? One could send a broadcast via LocalBroadcastManager whenever the status changes, but how does one update the about list stuff?

> 2. It shows me "Gerät mit ID %s registriert"

Yes, because I updated the English string only. From other projects I'd assume the localized strings either to be updated or removed with the next Crowdin sync, which is why I didn't remove them manually. Is that assumption wrong?

mueller-ma:
>No. Is this worth the effort?

Probably not.

> Yes, because I updated the English string only.

Right, changing translated strings causes merge conflicts.

mueller-ma:
@maniac103 Merging this broke the build: https://travis-ci.org/openhab/openhab-android/jobs/360904010#L1666

maniac103:
@mueller-ma Oops, that's a hidden merge conflict with #789. I can't fix it at the moment (no PC near me, only phone), but in case you want to fix it, you need to make the following changes in OpenHABMainActivity:

- move lines 545-550 to onCloudConnectionChanged()
- change the check to `if (mPendingOpenNotifications && connection != null)`
- optional, but probably nice: combine lines 540 and 541 to a single if condition again

PR #777, opened 2018-03-12
Clean up NFC write tag activity.

maniac103:


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWriteTagActivity.java
```diff
@@ -131,53 +126,50 @@ public void onResume() {
...
+            return;
+        }
+
         Tag tagFromIntent = intent.getParcelableExtra(NfcAdapter.EXTRA_TAG);
-        String openhabURI = "";
         //do something with tagFromIntent
         Log.d(TAG, "NFC TAG = " + tagFromIntent.toString());
-        Log.d(TAG, "Writing page " + sitemapPage + " to TAG");
-        TextView writeTagMessage = (TextView) findViewById(R.id.write_tag_message);
+        Log.d(TAG, "Writing page " + mSitemapPage + " to TAG");
```
mueller-ma: `to tag`

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWriteTagActivity.java
```diff
@@ -187,8 +179,7 @@ public void writeTag(Tag tag, String openhabUri) {
                 writeTagMessage.setText(R.string.info_write_tag_finished);
                 autoCloseActivity();
             } catch (IOException | FormatException e) {
-                if (e.getMessage() != null)
-                    Log.e(TAG, e.getMessage());
+                Log.e(TAG, "Writing to unformatted tag failed: " + e);
```
mueller-ma: `Log.e(TAG, "Writing to unformatted tag failed", e);`?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWriteTagActivity.java
```diff
@@ -206,14 +197,8 @@ public void writeTag(Tag tag, String openhabUri) {
...
-                } catch (IOException e) {
-                    // TODO Auto-generated catch block
-                    if (e != null)
-                        Log.e(TAG, e.getClass().getCanonicalName());
-                    writeTagMessage.setText(R.string.info_write_failed);
-                } catch (FormatException e) {
-                    // TODO Auto-generated catch block
-                    Log.e(TAG, e.getMessage());
+                } catch (IOException | FormatException e) {
+                    Log.e(TAG, "Writing to formatted tag failed: " + e);
```
mueller-ma: Same as above

mueller-ma:
I don't own a nfc device, so I cannot test it. If you tested it and feel comfortable, I can merge this.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWriteTagActivity.java
maniac103: Done.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWriteTagActivity.java
maniac103: I don't think the stack trace is actually useful here; that's why I chose this way.

maniac103:
> I don't own a nfc device, so I cannot test it. If you tested it and feel comfortable, I can merge this.

Yeah, I just tested this on a Nexus 4, it still works as well as before.

PR #779, opened 2018-03-12
Make app_name for beta builds translatable

mueller-ma:
There might be a language where the word "Beta" is usually translated.
Is it in English `openHAB Beta` or `openHAB beta`?


maniac103:
Does this actually work with Crowdin (a nontranslatable string overlaid with a translatable one)?

mueller-ma:
I think so, as both strings are in different files. See: https://crowdin.com/project/openhab-android/de#

maniac103:
OK, if that works, fine with me :-)

mueller-ma:
With the crowdin config you can wait for #775 as that pr introduces new strings files.

Issue #770, opened 2018-03-10
Connection to local url not possible


mueller-ma:
### Actual behaviour
App shows "Discovery failed" error

### Expected behaviour
Should connect to configured local url

### Can you reproduce the issue in demo mode?

No

#### Client
* Android version: 7.1.2
* Device model: Oneplus X
* Stock or customized ROM: 14.1
* App version : 2.0.25-beta
* Build flavor : 
* Device language: DE

#### App log
```
E/AbstractConnection: connect timed out
D/ContentController: Indicate missing configuration
D/PageConnectionHolderFragment: updateActiveConnections: URL list [], connection null
D/AbstractConnection: Checking reachability of http://openhab/
E/AbstractConnection: connect timed out
D/ContentController: Indicate missing configuration
D/PageConnectionHolderFragment: updateActiveConnections: URL list [], connection null
W/ResourceType: Failure getting entry for 0x7f0d0002 (t=12 e=2) (error -75)
I/AppCompatViewInflater: app:theme is now deprecated. Please move to using android:theme instead.
I/art: Do partial code cache collection, code=19KB, data=30KB
I/art: After code cache collection, code=14KB, data=26KB
I/art: Increasing code cache capacity to 128KB
D/OpenHABMainActivity: onSaveInstanceState
D/OpenHABMainActivity: onStop()
D/PageConnectionHolderFragment: onStop()
D/OpenHABMainActivity: onActivityResult requestCode = 1004, resultCode = -1
D/PageConnectionHolderFragment: onStart(), started false
D/OpenHABMainActivity: onResume()
D/ContentController: Indicate missing configuration
D/PageConnectionHolderFragment: updateActiveConnections: URL list [], connection null
D/AbstractConnection: Checking reachability of http://openhab/
E/AbstractConnection: connect timed out
D/ContentController: Indicate missing configuration
D/PageConnectionHolderFragment: updateActiveConnections: URL list [], connection null
```

mueller-ma:
I found the reason: I have AFWall+ (Firewall) running and allow the oh app only to connect to the local network, but not to the internet. When I allow internet connections, the app works.

FlorianSW:
I assume our configured local URL isn't http://openhab/? Because that's the URL the app tries to connect:
`D/AbstractConnection: Checking reachability of http://openhab/`

mueller-ma:
I replace my dns name with openhab, the log shows the correct domain. Should have done that more clear though.

[...]

Issue #271, opened 2017-03-02
Very slow local access if Internet is down


andredts:
Interface takes over 10 seconds to respond every time the local network is OK but there is no internet connection. Using the web interface in the same wifi does not seems to be slower, only the app.

mueller-ma:
Do you use openhab cloud?

maniac103:
@andredts Is this still an issue?

mueller-ma:
This should have been fixed by recent connection improvements.

PR #780, opened 2018-03-12
Wait up to 2 seconds for connection to become available

### LINKED ISSUE #755: Discovery failed error is shown after network reconnect
1. Open app
2. disable wifi
3. enable wifi
4. Discovery failed error is shown, even when a local url ist set

2.0.19-beta
Android 7.1.2

FlorianSW:
Especially for local connections, like WiFi or maybe VPN, too, a
connection may not be "available" (~ connected to the internet) once it is
"connected". To allow a connection to do whatever is needed to do, wait up
to 2 seconds (in 10 200ms chunks) for the connection to become available.
If it is not available after this timeframe, the connection is returned as
"not reachable".

Fixes #755


mueller-ma:
> Especially for local connections, like WiFi or maybe VPN, too, a connection may not be "available" (~ connected to the internet) once it is "connected".

What if I'm on my wifi and the internet link is down (Related #271)? Or internet access is blocked via a firewall (#770)?
Latter is a very rare case, but the former might happen sometimes.

mobile/src/main/java/org/openhab/habdroid/core/connection/Connection.java
```diff
@@ -54,5 +54,5 @@
     /**
      * @return Whether the this connection is currently reachable.
      */
-    boolean checkReachabilityInBackground();
+    boolean checkReachability();
```
maniac103: If you're changing this method name, please add a statement to the Javadoc saying this must not be called from the main thread.

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
```diff
@@ -10,15 +10,16 @@
...
+import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.net.MalformedURLException;
 import java.net.Socket;
 import java.net.URL;
 
 public abstract class AbstractConnection implements Connection {
     private static final String TAG = AbstractConnection.class.getSimpleName();
-
+    private static final int MAX_CONNECT_TRIES = 10;
```
maniac103: `MAX_CONNECT_ATTEMPTS`

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
```diff
@@ -97,24 +98,60 @@ public String getOpenHABUrl() {
...
+        }
+        return true;
+    }
+
+    private Socket createConnectedSocket() throws MalformedURLException {
+        for (int retries = 0; retries < MAX_CONNECT_TRIES; retries++) {
+            Socket s = new Socket();
+            InetSocketAddress socketAddress = getReachabilitySocketAddress();
+            try {
+                s.connect(socketAddress, 1000);
```
maniac103: In case of socket timeout, this now waits up to 12 seconds (10 x 1000ms timeout + 10 x 200ms sleep). I think that's a bit too much.

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
maniac103: IMHO this could also be local to `checkReachability` though, as the meaning of the constant is unclear outside of that method...one can't really derive from the constant name what its used for.

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
```diff
@@ -97,24 +98,60 @@ public String getOpenHABUrl() {
...
-            Socket s = new Socket();
-            s.connect(new InetSocketAddress(url.getHost(), checkPort), 1000);
-            Log.d(TAG, "Socket connected");
-            s.close();
-            return true;
-        } catch (Exception e) {
-            Log.e(TAG, e.getMessage());
+            s = createConnectedSocket();
+        } catch (MalformedURLException e) {
+            Log.d(TAG, "URL could not be parsed", e);
```
maniac103: This can't actually happen, as we validate the URL when creating the connection object.
How about using `android.net.Uri.parse()` and using `uri.getScheme()` below?

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
```diff
@@ -97,24 +98,60 @@ public String getOpenHABUrl() {
...
+
+        if (s == null) {
+            return false;
+        }
+
+        Log.d(TAG, "Socket connected");
+        try {
+            s.close();
+        } catch (IOException e) {
+            Log.d(TAG, "Could not close socket connection.", e);
```
maniac103: That log IMHO is useless. There's nothing we can do anyway, and it's not like we or the user would care.

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
```diff
@@ -97,24 +98,60 @@ public String getOpenHABUrl() {
...
+
+    @NonNull
+    private InetSocketAddress getReachabilitySocketAddress() throws MalformedURLException {
+        URL url = new URL(getOpenHABUrl());
+        int checkPort = url.getPort();
+        if (url.getProtocol().equals("http") && checkPort == -1)
+            checkPort = 80;
+        if (url.getProtocol().equals("https") && checkPort == -1)
+            checkPort = 443;
+
```
maniac103:
```
if (checkPort < 0) {
    checkPort = url.getProtocol().equals("https") ? 443 : 80;
}
```

FlorianSW:
> What if I'm on my wifi and the internet link is down (Related #271)? Or internet access is blocked via a firewall (#770)?
> Latter is a very rare case, but the former might happen sometimes.

That shouldn't matter, as the current check for connectivity doesn't check against a generic URL, like google.com or openhab.org. It is rather trying to open a connection to the connections target (the URL of the openhab instance configured in the settings). As long as the app is allowed to connect to this target, the check will most likely succeed.

What I meant with my comment was: A WiFi connection may enter the CONNECTED state, even before the connection is available to process connection requests (let's say, the DHCP request isn't finished and the adapter has no IP address assigned), in this case, we would already try the connection check, which will fail, as the adapter can not yet process any connections.

mueller-ma:
Thanks for clarifing this :+1:
I'll review soon.

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
```diff
@@ -97,24 +97,54 @@ public String getOpenHABUrl() {
...
         }
+        return true;
+    }
+
+    private Socket createConnectedSocket() {
+        for (int retries = 0; retries < 10; retries++) {
+            Socket s = new Socket();
+            InetSocketAddress socketAddress = getReachabilitySocketAddress();
+            try {
+                s.connect(socketAddress, 1000);
```
maniac103: Not: could return s here and save the continue below.

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
```diff
@@ -97,24 +97,54 @@ public String getOpenHABUrl() {
...
-            return false;
+        } catch (IOException ignored) {
         }
+        return true;
+    }
+
+    private Socket createConnectedSocket() {
+        for (int retries = 0; retries < 10; retries++) {
+            Socket s = new Socket();
+            InetSocketAddress socketAddress = getReachabilitySocketAddress();
```
maniac103: Should probably move this out of the loop, as it's the same every time anyway. Doing so probably also would avoid the need for a separate method for this, wouldn't it?

maniac103:
Mostly LGTM, just got some minor comments.

mueller-ma:
Superseded by #1232

PR #789, opened 2018-03-20
Defer main activity intent processing until onStart().

maniac103:
Updating the content controller while not started is not allowed.

Fixes a crash happening when clicking the notification while the app is not open.


mueller-ma:
:rocket:

PR #790, opened 2018-03-20
Simplify widget list selection highlight code.

### LINKED ISSUE #718: Click on chart shows outdated chart
In demo mode
1. Outside temperature
2. Select day
3. Short after select week
4. After the week chart is shown, click on it

Expected:
Nothing should happen

Acutall:
Day chart is shown for a short time

maniac103:
Don't temporarily set highlight on click, but only if really needed in
two-pane layout.

Closes #718


maniac103:
In case you're wondering how this change fixes #718:
- assume a single chart being bound with period 'D'
- change period to 'W'
- new data arrives from server with chart type 'W', this image is bound to a new ViewHolder, the old ViewHolder is recycled
- click on it -> setSelection is called, which causes notifyItemChange() to be called on the chart
- this in turn causes the chart to be re-bound, it turns out that the previously recycled ViewHolder is re-used and thus displays its previous content (the 'D' chart) until the new content (the 'W' chart) is loaded

PR #792, opened 2018-03-21
Fix OpenHABNotification object annotations.

maniac103:
The icon and severity members are not guaranteed to be present, while
the message member is expected to be present.

(Fixes NPE when opening notifications page)

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/OpenHABNotification.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/OpenHABNotification.java
@@ -14,6 +14,8 @@
 
 package org.openhab.habdroid.model;
 
+import android.support.annotation.Nullable;
+
 import com.google.auto.value.AutoValue;
 
 import org.json.JSONException;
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/model/OpenHABNotification.java
```diff
@@ -48,14 +52,14 @@ public static OpenHABNotification fromJson(JSONObject jsonObject) throws JSONExc
...
-                // keep created at null
+                // keep created at 0
             }
         }
 
         return new AutoValue_OpenHABNotification.Builder()
                 .icon(jsonObject.optString("icon", null))
                 .severity(jsonObject.optString("severity", null))
-                .message(jsonObject.optString("message", null))
+                .message(jsonObject.getString("message"))
```
mueller-ma: What happens now, when no message is present?

maniac103: A `JSONException` will be thrown ... IOW, message is required now. A notification without message doesn't make sense anyway, though.

mueller-ma: Maybe it's better to default to `""`. It doesn't make sense, but the app doesn't crash when the server sends invalid messages.

maniac103: The JSONException is catched, thus the app won't crash. That's the very point of this PR ;-) Currently, if message is missing, the Builder object throws a NPE.

PR #793, opened 2018-03-21
Fix stray slash in notification icon URL generation.

maniac103:
Causes 404 when fetching icon.


mueller-ma:
I haven't seen that `//` causes a 404. Can you provide an example link?

maniac103:
I have seen it during my tests with the URL `https://myopenhab.org:443//images/smiley.png` ... The proxying of the image from my own OH (1.8) instance failed in that case. Without the stray slash it worked fine.

maniac103:
Kai also stumbled over a similar issue: https://github.com/openhab/openhab-cloud/issues/158#issuecomment-364803103 ... looks like openhab-cloud is a bit picky about the URLs thrown at it.

PR #806, opened 2018-03-27
Don't persist highlighted page link to fragment arguments.

### LINKED ISSUE #791: Highlight in landscape mode not always resetted
### Steps to reproduce
1. Landscape
2. Go to a subpage (e.g. First Floor)
3. Rotate to portrait
4. Go back
5. First Floor is still marked

Yes

#### Client
* Android version: 5.1.1
* Device model: Fire Tablet 8
* Stock or customized ROM: Custom
* App version : 2940719b8a6f9a3675820129ef100eb5605f2b4c
* Build flavor : Foss
* Device language: DE

maniac103:
Otherwise, the highlight is used when two-pane controller is replaced by
one-pane controller after device rotation, but we never want one-pane
layout to have those highlights. As the two-pane controller reapplies
the highlight when restoring fragments, storing the link as instance
member is sufficient.

Fixes #791


mueller-ma:
Can you rebase this to pass the tests?

PR #817, opened 2018-04-01
Sort sitemaps alphabetically

### LINKED ISSUE #270: Order sitemaps alphabetically

mueller-ma:
but let the default sitemap always be the first one.

Fixes #270

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
@@ -784,12 +784,13 @@ public class OpenHABMainActivity extends AppCompatActivity implements
                 mPendingCall = null;
                 mInitState = InitState.DONE;
 
-                String defaultSitemapLabel = mSettings.getString(Constants.PREFERENCE_SITEMAP_LABEL, "");
+                String defaultSitemapName =
+                        mSettings.getString(Constants.PREFERENCE_SITEMAP_NAME, "");
 
                 // OH1 returns XML, later versions return JSON
                 List<OpenHABSitemap> result = mOpenHABVersion == 1
-                        ? loadSitemapsFromXml(responseBody, defaultSitemapLabel)
-                        : loadSitemapsFromJson(responseBody, defaultSitemapLabel);
+                        ? loadSitemapsFromXml(responseBody, defaultSitemapName)
+                        : loadSitemapsFromJson(responseBody, defaultSitemapName);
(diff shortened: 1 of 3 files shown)


mueller-ma:
@maniac103 May I ask for a review?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -784,10 +784,13 @@ public void onSuccess(Call call, int statusCode, Headers headers, byte[] respons
...
 
+                String defaultSitemapName =
+                        mSettings.getString(Constants.PREFERENCE_SITEMAP_NAME, "");
+
                 // OH1 returns XML, later versions return JSON
                 List<OpenHABSitemap> result = mOpenHABVersion == 1
-                        ? loadSitemapsFromXml(responseBody)
-                        : loadSitemapsFromJson(responseBody);
+                        ? loadSitemapsFromXml(responseBody, defaultSitemapName)
+                        : loadSitemapsFromJson(responseBody, defaultSitemapName);
```
maniac103: I don't like default sitemap name being passed to the load methods, as it's not required for loading. Suggestion: leave sorting out of the load methods completely and instead ...

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -784,10 +784,13 @@ public void onSuccess(Call call, int statusCode, Headers headers, byte[] respons
...
                 Log.d(TAG, "Server returned sitemaps: " + result);
                 mSitemapList.clear();
                 if (result != null) {
```
maniac103: ... sort `result` here (or `mSitemapList` below).

Separating those also makes for better unit testing.

mueller-ma:
> I'm not sure whether we should reorder the currently selected sitemap to the top at all

It's the default sitemap, not the currently open one.

maniac103:
> It's the default sitemap, not the currently open one. I think it's a good idea, since it is for the primary sitemap.

Fair enough. Fine with it then, I'd just (strongly) prefer splitting out loading and sorting of the sitemaps.

mueller-ma:
> Separating those also makes for better unit testing.

Indeed, much better for testing.

Issue #781, opened 2018-03-12
Dynamic Icons for switch-items not updated anymore


michaeljoos72:
Dynamic icons for switch-items are not updated anymore (local & remote).

I went back a few versions to see the history:

2.0.27-Beta --> OK
2.0.30-Beta --> Not OK
2.0.31-Beta --> Not OK
2.0.32-Beta --> Not OK

Tested on Android 8 as well on an older Samsung-Tablet with same result:

![image](https://user-images.githubusercontent.com/20517474/37304907-3631311a-2633-11e8-9a0e-be4f8dc26574.png)

![image](https://user-images.githubusercontent.com/20517474/37304940-4a2ea300-2633-11e8-8c88-c7889a5b73d1.png)

I see this even if I just use the app in demo-mode

openHAB: Latest 2.3-snapshot

maniac103:
Timing (thanks for the test!) suggests this regressed with #762.

I guess this is OH2? How does the respective sitemap item look like?

michaeljoos72:
Yes --> OH 2.3

Item:
Switch myStrom8_SW "Luftbefeuchter" <switch>

Sitemap:
Switch item=myStrom8_SW

maniac103:
Oh, I just saw it's also reproducible in demo mode...missed that before :-/
I'll have a look into it.

mueller-ma:
@michaeljoos72 thanks for your testing and @maniac103 thanks for fixing it.

michaeljoos72:
Working now. Thanks!

PR #818, opened 2018-04-02
Fix merge conflict introduced with #775/#781

maniac103:
PR #775 introduced code that should have been adapted to the
refactorings done in #781, but that went unnoticed.


maniac103:
@mueller-ma Don't want to merge it? ;-) It's the build fix you asked for earlier.

mueller-ma:
Yes, I only wanted to see the test passing and it failed two times.

Issue #810, opened 2018-03-31
Ignore SSL certificate and domain settings do not applies


magd1978:
2.0.43-beta seems to have a little bug: When you configure openHab beta settings, if you tell it to ignore SSL certificate and domain, the value is not applied until you completely close the application from the task manager and reopen it.

maniac103:
I know about this, and have prepared a fix already. Will submit a PR once #775 is merged.

maniac103:
This was fixed with #819.

PR #819, opened 2018-04-02
Refactor HTTP client code

### LINKED ISSUE #809: MemorizingTrustManager doesn't show up anymore
Since  #275 the user isn't asked anymore if he wants to trust _one_ certificate, if the app connects to server with an invalid certificate.
Once this is fixed, both "Ignore ... during ssl handshake" options should be removed, as they are a security risk, see #314

maniac103:
Avoid code duplication, use MemorizingTrustManager again and give sync client a sync API.

Fixes #809, #810


maniac103:
I also have https://github.com/maniac103/openhab.android/commit/27a3f2f8e5996018fd3ccc2c761259711a1d7d69 ... not sure whether to add it to this PR or to a separate one?

maniac103:
Additional question to the one above: do the removal of the SSL ignore prefs in this PR or a separate one?

mueller-ma:
It's always a good idea to keep PRs small, so please make PRs for both two commits

mobile/src/main/java/org/openhab/habdroid/util/HttpClient.java
```diff
@@ -168,6 +132,129 @@ protected Call prepareCall(String url, String method, Map<String, String> additi
...
+        public String[] getServerAliases(String keyType, Principal[] issuers) {
+            Log.d(TAG, "getServerAliases");
+            return null;
+        }
+
+        @Override
+        public PrivateKey getPrivateKey(String alias) {
+            Log.d(TAG, "getPrivateKey", new Throwable());
+            try {
+                return KeyChain.getPrivateKey(mContext, mAlias);
```
slawekjaranowski: Are you sure that this method will not be called from main thread?
https://developer.android.com/reference/android/security/KeyChain.html

mobile/src/main/java/org/openhab/habdroid/util/HttpClient.java
```diff
@@ -168,6 +132,129 @@ protected Call prepareCall(String url, String method, Map<String, String> additi
...
+        public String chooseServerAlias(String keyType, Principal[] issuers, Socket socket) {
+            Log.d(TAG, "chooseServerAlias");
+            return null;
+        }
+
+        @Override
+        public X509Certificate[] getCertificateChain(String alias) {
+            Log.d(TAG, "getCertificateChain", new Throwable());
+            try {
+                return KeyChain.getCertificateChain(mContext, alias);
```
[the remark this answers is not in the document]
maniac103: Yes. It will be called during socket communication from one of the OkHttp threads. Specifically, it's invoked from an OpenSSL callback which is fired when OpenSSL reads a client certificate request from the socket (I checked this in the sources: https://android.googlesource.com/platform/external/conscrypt/)

I wasn't able to actually check this though (callback wasn't called), probably due to my server configuration. Can you tell me what I'd need to configure on the server to enable client cert authentication - or alternatively - can you give this patch a try? The `new Throwable()` calls I accidentally left in should show the call stacktrace in logcat whenever it's called and should allow to clearly see from where it's called.

mobile/src/main/java/org/openhab/habdroid/util/HttpClient.java
maniac103: As above.

mobile/src/main/java/org/openhab/habdroid/util/HttpClient.java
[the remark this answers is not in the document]
slawekjaranowski: Unfortunately there is no tutorial how to use client certificate, you can see at https://github.com/openhab/openhab-docs/issues/656


When you finish your work, let me know I can test it.

maniac103:
@slawekjaranowski Should be ready for testing. As SmartImageView replacement and MemorizingTrustManager update will be done in a separate PR, I consider this one finished from my side (unless something pops up during further review and testing, of course).

mobile/src/main/java/org/openhab/habdroid/util/HttpClient.java
```diff
@@ -0,0 +1,252 @@
...
+    protected HttpClient(Context context, SharedPreferences prefs, String baseUrl) {
+        final Context appContext = context.getApplicationContext();
+        mClient = new OkHttpClient.Builder().build();
+        mBaseUrl = baseUrl != null ? HttpUrl.parse(baseUrl) : null;
+
+        mDefaultHostnameVerifier = mClient.hostnameVerifier();
+        mDefaultSocketFactory = mClient.sslSocketFactory();
+
+        applySslProperties(appContext, prefs);
+        prefs.registerOnSharedPreferenceChangeListener((prefsInstance, key) -> {
```
maniac103: Almost missed this one; that's the one item left to fix: With registering the listener we'd need to unregister it at some point. We could either add a `dispose()` (or similar) method to HttpClient and Connection which unregisters the listener, or we could call `applySslProperties()` from outside (e.g. ConnectionFactory) on pref changes.

@mueller-ma @FlorianSW Which variant would you consider more elegant? Both have the implication of the user of the class doing The Right Thing[TM] :-/

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
```diff
@@ -1140,6 +1066,85 @@ private void hideSnackbar() {
         }
     }
 
+    private void handleHttpFailure(Request request, int statusCode, Throwable error) {
```
mueller-ma: Please consider changing the name to `handleServerCommunicationFailure`, because it also handles non-http errors like SSL errors.

maniac103: Done.

mobile/src/main/java/org/openhab/habdroid/util/HttpClient.java
maniac103: Thinking about it it's probably not a big issue, since we're going to kill off the SSL options soon anyway. I already have prepared a patch for that; will submit once this PR + MTM update is done.

FlorianSW: The instance of HttpClient should be idempotent when it comes to it's settings (one reason, why the object shouldn't be cached in client/caller code). When the settings for preferences, used by the http client, changes, a new instance should be created and provided to the calling code. That all should be handled by the connection object (which already needs to track some settings, iirc).

A second thought: I'm not sure, if I like it, that the HttpClient takes a SharedPrefs object here. It would be more explicit if the actual values would be passed to the instance, as it would make clear for other code, what it really expects, instead of taking something, which is somewhere in the settings.

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -91,22 +88,13 @@ private void registerGcm(CloudConnection connection) throws IOException {
...
-            }
-
-            @Override
-            public void onSuccess(Call call, int statusCode, Headers headers, byte[] responseBody) {
-                Log.d(TAG, "GCM reg id success");
-                CloudMessagingHelper.sRegistrationFailureReason = null;
-            }
-        });
+        SyncHttpClient.HttpStatusResult result = connection.getSyncHttpClient().get(regUrl).asStatus();
+        if (result.error == null) {
```
FlorianSW: what do you think about a method for that, such "hasError" or even "isSuccessful"?

maniac103: Yeah, that's probably a good idea.

maniac103:
@FlorianSW can't reply to your comment above for whatever reason (?), so replying here: the shared prefs usage in the client is temporary to avoid the code duplication (and potentially inconsistent usage due to that). End goal is removing it again: https://github.com/maniac103/openhab.android/commit/7a73046d2ba54c236c72f6bf8fe097f44a2d1f8e
I can add that patch + the MTM update needed for it into this PR to get rid of the intermediate steps, but that will grow this PR quite a bit.

mueller-ma:
Toggling the demo mode checkbox needs some time:

````
04-07 15:58:50.922 14901-14901/org.openhab.habdroid.beta D/MemorizingTrustManager: getAcceptedIssuers()
04-07 15:58:51.608 14901-14901/org.openhab.habdroid.beta D/MemorizingTrustManager: getAcceptedIssuers()
04-07 15:58:52.266 14901-14901/org.openhab.habdroid.beta I/Choreographer: Skipped 79 frames!  The application may be doing too much work on its main thread.
````

mueller-ma:
If the certificate is not valid for a hostname an error is shown and the MTM does not show up. You can test it with https://wrong.host.badssl.com/

mueller-ma:
> Do you want me to do that?

You can do this in a different PR.

mueller-ma:
@maniac103 Since you didn't comment https://github.com/openhab/openhab-android/pull/819#issuecomment-379471814 I think you didn't see it, right?

mueller-ma:
> Is this behavior (log lines about lost frames when toggling demo mode, and no such lines without this PR) happening consistently for you?

Yes, it is. It also appears when I change a URL:
1. Open URL entering dialog
2. Change URL
3. Click OK
4. It needs some time until the changed URL is displayed + the "Skipping frames" message

That might be related to the cpu as I'm using a very cheap tablet.

mueller-ma:
No, I double checked.

maniac103:
I don't really have an explanation for this, given I didn't touch any code regarding prefs, and I don't expect the changes to have any relevant effect on timing either ¯\_(ツ)_/ ¯

mueller-ma:
> Not sure what to do here. We obviously want MTM to be enabled, so we'll likely have to take the performance penalty.

With the other optimizations I can live with the penalty.
I think this should be tested in beta versions before it goes in the play store and I want to release a new version soon, so I won't merge this right now. Anyway: the tests are still failing.

maniac103:
> Anyway: the tests are still failing.

I saw, but they are not failing for me when I run them locally (?)

Issue #826, opened 2018-04-07
Recycled image items are not cleared


mueller-ma:
### Actual behaviour
Old images are shown while the new image loads

### Expected behaviour
When an image is loading nothing should be shown

 
### Steps to reproduce
1. Go to demo mode multimedia page
2. Reload very often
3. The "320x240" image is shown twice

### Can you reproduce the issue in demo mode?

Yes, Multimedia

#### Client
* Android version: 5.1
* Device model: Fire Tablet 7
* Stock or customized ROM: Custom
* App version : 2.0.46-beta
* Build flavor : foss
* Device language: de

PR #830, opened 2018-04-10
Always try to load images from cache first.

### LINKED ISSUE #829: Wrong charts are shown while loading
Tested on 2.0.46-beta and seen on earlier betas as well (but not in 2.0.0 release):

When scrolling through a page with multiple charts, the app renders cached versions of the wrong chart until the new one is loaded - this causes a very irritating effect with charts being replaced by completely other charts:

img_0063

Similar to https://github.com/openhab/openhab-android/issues/826, but worse as it shuffles the charts.

maniac103:
Do this even if the useImageCache flag is set to false and use that flag
only to decide whether to do a new GET request or not.
When not doing so, images will be (temporarily) shown in wrong spots
when scrolling lists up and down.

Fixes #829 and probably also #826.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java
@@ -117,8 +117,10 @@ public abstract class OpenHABWidget implements Parcelable {
         public abstract Builder height(int height);
 
         public OpenHABWidget build() {
-            // Consider a minimal refresh rate of 100 ms
-            refresh(Math.max(refresh(), 100));
+            // Consider a minimal refresh rate of 100 ms, but 0 is special and means 'no refresh'
+            if (refresh() > 0 && refresh() < 100) {
+                refresh(100);
+            }
             // Default period to 'D'
             if (period() == null || period().isEmpty()) {
                 period("D");
@@ -258,7 +260,7 @@ public abstract class OpenHABWidget implements Parcelable {
                 .minValue((float) widgetJson.optDouble("minValue", 0))
                 .maxValue((float) widgetJson.optDouble("maxValue", 100))
                 .step((float) widgetJson.optDouble("step", 1))
-                .refresh(Math.max(widgetJson.optInt("refresh"), 100))
+                .refresh(widgetJson.optInt("refresh"))
                 .period(widgetJson.optString("period", "D"))
                 .service(widgetJson.optString("service", ""))
                 .legend(widgetJson.has("legend") ? widgetJson.getBoolean("legend") : null)
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -571,7 +571,7 @@ public class OpenHABWidgetAdapter extends RecyclerView.Adapter<OpenHABWidgetAdap
                     uri = Uri.parse(mConnection.getOpenHABUrl() + widget.url());
                 }
                 mImageView.setImageUrl(uri.toString(), mConnection.getUsername(),
-                        mConnection.getPassword(), false);
+                        mConnection.getPassword(), true);
                 mRefreshRate = widget.refresh();
             }
         }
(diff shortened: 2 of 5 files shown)


mueller-ma:
I tried some things, but I didn't see an improvement:
1. I had a sitemap with 5 charts, but scrolling didn't show the wrong charts even on master.
2. One page (A) with 2 and one page (B) with 3 charts: Go to A and wait until the charts are loaded, go to B and the charts from A appear until B's charts are loaded

maniac103:
Yes, this patch only will help for scenario 1, which actually seemed to happen for Kai. Fixing scenario 2 is something I plan to do after the HTTP client rework. I have a rewritten ImageView class already, but it depends on the HTTP rework.

kaikreuzer:
Let me know, if you want me to test any build.

maniac103:
The above commit should fix scenario 2.

mueller-ma:
@kaikreuzer You can try this apk: [mobile-foss-beta-debug.zip](https://github.com/openhab/openhab-android/files/1900168/mobile-foss-beta-debug.zip)

mueller-ma:
LGTM!

kaikreuzer:
> @kaikreuzer You can try this apk: mobile-foss-beta-debug.zip

Thanks! It is different, but not necessarily better 🙄 .
The effect now is that charts that are scrolled out of view are dropped and newly loaded once I scroll back to them - meaning that I cannot smoothly scroll through the page, especially as the page length changes since the invisible charts are also considered to be of 0 px height, i.e. the space on the page is only reserved once the chart is loaded again....

maniac103:
Damn, forgot to remove one `useCache` check ... which leads to charts not being cached :-/ Will fix it tomorrow.

mueller-ma:
[mobile-foss-beta-debug.zip](https://github.com/openhab/openhab-android/files/1903596/mobile-foss-beta-debug.zip)

kaikreuzer:
@mueller-ma Sorry, still same effect as described above :-(

maniac103:
I'll try to reproduce by duplicating the demo chart setup a few times.

mueller-ma:
@kaikreuzer What's the refresh rate of your charts?

kaikreuzer:
None!
```
		Text label="Charts" icon="chart" {
			Chart service="rrd4j" item=Leistung period=d
			Chart service="rrd4j" item=Wasser_Durchfluss period=d
			Chart service="rrd4j" item=Strom_Vortag period=W
			Chart service="rrd4j" item=Wasser_Vortag period=W
			Chart service="rrd4j" item=Chart_Temp period=d
			Chart service="rrd4j" item=Chart_WGT period=d
			Chart service="rrd4j" item=Weather_Wind period=W
			Chart service="rrd4j" item=Beregnung_Pegel period=W
		}
```

maniac103:
Found an issue with refresh parameter handling: refresh was set to 100ms if it wasn't given. While this shouldn't matter (except the excessive data + CPU usage), it might be releated here? Not sure either way, as I can't reproduce it :-( (I modified the demo setup to provide 5 different graphs, and no matter how much I scroll, the correct charts are always loaded.

mueller-ma:
Can you upload a debug apk?

maniac103:
It turns out that with more charts, the cache starts dropping images from memory. It (the WebImageCache class from loopj, that is) internally uses SoftReferences, whose documentation says it shouldn't be used for caching: https://developer.android.com/reference/java/lang/ref/SoftReference.html *rolleyes*

The perceived wrong image loading may thus simply be due to loading chart items having 0 height. The commit above fixes that by keeping them at full height in that case.

@kaikreuzer Can you please try 
[mobile-foss-beta-debug.zip](https://github.com/openhab/openhab-android/files/1924090/mobile-foss-beta-debug.zip) (rename to .apk of course)?

mueller-ma:
> The commit above fixes that by keeping them at full height in that case.

+1 for that. I think some loading indicator would look good.

maniac103:
> I think some loading indicator would look good.

Yes, I agree, I'll add that later when replacing SmartImageView.

kaikreuzer:
I doubt this theory - if you look at https://github.com/openhab/openhab-android/issues/829#issue-312550740, you will notice that charts are actually replaced, while others below stay at the exact same location as before.

> @kaikreuzer Can you please try
mobile-foss-beta-debug.zip (rename to .apk of course)?

Thanks - it actually looks quite good now. Images appear at the right place and their space is always reserved, so that scrolling works as expected (due to a fixed page height).
The only issue I still see is that when scrolling quickly, images are not immediately visible, but a blank space is scolling into the view and only a bit later the image is displayed. I think it is ok to live with that, so I'd fine to go with the current solution.

PR #834, opened 2018-04-11
Use python script to validate descriptions

mueller-ma:


mueller-ma:
We aren't allowed to say openHAB Android client: https://developer.android.com/distribute/marketing-tools/brand-guidelines.html

ThomDietrich:
> The following are guidelines for the Android brand and related assets.

You are saying we should call it "openHAB for Android"? Currently it's called "openHAB", I think that's just fine.

Another thing: Now that I loo at the Store description: I feel the "openHAB server needed" should be the first, not the last part of the description.

README.md
```diff
@@ -14,12 +14,12 @@
...
   <img alt="Logo" src="fastlane/metadata/android/en-US/images/icon.png" width="100">
   <br>
-  <b>openHAB Android Client</b>
+  <b>openHAB client for Android</b>
 </p>
 
 ## Introduction
 
-openHAB Android application is a native client for openHAB. It uses REST API of openHAB to render sitemaps of your openHAB.
+This app is a native client for openHAB. It uses REST API of openHAB to render sitemaps of your openHAB.
```
ThomDietrich: *the REST API

ThomDietrich: Also I think I said that before: A normal user shouldn't be confronted with this API thingy he doesn't know.

assets/store_descriptions/en-US/strings.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0"?>
+<strings>
+    <intro>openHAB - "empowering the smart home" - vendor and technology agnostic open source home automation</intro>
+    <whatis>openHAB is a Java-based open-source home automation platform that integrates and combines a vast range of different smart home systems and technologies into one single solution.</whatis>
+    <rules>On top of the unified abstraction layer all connected devices are available for the overarching automation rule engines and different user interfaces.</rules>
+    <supported>Supported Products</supported>
+    <bindings>Over 200 specific add-ons provide support for brands, devices, technologies and communication protocols. Examples are Z-Wave, Philips Hue, Amazon Echo, Chromecast and Sonos. Discover all available add-ons and supported devices and/or functions at: https://docs.openhab.org/addons/bindings.html</bindings>
+
+    <home_automation_solutions>• Home Automation Solutions: Z-Wave, EnOcean, Netatmo, Homematic, Insteon, ...</home_automation_solutions>
```
ThomDietrich: While at it, why not move those item markers into the script?

README.md
mueller-ma: `This app is a client for openHAB for Android rendering your sitemaps.`?

assets/store_descriptions/en-US/strings.xml
mueller-ma: Sry, I forgot to comment on that: RTL languages have the point on the other side.

README.md
maniac103: 'This app is a native client for the openHAB home automation system which allows easy [or quick] access to your sitemaps'?

assets/store_descriptions/en-US/strings.xml
ThomDietrich: Good point. I'd still say it's better to move it into the script. Your script can take care of the left-right-decision. 😃

mueller-ma: I think it's best to have them in the script. Otherwise I'd have to maintain a list in the script.

README.md
mueller-ma: In #831

mueller-ma:
@kaikreuzer Can you change the crowdin config? Please chane `.sh` to `.xml`: https://github.com/openhab/openhab-android/blob/master/crowdin.yml#L7

kaikreuzer:
Done!

PR #837, opened 2018-04-12
Improve Python script

mueller-ma:
Tested in my bot repo: https://github.com/mueller-ma-bot/openhab.android/pull/14/files

### DIFF
--- a/assets/store_descriptions/en-US/strings.xml
+++ b/assets/store_descriptions/en-US/strings.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0"?>
+<?xml version="1.0" encoding="utf-8"?>
 <strings>
     <intro>openHAB - "empowering the smart home" - vendor and technology agnostic open source home automation</intro>
     <whatis>openHAB is a Java-based open-source home automation platform that integrates and combines a vast range of different smart home systems and technologies into one single solution.</whatis>
--- a/assets/store_descriptions/en-US/strings.xml
+++ b/assets/store_descriptions/en-US/strings.xml
@@ -1,30 +1,30 @@
 <?xml version="1.0"?>
 <strings>
-    <intro>openHAB - "empowering the smart home" - vendor and technology agnostic open source home automation</intro>
-    <whatis>openHAB is a Java-based open-source home automation platform that integrates and combines a vast range of different smart home systems and technologies into one single solution.</whatis>
-    <rules>On top of the unified abstraction layer all connected devices are available for the overarching automation rule engines and different user interfaces.</rules>
-    <supported>Supported Products</supported>
-    <bindings>Over 200 specific add-ons provide support for brands, devices, technologies and communication protocols. Examples are Z-Wave, Philips Hue, Amazon Echo, Chromecast and Sonos. Discover all available add-ons and supported devices and/or functions at: https://docs.openhab.org/addons/bindings.html</bindings>
+    <string name="intro">openHAB - "empowering the smart home" - vendor and technology agnostic open source home automation</string>
+    <string name="whatis">openHAB is a Java-based open-source home automation platform that integrates and combines a vast range of different smart home systems and technologies into one single solution.</string>
+    <string name="rules">On top of the unified abstraction layer all connected devices are available for the overarching automation rule engines and different user interfaces.</string>
+    <string name="supported">Supported Products</string>
+    <string name="bindings">Over 200 specific add-ons provide support for brands, devices, technologies and communication protocols. Examples are Z-Wave, Philips Hue, Amazon Echo, Chromecast and Sonos. Discover all available add-ons and supported devices and/or functions at: https://docs.openhab.org/addons/bindings.html</string>
 
-    <home_automation_solutions>• Home Automation Solutions: Z-Wave, EnOcean, Netatmo, Homematic, Insteon, ...</home_automation_solutions>
-    <lighting>• Lighting: Philips Hue, Ikea Trådfri, LIFX, Lutron, Milight, ...</lighting>
-    <heating>• Heating: Max!, Nest, Vitotronic, Heatmiser, ...</heating>
-    <home_entertainment>• Home Entertainment: Samsung TV, LG TV, Sonos, Pioneer AVR, Squeezebox, Kodi, Plex, ...</home_entertainment>
-    <security>• Security: ZoneMinder, DSC, ...</security>
-    <open_protocols>• Open Protocols: HTTP, TCP/UDP, MQTT, Serial, ...</open_protocols>
-    <special_useCases>• Special UseCases: Minecraft, Tesla Car, Weather Services, ...</special_useCases>
-    <empty_point>• ...</empty_point>
+    <string name="home_automation_solutions">• Home Automation Solutions: Z-Wave, EnOcean, Netatmo, Homematic, Insteon, ...</string>
+    <string name="lighting">• Lighting: Philips Hue, Ikea Trådfri, LIFX, Lutron, Milight, ...</string>
+    <string name="heating">• Heating: Max!, Nest, Vitotronic, Heatmiser, ...</string>
+    <string name="home_entertainment">• Home Entertainment: Samsung TV, LG TV, Sonos, Pioneer AVR, Squeezebox, Kodi, Plex, ...</string>
+    <string name="security">• Security: ZoneMinder, DSC, ...</string>
+    <string name="open_protocols">• Open Protocols: HTTP, TCP/UDP, MQTT, Serial, ...</string>
+    <string name="special_useCases">• Special UseCases: Minecraft, Tesla Car, Weather Services, ...</string>
+    <string name="empty_point">• ...</string>
 
-    <oss_community>Open Source Community</oss_community>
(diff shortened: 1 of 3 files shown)


mueller-ma:
@gargamelonly Can you help me with this? RTL languages are too confusing for me.
https://github.com/mueller-ma-bot/openhab.android/pull/14/files#diff-32b9af6fc892cd3b6c5adbea015fc288R4
Is it normal, that the two quotes are differently escaped? One item it's `\"` and once `"\`.

mueller-ma:
Ok, the script fails on `<` signs, but this shouldnt be a dealbreaker.

assets/store_descriptions/en-US/strings.xml
```diff
@@ -1,30 +1,30 @@
...
-    <special_useCases>• Special UseCases: Minecraft, Tesla Car, Weather Services, ...</special_useCases>
-    <empty_point>• ...</empty_point>
+    <string name="home_automation_solutions">• Home Automation Solutions: Z-Wave, EnOcean, Netatmo, Homematic, Insteon, ...</string>
+    <string name="lighting">• Lighting: Philips Hue, Ikea Trådfri, LIFX, Lutron, Milight, ...</string>
+    <string name="heating">• Heating: Max!, Nest, Vitotronic, Heatmiser, ...</string>
+    <string name="home_entertainment">• Home Entertainment: Samsung TV, LG TV, Sonos, Pioneer AVR, Squeezebox, Kodi, Plex, ...</string>
+    <string name="security">• Security: ZoneMinder, DSC, ...</string>
+    <string name="open_protocols">• Open Protocols: HTTP, TCP/UDP, MQTT, Serial, ...</string>
+    <string name="special_useCases">• Special UseCases: Minecraft, Tesla Car, Weather Services, ...</string>
+    <string name="empty_point">• ...</string>
```
ThomDietrich: I'd once again suggest to break those into their components. Imho the latter parts ("ZoneMinder, DSC, ...") are non-translatable and the item marker should imho not be part of the translation. There are just a few languages that are RTL and it shouldn't be a big task to add a list of them. While at it I'd suggest to indent the item marker with an enspace character (`&ensp;`).

assets/store_descriptions/en-US/strings.xml
```diff
@@ -1,30 +1,30 @@
...
-    <translation>If you want to translate the app into your language join us at https://crowdin.com/project/openhab-android</translation>
-    <foundation>openHAB Foundation</foundation>
-    <about_foundation>The openHAB Foundation e.V. is a nonprofit organization with the mission to educate the public about the possibilities and benefits of free and open smart home solutions. Learn about the mission and the services of the foundation under https://www.openhabfoundation.org</about_foundation>
-    <important_note>Important note</important_note>
-    <oh_server>You need an openHAB server for this app</oh_server>
-    <short_description>openHAB client for Android</short_description>
-    <fdroid>The builds on F-Droid have map view support, GCM and crash reporting removed and will not be able to receive push notifications from openHAB Cloud.</fdroid>
-    <fdroid_beta>You can install the beta version alongside the stable version: [[org.openhab.habdroid]]</fdroid_beta>
-</strings>
+    <string name="report_issues">When you found an issue please report it in the forum or at https://github.com/openhab/openhab-android/issues</string>
```
ThomDietrich: "find"?

assets/store_descriptions/generate_and_validate.py
```diff
@@ -3,20 +3,27 @@
...
+import os
 
 
 en_tree = ET.parse('assets/store_descriptions/en-US/strings.xml')
 en_root = en_tree.getroot()
 
 def getString(key):
     try:
-        string = root.findall(key)[0].text
+        string = root.findall(".//string[@name='" + key + "']")[0].text
```
ThomDietrich: The use of `root` is ambiguous. I'd recommend to build a class.

assets/store_descriptions/en-US/strings.xml
mueller-ma: > are non-translatable

Are they? In non english familiar languages company and product names may be translated, e.g. the following image says "Overwatch Retribution" (Video game by blizzard) in Chinese:
http://s.newsweek.com/sites/www.newsweek.com/files/styles/embed-lg/public/2018/04/03/overwatch-chinese-venice-incident.png
Another example is the word firewall in windows. In German it's "Windows-Firewall", but in the Russian version it is literary translated.
I would put everything but the `<b>` tags in the translatable strings.

assets/store_descriptions/generate_and_validate.py
```diff
@@ -3,20 +3,27 @@
...
+    string = string.replace('\\"', '"')
+    string = string.replace("\\'", "'")
     return(string)
 
 strings_files=glob.glob('assets/store_descriptions/*/strings.xml')
 for file in strings_files:
+    print(file)
     tree = ET.parse(file)
     root = tree.getroot()
     lang=file[26:-12]
```
ThomDietrich: This can't be stable :) How about a regex?

assets/store_descriptions/en-US/strings.xml
ThomDietrich: Edge cases I'd say. The only ones I'd expect to be translatable are "Tesla Car, Weather Services".
It would be fine with me to leave the category and string as one string but I'd remove all formatting elements from the string (the item marker). Anyhow, as this project is your turf, I'll leave this decision with you.

gargamelonly:
should be the same in RTL as in non-RTL. \" always.
There is something strange in the xml "whatis" string, one \" appears at
the end and the other at the beginning. It should match the original text,
the first 3 words after openHAB should be encapsulated with \".
Also, 1 letter is missing from the first word on the right side in "rules",
"supported" and "bindings". These should be identical to the 3 words that
are encased in \" in the "whatis"

assets/store_descriptions/en-US/strings.xml
mueller-ma: Changed.

assets/store_descriptions/generate_and_validate.py
mueller-ma: I was thinking about a regex as well, but something like `assets/store_descriptions/*/strings.xml` is as stable as the substring command I use.

mueller-ma:
@gargamelonly Can you check if I submitted the string "intro" correctly here: https://crowdin.com/translate/openhab-test/118/enus-he

gargamelonly:
@mueller-ma The "intro" string is correct.

mueller-ma:
When I compare those two outputs of crowdin of the same hebrew string, I think the old version works quite better.
New: https://github.com/mueller-ma-bot/openhab.android/pull/14/files#diff-32b9af6fc892cd3b6c5adbea015fc288R4
Old: https://github.com/openhab/openhab-android/pull/835/files#diff-32b9af6fc892cd3b6c5adbea015fc288R3

gargamelonly:
I agree with your view.

mueller-ma:
I think the only improvement I should make is to catch the crash on empty strings.

assets/store_descriptions/generate_and_validate.py
mueller-ma: I'm not sure if I understand you correctly. The var `root` is overwritten each time a new language is loaded. If I create a class I have to save the instance which I will do in a variable. This instance will be lost when a new language is loaded, so I have the exact same thing as before. Right?

mueller-ma:
@gargamelonly Can you verify that the Hebrew description looks good: https://play.google.com/store/apps/details?id=org.openhab.habdroid&hl=iw

gargamelonly:
@mueller-ma The description when viewing with a browser on a PC looks good but when viewing with mobile device through the Play Store the sentence starting with an English word then continue with Hebrew words is not ordered correctly. This is a problem with RTL languages with sentences starting with an English word, the tools do not know how to handle it. When tools don't do it correctly we start with a Hebrew word first then English. Attached good and wrong screen captures.

BTW, the new 2.1 app looks good. Thank you for the hard work!

Wrong:
mobile view wrong

Good:
pc view good except amazon echo

mueller-ma:
I also noticed that the text isn't bold in the browser. This is how it's displayed in the play console:
play console

gargamelonly:
@mueller-ma That's the trick with RTL language. Parsing characters is still done from left to right just the paragraph/sentence are right aligned on the page. So to have the text in bold you need to use the same LTR signs, open with <\b> and end with <b>.

mueller-ma:
Hey @gargamelonly, I have some news on this issue: I contacted the google support and they told me that bi-directional text is not support in the Play Store. For now I removed the Hebrew description from the play store.
I have seen that many popular apps don't have a Hebrew description. One exception is [Skype](https://play.google.com/store/apps/details?id=com.skype.raider&hl=he). They have even _translated_ the app name to Hebrew. @gargamelonly, is this a common thing in Hebrew?
Whatsapp is the only app I found with a bi-directional description. Is it displayed correctly both in app and on the website?

gargamelonly:
Hi @mueller-ma ,
The app description language is the developers decision. It is true that some apps have it in Hebrew and most do not. I have seen also a mixed approach. Have both descriptions, first one partial, main features in Hebrew and then a full description in English. I opened the OpenHAB play store description (on mobile) and now it is in English but strangely right side aligned, I guess it is an error. 
The Hebrew description looked great with the only improvement needed in my previous message. 
If it is too complicated to automate to look good it is up to you as the developer to select the preferred language for the app description. It is ok.

PR #840, opened 2018-04-13
Lint yml files with yamllint

mueller-ma:
@maniac103 Do you think I should enable checkstyle as well?


maniac103:
@mueller-ma Checkstyle for our Java code? I wouldn't do this yet, but we're getting near. I have some cleanups (e.g. using lambdas, and using consistent variable identifiers) in my pipeline already, I'm just holding them back for now due to merge conflicts. Once we've done that cleanup, I'm :+1: for enabling checkstyle, but at the moment I'm not sure about it.

PR #841, opened 2018-04-15
Show ssl client howto hint again

mueller-ma:


mueller-ma:
One option and its help seems too less for a category, so it's a good idea to merge these two points.

PR #843, opened 2018-04-20
Improve chart scaling.

maniac103:
The server's density calculations are center around a default density of
96 dpi. In Android, the default is 160 dpi; thus the charts are scaled
too much (fonts and lines are disproportionally large/thick).

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -891,7 +891,7 @@ public class OpenHABWidgetAdapter extends RecyclerView.Adapter<OpenHABWidgetAdap
                 }
                 chartUrl.append("&period=").append(widget.period())
                         .append("&random=").append(mRandom.nextInt())
-                        .append("&dpi=").append(mDensity);
+                        .append("&dpi=").append(mDensity * 96 / 160);
                 if (!TextUtils.isEmpty(widget.service())) {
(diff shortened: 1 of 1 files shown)


maniac103:
Screenshots of my 6,0", 1080x1920 (360 dpi) device - before:
screenshot_20180420-094101

After:
screenshot_20180420-094149

Maybe an even better option would be a preference which selects a fixed scaling multiplier? For my taste, the labels and lines could be even thinner/smaller than with this patch ... and I think ultimately it boils down to personal preference?

maniac103:
Also tagging @kaikreuzer ... you usually have good input on UI aspects :-)

kaikreuzer:
I definitely prefer your "after" version as well - it looks like a much better fit wrt font and line sizes.

> The server's density calculations are center around a default density of
96 dpi.

Is this something that should be changed on the server side then?

maniac103:
> Is this something that should be changed on the server side then?

If anything, the dpi calculations of the chart themes need to be re-thought. The monitors I use pretty much match 96dpi (1920x1200 at 24 inch => 94 dpi), and the physical resolution of my phone is almost precisely the reported 360dpi (1920x1080 at 6 inch => 367dpi), so the perceived sizes of the fonts when looking at the charts should pretty much the same between those two devices, but they're far from being equal. The calculations look fine on first look though; I haven't found any obvious issue with them :-/

Given the above and thinking about it, the re-scaling I do in this PR probably is a hack and the root cause is elsewhere ... perhaps the scaling done on the server side shouldn't be so aggressive? How does this look on the iOS side?

PR #847, opened 2018-04-20
Implement support for Server Sent Events (SSE)

### LINKED ISSUE #504: Refresh using Server Sent Events
I would like to discuss about the move to SSE to refresh the app.

Unfortunately, SSE client is not part of the Android SDK ?
So should we use an existing library or try to implement a SSE client in our app ?
What do you thing about this library: https://github.com/tylerjroach/eventsource-android ? It is apparetnly not updated since a long time.

maniac103:
Closes #504.

Certainly needs thorough testing though, especially for corner cases like changing network connections.

@lolodomo


mobile/build.gradle
```diff
@@ -161,9 +162,26 @@ dependencyVerification {
...
+            'com.squareup.okhttp3:okhttp:a0d01017a42bba26e507fc6d448bb36e536f4b6e612f7c42de30bbdac2b7785e',
+            'com.github.loopj:android-smart-image-view:2437b0e868e8296e1578c25cf7aa0d125c72539e29e45d08ba01a41bfbd969fa',
+            'com.github.BigBadaboom:androidsvg:cef17e77b14a1b8579eab045c74bc97d9aa7ef48624b45bce80f943c2248b0a0',
+            'com.android.support:cardview-v7:7ea56ed5560b629ee1c0f24af6693e32974fbc8b91b544052cd2c14b176c85e0',
+            'com.mikepenz:aboutlibraries:8cc8aec1760f785f65bfe40bb164f963e59d3f7a1f829b2763ece15c3c883206',
+            'com.android.support:animated-vector-drawable:d5905aee3c8a4ac75e069a73b914c0a41b9b36b0e6b04126719fca22659d3cc8',
+            'com.android.support:support-fragment:a0ab3369ef40fe199160692f0463a5f63f1277ebfb64dd587c76fdb128d76b32',
+            'com.android.support:support-core-ui:82f538051599335ea881ec264407547cab52be750f16ce099cfb27754fc755ff',
+            'com.android.support:support-vector-drawable:1151b7f0ea29c9a9a8fee042a1dbe82f196632d801c438d08b279e131c767118',
+=======
```
mueller-ma: Unresolved merge conflict here

maniac103: Thanks, fixed.

mobile/build.gradle
```diff
@@ -100,6 +100,7 @@ dependencies {
     implementation 'com.android.support.constraint:constraint-layout:1.0.2'
     implementation 'org.jmdns:jmdns:3.5.3'
     implementation 'com.squareup.okhttp3:okhttp:3.9.1'
+    implementation 'com.github.heremaps:oksse:master-SNAPSHOT'
```
mueller-ma: Please replace `master-SNAPSHOT` with the git hash sum of the latest commit.

mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
```diff
@@ -0,0 +1,176 @@
...
+            @Override
+            public void onSuccess(String response, Headers headers) {
+                try {
+                    JSONObject result = new JSONObject(response);
+                    // If this succeeded, we're talking to OH2
+                    int flags = SERVER_FLAG_JSON_REST_API;
+                    client.removeHeader("Accept");
+                    try {
+                        String versionString = result.getString("version");
+                        int versionNumber = Integer.parseInt(versionString);
```
mueller-ma: Demo server and my local servers cause a nfe here.
My test server (2.?-SNAPSHOT) returns `0.10.0.201712141004`, demo server `0.10.0.b1`.

maniac103: A 2.3 snapshot is needed for it to work. See https://github.com/eclipse/smarthome/pull/5032 for the numbering change.

mueller-ma:
Icons do not change when the item value has been changed via the app, but it works when the value is changed via e.g. Basic UI.

mueller-ma:
@maniac103 I guess you have overlooked my comment above?

maniac103:
Indeed I did. 99.9% it's just the icon path that needs to be updated.

mobile/src/main/java/org/openhab/habdroid/model/OpenHABWidget.java
```diff
@@ -290,6 +290,14 @@ public static void parseJson(List<OpenHABWidget> allWidgets, OpenHABWidget paren
         }
     }
 
+    public static OpenHABWidget updateFromEvent(OpenHABWidget source, JSONObject eventPayload)
+            throws JSONException {
+        return source.toBuilder()
+                .label(eventPayload.optString("label", source.label()))
+                .item(OpenHABItem.updateFromEvent(source.item(), eventPayload.getJSONObject("item")))
```
maniac103: Need to update iconPath here.

mueller-ma:
When I change "Widget Overview" => "Temperature" via the browser the text in the app changes for a short time to black and then back to grey. Otherwise looks good.

maniac103:
I've seen it, but forgot about it between reading it and looking at the code for the next time ;-) I'll check it, even though that issue sounds slightly weird.

maniac103:
@mueller-ma Do you have/can you make a screencast of that effect? I don't seem to see this problem myself (testing with API 26 emulator and demo server).

mueller-ma:
I've test it again, but this sounds like the effect I saw. It was only for a very short time.

mueller-ma:
I'm going to merge this so we can get feedback from the beta users.

PR #849, opened 2018-04-20
Update MTM and remove SSL ignore prefs

### LINKED ISSUE #314: Ignoring the ssl certificate should give a warning
At the moment, if the user checks the checkbox to ignore all ssl certificates (better: trust all certificates), it is applied without any warning. This setting is more or less a security problem when used in the wrong way and we should warn the user, that he should use this option only, if he knows what he's doing and probably say, that he should do it only, if he wants to connect to his newly installed local Openhab (which comes with a self-signed certificate).

maniac103:
Accepting invalid certificates can be done solely via MTM now.

Closes #314


maniac103:
Indeed. In my dev branch I had the image fetch refactor before this one. I'll fix it.

mueller-ma:
No problem, I can review the image refactor before.

maniac103:
Actually, I think it's a good idea to handle this one after the image refactor. I can't think of a good solution to the compile problem without it:
- I could simply pass null as client cert, but if images are loaded directly from the OH server (think charts) this will fail
- I could query the prefs, but if the client cert is changed at runtime later, image loading won't update for that
- I could query the prefs and additionally add a shared pref update listener, but that's kinda ugly.

The 'right' solution ultimately is to pass in the connection object and use the HTTP client of that one, which incidently is what #850 does ;-)

maniac103:
@mueller-ma This one should be review ready now.

mueller-ma:
I tested it with my local oh servers and the demo server.


1. > When restoring this, the 'once' option should be killed though

from https://github.com/openhab/openhab-android/issues/809#issuecomment-377676555

2. Please adapt all ssl error string (`s/disable certificate validation/accept certificate`)

3. Discover the https port via mDNS instead of the http one?

4. Sometimes the MTM opens twice at once and clicking on "always" doesn't close the second one. When it occurs the next time I can have a look at the log.

5. I clicked on "Cancel" on the MTM and the app crashed:

````
06-05 21:28:13.987 30341-30341/org.openhab.habdroid.beta E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.openhab.habdroid.beta, PID: 30341
        at android.support.v4.app.FragmentManagerImpl.checkStateLoss(FragmentManager.java:2053)
        at android.support.v4.app.FragmentManagerImpl.ensureExecReady(FragmentManager.java:2191)
        at android.support.v4.app.FragmentManagerImpl.execSingleAction(FragmentManager.java:2211)
        at android.support.v4.app.BackStackRecord.commitNow(BackStackRecord.java:643)
        at org.openhab.habdroid.ui.activity.ContentControllerTwoPane.updateFragmentState(ContentControllerTwoPane.java:73)
        at org.openhab.habdroid.ui.activity.ContentController.indicateServerCommunicationFailure(ContentController.java:252)
        at org.openhab.habdroid.ui.OpenHABMainActivity.handleServerCommunicationFailure(OpenHABMainActivity.java:1110)
        at org.openhab.habdroid.ui.OpenHABMainActivity.access$300(OpenHABMainActivity.java:119)
        at org.openhab.habdroid.ui.OpenHABMainActivity$2.onFailure(OpenHABMainActivity.java:325)
        at org.openhab.habdroid.util.AsyncHttpClient$1.lambda$onFailure$0$AsyncHttpClient$1(AsyncHttpClient.java:127)
        at org.openhab.habdroid.util.AsyncHttpClient$1$$Lambda$0.run(Unknown Source)
        at android.os.Looper.loop(Looper.java:135)
        at android.app.ActivityThread.main(ActivityThread.java:5254)
        at java.lang.reflect.Method.invoke(Method.java:372)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
````

6. Don't accept cert, go to settings, go back and cancel again => Loading forever

File: mobile/src/main/res/values/strings.xml [hunk 1]
```diff
@@ -4,7 +4,7 @@
     <!-- See: https://github.com/openhab/openhab-android/blob/master/README.md#localization -->
 
     <!-- General app strings -->
-    <string name="openhab_service_type" translatable="false">_openhab-server._tcp.local.</string>
+    <string name="openhab_service_type" translatable="false">_openhab-server-ssl._tcp.local.</string>
```
mueller-ma: This doesn't seem to be complete. You need to change `http` to `https` where the url is built, similar to https://github.com/openhab/openhab-android/pull/496/files#diff-6f29e9c52ba079a271b8efc27cfe5468L156

maniac103: Dann, right. I'll fix it.

maniac103:
@mueller-ma Your issues should be fixed now - please give it a try. Looks like the activity lifecycle is different in L (your tablet) compared to N (my phone) and O (the emulator I'm testing with) in that L calls onSaveInstanceState() when pausing the main activity due to the MTM activity being in the foreground.

mueller-ma:
Looks good so far.
When the MTM is open and you go to the recent apps view, the openHAB header has a black background. This isn't a big issue though and I would merge it with this bug.

maniac103:
That's probably a theme problem (missing attribute). I'll check it, shouldn't be a big issue to fix.

maniac103:
Theme issue fixed.

PR #855, opened 2018-04-23
Make absolutely sure fragment updates are deferred until onStart().

### LINKED ISSUE #853: Notification received while app is running kills the app
1m
Just tested the restarted notification service and found a problem.
If the app is running in background while the message is received followed by the message clicked OK, the app gets either stopped or a selection window gets displayed asking for stopping the app or giving fedback (checking on two different android devices)

App Version 2.1.0

maniac103:
PR #789 attempted to fix this already, but was incomplete: if a
notification was clicked and the cloud connection already was resolved
at the time the activity was created (which may or may not have
happened, as activity creation in main thread and cloud connection
resolving in update thread race against each other), fragment updates
happened even though the activity wasn't started yet. Fix this by
tracking explicitly whether the activity is started, and deferring
updates until then.

Closes #853

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
@@ -169,6 +169,7 @@ public class OpenHABMainActivity extends AppCompatActivity implements
     private ContentController mController;
     private InitState mInitState = InitState.QUERY_SERVER_PROPS;
     private Call mPendingCall;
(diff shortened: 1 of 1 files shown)


mueller-ma:
I wasn't able to reproduce the bug, so I cannot really test this. If you tested it, I'm going to merge this.

maniac103:
Yes, I tested it. A reliable reproducer should be
- configure remote to myopenhab
- go to app, let it connect
- press home
- send test notification via web
- click notification

PR #858, opened 2018-04-24
Implement a preference controlling chart scaling

maniac103:
As mentioned in #843.

Screenshot of the pref:
screenshot_1524561083


mueller-ma:
Android's font and display size settings include the option "default". I think you should add this or rename 1.5 to default.

mueller-ma:
Looks good so far. Two improvements, that would be nice to have, but not a must:
1. Having a preview like the font size setting in Android (7+)
2. Reading the Android font size and set the default chart scalling accordingly

maniac103:
> Looks good so far. Two improvements, that would be nice to have, but not a must:

I'd prefer to leave out both, because ...

> Having a preview like the font size setting in Android (7+)

... doing this one will be hard (as chart rendering is done server side), unless we resort to hacks like shipping preview images in the app and

> Reading the Android font size and set the default chart scalling accordingly

... I'm not sure whether implementing more or less complicated logic here just to set a default value makes too much sense. It's purely speculation on my part, but I think the vast majority of users won't ever touch the Android font scaling. Seeing that a lot of apps aren't prepared to properly deal with changed font sizes is another data point backing that assumption ;-)

mobile/src/main/res/values/strings.xml
```diff
@@ -46,6 +46,12 @@
     <string name="settings_openhab_icon_format_value_svg" translatable="false">SVG</string>
     <string name="settings_openhab_icon_format_png">Bitmap</string>
     <string name="settings_openhab_icon_format_svg">Vector</string>
+    <string name="settings_chart_scaling">Chart scaling</string>
+    <string name="settings_chart_scaling_summary">Note: Smaller scaling shrinks fonts and line widths, so there\'s more space for the graph, but font readability may suffer.</string>
```
mueller-ma: I've seen that Android itself also uses abbreviation like `there's`, so it should be changed in openHAB to be consistently. WDYT @maniac103 @digitaldan?

PR #861, opened 2018-04-27
Fix URL handling for cases where server URL contains path segments.

maniac103:
Previously those path segments were simply stripped off, so a base URL
in the form 'http://server.domain/some/path' with the endpoint
'/rest/bindings' became 'http://server.domain/rest/bindings'.
As the newly called method interprets a trailing slash in the URL as an
empty path segment, we have to strip off those trailing slashes from the
passed-in URLs, so that the example above resolves to
'http://server.domain/some/path/rest/bindings'.

Fixes issue reported at https://community.openhab.org/t/version-2-1-0-broken-wrong-url/43999


maniac103:
One thing I forgot to test (and after some thinking I have a gut feeling it won't work) is URL query parameters ... IOW, we'll need to test whether e.g. cloud device registration still yields the correct URL.

PR #871, opened 2018-05-02
Fix voice interpreter API URL.

### LINKED ISSUE #869: VoiceCommand not working for me
Hi,

I've just tried to use VoiceCommand with the current beta 2.1.5b and Openhab Snapshot 2.3.0 and noticed VoiceCommand is not triggered anymore.

Looking at the source, the error is here:
https://github.com/openhab/openhab-android/blob/c622102c556de8b791c207e47a280f5f7d7660ad/mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java#L90

URL should be "rest/voice/interpreters" which works fine using the openHAB REST API website.

Sure I can correct this one but I'd like some feedback if this is the way to go. Thanks!

Concerned PR should be #749 which adapted the voice service to the new scheme.

Best,
Sascha

maniac103:
Closes #869

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
@@ -87,7 +87,7 @@ public class OpenHABVoiceService extends IntentService {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


mueller-ma:
@maniac103 Can you upload an apk with this fix and #872? People having these issues can check if both bugs are fixed. Thanks

PR #872, opened 2018-05-02
Fix voice service toast generation.

### LINKED ISSUE #866: Spoken text not displayed anymore when using speech input
### Actual behaviour
When using speech input via the microphone icon (both in-app and widget), the recognized text is not displayed in a toast anymore. Started to occur after the latest app update. 

### Expected behaviour
The recognized text should be shown in a toast (as before the update), as else it is impossible to know if the correct command was recognized.

 
### Steps to reproduce
* Use microphone widget or microphone icon on top right corner of app
* Say any text
* -> No feedback

#### Client
* Android version: 7.1.1
* Device model: BQ Aquaris X Pro
* Stock or customized ROM: Stock
* App version Version 2.1.0 - 20.03.2018
* Device language: German

#### Server
* Server version: openHAB 2.3.0 Build #1246
* Reverse Proxy: no
* Authentication method: None

#### openHAB Cloud
* openHAB Cloud version: 2.3.0-SNAPSHOT
* Self hosted cloud or myopenhab.org: myopenhab.org
* Authentication method User and Password

maniac103:
This service runs in a background thread now, so UI updates need to be
explicitly scheduled in the main thread.

Closes #866

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHABVoiceService.java
@@ -11,6 +11,8 @@ package org.openhab.habdroid.core;
 
 import android.app.IntentService;
 import android.content.Intent;
+import android.os.Handler;
+import android.os.Looper;
 import android.speech.RecognizerIntent;
 import android.support.annotation.Nullable;
 import android.util.Log;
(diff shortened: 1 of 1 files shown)


mueller-ma:
[mobile-full-stable-debug.zip](https://github.com/openhab/openhab-android/files/1989161/mobile-full-stable-debug.zip)

PR #878, opened 2018-05-15
 Add note about privacy policy

mueller-ma:
## In-app
A link to the privacy policy has been added to the about page:
screenshot_2018-05-15-16-56-26

## F-Droid
In the Play Store it is possible to add a link to the privacy policy as
metadata, but in F-Droid you cannot, so we have to add it to the
description.

The note about the anti features is nice to have. It explains why the
app entry contains the line "The app has features you may not like."

@kaikreuzer Can I use https for all links to openhabfoundation.org?


kaikreuzer:
> Can I use https for all links to openhabfoundation.org?

Yes!

mueller-ma:
@kaikreuzer Done. http://www.openhabfoundation.org should be configured to redirect to https by default.

Issue #403, opened 2017-10-11
Icon of beta version should contain a badge with "beta"


mueller-ma:
The icon of the beta version should contain a badge with "beta".

----------
Previous post:
Hey,
it is possible and sometimes very usefull to distribute a beta version on play store, e.g. when a user, who reported a bug, should try a possible fix and doesn't have a building enviroment.

FlorianSW:
+1 for that idea! Additionally to that I would even love to see a release in beta more often. There's probably also a way to automatically publish a new beta release from cloudbees, so that there could be a scheduled job for, say, each week to publish a new version?

mueller-ma:
https://github.com/fastlane/fastlane might be worth a try.

Also https://gitlab.com/fdroid/repomaker is an application to built apps locally and publish them on e.g. github. I think I will setup a fdroid openhab repo for nightly builts and the official full built, which cannot make it to the official fdroid repo since it contains non-foss.

mueller-ma:
@kaikreuzer @digitaldan Should I upload beta builds via the beta track or via a separate "app"? I preffer the latter, Google also does this: https://play.google.com/store/apps/details?id=com.chrome.beta

kaikreuzer:
I cannot tell, what is the better way - up to you and @digitaldan to decide...

mueller-ma:
Ok, I added it as separate app to the play store (https://play.google.com/store/apps/details?id=org.openhab.habdroid.beta), however it need to pass Google's review first.
A small todo list:

- [ ] Add note about the beta version to the description
- [ ] Add "Beta" to the app icon

maniac103:
I'm not sure the transparency will work very well. I think it's better to make a 'cutout' in the logo to make room for the text.

mueller-ma:
I prefer the version with a border:

![grafik](https://user-images.githubusercontent.com/22525368/204321345-2c4f991a-5b73-4a39-b172-33954c4b8f9b.png)

maniac103:
In this size I agree, but in normal icon size?

mueller-ma:
In icon size it looks indeed good, see #3107

PR #879, opened 2018-05-17
Add beta note to play store description

mueller-ma:
See https://github.com/openhab/openhab-android/issues/403#issuecomment-389934693

@kaikreuzer @digitaldan I created a developer page in the play store. It's basicly an image and website link added to the [default app listing](https://play.google.com/store/apps/developer?id=Twitter,+Inc.) of our apps.
For the preview I have taken an image from mysplash, so we don't have to credit the author and can use it for free: https://play.google.com/store/apps/dev?id=7475306295333689578 I think the usage of the openHAB logo for this image would be an over-use of it.


kaikreuzer:
Cool, thanks! I assume the released app version also needs to be associated to this developer page? Currently, I only see the beta app listed there.

mueller-ma:
I guess it needs some time until new released apps appear everywhere in the play store.

mueller-ma:
The stable version is "Featured" while the beta is under "All apps" now.

PR #881, opened 2018-05-18
Update short description

keverets:
The short description shows up in F-Droid as:

openHAB - openHAB client for Android

It's more useful to describe what the application is.

### DIFF
--- a/fastlane/metadata/android/en-US/short_description.txt
+++ b/fastlane/metadata/android/en-US/short_description.txt
@@ -1,1 +1,1 @@
-openHAB client for Android
+Home Automation client


mueller-ma:
Hey @keverets,
thanks for your suggestion. Indeed, "Home automation client" is more descriptive than "openHAB client for Android". For various reasons we generate all fastlane metadata from https://github.com/openhab/openhab-android/blob/master/assets/store_descriptions/en-US/strings.xml#L27, so please make the change there.

mueller-ma:
@ThomDietrich was involved in creating the app store descriptions. What do you think about this change?

keverets:
Closing in favor of #895

PR #895, opened 2018-05-21
Update short description

keverets:
As a replacement for #881, update the description in the source rather than the intermediary file.

As noted in the original PR, the short description shows up in F-Droid as:

openHAB - openHAB client for Android

It's more useful to describe what the application is.

### DIFF
--- a/assets/store_descriptions/en-US/strings.xml
+++ b/assets/store_descriptions/en-US/strings.xml
@@ -24,7 +24,7 @@
     <about_foundation>The openHAB Foundation e.V. is a nonprofit organization with the mission to educate the public about the possibilities and benefits of free and open smart home solutions. Learn about the mission and the services of the foundation under https://www.openhabfoundation.org</about_foundation>
     <important_note>Important note</important_note>
     <oh_server>You need an openHAB server for this app</oh_server>
-    <short_description>openHAB client for Android</short_description>
+    <short_description>Home Automation client</short_description>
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Imho misleading. It is not a home automation client per se. How about something like

> "Control your openHAB system [from your mobile device]. openHAB - empowering the smart home"

mueller-ma:
The short description has a limit of 80 chars, maybe go for

> Control your openHAB system - empowering the smart home

keverets:
"openHAB" should not be mentioned again in the short description, as it's already mentioned in the title right next to it. Similarly, the only place this is observed is in a market devoted to mobile devices, so that is also unneeded to mention.

Most users seeing this will not know what openHAB is, so the short description needs to capture that to give them a reason to check out the project. "Empowering the smart home" doesn't give enough context.

"a vendor and technology agnostic open source automation software for your home" is much more useful, clocks in at 78 chars, and appears to be the official byline on https://www.openhab.org/

How about that (or a small variation of it)?

ThomDietrich:
Good point. I thought it might be good to mention the fact that the app is just a client. It is mentioned in the full description though. I guess that does it.

https://alternativeto.net/software/openhab/

> vendor and technology agnostic open source home automation

mueller-ma:
I updated this PR to contain @ThomDietrich's suggestion. I'll merge this soon.

PR #905, opened 2018-05-27
Add arrow to widget when clicking opens a sub page

mueller-ma:
screenshot_2018-05-27-13-53-27


maniac103:
LGTM. Only exception is the color: I'd probably make it the same color as the text ('secondary' instead of 'primary').

PR #906, opened 2018-05-27
Update documentation links

wborn:
Update documentation links to the new website layout.

Should I also update the links in the fastlane and assets directories or should that be done via Crowdin?


mueller-ma:
> Should I also update the links in the fastlane and assets directories or should that be done via Crowdin?

The fastlane metadata is build via a CI job, so we can have multiple versions (Play/F-Droid and Stable/Beta). The source file is here: https://github.com/openhab/openhab-android/blob/master/assets/store_descriptions/en-US/strings.xml The strings in the other languages are changed via crowdin.

Issue #767, opened 2018-03-10
Sliders not displaying stateformat


kvolden:
I have Dimmer items with the label "Lights [%d %%]", and Slider elements in my sitemap linking to these items. The stateformat is rendered correctly in Basic UI, showing the percentage of the dimmer, but ignored in the Android app which only show the name "Lights" and the slider widget.

maniac103:
@mueller-ma I wonder whether we shouldn't handle labels + values in the base ViewHolder class. I noticed this kind of inconsistent handling of states during #650 already.

kvolden:
I have also noticed that when the app does show the status, it is shown aligned next to the label in square brackets, while Basic UI shows it aligned with the control without square brackets. This is not directly related (and perhaps you prefer it like this, though I personally prefer the Basic UI way), but maybe this is something to consider when you're doing whatever change is necessary for fixing this bug.

mueller-ma:
@maniac103 We also have to check if there are more elements with missing value

maniac103:
Yeah, that's why I suggested to introduce common handling ;-)

PR #916, opened 2018-06-07
Add a common base class for view holders displaying label + value.

### LINKED ISSUE #233: Formatting label text output for Android app
Hi,

I am using a the following item definitions to create a couple of switches:

Switch  LRBlindsetting  "Blinds are [MAP(blinds.map):%s]"   <rollershutter> (GF_Living) 
Switch  LRBlindmode "Auto [%s]" <light>

The web view of these switches look as desired:

webpage

However, the Android app view has square brackets around the output status (Open/Closed and Off/On):

android

Can anyone explain how I can get rid of the square brackets on the Android app display without messing up the web view?

I am using OpenHAB 1.8.3 and Android 6.0.1

Thanks,
Gibber
##

maniac103:
More precisely, use that base class for all cases where the server
provided label is likely to use the 'label [value]' syntax.
While doing so, add a separate value view for switch, slider and
rollershutter items.

Closes #233, #767


maniac103:
The following view holder types are left with (potentially) showing the value in square brackets:
- FrameViewHolder
- ColorViewHolder
- MapViewHolder
- GenericViewHolder

As the first three don't have values that can meaningfully be shown and the latter is a catch-all fallback for items unknown at the time of compiling the used app version, this seems fine to me?

mueller-ma:
> this seems fine to me?

Seems also fine to me, but I haven't looked at the code yet.

mueller-ma:
For color and mapview items the value part shouldn't be shown when it's present in the label to have a consistent behavior with the basic ui.

maniac103:
Done.

mueller-ma:
@maniac103 Can you resolve the conflict?

maniac103:
Sure. Will do that tomorrow.

maniac103:
@mueller-ma Something's up with travis ... license for android-27 not accepted (?)

mueller-ma:
It seems that the removed sdkmanager call accepted the license. According to the travis docs it should accept all licenses, but that doesn't work. Can you re-add the line in travis.yml?

maniac103:
Oops. I'll update the tests.

PR #917, opened 2018-06-07
Use upstream HoloColorPicker instead of rolling our own.

maniac103:
Screenshot:
screenshot_1528366857

Main difference is that the saturation/value bars were in the middle of the wheel before and now are below it, but doing so leaves room for a preview circle which, when clicked, resets the color to the previous one.
I wonder whether we shouldn't convert the dialog to an AlertDialog with OK/cancel buttons?


kaikreuzer:
> I wonder whether we shouldn't convert the dialog to an AlertDialog with OK/cancel buttons?

That wouldn't make sense to me - this widget should not just be a "preview" and send the value on closing, but it should continuously send commands, best even while still dragging the knobs (of course not more than every 300ms or so).

maniac103:
> That wouldn't make sense to me - this widget should not just be a "preview" and send the value on closing, but it should continuously send commands

In fact that's what it does, but there currently is no obvious way to close the dialog. Maybe having a single 'Close' button like for MapView would be sufficient already.

kaikreuzer:
Yes, having a "close" button makes sense - my comment was mainly about "cancel", which would be pretty misleading.

mobile/build.gradle
```diff
@@ -102,6 +102,8 @@ dependencies {
     implementation 'com.android.support.constraint:constraint-layout:1.1.1'
     implementation 'org.jmdns:jmdns:3.5.3'
     implementation 'com.squareup.okhttp3:okhttp:3.9.1'
+    implementation 'com.github.heremaps:oksse:c92d0556f01e769d7c06c650941107642ce98fb5'
```
mueller-ma: Seems wrong in this PR

mueller-ma:
Since it's no preview I think the color circle in the middle shouldn't show the old value. Otherwise it looks good

mobile/build.gradle
maniac103: Rebase fail, I'll fix it.

mueller-ma:
When using the Nexus 5 emulator and turning the device in landscape, the two sliders aren't shown anymore.

maniac103:
I guess we need a -land layout then. Will check it.

maniac103:
Yep, I think so.

PR #920, opened 2018-06-09
Add vibration and led light to notifications

mueller-ma:

### DIFF
--- a/mobile/src/full/java/org/openhab/habdroid/core/GcmMessageListenerService.java
+++ b/mobile/src/full/java/org/openhab/habdroid/core/GcmMessageListenerService.java
@@ -186,6 +186,23 @@ public class GcmMessageListenerService extends GcmListenerService {
     }
 
     private NotificationCompat.Builder makeNotificationBuilder(String channelId, long timestamp) {
+        long[] vibrationPattern;
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(this);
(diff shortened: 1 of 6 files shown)


mobile/src/full/java/org/openhab/habdroid/core/GcmMessageListenerService.java
```diff
@@ -186,6 +186,23 @@ private Notification makeSummaryNotification(int subNotificationCount, long time
     }
 
     private NotificationCompat.Builder makeNotificationBuilder(String channelId, long timestamp) {
+        long[] vibrationPattern;
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(this);
+        switch (prefs.getString(Constants.PREFERENCE_NOTIFICATION_VIBRATION, "")) {
+            case "short":
```
maniac103: Should probably use the defined string resources here.

mueller-ma: I used strings here, because of the following error message. The only fix I found is to convert the switch to an if.
````
:mobile:compileFullStableDebugJavaWithJavacC:\Users\user\StudioProjects\openhab.android\mobile\src\full\java\org\openhab\habdroid\core\GcmMessageListenerService.java:192: error: constant string expression required
            case getString(R.string.settings_notification_vibration_value_off):
                          ^
````

PR #923, opened 2018-06-10
Don't show "Could not discover" message when on mobile data

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
@@ -231,7 +231,9 @@ public abstract class ContentController implements PageConnectionHolderFragment.
     }
 
     /**
-     * Indicate to the user that server configuration is missing
+     * Indicate to the user that server configuration is missing.
+     *
+     * @param resolveAttempted Indicate if discovery was attempted, but not successful
      */
     public void indicateMissingConfiguration(boolean resolveAttempted) {
         Log.d(TAG, "Indicate missing configuration (resolveAttempted " + resolveAttempted + ")");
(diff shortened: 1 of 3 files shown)


maniac103:
Showing 'no network available' doesn't seem right to me in that case. I'd rather suggest giving the `indicateMissingConfiguration()` call a boolean parameter (`resolveAttempted` or something like that) which is then used for selecting between different texts. In the mobile network case the text would need to hint at the remote server needing to be configured.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -566,9 +566,16 @@ public void onClick(View view) {
...
+                                                               boolean resolveAttempted) {
             MissingConfigurationFragment f = new MissingConfigurationFragment();
-            f.setArguments(buildArgs(context.getString(R.string.configuration_missing),
+            String text;
+            if (resolveAttempted) {
+                text = context.getString(R.string.configuration_missing);
+            } else {
+                text = context.getString(R.string.no_remote_server);
+            }
+            f.setArguments(buildArgs(text,
```
maniac103: I realize it's a total nitpick, but as a suggestion:

```
@StringRes int textResId = resolveAttempted ? R.string.configuration_missing : R.string.no_remoteServer;
f.setArguments(buildArgs(context.getString(textResId), ...
```

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -233,10 +233,10 @@ public void indicateNoNetwork(CharSequence message) {
     /**
      * Indicate to the user that server configuration is missing
      */
-    public void indicateMissingConfiguration() {
-        Log.d(TAG, "Indicate missing configuration");
+    public void indicateMissingConfiguration(boolean resolveAttempted) {
```
maniac103: Oh, and should add Javadoc for parameter here.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -565,10 +565,18 @@ public void onClick(View view) {
         }
     }
 
+    /**
+     * Shows a fragment that indicates a missing configuration. Displays a hint about no remote
+     * connection when resolveAttempted is set to false.
+     * @param context
+     * @param resolveAttempted Indicate if discovery was attempted, but not successful
+     */
```
maniac103: That Javadoc should go to the indicateMissingConfiguration method (since that's what's called into by the activity) ... just add the `@param` line there.

Issue #911, opened 2018-06-01
Selection dialog closes when items update 


PartTimeDev:
### Actual behaviour
The selection dialog of a selection item closes if (another) items state gets updated. For short refresh cycle it's close to impossible to hit the desired state. 

### Expected behaviour
Don't close dialog on item update 

 
### Steps to reproduce
In the screencast you can see, the setpoint dialog stays open, while the selection dialog immediately closes.
![screencast_2018-05-30-20-00-22](https://user-images.githubusercontent.com/39832934/40845417-f02f0e26-65b6-11e8-97da-7ef7f82122c1.gif)

Single selection dialog works:
```
sitemap default label="Zuhause" {

	Frame label="Lueftung" {
		Selection item=ventilationStateOpMode label="Betriebsart" mappings=[0=Automatik, 1=Stosslueften, 2=Party, 3=Ferien,4=Aus]
	}
		}
```

This already has the closing behavior:

```
sitemap default label="Zuhause" {
        Frame label="Lueftung" {               
                Selection item=ventilationStateOpMode label="Betriebsart" mappings=[0=Automatik, 1=Stosslueften, 2=Party, 3=Ferien,4=Aus]
        }

        Frame label="Strom" {
                Text item=Production label="PV-Leistung [%d W]" }
        }

```

### Can you reproduce the issue in demo mode?

There is no refresh in demo mode... 

#### Client
* Android version: 7.1
* Device model: Sony z5c
* Stock or customized ROM: lineageos
* App version : 2.2.0
* Build flavor : play store
* Device language: 

#### Server
* Server version: 2.2.0 release on Raspi
* Reverse Proxy: 
* Authentication method :

PR #924, opened 2018-06-11
Optimize widget list update handling.

### LINKED ISSUE #586: Refreshing only updated widgets, not all the fragments
I open this issue to discuss how we could improve the refresh of data in the app.
Currently when we receive a new information from the server, `notifyDataSetChanged` is called which leads to the refresh of all the fragments. The best would be to trigger the refresh of only updated widgets (like it is done in Basic UI for example). We are able to easily identify them when using SSE.
@9a4gl commented with a possible solution: https://github.com/openhab/openhab.android/issues/217#issuecomment-354495701

I link to my issue #555 but the new goal would be a more ambitious fix ;)

maniac103:
Only call notifyDataSetChanged() if needed, that is, if the newly sent
list of widgets is structurally different from the old list of widgets.

Closes #586

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -126,9 +126,30 @@ public class OpenHABWidgetAdapter extends RecyclerView.Adapter<OpenHABWidgetAdap
     }
 
     public void update(List<OpenHABWidget> widgets) {
-        mItems.clear();
-        mItems.addAll(widgets);
-        notifyDataSetChanged();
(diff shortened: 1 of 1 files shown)


maniac103:
BTW, rationale for this is avoiding rebinding views with open popup windows (that will close because of that) if other unrelated widgets are updated.

mueller-ma:
So it also fixes #911?

maniac103:
> So it also fixes #911?

I hope so, even though I didn't test this specifically (wasn't sure about the easiest method to generate continuous item updates in the server)

mueller-ma:
I couldn't reproduce #911 anymore with this PR.

mueller-ma:
Travis is always failing because of a not found jar file. Can take some retries until it works.

PR #929, opened 2018-06-18
Fix image item refresh.

### LINKED ISSUE #928: Image refresh does not work anymore (cached forever)
### Actual behaviour
After upgrade from 2.2.0 (release) to 2.2.5-beta (and in the meantime to 2.2.7-beta), images (Image type in sitemap) are no longer refreshed but seem to be cached forever.

### Expected behaviour
I use the refresh= parameter in the sitemap Image set to 60000 (60.000 ms=60 s) and now expect, that the images are updated once per minute, at least if I refresh the page by swiping down.  But nothing happens, the images are still cached.

Only way to refresh the images is by invalidating the cache via menu entry clear image cache ("Bilder-Cache leeren" in German).
 
### Steps to reproduce
1. Add something like this to your sitemap:
`Image url="http://www.uedorf.de/webcam/images/latest.jpg" refresh=60000
` 
  This is a webcam that refreshes its content every minute.
2. Open the Android app with this sitemap entry the first time. You see a little image icon with an hour glass, which is replaced by the webcam image after a second.
3. Wait some minutes.
4. Open the Android app with the sitemap entry again.  This time the webcam image is shown without delay or the little hour glass icon. But as you can see in the timestamp of the webcam, there was no refresh of the image.

#### Client
* Android version: 7.1.1
* Device model: Moto Z Play
* Stock or customized ROM: stock
* App version : 2.2.7-beta (2.2.0 release does **not** have this issue!)
* Build flavor : Full
* Device language: German

#### Server
* Server version: 2.2.0
* Reverse Proxy: Apache
* Authentication method : user/password

Greetings
Roland

maniac103:
We must only use the cache for the initial fetch (if the forceLoad
parameter wasn't passed); refreshes always should avoid using the cache.

Fixes #928


mueller-ma:
This doesn't seem to work for me at all.

maniac103:
Manual refresh as in pull-to-refresh? Did the automatic refresh also not work for you?

mueller-ma:
> Manual refresh as in pull-to-refresh?

Yes

> Did the automatic refresh also not work for you?

Nothing at all. I double checked everything (correct url for the image item, correct commit)

maniac103:
Now this should do it.

mueller-ma:
Image items with `refresh=` work now, but pull-to-refresh doesn't refresh items w/o `refresh=`.

maniac103:
Right, but I have no idea how I should/could implement that in a proper way, as pull-to-refresh only reloads the sitemap contents. From the view/adapter perspective, there isn't any difference between that kind of reload and a normal data change.
Only thing I can think of is the sledge hammer approach of clearing the image cache whenever pull-to-refresh is done, but that cache clear then obviously also applies to other pages.

maniac103:
@mueller-ma Any further comments/thoughts on this?

mueller-ma:
I think clearing all caches of image items is ok. But this will also clear the icon cache, right?

maniac103:
Yes. We can selectively clear the in-memory bitmap cache, but there's no way to clear a single URL from the HTTP cache AFAICT, so it's an all-or-nothing thing.

PR #930, opened 2018-06-18
Support using item options for selection widget

### LINKED ISSUE #912: Consider item options for selection widget when no mapping is defined
As discussed in https://github.com/eclipse/smarthome/issues/4942 it would be cool to implement this feature in the Android app too.
I just implemented it in Basic UI and Classic UI; it was already present in Paper UI.

maniac103:
Fixes #912


maniac103:
Please let me know whether you prefer to have the two latter (cleanup) commits in a separate PR. I stumbled upon this code during implementation of this feature, and found it handy to include it here as they'll conflict with the new code.

mobile/src/main/java/org/openhab/habdroid/ui/widget/ExtendedSpinner.java
```diff
@@ -0,0 +1,42 @@
...
+    public ExtendedSpinner(Context context, AttributeSet attrs) {
+        super(context, attrs);
+    }
+
+    public ExtendedSpinner(Context context, AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+    }
+
+    public void setOnSelectionUpdatedListener(OnSelectionUpdatedListener l) {
+        mOnSelectionUpdatedListener = l;
```
mueller-ma: This really looks like `mOnSelectionUpdatedListener` is an int and set to `one`.

maniac103: I can rename it to `listener`.

mueller-ma: Go for it :+1:

PR #933, opened 2018-06-22
Fix fastlane description

mueller-ma:


mueller-ma:
@ThomDietrich This is really ugly, but I couldn't find a better way. Ideas?

maniac103:
Isn't CDATA meant for situations like this?
https://en.m.wikipedia.org/wiki/CDATA

mueller-ma:
@maniac103 Thanks for the hint, it's a much better solution.

PR #944, opened 2018-07-06
Several preference screen improvements

### LINKED ISSUE #934: Hide preferences when they are unused in a flavor
The full screen setting is (or should be, it doesn't work) hidden when running a device with Android 4.4 or lower: https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java#L230
This should be done for the ringtone settings as well.

mueller-ma:
* Disable preferences if unsupported:
  * Android version too low
  * FOSS flavor
  * openHAB 1
* Show summary for connection subscreens indicating whether the connection is unsecure
* Show summary for notification ring tone
* Rename removeProtocollFromUrl to getHostFromUrl

Fixes #934

TODO:

- [x] update connection summary when connection changes.


mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -224,17 +243,86 @@ public boolean onPreferenceClick(Preference preference) {
...
+            });
+
+            final PreferenceScreen ps = getPreferenceScreen();
+
+            // Fullscreen is not supoorted in Android < 4.4
+            if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.KITKAT) {
+                Preference fullscreenPreference = ps.findPreference(Constants.PREFERENCE_FULLSCREEN);
+                fullscreenPreference.setEnabled(false);
+                fullscreenPreference.setSummary(R.string.settings_disabled_kitkat);
+            }
```
maniac103: I'd keep this removed. There's nothing the user can do anyway.

mueller-ma: Removing settings didn't work for me at all and I think it's better to give the user feedback why it's not supported.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -224,17 +243,86 @@ public boolean onPreferenceClick(Preference preference) {
...
+
+            if (mOpenhabVersion == 1) {
+                Preference iconFormatPreference =
+                        ps.findPreference(Constants.PREFERENCE_ICON_FORMAT);
+                iconFormatPreference.setEnabled(false);
+                iconFormatPreference.setSummary(R.string.settings_disabled_openhab_1);
+                Preference chartScalingPreference =
+                        ps.findPreference(Constants.PREFERENCE_CHART_SCALING);
+                chartScalingPreference.setEnabled(false);
+                chartScalingPreference.setSummary(R.string.settings_disabled_openhab_1);
```
maniac103: Same applies here, just hide those.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -224,17 +243,86 @@ public boolean onPreferenceClick(Preference preference) {
...
+            final PreferenceScreen ps = getPreferenceScreen();
+
+            // Fullscreen is not supoorted in Android < 4.4
+            if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.KITKAT) {
+                Preference fullscreenPreference = ps.findPreference(Constants.PREFERENCE_FULLSCREEN);
+                fullscreenPreference.setEnabled(false);
+                fullscreenPreference.setSummary(R.string.settings_disabled_kitkat);
+            }
+
+            removeUnsupportedPrefs(ps);
```
maniac103: Why not move the remove code here and base it on CloudMessagingHelper.isSupported()?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -224,17 +243,86 @@ public boolean onPreferenceClick(Preference preference) {
...
+
+        public void updateConnectionSummary(boolean isRemote) {
+            String subscreenPrefValue = isRemote ? Constants.SUBSCREEN_REMOTE_CONNECTION
+                    : Constants.SUBSCREEN_LOCAL_CONNECTION;
+            String urlPrefValue = isRemote ? Constants.PREFERENCE_REMOTE_URL
+                    : Constants.PREFERENCE_LOCAL_URL;
+            String userPrefValue = isRemote ? Constants.PREFERENCE_REMOTE_USERNAME
+                    : Constants.PREFERENCE_LOCAL_USERNAME;
+            String passwordPrefValue = isRemote ? Constants.PREFERENCE_REMOTE_PASSWORD
+                    : Constants.PREFERENCE_LOCAL_PASSWORD;
```
maniac103: Could pass the 4 key names as parameters here ... that way it would be a bit more clear at the caller side what's going on.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -224,17 +243,86 @@ public boolean onPreferenceClick(Preference preference) {
...
+            Preference pref = findPreference(subscreenPrefValue);
+            String url = pref.getSharedPreferences().getString(urlPrefValue, "");
+            String user = pref.getSharedPreferences().getString(userPrefValue, "");
+            String password = pref.getSharedPreferences().getString(passwordPrefValue, "");
+            String summary;
+            if (TextUtils.isEmpty(url)) {
+                summary = getString(R.string.info_not_set);
+            } else {
+                if (url.startsWith("https://") && ! TextUtils.isEmpty(user)
+                        && ! TextUtils.isEmpty(password)) {
```
maniac103: I'd suggest differentiating between 'secure' (http vs. https) and 'authenticated'. E.g. `Connected securely to xyz.com, using authentication`

mobile/src/main/res/values/strings.xml
```diff
@@ -61,6 +59,10 @@
     <string name="settings_notification_vibration_value_short" translatable="false">short</string>
     <string name="settings_notification_vibration_value_long" translatable="false">long</string>
     <string name="settings_notification_vibration_value_twice" translatable="false">twice</string>
+    <string name="settings_disabled_kitkat">Fullscreen is only available in Android 4.4 and above</string>
+    <string name="settings_disabled_openhab_1">Preference isn\'t available for openHAB 1</string>
+    <string name="settings_connection_summary">Connected to %s</string>
+    <string name="settings_insecure_connection_summary">Insecurly connected to %s</string>
```
maniac103: insecurely

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
mueller-ma: Initially I wanted to keep the "Not supported in foss" string in the foss strings file.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
mueller-ma: Done

mobile/src/main/res/values/strings.xml

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
mueller-ma: IMO a non-authenticated https connection is pretty useless. An attacker can see the certificate cn during the handshake and visit the page. I guess it's not clear for every user why the "secure" connection isn't secure.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
maniac103: > I think it's better to give the user feedback why it's not supported.

And what is the user supposed to do with that information? I'm all for informing the user, but if he can't do anything about it, it IMHO just adds clutter.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
maniac103: Fair enough, but like mentioned in the other comment, I'd rather remove the preference completely. IMHO it's sufficient to inform the user about the existence of the full build in one place.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -224,17 +250,111 @@ public boolean onPreferenceClick(Preference preference) {
...
+                    return root;
+                if (PreferenceGroup.class.isInstance(p))
+                {
+                    PreferenceGroup parent = getParent((PreferenceGroup)p, preference);
+                    if (parent != null)
+                        return parent;
+                }
+            }
+            return null;
+        }
```
maniac103: Nit: should fix curly braces in those two methods (put starting ones on old line, add missing ones)

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -224,17 +250,111 @@ public boolean onPreferenceClick(Preference preference) {
...
+         * @author https://stackoverflow.com/a/17633389
+         */
+        private PreferenceGroup getParent(PreferenceGroup root, Preference preference)
+        {
+            for (int i = 0; i < root.getPreferenceCount(); i++)
+            {
+                Preference p = root.getPreference(i);
+                if (p == preference)
+                    return root;
+                if (PreferenceGroup.class.isInstance(p))
```
maniac103: `p instanceof PreferenceGroup`?

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
mueller-ma: Done

mueller-ma:
The connection summary isn't updated when the connection setting changes. Any ideas how to implement that? Setting `setOnPreferenceChangeListener` in `updateAndInitPreferences()` for the url, user and pw fields doesn't work as they are on a subpage.

maniac103:
Update them in 'onResume()'?

PR #952, opened 2018-07-13
Add icons to settings menu

### LINKED ISSUE #731: Add icons to settings menu
Like in the android settings: https://cdn57.androidauthority.net/wp-content/uploads/2016/03/settings-revamp.jpg

mueller-ma:
* Add icons to settings menu 
* rename icons to include color and dp and make the grey icons a bit darker
* Dynamically change ring tone and vibration icon 
* Color url, user and pw icon 

Fixes #731 
Fixes #944 

@maniac103 Any ideas how to add `set[Url,User,Password]IconColor()` to the `initEditorPreference()` in a smart way?


maniac103:
@mueller-ma Define an interface mapping a pref key (or a EditTextPreference object) to a color, pass that interface to the respective init method and then apply the color using the DrawableCompat tint methods instead of using separate drawables. Can also register the pref change listener inside the init method that way.

mueller-ma:
No url set:

No user and weak password:

User and secure password:


Any suggestion for a better demo mode icon?

/cc @kaikreuzer

maniac103:
@mueller-ma I've simplified the code a bit, maybe you want to pick? https://github.com/maniac103/openhab.android/commits/731-pref-icons
Otherwise mostly LGTM, I'm just wondering whether we should omit the green tint for user/password in favor of keeping them grey.

mueller-ma:
> I'm just wondering whether we should omit the green tint for user/password in favor of keeping them grey.

I have seen that some icons in Android P DP 3 (or 4, I lost the link) are tinted green when security features are enabled, e.g. Play Protect.

mueller-ma:
> maybe you want to pick?

Yes, I did. It seems that you have swaped the logic for user and password and the initial (when the settings are opened) tinting doesn't work.

mobile/src/main/java/org/openhab/habdroid/ui/OpenHABPreferencesActivity.java
```diff
@@ -323,70 +387,140 @@ private void onNoDefaultSitemap(Preference pref) {
...
+                    }
+                }
+                return null;
+            });
+        }
+
+        private void updateIconColor(Preference pref, IconColorGenerator colorGenerator) {
+            Drawable icon = pref.getIcon();
+            if (icon == null) {
+                return;
```
mueller-ma: The icon will be null here when the subpage will be opend, at least on api 22. Instead of getting the icon via `pref.getIcon()` there could be a mapping of `Preference` to icon name.

maniac103: I'll look at the Preference.getIcon implementation again.

maniac103: OK, this behavior indeed was changed in Android 7.1: https://github.com/LineageOS/android_frameworks_base/commit/f049e24c644912a308ca283ba1c75d678a09f4a8
I'll hardcode the resource IDs (make them a parameter of initEditorPreference).

maniac103: Fix: https://github.com/maniac103/openhab.android/commit/dddfc22070cb11873b5fa4181a20860a950417ce ... tested down to API 16.

PR #982, opened 2018-08-19
Clear webviews before being re-used

### LINKED ISSUE #948: Problems accessing Webview of own domain
### Actual behaviour
When trying to view Webview items with the beta app, I no longer can see Webview elements on my own server using a special port of the format:
`Webview url="https://my.own.domain:64443/grafana/d-solo/000000001/openhab-data?orgId=1&panelId=16&from=now-24h&to=now" height=15`
I am not sure if the reason is the own domain or the special port. The same sitemap loads fine in the basicui website (both accessed locally and remotely) and the stable Android app (both accessed locally and remotely). In the beta app I sometimes get the content of another Webview item previously looked at from the same sitemap, as if that is still in the cache. Other Webview elements (not on my own domain and without a special port) also work fine, even in the beta app.

### Expected behaviour
The web content should be loaded in the app.

 
### Steps to reproduce
1. Access the part of the sitemap that has the Webview
2. Wait (nothing happens)

n/a

#### Client
* Android version: 7.1.2
* Device model: Samsung Galaxy S5
* Stock or customized ROM: LineageOS 14.1
* App version: 2.2.13-beta
* Build flavor: Full and Foss
* Device language: English

#### Server
* Server version: 2.3 stable
* Reverse Proxy: yes, nginx
* Authentication method: both user and password and SSL Client certificate

#### openHAB Cloud
* openHAB Cloud version: n/a
* Self hosted cloud or myopenhab.org: n/a
* Authentication method: n/a

#### App log
```
07-11 20:43:52.831 19082 19082 D AnchorWebViewClient: Did not find anchor from url https://my.own.domain:64443/grafana/d-solo/000000001/openhab-data?orgId=1&panelId=16&from=now-24h&to=now
```

#### openHAB Server log
```
Insert your log here
```

#### Reverse Proxy log
```
Insert your reverse proxy error log here
```
```
Insert your reverse proxy access log here
```

mueller-ma:
Webviews are recycled and need to be cleared before being re-used.
Otherwise the old webview content is shown until the new one is loaded.

Closes #948

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABWidgetAdapter.java
@@ -970,6 +970,7 @@ public class OpenHABWidgetAdapter extends RecyclerView.Adapter<OpenHABWidgetAdap
         @SuppressLint("SetJavaScriptEnabled")
         @Override
         public void bind(OpenHABWidget widget) {
+            mWebView.loadUrl("about:blank");
             ViewGroup.LayoutParams lp = mWebView.getLayoutParams();
             int desiredHeightPixels = widget.height() > 0
                     ? widget.height() * mRowHeightPixels : ViewGroup.LayoutParams.WRAP_CONTENT;


maniac103:
LGTM. One could move clearing the view into an implementation of the view holder's `stop()` method, but it's probably not much of a difference here.

PR #984, opened 2018-08-24
Revert "Remove unneeded try catch in voice command funtionality (#899)"

mueller-ma:
In some case the try catch is needed. See
https://console.firebase.google.com/project/openhab-mobile/crashlytics/app/android:org.openhab.habdroid/issues/5b755c156007d59fcdf12bc8
(only visible to the openHAB staff).
I updated the code to open the Google app in any app store. Some users
might use YALP store or any other custom Play Store (or Amazon app
store). For users without any additional app store the Play Store is opened.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/OpenHABMainActivity.java
@@ -60,6 +60,7 @@ import android.view.View;
 import android.view.ViewStub;
 import android.view.WindowManager;
 import android.widget.ProgressBar;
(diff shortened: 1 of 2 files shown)


maniac103:
I'm not sure whether pointing users towards the Google app will help much. We guard showing the menu item by checking `SpeechRecognizer.isRecognitionAvailable()`, so if that one returns true and launching the activity fails this suggests some manufacturer (here: Huawei) fuckery.

mueller-ma:
> Probably need to catch Play Store missing and show a toast in that case though

Right, the user can't install the app from the website anyway.

PR #987, opened 2018-09-03
Fix image refresh regressions

### LINKED ISSUE #985: Image refresh broken
### Actual behaviour
Image items without the refresh parameter doen't refresh, when swiped-to-refresh.

### Expected behaviour
They should refresh.

#### Client
* Android version: Android 7.1.2
* Device model: Moto G4
* Stock or customized ROM: LineageOS 14
* App version : 2.2.19
* Build flavor : Foss
* Device language: de

@maniac103 I experience this issue on multiple devices, but have no time to debug it right now.

maniac103:
- Correctly refresh images after pull-to-refresh
- Correctly resume image refresh when reusing views

Fixes #985

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.java
@@ -197,6 +197,7 @@ public class WidgetImageView extends AppCompatImageView {
     protected void onDetachedFromWindow() {
         super.onDetachedFromWindow();
         cancelCurrentLoad();
+        mLastRequest = null;
     }
 
     public void setRefreshRate(int msec) {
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.java
```diff
@@ -308,12 +304,15 @@ public void execute(boolean avoidCache) {
...
             }
         }
 
+        public boolean hasCompleted() {
+            return mCall == null;
+        }
+
         public boolean isActiveForUrl(HttpUrl url) {
-            return mCall != null && mCall.request().url().equals(url);
+            return mCall != null && mCall.request().url().equals(url) && !mCall.isCanceled();
```
mueller-ma: `return !hasCompleted() ...`?

maniac103: I went a bit back and forth about this (started splitting the call to `isActiveForUrl()` into `!hasCompleted() && matchesUrl(url)`), but ultimately figured keeping it the way it is is the most readable/simple solution. If you have strong feelings about it though, I can change it.

PR #1020, opened 2018-09-26
Update dependencies to Api 28

mueller-ma:


mueller-ma:
@maniac103 I've reordered the imports automatically with Android Studio and now Checkstyle complains about a ton of import order issue. Did you sort them manually in #1029?

maniac103:
Yes, I did. The checkstyle complaints are completely valid, so I guess we'll need a fixed Android Studio formatting setup.

Thinking about it, we probably should add the androidx.* imports to the special imports group in the checkstyle config, as they belong to the android.* imports the same way javax belongs to java imports. The way I set up the import rules follows AOSP coding style (android.* imports first, java/javax last, everything else in between).

maniac103:
Oh, and completely ignored checkstyle issues in MTM code, as I wanted to keep it as close as possible to upstream. Not sure whether we can exclude files from Stickler?

mueller-ma:
I haven't found a way to adjust the default Android Studio import sort. I would rather disable the rule in checkstyle than resign the automatic sort.
I think we can just ignore the checkstyle errors in MTM code.

maniac103:
I'll check AS as well. I can't imagine there's no way to make it sort imports lexicographically.

mueller-ma:
It's configurable in Settings => Editor => Code Style => Java => Imports.
grafik
and a black line at the end.

PR #1021, opened 2018-09-26
Make beautifyUrl() null safe

mueller-ma:
Fixes https://console.firebase.google.com/project/openhab-mobile/crashlytics/app/android:org.openhab.habdroid/issues/5bab56f2f8b88c29635b445a


mueller-ma:
@maniac103 Any idea how this crash happened? When `url` is null it is catched in line 415. Or did I miss something?

maniac103:
AFAICT it can happen for invalid URLs without host ... `getHostFromUrl()` will return null in that case.

PR #1028, opened 2018-10-01
Run Checkstyle with Stickler CI

### LINKED ISSUE #675: Code formatting
> That may be true, but it's hard to follow code style if large parts of the existing code base don't follow it.
I can update my changes to fully conform, but I wonder whether it wouldn't be a better idea to turn off Stickler for now, merge the large, invasive PRs and afterwards do a cleanup of the code base first: fix whitespace, name variables consistently, remove unused code etc. ... then enable Stickler again afterwards. I volunteer to help with the cleanup, we just need to coordinate when to do it as PR rebase will become painful afterwards.

From https://github.com/openhab/openhab-android/pull/669#issuecomment-360900728

We should wait for

- [ ] #556 
- [x] #562 
- [x] #638 
- [x] #640 
- [x] #650 
- [x] #688

mueller-ma:
Closes #675

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -1,3 +1,5 @@
 linters:
+    checkstyle:
+        config: ./stickler/checkstyle.xml
     yamllint:
         config: ./stickler/yamllint.yml


maniac103:
This probably was a bit early? We still have about 1600 checkstyle errors, some of which are legit (line lengths), some of which are checkstyle config issues (import order, variable naming).

PR #1049, opened 2018-10-16
Move app string validation to separate file

mueller-ma:


mueller-ma:
@maniac103 Do you think it's better to show both validations (app store and app strings) as two checks or as one?

maniac103:
@mueller-ma I don't really have a strong opinion either way, so just do it the way it fits you best.

PR #1052, opened 2018-10-16
Add up/down buttons to setpoint widgets

mueller-ma:
Edit: This adds a setting that allows switching between the old (with up/down
buttons) and new (with dialog) setpoint widgets.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
@@ -84,7 +84,7 @@ public class WidgetAdapter extends RecyclerView.Adapter<WidgetAdapter.ViewHolder
     private static final String TAG = WidgetAdapter.class.getSimpleName();
 
     public interface ItemClickListener {
-        void onItemClicked(Widget item);
+        boolean onItemClicked(Widget item); // returns whether click was handled
         void onItemLongClicked(Widget item);
     }
 
@@ -335,8 +335,8 @@ public class WidgetAdapter extends RecyclerView.Adapter<WidgetAdapter.ViewHolder
         ViewHolder holder = (ViewHolder) view.getTag();
         int position = holder.getAdapterPosition();
         if (position != RecyclerView.NO_POSITION) {
-            if (!holder.handleRowClick()) {
-                mItemClickListener.onItemClicked(mItems.get(position));
+            if (!mItemClickListener.onItemClicked(mItems.get(position))) {
+                holder.handleRowClick();
             }
         }
     }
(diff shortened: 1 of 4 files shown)


mueller-ma:
I want to keep this PR small, so only the setpoint item will be changed. Please note, that the current commit doesn't support switching between the different widget types.
My first approach was to set the visibility of the buttons, but it turned out to be pretty ugly, because I had to change the padding in the code. I created additional layouts. The only way to select the layout I see is to get the correct layout in onCreateViewHolder() and pass it via SetpointViewHolder(), but then I don't know how the change the layout w/o restarting the app.

maniac103:
You can't change the layout of a view holder at runtime. Best idea is to create two different view holder types (as in number constants); you can use the same class for both.

mueller-ma:
> (as in number constants)

I'm not sure where exactly you reffer to.

maniac103:
I'm referring to the numbers returned by `getItemViewType()`

mueller-ma:
@maniac103 Do you prefer having three xml layout files or one and hide the unneeded components in code?

maniac103:
Didn't you say you need different layout files because of different paddings? If it's just visibility I'd prefer one layout file, I think, as we then don't need multiple item view types.

mueller-ma:
Padding could be done in code as well, but that would be quite hacky.
Thinking about the preference to change the type: I'm not sure if a per user setting is really useful. For some items the dialog might be better, for some other ones the plus/minus buttons. What about dropping the preference entirely and having always dialog and buttons?

maniac103:
> What about dropping the preference entirely and having always dialog and buttons?

Sounds fine to me. Move dropdown button to the left of the up/down buttons?

mueller-ma:
I assigned an ID to the linear layout and assigned an onClickListener to this ID. However if I click on light blue areas, the default click animation is shown, but the dialog isn't opened.
Screenshot_2019-03-31-13-07-14

maniac103:
@mueller-ma Please see my 4 commits on top of your additions at https://github.com/maniac103/openhab.android/commits/550-selection

mueller-ma:
I had the same idea to implement an onRowClicked() function to fix long-clicking on setpoints (for nfc) not working anymore. I'll review soon.

mueller-ma:
Using the demo package => "Widget Overview" => Temperature setpoint: If I increment it to `28.0°C` and then click on the down button, nothing happens. This is, because `state` in line 840 is 28 as int, not float. Side effect of  #1269?

maniac103:
Sounds like it ... did you check the state in rest API response and/or SSE event?

maniac103:
I've added another commit for the click handling, please pull: https://github.com/maniac103/openhab.android/commit/733cf7d09c61edcca6def47640a590c2e370fe5e ... it gives page opening precedence over dialog opening in the (admittedly unlikely) case a page link exists for the setpoint widget ... I'm not sure it's actually possible to achieve that, but better be safe than sorry ;-)

As for the float vs. int thing, it seems to work just fine for me using the current code in the branch. I'd suspect https://github.com/openhab/openhab-android/pull/1052/commits/a1e4ce15f0240826b7ff4dfe501b43b15d28badd has fixed it.

mueller-ma:
> I'd suspect a1e4ce1 has fixed it.

Strange, it's fixed for me now. Maybe I just had an old branch.

PR #1057, opened 2018-10-21
Decrease font size in selection widget dialog

mueller-ma:
IMO the font was always too large in the dialog.
Old:
screenshot_2018-10-21-18-39-20
New:
screenshot_2018-10-21-18-46-16


maniac103:
+1, but I'd love to see this taken this as a chance to modernize the dialog, in particular the radio buttons (either replacing them by a checkmark for the selected item, or by moving them to the left like standard selection dialogs in AOSP behave).

mueller-ma:
I moved the radio buttons to the left. Looks way better now.

Issue #1018, opened 2018-09-25
Camera image is not refreshing after phone goes to sleep


dakipro:
Hi,
After last update camera images (timed images?) are not refreshing any more.
F.eks i have image with 2s refresh and when i open the app it works fine. Then phone goes to sleep and after unlock image stops refreshing.
Using it mostly to check if toddler is sleeping.
Phone is samsung note 8, Android 8.0

mueller-ma:
I guess you have version 2.3.0?

dakipro:
yes, last stable app from playstore 2.3.0, latest stable OpenHab 2.3.0 with all latest stable plugins (although camera image is only defined as plain image in sitemap with 2s refresh).
Previous app version worked fine with openhab 2.3 some time, after upgrade this started happening.

Thanks for looking into it!

mueller-ma:
I'll look into this. For now the distribution of 2.3.0 is paused, so the number of users with the broken version doesn't increase.

maniac103:
OTOH I'd say we need to switch last refresh timestamp in WidgetImageView from `SystemClock.uptimeMillis()` to `SystemClock.elapsedRealtime()`, as the former, which is what we're currently using, doesn't count time the device spent in deep sleep. Can't test myself at the moment though.

mueller-ma:
I couldn't reproduce this bug. @dakipro can you try to pull down the view to refresh it manually? Does this re-activate auto refresh?

dakipro:
Yes it does. After waking and unlocking the phone (also if i switch to another app and come back) the camera image is frozen. Pull down for manual refresh starts it working again

mueller-ma:
I've made the change that @maniac103 suggested. @dakipro can you try the apk: [mobile-full-stable-debug.zip](https://github.com/openhab/openhab-android/files/2473780/mobile-full-stable-debug.zip)?

dakipro:
I could not update existing app so I uninstalled previous and installed this version you attached, but now it is sort of not working at all. When I load dashboard, image refreshes once (i see the time on camera change one second) but then it freezes and doesn't refresh anymore at all. On manual refresh it is the same, I get a new image and one refresh after that, and then it stops again.

mueller-ma:
I see, my patch isn't complete. I'll look into this tomorrow.

mueller-ma:
@dakipro does the apk work for you?

dakipro:
[...]

PR #1059, opened 2018-10-23
 Use elapsedRealtime() instead of uptimeMillis() to time image refreshs 

mueller-ma:
See https://github.com/openhab/openhab-android/issues/1018#issuecomment-425465121


mueller-ma:
An image refresh is triggered whenever the image comes to foreground (e.g. when unlocking the device), so it's not necessary to check how much time has been elapsed since the device lock.

PR #1061, opened 2018-10-24
Fix slow chart loading issue

### LINKED ISSUE #440: Time to display a chart
It takes around 7 seconds to render a chart. Why is it so long while it is fast in Basic UI ?

mueller-ma:
Fixes #440

This issue can already solved server-side by setting a max width for
charts in Paper UI, but it's not easy to find for the user.

@lolodomo Do you want to test this?

Low res. Please note, that my device has a quite low resolution.
screenshot_2018-10-24-09-13-16

High res:
screenshot_2018-10-24-09-13-31

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
@@ -74,6 +74,8 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Random;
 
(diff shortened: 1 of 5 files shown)


mueller-ma:
@kaikreuzer WDYT about this PR?

> This issue can already solved server-side by setting a max width for charts in Paper UI, but it's not easy to find for the user.

kaikreuzer:
@mueller-ma Sounds like a pragmatic solution to me.

PR #1075, opened 2018-10-30
Support for changing StringItems via UI.

paphko:
I always wondered why it is not possible to change string items via the Android UI - at least it is possible via the Paper UI!

So I added a listener on string items which prompts for a new value.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
@@ -11,6 +11,7 @@ package org.openhab.habdroid.ui;
 
 import android.annotation.SuppressLint;
 import android.content.Context;
+import android.content.DialogInterface;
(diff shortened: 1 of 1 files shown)


mueller-ma:
Changing text items is not possible in Basic UI, so I highly doubt it should be possible in the app. Do you know if there are any plans to add this to Basic UI?

paphko:
I don't know about basic ui, but I already use this feature in the Android app :-)
It works nicely.

maniac103:
Without a way to distinguish changeable values from non-changeable ones, I don't think we should add this. Offering a way to change e.g. a displayed date or temperature value doesn't make any sense to me.

(The patch also has some other issues like hardcoded strings and no use of lambdas, but I don't think it makes sense to make you fix them as long as the general question is unresolved)

mueller-ma:
As long it's not possible to change string items in Basic UI this PR shouldn't be merged.

PR #1089, opened 2018-11-10
Use switches instead of checkboxes in settings

mueller-ma:
> Use switches to:
>  * Toggle a single option on or off, on mobile and tablet
>  * Immediately activate or deactivate something

from https://material.io/design/components/selection-controls.html#usage


mueller-ma:
I should also change to demo mode summary to something more informative.

PR #1114, opened 2018-12-01
Remove pullapprove

mueller-ma:
Pullapprove's v2 api has been depricated and v3 doesn't offer a
sign-off-by check.
Replaced by https://probot.github.io/apps/dco/


mueller-ma:
@kaikreuzer Can you accept the request to add https://probot.github.io/apps/dco/ and remove pullapprove?

kaikreuzer:
Accepted DCO, but could not see Pullapprove being used anywhere...

mueller-ma:
Pullapprove should be in your account

kaikreuzer:
I guess Pullapprove simply won't be active anymore, when this PR is merged, because it is looking for the yml file. We still use it on other repos, so I don't want to revoke access from my account right now.

PR #1123, opened 2018-12-07
Save server properties for later use

mueller-ma:
When the app is offline and the user opens the settings, some settings
like chart scaling are missing and it might be confusing to have that
settings "just sometimes". Since we don't support multiple servers
(except demo server, which is excluded), saving the flags to prefs
seems save to me.


maniac103:
An alternative to this approach would be just showing all options if the server version is unknown ... this would also cover the case where we never successfully connected to a server so far.

If we're going to save the flags instead, I'd do the saving in the successCb in MainActivity; that way we don't need any extra parameters and need to add the code only once.

PR #1124, opened 2018-12-07
Display HABPanel

### LINKED ISSUE #426: FR: Display habpanel
This app has some features for wallmounted tablets, e.g. wakelock and full screen. Do you think it might be good to have a webview, that displays habpanel and make profit of these features?

mueller-ma:
Fixes #426

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
@@ -34,6 +34,7 @@ public abstract class ServerProperties implements Parcelable {
     public static final int SERVER_FLAG_SSE_SUPPORT           = 1 << 1;
     public static final int SERVER_FLAG_ICON_FORMAT_SUPPORT   = 1 << 2;
     public static final int SERVER_FLAG_CHART_SCALING_SUPPORT = 1 << 3;
+    public static final int SERVER_FLAG_HABPANEL_INSTALLED    = 1 << 4;
 
     public static class UpdateHandle {
         public void cancel() {
@@ -64,6 +65,10 @@ public abstract class ServerProperties implements Parcelable {
 
     public boolean hasSseSupport() {
         return (flags() & SERVER_FLAG_SSE_SUPPORT) != 0;
+    }
+
+    public boolean hasHabpanelInstalled() {
+        return (flags() & SERVER_FLAG_HABPANEL_INSTALLED) != 0;
     }
 
     abstract Builder toBuilder();
@@ -119,6 +124,15 @@ public abstract class ServerProperties implements Parcelable {
                     } catch (NumberFormatException nfe) {
                         // ignored: older versions without SSE support didn't return a number
                     }
+
+                    JSONArray jsonArray = result.getJSONArray("links");
+                    for (int i = 0; i < jsonArray.length(); i++) {
+                        JSONObject extensionJson = jsonArray.getJSONObject(i);
+                        if (extensionJson.getString("type").equals("habpanel")) {
+                            flags |= SERVER_FLAG_HABPANEL_INSTALLED;
+                        }
+                    }
+
                     handle.builder.flags(flags);
                     fetchSitemaps(client, handle, successCb, failureCb);
                 } catch (JSONException e) {
--- a/mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
@@ -125,16 +125,16 @@ public abstract class ServerProperties implements Parcelable {
                         // ignored: older versions without SSE support didn't return a number
                     }
 
-                    try {
-                        JSONArray jsonArray = result.getJSONArray("links");
+                    JSONArray jsonArray = result.optJSONArray("links");
+                    if (jsonArray == null) {
+                        Log.e(TAG, "No 'links' array available");
+                    } else {
                         for (int i = 0; i < jsonArray.length(); i++) {
                             JSONObject extensionJson = jsonArray.getJSONObject(i);
-                            if (extensionJson.getString("type").equals("habpanel")) {
+                            if ("habpanel".equals(extensionJson.getString("type"))) {
                                 flags |= SERVER_FLAG_HABPANEL_INSTALLED;
                             }
                         }
-                    } catch (JSONException e) {
-                        Log.e(TAG, "Failed checking for installed extensions", e);
                     }
 
                     handle.builder.flags(flags);
--- a/mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
@@ -125,14 +125,15 @@ public abstract class ServerProperties implements Parcelable {
                         // ignored: older versions without SSE support didn't return a number
                     }
 
-                    JSONArray jsonArray = result.optJSONArray("links");
-                    if (jsonArray == null) {
+                    JSONArray linksJsonArray = result.optJSONArray("links");
+                    if (linksJsonArray == null) {
                         Log.e(TAG, "No 'links' array available");
                     } else {
-                        for (int i = 0; i < jsonArray.length(); i++) {
-                            JSONObject extensionJson = jsonArray.getJSONObject(i);
+                        for (int i = 0; i < linksJsonArray.length(); i++) {
+                            JSONObject extensionJson = linksJsonArray.getJSONObject(i);
                             if ("habpanel".equals(extensionJson.getString("type"))) {
                                 flags |= SERVER_FLAG_HABPANEL_INSTALLED;
+                                break;
                             }
                         }
                     }
--- a/mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
@@ -125,12 +125,16 @@ public abstract class ServerProperties implements Parcelable {
                         // ignored: older versions without SSE support didn't return a number
                     }
 
-                    JSONArray jsonArray = result.getJSONArray("links");
-                    for (int i = 0; i < jsonArray.length(); i++) {
-                        JSONObject extensionJson = jsonArray.getJSONObject(i);
-                        if (extensionJson.getString("type").equals("habpanel")) {
-                            flags |= SERVER_FLAG_HABPANEL_INSTALLED;
+                    try {
+                        JSONArray jsonArray = result.getJSONArray("links");
+                        for (int i = 0; i < jsonArray.length(); i++) {
+                            JSONObject extensionJson = jsonArray.getJSONObject(i);
+                            if (extensionJson.getString("type").equals("habpanel")) {
+                                flags |= SERVER_FLAG_HABPANEL_INSTALLED;
+                            }
                         }
+                    } catch (JSONException e) {
+                        Log.e(TAG, "Failed checking for installed extensions", e);
                     }
 
                     handle.builder.flags(flags);
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
@@ -558,7 +558,10 @@ public class MainActivity extends AppCompatActivity implements
                 if (mServerProperties != null && mPropsUpdateHandle == null) {
                     mPropsUpdateHandle = ServerProperties.updateSitemaps(mServerProperties,
                             mConnection,
-                            props -> { mServerProperties = props; updateSitemapAndHabpanelDrawerItems(); },
+                            props -> {
+                                mServerProperties = props;
+                                updateSitemapAndHabpanelDrawerItems();
+                            },
                             MainActivity.this::handlePropertyFetchFailure);
                 }
             }
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
@@ -530,7 +530,7 @@ public class MainActivity extends AppCompatActivity implements
         mController.triggerPageUpdate(pageUrl, forceReload);
     }
 
-    public void setupToolbar() {
+    private void setupToolbar() {
         Toolbar toolbar = findViewById(R.id.openhab_toolbar);
         setSupportActionBar(toolbar);
         ActionBar actionBar = getSupportActionBar();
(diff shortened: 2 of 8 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -605,6 +619,90 @@ public void onClick(View view) {
...
+                updateViewVisibility(true, false, view);
+                return;
+            }
+
+            if (mConnection == null) {
+                updateViewVisibility(true, false, view);
+                return;
+            } else {
+                updateViewVisibility(false, true, view);
+            }
```
maniac103: Could omit the else clause here:
```
            if (mConnection == null) {
                updateViewVisibility(true, false, view);
                return;
            }

            updateViewVisibility(false, true, view);
```

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -605,6 +619,90 @@ public void onClick(View view) {
...
+
+            HttpUrl url = mConnection.getAsyncHttpClient().buildUrl("/habpanel/index.html");
```
maniac103: `String url = mConnection.getAsyncHttpClient().buildUrl(...).toString()`

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -605,6 +619,90 @@ public void onClick(View view) {
...
+                @Override
+                public void onPageFinished(WebView view, String url) {
+                    updateViewVisibility(false, false, getView());
+                }
+
+                @Override
+                public void onReceivedError(WebView view, WebResourceRequest request,
+                        WebResourceError error) {
+                    updateViewVisibility(true, false, getView());
+                }
```
maniac103: From https://developer.android.com/reference/android/webkit/WebViewClient.html#onReceivedError(android.webkit.WebView,%20android.webkit.WebResourceRequest,%20android.webkit.WebResourceError):
```
Note that unlike the deprecated version of the callback, the new version will be called for any resource (iframe, image, etc.), not just for the main page.
```
So we probably don't want to do this unconditionally?

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -605,6 +619,90 @@ public void onClick(View view) {
...
+
+                @Override
+                public void onReceivedHttpError(WebView view, WebResourceRequest request,
+                        WebResourceResponse errorResponse) {
+                    updateViewVisibility(true, false, getView());
+                }
```
maniac103: As above

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -605,6 +619,90 @@ public void onClick(View view) {
...
+        }
+
+        @Override
+        public void onAvailableConnectionChanged() {
+            loadHabpanel(getView());
+        }
+
+        @Override
+        public void onCloudConnectionChanged(CloudConnection connection) {
+            loadHabpanel(getView());
```
maniac103: Why this? Habpanel doesn't care about the cloud connection...

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -605,6 +619,90 @@ public void onClick(View view) {
...
+
+            TextView retryButton = view.findViewById(R.id.retry_button);
+            retryButton.setOnClickListener(v -> loadHabpanel(view));
+            loadHabpanel(view);
+
+            return view;
+        }
+
+        private void loadHabpanel(View view) {
+            updateViewVisibility(false, true, view);
```
maniac103: This call seems pointless, since you'll call `updateViewVisibility` again in any case below.

mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
```diff
@@ -119,6 +124,15 @@ public void onSuccess(String response, Headers headers) {
                     } catch (NumberFormatException nfe) {
                         // ignored: older versions without SSE support didn't return a number
                     }
+
+                    JSONArray jsonArray = result.getJSONArray("links");
```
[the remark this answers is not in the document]
mueller-ma: I haven't checked it so far.

mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
```diff
@@ -119,6 +124,15 @@ public void onSuccess(String response, Headers headers) {
...
+                    for (int i = 0; i < jsonArray.length(); i++) {
+                        JSONObject extensionJson = jsonArray.getJSONObject(i);
+                        if (extensionJson.getString("type").equals("habpanel")) {
```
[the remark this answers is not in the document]
mueller-ma: I haven't checked it so far.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
```diff
@@ -511,12 +512,13 @@ public void triggerPageUpdate(String pageUrl, boolean forceReload) {
         mController.triggerPageUpdate(pageUrl, forceReload);
     }
 
-    private void setupToolbar() {
+    public void setupToolbar() {
```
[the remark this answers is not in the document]
mueller-ma: Sure

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: Will do

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: Will do

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: I didn't see that. I need to find a better way to hide the webview if there are connection errors.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: HABPanel should only be refresh on onAvailableConnectionChanged.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: This sets the UI in the "loading" state while the connection is fetched.

maniac103: This whole method runs on the UI thread, so during the execution of this method no UI will be updated anyway (and `getUsableConnection()` executes fast anyway, as otherwise it wouldn't be OK to run on the main thread) ... so I still am pretty sure this call has no effect at all.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
maniac103: Could check `request.getUrl()`?

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: Right, I remove it

mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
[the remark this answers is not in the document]
mueller-ma: Since the links array contains links to elementary parts of openHAB, e.g. items, I'm pretty sure it's always present.

mueller-ma:
@maniac103 Do you think this is worth a static shortcut?

maniac103:
@mueller-ma What do you mean exactly?

mueller-ma:
https://developer.android.com/guide/topics/ui/shortcuts/
Anyway I don't want to add this in this PR.

maniac103:
Ah, app shortcuts. Yeah, that would make sense to me, same as for notification list.

mueller-ma:
How should the notification shortcut behave if no remote connection is set?
1. Not present at all
2. Present, but shows toast that connection must be set
3. Present and opens notification list, but shows error there

maniac103:
I'd say 1, given we also hide the notification list from the drawer if remote connection is unset.

mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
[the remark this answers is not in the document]
mueller-ma: I added a safeguard against a missing links array.

mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
```diff
@@ -119,6 +124,19 @@ public void onSuccess(String response, Headers headers) {
...
+
+                    try {
+                        JSONArray jsonArray = result.getJSONArray("links");
+                        for (int i = 0; i < jsonArray.length(); i++) {
+                            JSONObject extensionJson = jsonArray.getJSONObject(i);
+                            if (extensionJson.getString("type").equals("habpanel")) {
+                                flags |= SERVER_FLAG_HABPANEL_INSTALLED;
+                            }
+                        }
+                    } catch (JSONException e) {
```
maniac103: This feels a bit ... unclean to me. Why not use `optJSONArray()` and check for null?

Similarly for the if check, just do `if ("habpanel".equals(extensionJson.optString("type"))`

mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
```diff
@@ -119,6 +124,19 @@ public void onSuccess(String response, Headers headers) {
                     } catch (NumberFormatException nfe) {
                         // ignored: older versions without SSE support didn't return a number
                     }
+
+                    JSONArray jsonArray = result.optJSONArray("links");
```
maniac103: Nit: maybe name this variable `links` or something?

mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
```diff
@@ -119,6 +124,19 @@ public void onSuccess(String response, Headers headers) {
...
+                    if (jsonArray == null) {
+                        Log.e(TAG, "No 'links' array available");
+                    } else {
+                        for (int i = 0; i < jsonArray.length(); i++) {
+                            JSONObject extensionJson = jsonArray.getJSONObject(i);
+                            if ("habpanel".equals(extensionJson.getString("type"))) {
+                                flags |= SERVER_FLAG_HABPANEL_INSTALLED;
```
maniac103: Can `break;` here.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -654,6 +668,95 @@ public void onClick(View view) {
...
+                    if (url.endsWith("/rest/events")) {
+                        updateViewVisibility(true, false, getView());
+                    }
+                }
+
+                @Override
+                public void onReceivedError(WebView view, int errorCode, String description,
+                        String failingUrl) {
+                    Log.e(TAG, "onReceivedError() (old) on URL: " + failingUrl);
+                    updateViewVisibility(true, false, getView());
```
maniac103: Why don't you check the failing URL here like you do above?

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -654,6 +668,95 @@ public void onClick(View view) {
...
+                }
+            });
+            webView.getSettings().setDomStorageEnabled(true);
+            webView.getSettings().setJavaScriptEnabled(true);
+            webView.loadUrl(url);
+        }
+
+        private void updateViewVisibility(boolean error, boolean loading, @NonNull View view) {
```
maniac103: NonNull Annotation doesn't match the check below.
Also can remove the `view` parameter from the method signature and just call getView() inside the method.

mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
mueller-ma: `linksJsonArray`? I want to keep the fact that it's a json object in the name.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: https://github.com/openhab/openhab-android/pull/1124#discussion_r241302142
This is the deprecated version, so it's only called for the main page, right?

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: updateViewVisibility() is called by loadHabpanel() which is called during onCreateView(), so `view` (which is non-null) != `getView()` (which is null), right?

mobile/src/main/java/org/openhab/habdroid/model/ServerProperties.java
maniac103: Sounds fine.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
maniac103: Just re-checked the docs; you're right, so never mind.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
maniac103: Just do it from onViewCreated() and that point will be moot :-)

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -654,6 +668,96 @@ public void onClick(View view) {
...
+        }
+
+        @Override
+        public View onCreateView(@NonNull LayoutInflater inflater,
+                ViewGroup container, Bundle savedInstanceState) {
+            View view = inflater.inflate(R.layout.fragment_habpanel, container, false);
+
+            TextView retryButton = view.findViewById(R.id.retry_button);
+            retryButton.setOnClickListener(v -> loadHabpanel(view));
+            loadHabpanel(view);
```
maniac103: Should still move those 3 lines to onViewCreated(), so `updateViewVisibility()` from `loadHabpanel()` works. Can omit the `view` parameter from `loadHabpanel` in that case as well.

mueller-ma: Done :)

PR #1131, opened 2018-12-09
Enhance status fragments

### LINKED ISSUE #1115: Enhance "No connection" fragments
1. Add "Turn on Wifi" button if wifi is turned off and no remote connection is configured.
2. Add button and info text to enable demo mode if discovery failed.

mueller-ma:
* Add ability to show a second hint and/or button.

Discorvery failed:
* Discorvery failed hint and open settings button (old)
* Demo mode hint and enable demo mode button (added)

No network available:
* Enable wifi button (added)

No remote server:
If wifi is enabled:
* "No remote server configured" hint and open settings button (old)

If wifi is disabled:
* "No remote server configured" hint and open settings button (old)
* Enable wifi button (added)

Fixes #1115

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
@@ -138,7 +138,7 @@ public class MainActivity extends AppCompatActivity implements
     private Uri mPendingNfcData;
     private String mPendingOpenedNotificationId;
     private Sitemap mSelectedSitemap;
-    public ContentController mController;
+    private ContentController mController;
     private ServerProperties mServerProperties;
     private ServerProperties.UpdateHandle mPropsUpdateHandle;
     private boolean mStarted;
@@ -273,6 +273,14 @@ public class MainActivity extends AppCompatActivity implements
             }
         }
         queryServerProperties();
+    }
+
+    public void enableWifiAndIndicateStartup() {
+        WifiManager wifiManager =
+                (WifiManager) getApplicationContext().getSystemService(Context.WIFI_SERVICE);
+        wifiManager.setWifiEnabled(true);
+        mController.updateConnection(null, getString(R.string.waiting_for_wifi),
+                R.drawable.ic_signal_wifi_0_bar_black_24dp);
     }
 
     public void retryServerPropertyQuery() {
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
@@ -445,7 +445,8 @@ public class MainActivity extends AppCompatActivity implements
                     mController.indicateMissingConfiguration(false);
                 }
             } else if (failureReason != null) {
-                WifiManager wifiManager = (WifiManager) getSystemService(Context.WIFI_SERVICE);
+                WifiManager wifiManager = (WifiManager)
+                        getApplicationContext().getSystemService(Context.WIFI_SERVICE);
                 if (failureReason instanceof NetworkNotSupportedException) {
                     NetworkInfo info =
                             ((NetworkNotSupportedException) failureReason).getNetworkInfo();
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
@@ -445,18 +445,19 @@ public class MainActivity extends AppCompatActivity implements
                     mController.indicateMissingConfiguration(false);
                 }
             } else if (failureReason != null) {
+                WifiManager wifiManager = (WifiManager) getSystemService(Context.WIFI_SERVICE);
                 if (failureReason instanceof NetworkNotSupportedException) {
                     NetworkInfo info =
                             ((NetworkNotSupportedException) failureReason).getNetworkInfo();
                     mController.indicateNoNetwork(
-                            getString(R.string.error_network_type_unsupported, info.getTypeName()));
+                            getString(R.string.error_network_type_unsupported, info.getTypeName()), false);
                 } else if (failureReason instanceof NetworkNotAvailableException
-                        && !((WifiManager) getSystemService(Context.WIFI_SERVICE))
-                        .isWifiEnabled()) {
-                    mController.indicateEnableWifiNetwork(
-                            getString(R.string.error_wifi_not_available));
+                        && !wifiManager.isWifiEnabled()) {
+                    mController.indicateNoNetwork(
+                            getString(R.string.error_wifi_not_available), true);
                 } else {
-                    mController.indicateNoNetwork(getString(R.string.error_network_not_available));
+                    mController.indicateNoNetwork(getString(R.string.error_network_not_available),
+                            false);
                 }
             } else {
                 mController.updateConnection(null, null);
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
@@ -451,7 +451,8 @@ public class MainActivity extends AppCompatActivity implements
                     mController.indicateNoNetwork(
                             getString(R.string.error_network_type_unsupported, info.getTypeName()));
                 } else if (failureReason instanceof NetworkNotAvailableException
-                        && !((WifiManager) getSystemService(Context.WIFI_SERVICE)).isWifiEnabled()) {
+                        && !((WifiManager) getSystemService(Context.WIFI_SERVICE))
+                        .isWifiEnabled()) {
                     mController.indicateEnableWifiNetwork(
                             getString(R.string.error_wifi_not_available));
                 } else {
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
@@ -450,7 +450,8 @@ public class MainActivity extends AppCompatActivity implements
                     NetworkInfo info =
                             ((NetworkNotSupportedException) failureReason).getNetworkInfo();
                     mController.indicateNoNetwork(
-                            getString(R.string.error_network_type_unsupported, info.getTypeName()), false);
+                            getString(R.string.error_network_type_unsupported, info.getTypeName()),
+                            false);
                 } else if (failureReason instanceof NetworkNotAvailableException
                         && !wifiManager.isWifiEnabled()) {
                     mController.indicateNoNetwork(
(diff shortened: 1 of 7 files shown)


maniac103:
Do you have a screenshot of how it looks now?

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
```diff
@@ -443,16 +445,19 @@ public void onAvailableConnectionChanged() {
...
-                final String message;
                 if (failureReason instanceof NetworkNotSupportedException) {
                     NetworkInfo info =
                             ((NetworkNotSupportedException) failureReason).getNetworkInfo();
-                    message = getString(R.string.error_network_type_unsupported,
-                            info.getTypeName());
+                    mController.indicateNoNetwork(
+                            getString(R.string.error_network_type_unsupported, info.getTypeName()));
+                } else if (failureReason instanceof NetworkNotAvailableException
+                        && !((WifiManager) getSystemService(Context.WIFI_SERVICE))
```
maniac103: I'd prefer to fetch the WifiManager instance out of the if-else cascade (e.g. in line 448), in order to make the code more clear/readable.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -232,16 +235,31 @@ public void indicateNoNetwork(CharSequence message) {
         mActivity.updateTitle();
     }
 
+    /**
+     * Indicate to the user that no network connectivity is present and Wi-Fi is disabled.
+     *
+     * @param message Error message to show
+     */
+    public void indicateEnableWifiNetwork(CharSequence message) {
```
maniac103: Couldn't this be solved by an additional parameter to `indicateNoNetwork` (e.g. `shouldSuggestEnablingWifi`)

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -543,99 +561,125 @@ protected static int determineTransition(FragmentUpdateReason reason) {
...
-            CharSequence message = arguments.getCharSequence("message");
-            if (!TextUtils.isEmpty(message)) {
-                descriptionText.setText(message);
-            } else {
-                descriptionText.setVisibility(View.GONE);
+            boolean needsOr = false;
+            if (setDescription(arguments, view, R.id.description1, KEY_MESSAGE_1)
+                    & setDescription(arguments, view, R.id.description2, KEY_MESSAGE_2)) {
+                needsOr = true;
             }
```
maniac103: boolean needsOr = setDescription(...) && setDescription(...);

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -646,16 +690,77 @@ public View onCreateView(@NonNull LayoutInflater inflater,
                 watermark.setVisibility(View.GONE);
             }
 
-            final Button button = view.findViewById(R.id.button);
-            int buttonTextResId = arguments.getInt("buttontext");
+            if (initButton(arguments, view, R.id.button1, KEY_BUTTON_1_TEXT, KEY_BUTTON_1_TAG)
+                    & initButton(arguments, view, R.id.button2, KEY_BUTTON_2_TEXT,
+                    KEY_BUTTON_2_TAG)) {
+                needsOr = true;
+            }
```
maniac103: As above.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -646,16 +690,77 @@ public View onCreateView(@NonNull LayoutInflater inflater,
...
+                WifiManager wifiManager =
+                        (WifiManager) getContext().getSystemService(Context.WIFI_SERVICE);
+                wifiManager.setWifiEnabled(true);
+            } else if (view.getTag().equals(BUTTON_TAG_RETRY_NETWORK)) {
+                ConnectionFactory.restartNetworkCheck();
+                getActivity().recreate();
+            } else if (view.getTag().equals(BUTTON_TAG_RETRY_SERVER_PROP_FETCH)) {
+                ((MainActivity) getActivity()).retryServerPropertyQuery();
+            }
+        }
```
maniac103: Could use a lambda for each of the cases, assign the lambda as tag and execute it in onClick. Would save the indirection.

mueller-ma: Some actions are used more than once, e.g. enable wifi and open settings. This is why I put everything in this function.

maniac103:
Having now looked at the new layout, I fail to like it :-/ The 'or' text looks out of place IMHO, and so does the second description text.
I'd prefer if we dropped the 'or' text and keep a single help text for each of the cases, which just becomes longer if there's a second button, and then place the buttons right below each other (phone) or next to each other (tablet?).

To give a wording example, in the 'discovery failed' case I'd suggest something like `We didn't find an openHAB server in your network. If you already have a server, please configure its IP address or host name in the server settings. If you don't yet have a server, you can enable demo mode to see what openHAB is capable of.` (and then have the two buttons below that).

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
```diff
@@ -136,7 +138,7 @@
     private Uri mPendingNfcData;
     private String mPendingOpenedNotificationId;
     private Sitemap mSelectedSitemap;
-    private ContentController mController;
+    public ContentController mController;
```
maniac103: Should keep this private.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -543,99 +554,115 @@ protected static int determineTransition(FragmentUpdateReason reason) {
     public static class CommunicationFailureFragment extends StatusFragment {
```
maniac103: The whole reason for the existence of those classes is/was the different onClick implementation. Could nuke them with the current impl. What I think I'd prefer, though, is
- having action implementation methods in the base class (`enableWifi()`, `enableDemoMode()` etc.)
- keep the onClick implementations in the sub classes and make those call the base class methods
- nuke the button tags

That way it's more obvious what's going on (less indirections), and the argument list for the fragment becomes shorter.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -543,99 +554,115 @@ protected static int determineTransition(FragmentUpdateReason reason) {
...
             View view = inflater.inflate(R.layout.fragment_status, container, false);
-
-            TextView descriptionText = view.findViewById(R.id.description);
-            CharSequence message = arguments.getCharSequence("message");
-            if (!TextUtils.isEmpty(message)) {
-                descriptionText.setText(message);
-            } else {
-                descriptionText.setVisibility(View.GONE);
-            }
+            setDescription(arguments, view, R.id.description, KEY_MESSAGE);
```
maniac103: Could restore original code here now that we have only one description.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -646,16 +673,74 @@ public View onCreateView(@NonNull LayoutInflater inflater,
...
+            } else if (view.getTag().equals(BUTTON_TAG_ENABLE_DEMO_MODE)) {
+                PreferenceManager.getDefaultSharedPreferences(getContext())
+                        .edit()
+                        .putBoolean(Constants.PREFERENCE_DEMOMODE, true)
+                        .apply();
+            } else if (view.getTag().equals(BUTTON_TAG_ENABLE_WIFI)) {
+                WifiManager wifiManager = (WifiManager)
+                        getContext().getApplicationContext().getSystemService(Context.WIFI_SERVICE);
+                wifiManager.setWifiEnabled(true);
+                ((MainActivity) getActivity()).mController.updateConnection(null,
```
maniac103: `((MainActivity) getActivity()).mController` equals `this`, so you can omit it and just call `updateConnection`

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: Will do

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: Without refactoring `this` is the onClickListener.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: I added the tags, because in MissingConfigurationFragment the buttons can have different actions based on hasWifiEnabled and resolvedAttempted. I can re-add the onClick() functions in all fragments, but then I need to split MissingConfigurationFragment in three fragments.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
maniac103: Oops, right. I thought I had attempted that refactoring, but it looks like I failed at it.
I'll have a look at how to refactor this, I really don't like making `mController` public.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
maniac103: You don't ;-) I'll upload a suggestion to my repo.

mobile/src/main/res/layout/fragment_status.xml
```diff
@@ -1,48 +1,66 @@
...
-        android:id="@+id/image"
-        android:layout_width="100dp"
-        android:layout_height="wrap_content"
-        android:layout_gravity="center"
-        android:adjustViewBounds="true"
-        android:gravity="center"
-        android:scaleType="fitXY"
-        android:src="@drawable/ic_signal_cellular_off_black_24dp"
-        tools:ignore="contentDescription"/>
+<ScrollView
```
maniac103: Is the ScrollView still needed? I still think we should avoid the need for scrolling in those views.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
maniac103: https://github.com/maniac103/openhab.android/commit/dee4267e5022a335df14f3f442649fe41da5ce48 would be my suggestion.

mobile/src/main/res/layout/fragment_status.xml
mueller-ma: I have no device where the view is scrollable, but IMO it's better to have a safeguard for extra small devices.

maniac103: What's the smallest screen size we (want to) support? Since our min API level is ICS I'd say 4 or 4.3 inch?
(I want to double check with emulator)

mueller-ma: Even if it's ok in English, it might be too much text in other languages.

PR #1135, opened 2018-12-11
Add activity to view and share adb log

### LINKED ISSUE #565: Custom log functions
When we have our own log function, we can enhance the logging. They basicly pass the message to Android's `Log`, but have some more features, e.g.

- Only log debug and verbose, when running a debug build (like Android Studio suggests), but without needing to write the condition every time Log.d is used
- When an exception is caught, show stactraces in dialog, when debug mode is enabled. Most users are not able to capture a log, but they can easily take screenshots.
- Write a user accessable log file, when in debug mode

mueller-ma:
It's only visible if debug mode is enabled.

Closes #565


mobile/build.gradle
```diff
@@ -17,7 +17,7 @@ buildscript {
 }
 
 apply plugin: 'com.android.application'
-if(!isFoss) {
+if (!isFoss) {
```
mueller-ma: The formatting changes here and in the manifest were done by Android Studio when I added the new activity.

mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.java
```diff
@@ -0,0 +1,137 @@
...
+            sendIntent.putExtra(Intent.EXTRA_TEXT, logView.getText());
+            sendIntent.setType("text/plain");
+            startActivity(sendIntent);
+        });
+    }
+
+    @Override
+    protected void onResume() {
+        super.onResume();
+        setProgressIndicatorVisible(true);
```
mueller-ma: For some reason I never see the progress indicator.

mobile/build.gradle
maniac103: In a first glance they look correct anyway. One could put them to a separate commit, but since we squash PRs anyway...

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
```diff
@@ -575,6 +578,9 @@ public void onDrawerOpened(View drawerView) {
                 case R.id.about:
                     openAbout();
                     return true;
+                case R.id.log:
+                    openLog();
+                    return true;
```
maniac103: Wouldn't it be a better idea to add an entry to the prefs list (below the debug toggle) instead of the drawer? IMHO it's not that important to justify a drawer item; I'll guess we'll need to point most people towards it either way.

mobile/src/main/res/layout/activity_log.xml
```diff
@@ -0,0 +1,84 @@
...
+        android:layout_marginRight="16dp"
+        android:layout_marginBottom="16dp"
+        app:fabSize="normal"
+        android:src="@drawable/ic_share_white_24dp"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent" />
+
+    <ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+        xmlns:app="http://schemas.android.com/apk/res-auto"
+        xmlns:tools="http://schemas.android.com/tools"
```
maniac103: The tools namespace should be moved to the top level item, the others are duplicated.

mobile/src/main/res/layout/activity_log.xml
```diff
@@ -0,0 +1,84 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="0dp"
+        android:fillViewport="true"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/openhab_toolbar">
+
+        <androidx.constraintlayout.widget.ConstraintLayout
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            tools:context=".ui.LogActivity">
```
maniac103: Shouldn't / couldn't this be a FrameLayout?

mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.java
```diff
@@ -0,0 +1,137 @@
...
+        new GetLogFromAdbTask().execute(false);
+    }
+
+    private void setProgressIndicatorVisible(boolean visible) {
+        findViewById(R.id.progressBar).setVisibility(visible ? View.VISIBLE : View.GONE);
```
maniac103: Save a reference to the progress bar in onCreate? Should probably also toggle visibility of the log view here, and either toggle visibility or enabled state of the FAB.

mobile/src/main/res/menu/log_menu.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <item
+        android:id="@+id/delete_log"
+        android:icon="@drawable/ic_delete_white_24dp"
+        android:title="@string/delete_log"
+        app:showAsAction="always" />
```
maniac103: Isn't this action rather 'clear and update' instead of 'delete'? In that case, a refresh icon would also make a bit more sense to me.

maniac103:
Very good idea :+1:

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
mueller-ma: Makes sense, I'll change it.

mobile/src/main/res/layout/activity_log.xml
mueller-ma: I'll remove it.

mobile/src/main/res/layout/activity_log.xml
mueller-ma: It could be, but "Note: For better performance and tooling support, you should instead build your layout with ConstraintLayout." (from Linear Layout page, but there is no page for frame) https://developer.android.com/guide/topics/ui/layout/linear.

mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.java
mueller-ma: Done.

mobile/src/main/res/menu/log_menu.xml
mueller-ma: IMO a refresh icon would be incorrect as the user thinks it will add new entries, but it removes all entries.

mueller-ma:
TODO:

- [ ] Change issue template
- [ ] Add formatting to logcat call (default format differs between devices)

mobile/src/main/res/layout/activity_log.xml
maniac103: I don't think FrameLayout can be beaten in terms of performance, since it's among the simplest containers imaginable. It's significantly simpler than LinearLayout, and since you want to only show one of the children at a time, I think it's the perfect tool for the job.
NB: ConstraintLayout's performance improvements only become relevant if (can) use it to avoid nesting other layouts (e.g. replacing nested horizontal/vertical LinearLayouts by one ConstraintLayout). When doing 1:1 replacement there's no performance difference.

mobile/src/main/res/menu/log_menu.xml
maniac103: Fair enough. MatLog uses a 'close' icon for that purpose ... how about that one?

mueller-ma: IMO the delete icon is the best choice here.

mobile/src/main/res/layout/activity_log.xml
mueller-ma: I changed the inner ConstraintLayout to a FrameLayout.

mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.java
```diff
@@ -103,15 +109,16 @@ public boolean onOptionsItemSelected(MenuItem item) {
 
     private class GetLogFromAdbTask extends AsyncTask<Boolean, Void, String> {
         @Override
-        protected String doInBackground(Boolean... clearBeforeRead) {
+        protected String doInBackground(Boolean... clear) {
             StringBuilder logBuilder = new StringBuilder();
             String separator = System.getProperty("line.separator");
             try {
-                if (clearBeforeRead[0]) {
+                if (clear[0]) {
```
mueller-ma: @maniac103 The doInBackground does two indipendent thing now. Should I split them in two AsyncTasks?

maniac103: No, I think it's fine as is. Chaining two threads won't gain us anything as dar as I can tell.

mobile/src/main/res/values/strings.xml
```diff
@@ -200,6 +200,12 @@
     <string name="no_remote_server">Local network is unavailable and you haven\'t configured a remote server</string>
     <string name="resolving_openhab">Trying to discover openHAB server instance\u2026</string>
 
+    <!-- Log -->
+    <string name="log">Log</string>
+    <string name="view_log">View log</string>
+    <string name="delete_log">Delete log</string>
```
maniac103: 'Clear log'?

mobile/src/main/res/layout/activity_log.xml
```diff
@@ -0,0 +1,107 @@
...
+                android:clickable="true"
+                android:focusable="true"
+                android:textSize="12sp"
+                android:typeface="monospace"
+                app:fontFamily="monospace"
+                app:layout_constraintBottom_toBottomOf="parent"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toTopOf="parent"
+                app:layout_constraintVertical_bias="0.0"
```
maniac103: Should nuke the ConstraintLayout specific stuff here and below.

mobile/src/main/res/values/strings.xml

mobile/src/main/res/layout/activity_log.xml
mueller-ma: Done

PR #1137, opened 2018-12-12
Send alarm clock to openHAB

mueller-ma:
and also add a framework for adding more background tasks like this in
the future.

Thanks to @ekahlenberger for this idea.

TODO:

- [x] Add feature with example to Readme/Docs


mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
```diff
@@ -0,0 +1,151 @@
...
+                .setSmallIcon(R.drawable.ic_openhab_appicon_white_24dp)
+                .setContentTitle(context.getString(R.string.app_name))
+                .setWhen(System.currentTimeMillis())
+                .setStyle(new NotificationCompat.BigTextStyle().bigText(msg))
+                .setContentText(msg)
+                .setAutoCancel(true)
+                .setContentIntent(intent)
+                .setColor(ContextCompat.getColor(context, R.color.openhab_orange))
+                .setCategory(isError ? NotificationCompat.CATEGORY_ERROR
+                        : NotificationCompat.CATEGORY_PROGRESS)
```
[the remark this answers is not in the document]
mueller-ma: @maniac103 I think we have to remove some rules. Setting `:` to 16 would look like this:
````
                .setColor(ContextCompat.getColor(context, R.color.openhab_orange))
                .setCategory(isError ? NotificationCompat.CATEGORY_ERROR
                : NotificationCompat.CATEGORY_PROGRESS)
                .setOngoing(!isError)
````
This seems wrong to me

**Comment by mueller-ma** [comment]
APK for testing: [mobile-foss-beta-debug.zip](https://github.com/openhab/openhab-android/files/2673549/mobile-foss-beta-debug.zip)

File: mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java (hunk 1 above)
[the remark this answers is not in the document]
Comment (maniac103) [reply mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java]: Yeah, what you have definitely is correct.

File: mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java [hunk 2]
```diff
@@ -0,0 +1,72 @@
...
+        Log.d(TAG, "onReceive() with intent " + intent.getAction());
+        if (!intent.getAction().equals("android.app.action.NEXT_ALARM_CLOCK_CHANGED")
+                || Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP
+                || ! PreferenceManager.getDefaultSharedPreferences(context)
+                .getBoolean(PREFERENCE_ALARM_CLOCK_ENABLED, true)) {
+            Log.d(TAG, "Don't send data");
+            return;
+        }
+
+        startAlarmChangedWorker(context);
```
maniac103: I'd suggest inverting the logic here, and use a constant for the action:
```
if (AlarmManager.ACTION_NEXT_ALARM_CLOCK_CHANGED.equals(intent.getAction())
        && Build.VERSION_SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
    SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
    if (prefs.getBoolean(PREFERENCE_ALARM_CLOCK_ENABLED, true))) {
        startAlarmChangedWorker(context);
    }
}
```

mobile/src/main/AndroidManifest.xml
```diff
@@ -95,5 +95,20 @@
...
+
+        <receiver
+            android:name=".background.AlarmChangedReceiver"
+            android:enabled="true"
+            android:exported="true">
+            <intent-filter>
+                <action android:name="android.app.action.NEXT_ALARM_CLOCK_CHANGED" />
+            </intent-filter>
+        </receiver>
+        <receiver android:name=".background.BackgroundTaskRetryBroadcastReceiver" />
```
maniac103: Could maybe unify those two into one receiver?

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
```diff
@@ -0,0 +1,72 @@
...
+import org.openhab.habdroid.R;
+
+import static android.content.Context.NOTIFICATION_SERVICE;
+import static org.openhab.habdroid.background.BackgroundUtils.NOTIFICATION_ID_SEND_ALARM_CLOCK;
+import static org.openhab.habdroid.background.BackgroundUtils.NOTIFICATION_TAG_BACKGROUND;
+import static org.openhab.habdroid.background.BackgroundUtils.NOTIFICATION_TAG_BACKGROUND_ERROR;
+import static org.openhab.habdroid.background.BackgroundUtils.WORKER_TAG_SEND_ALARM_CLOCK;
+import static org.openhab.habdroid.background.BackgroundUtils.createNotificationChannels;
+import static org.openhab.habdroid.background.BackgroundUtils.makeBackgroundNotification;
+import static org.openhab.habdroid.util.Constants.PREFERENCE_ALARM_CLOCK_ENABLED;
```
maniac103: I'm not sure, but IMHO we should avoid static imports or at least constrain it to constants, so it's easier to track in the code where stuff is coming from.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
```diff
@@ -0,0 +1,72 @@
...
+        nm.cancel(NOTIFICATION_TAG_BACKGROUND_ERROR, NOTIFICATION_ID_SEND_ALARM_CLOCK);
+
+        Log.d(TAG, "Schedule worker");
+        workManager.enqueue(sendAlarmClockWorker);
+
+        createNotificationChannels(context);
+        Notification notification = makeBackgroundNotification(context,
+                R.string.waiting_for_network_to_send_alarm_clock, R.drawable.ic_alarm_grey_24dp,
+                false, null);
+        nm.notify(NOTIFICATION_TAG_BACKGROUND, NOTIFICATION_ID_SEND_ALARM_CLOCK, notification);
```
maniac103: Is that notification really necessary? I think notifying on failure (for retry) is OK, even though I'd probably not even do that (I'd just continue retrying), but notifying during the upload operation seems a bit much to me.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedWorker.java
```diff
@@ -0,0 +1,134 @@
...
+            Log.e(TAG, "Got no connection "
+                    + (failureReason != null ? failureReason.getMessage() : null));
+            return retryOrFail(
+                    context.getString(R.string.error_sending_alarm_clock_no_connection),
+                    context, nm);
+        }
+
+        AlarmManager alarmManager = (AlarmManager) context.getSystemService(Context.ALARM_SERVICE);
+        String nextAlarm;
+        if (alarmManager.getNextAlarmClock() == null) {
```
maniac103: Should use a local variable for this, as this is an IPC call.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedWorker.java
```diff
@@ -0,0 +1,134 @@
...
+            nextAlarm = "0";
+        } else {
+            nextAlarm = String.valueOf(alarmManager.getNextAlarmClock().getTriggerTime());
```
maniac103: Is there already some predecent for this format? IOW, will consumers of this item understand it? Does this need to be documented?

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedWorker.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+        String item = PreferenceManager.getDefaultSharedPreferences(context)
+                .getString(PREFERENCE_ALARM_CLOCK_ITEM, "");
+        if (TextUtils.isEmpty(item.trim())) {
+            return retryOrFail(
+                    context.getString(R.string.error_sending_alarm_clock_item_empty),
+                    context, nm);
```
maniac103: Could just fail right away (or, even better, don't start the worker in the first place - could have a static method `startIfNeeded(context)` that checks for the preconditions of the toggle being enabled and item name being non-empty.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
```diff
@@ -322,6 +344,25 @@ protected void updateAndInitPreferences() {
...
+            } else {
+                updateAlarmClockEnabledPreferenceIcon(alarmClockEnabledPref,
+                        getPreferenceBool(alarmClockEnabledPref, false));
+                alarmClockEnabledPref.setOnPreferenceChangeListener((preference, newValue) -> {
+                    updateAlarmClockEnabledPreferenceIcon(preference, newValue);
+                    return true;
+                });
+
+                setEditorSummary(alarmClockItemPref, getPreferenceString(alarmClockItemPref,
+                        ""));
```
maniac103: Nit: The line wrap is slightly weird, should wrap after `alarmClockItemPref,` for better readability.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
```diff
@@ -322,6 +344,25 @@ protected void updateAndInitPreferences() {
...
             }
 
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+                Log.d(TAG, "Removing alarm clock prefs");
+                getPreferenceScreen().removePreference(alarmClockPrefCat);
+            } else {
+                updateAlarmClockEnabledPreferenceIcon(alarmClockEnabledPref,
+                        getPreferenceBool(alarmClockEnabledPref, false));
+                alarmClockEnabledPref.setOnPreferenceChangeListener((preference, newValue) -> {
+                    updateAlarmClockEnabledPreferenceIcon(preference, newValue);
```
maniac103: Also start the worker here if `newValue` is true (and the item text is non-empty)?

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
```diff
@@ -322,6 +344,25 @@ protected void updateAndInitPreferences() {
...
+                    return true;
+                });
+
+                setEditorSummary(alarmClockItemPref, getPreferenceString(alarmClockItemPref,
+                        ""));
+                alarmClockItemPref.setOnPreferenceChangeListener((preference, newValue) -> {
+                    setEditorSummary(preference, newValue);
+                    return true;
```
maniac103: As above.

mobile/src/main/res/values/strings.xml
```diff
@@ -122,6 +122,22 @@
...
+    <string name="settings_alarm_clock">Send alarm clock to openHAB</string>
+    <string name="settings_alarm_clock_summary">Sends alarm clock time to a configured item</string>
+    <string name="settings_alarm_clock_item">Item</string>
+    <string name="waiting_for_network_to_send_alarm_clock">Waiting for network to send alarm clock</string>
+    <string name="notification_channel_background">Background</string>
+    <string name="notification_channel_background_description">Used to display pending background tasks</string>
+    <string name="notification_channel_background_error">Error in background tasks</string>
+    <string name="notification_channel_background_error_description">Used to notify about errors in background tasks</string>
+    <string name="error_sending_alarm_clock">Error sending alarm clock. Got HTTP error %d</string>
+    <string name="error_sending_alarm_clock_retry">Error sending alarm clock, but it\'ll be retried later</string>
```
maniac103: This text cries for 'I should not be a notification` ;-)

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
mueller-ma: IMO `false` is the better default for getBoolean, but otherwise I'll use it.

mobile/src/main/AndroidManifest.xml
mueller-ma: Both receivers might be doing different things in the future, e.g. the LocaleChange one could recreate the cloud notification channels, so I want to keep them as two receivers. Also the BackgroundTaskRetryBroadcastReceiver isn't exported.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
mueller-ma: Yes, I'll remove all static imports for functions.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
mueller-ma: When I added this I thought at the following scenario: A user accidentally disables network, sets up alarm and will never be notified, because Android doesn't run the worker at all.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedWorker.java
mueller-ma: Will do

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedWorker.java
mueller-ma: This needs documentation and a sample rule to parse the time.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedWorker.java
mueller-ma: Will do

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
mueller-ma: Will do

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
mueller-ma: Will do

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
mueller-ma: Will do

mobile/src/main/res/values/strings.xml
mueller-ma: Same reason as above.

mobile/src/main/AndroidManifest.xml
maniac103: Exported is a fair point, but LocaleChangedReceiver and AlarmChangedReceiver could be unified, couldn't they? I guess I just don't like the introduction of 3 (!) manifest registered receivers for such a small feature.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+    }
+
+    public static void startAlarmChangedWorker(Context context) {
+        final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+        startAlarmChangedWorker(context, prefs.getString(PREFERENCE_ALARM_CLOCK_ITEM, ""));
+    }
+
+    public static void startAlarmChangedWorker(Context context, String itemName) {
```
maniac103: Maybe pass in the enabled flag here as well and rename the function appropriately?
```
public static void startAlarmChangedWorkerIfNeeded(Context context, boolean enabled, String itemName) {
    if (!enabled || TextUtils.isEmpty(itemName.trim()) {
        return;
    }
    [...]
}

That way you don't need to check the enabled flag in all calling places.

File: mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java [hunk 14]
```diff
@@ -0,0 +1,93 @@
...
+        final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+
+        Log.d(TAG, "Cancel previous worker");
+        workManager.cancelAllWorkByTag(WORKER_TAG_SEND_ALARM_CLOCK);
+
+        if (TextUtils.isEmpty(itemName.trim())) {
+            Log.d(TAG, "Empty item name");
+            Notification notification = BackgroundUtils.makeBackgroundNotification(context,
+                    R.string.error_sending_alarm_clock_item_empty,
+                    R.drawable.ic_alarm_grey_24dp, true, null);
```
maniac103: Instead of this notification, I'd suggest updating the item name pref summary to reflect the fact that nothing is sent while it's empty.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
maniac103: OK, makes sense.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedWorker.java
```diff
@@ -0,0 +1,130 @@
...
+            failureReason = null;
+        } catch (ConnectionException e) {
+            Log.e(TAG, "ConnectionException", e);
+            connection = null;
+            failureReason = e;
+        }
+
+        if (connection == null) {
+            Log.e(TAG, "Got no connection "
+                    + (failureReason != null ? failureReason.getMessage() : null));
```
maniac103: This should be equivalent: `Log.e(TAG, "Got no connection: " + failureReason);`

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
```diff
@@ -0,0 +1,169 @@
...
+     */
+    public static void createNotificationChannels(Context context) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            NotificationManager notificationManager =
+                    context.getSystemService(NotificationManager.class);
+
+            String name = context.getString(R.string.notification_channel_background);
+            String description =
+                    context.getString(R.string.notification_channel_background_description);
+            int importance = NotificationManager.IMPORTANCE_LOW;
```
maniac103: IMPORTANCE_LOW or IMPORTANCE_MIN? IOW, do we need a status bar icon here?

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
```diff
@@ -0,0 +1,169 @@
...
+                .setContentText(msg)
+                .setAutoCancel(true)
+                .setContentIntent(intent)
+                .setColor(ContextCompat.getColor(context, R.color.openhab_orange))
+                .setCategory(isError ? NotificationCompat.CATEGORY_ERROR
+                        : NotificationCompat.CATEGORY_PROGRESS)
+                .setOngoing(!isError)
+                .setGroup(isError ? CHANNEL_ID_BACKGROUND_ERROR : CHANNEL_ID_BACKGROUND)
+                .setPriority(isError ? NotificationCompat.PRIORITY_DEFAULT
+                        : NotificationCompat.PRIORITY_LOW);
```
maniac103: As above.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
```diff
@@ -0,0 +1,169 @@
...
+                .setWhen(System.currentTimeMillis())
+                .setStyle(new NotificationCompat.BigTextStyle().bigText(msg))
+                .setContentText(msg)
+                .setAutoCancel(true)
+                .setContentIntent(intent)
+                .setColor(ContextCompat.getColor(context, R.color.openhab_orange))
+                .setCategory(isError ? NotificationCompat.CATEGORY_ERROR
+                        : NotificationCompat.CATEGORY_PROGRESS)
+                .setOngoing(!isError)
+                .setGroup(isError ? CHANNEL_ID_BACKGROUND_ERROR : CHANNEL_ID_BACKGROUND)
```
maniac103: I don't think we need group here, given we don't have a group summary.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
```diff
@@ -0,0 +1,169 @@
...
+                .setContentTitle(context.getString(R.string.app_name))
+                .setWhen(System.currentTimeMillis())
+                .setStyle(new NotificationCompat.BigTextStyle().bigText(msg))
+                .setContentText(msg)
+                .setAutoCancel(true)
+                .setContentIntent(intent)
+                .setColor(ContextCompat.getColor(context, R.color.openhab_orange))
+                .setCategory(isError ? NotificationCompat.CATEGORY_ERROR
+                        : NotificationCompat.CATEGORY_PROGRESS)
+                .setOngoing(!isError)
```
maniac103: IMHO error notifications aren't ongoing either, as
- they don't represent an ongoing action
- there's no reason to prevent them from being swiped away

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
```diff
@@ -0,0 +1,169 @@
...
+            @DrawableRes int iconId, boolean isError, NotificationCompat.Action action) {
+        Intent notificationIntent = new Intent(context, MainActivity.class);
+        notificationIntent.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP
+                | Intent.FLAG_ACTIVITY_SINGLE_TOP);
+        PendingIntent intent = PendingIntent.getActivity(context, 0,
+                notificationIntent, 0);
+
+        NotificationCompat.Builder nb = new NotificationCompat.Builder(context,
+                isError ? CHANNEL_ID_BACKGROUND_ERROR : CHANNEL_ID_BACKGROUND)
+                .setLargeIcon(drawableToBitmap(context.getResources().getDrawable(iconId)))
```
maniac103: I wouldn't do this, TBH. Large icon is for things like user images, I haven't seen it being used for icons. The fact you only can pass in bitmaps to it reflects that.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
```diff
@@ -0,0 +1,169 @@
...
+     * @return
+     */
+    public static NotificationCompat.Action makeRetryAction(Context context, int notificationId) {
+        Intent retryIntent = new Intent(context, BackgroundTaskRetryBroadcastReceiver.class);
+        retryIntent.putExtra(OH_EXTRA_NOTIFICATION_ID, notificationId);
+        PendingIntent retryPendingIntent =
+                PendingIntent.getBroadcast(context, 0, retryIntent, 0);
+        return new NotificationCompat.Action(
+                R.drawable.ic_refresh_grey_24dp,
+                context.getString(R.string.error_sending_alarm_clock_action_retry),
```
maniac103: The method is generic, but the string resource name isn't ... should probably adapt the latter.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
```diff
@@ -391,6 +446,18 @@ private void updateVibrationPreferenceIcon(Preference pref, Object newValue) {
...
         }
 
+        private void updateAlarmClockEnabledPreferenceIcon(Preference pref, Object newValue) {
+            boolean enabled = (boolean) newValue;
+            pref.setIcon(enabled ? R.drawable.ic_alarm_grey_24dp
+                    : R.drawable.ic_alarm_off_grey_24dp);
+        }
+
+        private void setEditorSummary(Preference pref, Object newValue) {
+            String itemName = (String) newValue;
```
maniac103: Generic method name, not-so-generic variable name ;-)

mobile/src/main/AndroidManifest.xml
mueller-ma: I merged both exported one

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
mueller-ma: Done, but the notification will still be shown if the alarm changes, but not anymore if the settings are changed.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedWorker.java
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
mueller-ma: I'm fine with IMPORTANCE_MIN. Actually IMPORTANCE_LOW was just for testing.

> do we need a status bar icon here?

No

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
mueller-ma: Errors aren't set to ongoing here. It's the "waiting for network" notification.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
mueller-ma: Sounds fair. I'll remove it

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
maniac103: Oops, right. The connection to the notification type is a bit on-obvious here though...maybe make this explicit through a parameter to the function?

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
maniac103: I don't understand ... what exactly do you mean by 'if the alarm changes'? IMHO the worker shouldn't start if the item name is empty, so how would this notification happen?

mueller-ma: Enabling the feature in the settings or change the item name triggers `startAlarmChangedWorker()`, but won't show any notification if the item name is empty.
Changing the alarm clock time also triggers `startAlarmChangedWorker()` and will show a notification if the item name is empty. IMO a user can overlook the item name preference summary.
In both cases the worker isn't started.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
mueller-ma: I would like keep the function signature as small as possible. If I add everything where `isError` is used to the signature, it'll add a lot of duplicate code to the function's call.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
maniac103: I don't think we need to optimize on 'user overlooked item name pref' ... after all, where is the alarm time uploaded to if it's not set? I think we can assume the user knows about the feature's implications when enabling it.

Alternative: make a custom pref with a dialog that includes both switch and editor.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
maniac103: Do you want to create a generic framework or not? ;-) If it's supposed to be generic it must not cater to specific behavior IMHO.
At the least the Javadoc needs to point out that isError == false means the notification reason is expected to be an ongoing process.

docs/USAGE.md
```diff
@@ -55,6 +56,35 @@ There are a number of strategies available to provide [secure remote access]({{b
 
 If you want to use openHAB Android on a wall mounted tablet, go to settings and select `Disable display timer` and `Fullscreen`.
 
+## Alarm Clock
```
mueller-ma: @Confectrician Do you want to have a look at the docs change?

cribskip:
Thanks for the idea and code! Please add a setting for the alarm clock item to be configurable thus to allow multiple devices /persons alarm to be distinguished.

 thank you!

mueller-ma:
> Please add a setting for the alarm clock item to be configurable thus to allow multiple devices /persons alarm to be distinguished.

Already done :)

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedReceiver.java
mueller-ma: I want to keep the settings for now as is and move them to a new page when more intents can be send to the server, e.g. receiving a phone call.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -254,4 +255,22 @@ public static void applySwipeLayoutColors(SwipeRefreshLayout swipeLayout,
         }
         swipeLayout.setColorSchemeColors(colors);
     }
+
+    public static long[] getVibrationPattern(Context context) {
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+        String vibration =
+                prefs.getString(Constants.PREFERENCE_NOTIFICATION_VIBRATION, "");
```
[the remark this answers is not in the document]
maniac103: Why not? I surely hope we're playing neither a sound nor vibration for the background action notifications.

docs/USAGE.md
```diff
@@ -55,6 +56,35 @@ There are a number of strategies available to provide [secure remote access]({{b
...
+
+The openHAB app will send the next wake-up time from your alarm clock app to the server. Please configure an Item name in the settings.
```
maniac103: Probably should mention the format used for sending  (`The time is sent as a number containing the number of milliseconds since the epoch` or something like that)

docs/USAGE.md
```diff
@@ -55,6 +56,35 @@ There are a number of strategies available to provide [secure remote access]({{b
...
+when
+    Time cron "*/10 * * * * ?"
+then
+    if (AndroidAlarm.state as Number == 0) {
+        // Alarm is turned off
+        return
+    }
+    val diff = AndroidAlarm.state as Number - now().millis
+    if (diff <= 15000) {
+        // Turn on stuff
```
maniac103: What does this comment mean? It's a bit unclear what exactly the if condition is supposed to mean, at least if you don't know the implementation)

mobile/src/main/java/org/openhab/habdroid/util/Util.java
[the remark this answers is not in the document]
mueller-ma: I think for errors we should play a sound.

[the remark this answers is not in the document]
maniac103: Looks like we agree to disagree then. IMHO even a failure here doesn't need immediate attention.

mobile/src/main/res/values/strings.xml
```diff
@@ -124,6 +124,25 @@
     <string name="info_not_set">Not set</string>
     <string name="empty_page">This page doesn\'t contain any widgets that are set to be visible</string>
 
+    <!-- Background tasks -->
+    <string name="alarm_clock">Alarm clock</string>
+    <string name="settings_alarm_clock">Send alarm clock to openHAB</string>
+    <string name="settings_alarm_clock_summary">Sends alarm clock time to a configured item</string>
+    <string name="settings_alarm_clock_howto_summary">Tap here to get some hints and explanations on how to set up SSL client certificate authentication</string>
```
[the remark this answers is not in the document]
mueller-ma: Right, I'll correct this

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
```diff
@@ -0,0 +1,191 @@
...
+                .setOngoing(!isError)
+                .setPriority(isError ? NotificationCompat.PRIORITY_DEFAULT
+                        : NotificationCompat.PRIORITY_MIN);
+
+        if (isError) {
+            SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+            nb.setLights(ContextCompat.getColor(context, R.color.openhab_orange),
+                    3000, 3000)
+                    .setSound(Uri.parse(prefs.getString(Constants.PREFERENCE_TONE, "")))
+                    .setVibrate(Util.getVibrationPattern(context));
```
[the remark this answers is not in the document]
mueller-ma: I'll split up `isError` in `isOngoing`, `hasSound` and `hasHighImportance` and set the alarm clock error notifications to not ongoing, no sound, but high importance. I want to keep the ability to show notifications with sound for later use, e.g. monitoring of a connection.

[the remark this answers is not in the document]
maniac103: Sounds fine.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
```diff
@@ -0,0 +1,191 @@
...
+     */
+    public static void createNotificationChannels(Context context) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            NotificationManager notificationManager =
+                    context.getSystemService(NotificationManager.class);
+
+            String name = context.getString(R.string.notification_channel_background);
+            String description =
+                    context.getString(R.string.notification_channel_background_description);
+            int importance = NotificationManager.IMPORTANCE_MIN;
```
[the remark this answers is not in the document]
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedWorker.java
```diff
@@ -0,0 +1,99 @@
...
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    @Override
+    public Result doWork() {
+        Context context = getApplicationContext();
+        NotificationManager nm =
+                (NotificationManager) context.getSystemService(NOTIFICATION_SERVICE);
+
+        Connection connection;
+        ConnectionException failureReason;
+        try {
```
maniac103: Should call `ConnectionFactory.waitForInitialization()` first, I think.

docs/USAGE.md
```diff
@@ -55,6 +56,35 @@ There are a number of strategies available to provide [secure remote access]({{b
 
 If you want to use openHAB Android on a wall mounted tablet, go to settings and select `Disable display timer` and `Fullscreen`.
 
+## Alarm Clock
+
+The openHAB app will send the next wake-up time from your alarm clock app to the server. Please configure an Item name in the settings.
+
+Example item definition:
+```
+Number AndroidAlarm
```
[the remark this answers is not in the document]
mueller-ma: I didn't get it working with a DateTime item.

docs/USAGE.md
```diff
@@ -55,6 +56,36 @@ There are a number of strategies available to provide [secure remote access]({{b
 
 If you want to use openHAB Android on a wall mounted tablet, go to settings and select `Disable display timer` and `Fullscreen`.
 
+## Alarm Clock
+
+The openHAB app will send the next wake-up time from your alarm clock app to the server. The time is sent as a number containing the number of milliseconds since the epoch. The Item name's default is `AndroidAlarmClock`, but you can change it in the settings.
```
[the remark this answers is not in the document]
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
```diff
@@ -0,0 +1,191 @@
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            NotificationManager notificationManager =
+                    context.getSystemService(NotificationManager.class);
+
+            String name = context.getString(R.string.notification_channel_background);
+            String description =
+                    context.getString(R.string.notification_channel_background_description);
+            NotificationChannel channel =
+                    new NotificationChannel(CHANNEL_ID_BACKGROUND, name,
+                            NotificationManager.IMPORTANCE_MIN);
```
maniac103: Have to agree to Stickler here, better format like line 67.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundUtils.java
```diff
@@ -0,0 +1,191 @@
...
+                .setCategory(isError ? NotificationCompat.CATEGORY_ERROR
+                        : NotificationCompat.CATEGORY_PROGRESS)
+                .setOngoing(!isOngoing)
+                .setPriority(isError ? NotificationCompat.PRIORITY_DEFAULT
+                        : NotificationCompat.PRIORITY_MIN);
+
+        if (hasSound) {
+            SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+            nb.setLights(ContextCompat.getColor(context, R.color.openhab_orange),
+                    3000, 3000)
```
maniac103: Does this really not fit on one line?

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
```diff
@@ -322,13 +346,39 @@ protected void updateAndInitPreferences() {
...
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+                Log.d(TAG, "Removing alarm clock prefs");
+                getPreferenceScreen().removePreference(alarmClockPrefCat);
+            } else {
+                updateAlarmClockEnabledPreferenceIcon(alarmClockEnabledPref,
+                        getPreferenceBool(alarmClockEnabledPref, false));
+                alarmClockEnabledPref.setOnPreferenceChangeListener((preference, newValue) -> {
+                    updateAlarmClockEnabledPreferenceIcon(preference, newValue);
+                    BackgroundTasksBroadcastReceiver.startAlarmChangedWorker(getActivity(),
+                            getPreferenceString(PREFERENCE_ALARM_CLOCK_ITEM, ""),
```
maniac103: The default value specified here doesn't match the `android:defaultValue` for that pref in XML.

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedWorker.java
```diff
@@ -0,0 +1,101 @@
...
+        AlarmManager alarmManager = (AlarmManager) context.getSystemService(Context.ALARM_SERVICE);
+        String nextAlarm;
+        AlarmManager.AlarmClockInfo alarmClockInfo = alarmManager.getNextAlarmClock();
+        if (alarmClockInfo == null) {
+            nextAlarm = "0";
+        } else {
+            nextAlarm = String.valueOf(alarmClockInfo.getTriggerTime());
+        }
+        String item = PreferenceManager.getDefaultSharedPreferences(context)
+                .getString(PREFERENCE_ALARM_CLOCK_ITEM, "");
```
maniac103: As below, should specify correct default value.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksBroadcastReceiver.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+        } else if (Intent.ACTION_LOCALE_CHANGED.equals(intent.getAction())) {
+            Log.d(TAG, "Locale changed, recreate notification channels");
+            BackgroundUtils.createNotificationChannels(context);
+        }
+    }
+
+    public static void startAlarmChangedWorker(Context context) {
+        final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+        startAlarmChangedWorker(context, prefs.getString(PREFERENCE_ALARM_CLOCK_ITEM, ""),
```
maniac103: Should specify correct default value.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksBroadcastReceiver.java
```diff
@@ -0,0 +1,107 @@
...
+                prefs.getBoolean(PREFERENCE_ALARM_CLOCK_ENABLED, false), true);
+    }
+
+    public static void startAlarmChangedWorker(Context context, String itemName, boolean enabled,
+            boolean showItemEmptyNotification) {
```
maniac103: How about we kill this parameter (and notification) and just prevent the user from clearing the item name by not allowing pref saving in that case?

Requires subclassing EditTextPreference, adding a `TextWatcher` to its EditText and graying out the OK button in that case - `UrlInputPreference` gives an example of that. I'm still in favor of a combined pref, though ;-) But I guess we can refactor that in a separate PR.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
mueller-ma: The pref is saved when the app is started: https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java#L174. So it can never be non-existent and the defValue is never used, right?

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksBroadcastReceiver.java
mueller-ma: > prevent the user from clearing the item name by not allowing pref saving in that case?

Yes, that's a good idea. I should also catch whitespace and newline from the item name.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
maniac103: TIL I learned about that API ... I'll take a look into its implementation tomorrow.

maniac103: As `readAgain` is set to false there, I don't think it covers any prefs that were added after the user opened the app for the first time. If it's set to true I can confirm the default values are put into the shared prefs map, so yes, defValue is not used in that case. Setting it to true will probably slightly delay app startup, though...

maniac103: Maybe run that call with `readAgain` set to true in `OnUpdateBroadcastReceiver.updateComparableVersion()` instead if the version stored in prefs doesn't match `BuildConfig.VERSION_CODE`?

mueller-ma: I installed a version without this patch, started the app, applied this patch and started the app again. The default value will be set as soon as I open the preference activity. W/o opening the prefs the user cannot enable this feature, so using `""` as default seems save to me.

maniac103: > The default value will be set as soon as I open the preference activity

How exactly did you check/verify that?

maniac103: OK, fair enough. I still think I'd prefer doing it explicitly on update, but I can also make a separate PR for that.

mueller-ma: Yes, it should also load the local and remote settings.

mueller-ma:
Should be fine now.

mobile/src/main/java/org/openhab/habdroid/ui/widget/HelpIconShowingPrefernceUtil.java
```diff
@@ -0,0 +1,32 @@
...
+
+import android.content.Context;
+import android.content.Intent;
+import android.net.Uri;
+import android.view.View;
+import android.widget.ImageView;
+import androidx.annotation.StringRes;
+import androidx.appcompat.widget.TooltipCompat;
+
+public class HelpIconShowingPrefernceUtil {
```
maniac103: Typo in class name

mobile/build.gradle
```diff
@@ -102,6 +102,7 @@ dependencies {
     implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
     implementation 'com.google.android.material:material:1.0.0'
     implementation 'androidx.multidex:multidex:2.0.1'
+    implementation 'android.arch.work:work-runtime:1.0.0-alpha12'
```
maniac103: There's a 1.0.0-beta02 now, but we can also leave that up to dependabot and/or update once it's final.

mobile/src/main/res/values/strings.xml
```diff
@@ -124,6 +124,24 @@
...
+    <string name="settings_alarm_clock_howto_url" translatable="false">https://www.openhab.org/docs/apps/android.html#alarm-clock</string>
+    <string name="settings_alarm_clock_item">Item</string>
+    <string name="waiting_for_network_to_send_alarm_clock">Waiting for network to send alarm clock</string>
+    <string name="notification_channel_background">Background</string>
+    <string name="notification_channel_background_description">Used to display pending background tasks</string>
+    <string name="notification_channel_background_error">Error in background tasks</string>
+    <string name="notification_channel_background_error_description">Used to notify about errors in background tasks</string>
+    <string name="error_sending_alarm_clock">Error sending alarm clock. Got HTTP error %d</string>
+    <string name="error_sending_alarm_clock_retry">Error sending alarm clock, but it\'ll be retried later</string>
+    <string name="error_sending_alarm_clock_item_empty">Please enter a valid Item name</string>
```
maniac103: Item -> item?

maniac103:
What just occurred to me:

> I think we should at least make the change to pass item name into the worker instead of fetching it again there.

If we also pass in the value (here: alarm clock setting as string), the worker becomes a totally generic 'upload item value' worker ... maybe a worthwhile goal?

mobile/build.gradle
mueller-ma: Let the bot do that :)

mobile/src/main/res/values/strings.xml
mueller-ma: Item should be written this way. I think we should discuss this, but rather in the forum and with @ThomDietrich. Similar issue: https://community.openhab.org/t/translation-of-openhab-terms/59818

mobile/src/main/java/org/openhab/habdroid/background/AlarmChangedWorker.java
```diff
@@ -0,0 +1,101 @@
...
+
+        if (result.isSuccessful()) {
+            Log.d(TAG, "Alarm clock successfully sent: " + nextAlarm);
+            nm.cancel(NOTIFICATION_TAG_BACKGROUND, NOTIFICATION_ID_SEND_ALARM_CLOCK);
+            return Result.success();
+        } else {
+            Log.e(TAG, "Error sending alarm clock. Got HTTP error "
+                    + result.statusCode, result.error);
+
+            return BackgroundUtils.retryOrFail(
```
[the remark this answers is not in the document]
mueller-ma: The device could be connected to a wifi that has a webserver at the same ip/dns than the openHAB server in the home wifi. The app connects to this server, but gets a 404, because the rest api isn't present at all.

mueller-ma:
> If we also pass in the value (here: alarm clock setting as string), the worker becomes a totally generic 'upload item value' worker ... maybe a worthwhile goal?

Yes, it is.

> I've refactored the pref storage so enabled flag + item name is stored in one pref. Doing so helps in adding additional similar prefs later.

Thinking about this issue: We could allow the user to add a custom prefix, e.g. device hostname, username, etc., and have a fixed sufix (e.g. AlarmClock). IMO it's bad UI if the user has to type a lot of item names in his phone. It's easier to copy/paste the server-side items and rules than editing all item names on multiple devices.
This way we can add one text pref for the prefix and simple switch prefs for the features.

maniac103:
I agree about the UI issue, but that could also be solved by adding another pref ('Auto-assign item names'). I'm not sure whether we really should put a hard item name restriction here...but dropping the item name complexity indeed also has its merits.

mueller-ma:
A compromise would be to set good default values for prefix (fetch device model or name of user account) and the item name (like in this PR).

maniac103:
> > If we also pass in the value (here: alarm clock setting as string), the worker becomes a totally generic 'upload item value' worker ... maybe a worthwhile goal?
> Yes, it is.

Please pull :-) https://github.com/maniac103/openhab.android/commit/ff3d6a6f388e406c055caab9ab25d9b1801dca0d

mueller-ma:
I'll look into this soon. One small note: I'd like to keep `makeRetryAction()` and `retryOrFail()` in BackgroundUtils, so other non-item-send workers can use them, e.g. monitoring.

IMO the BackgroundTaskRetryBroadcastReceiver needs some rework. Imagine we have 10 different values send to the server and all fail. In this case having 10 ungrouped notifications and 10 retry buttons are really bad ux.

maniac103:
That's a good point. Sounds like we need some different approach then ... perhaps 'some' kind of queue of things to be updated + their value; and one worker which then empties the queue, so we have a single notification + a single success/failure result.

Not sure what the best way to get there is though ... can one get the worker params (and thus input data) for an already scheduled worker? If yes, when we notice we need to update something, we can fetch the data, cancel the worker, amend to the data and reschedule the worker.

maniac103:
> so other non-item-send workers can use them, e.g. monitoring.

What kind of monitoring worker do you envision though? I can't think of any potential workers that don't end up sending some item update to the server.

mueller-ma:
> What kind of monitoring worker do you envision though?

A periodic worker that makes a simple http get call to `/rest/` and shows an error if it fails for a few times. Since we cannot determine if a local server should be available in a specific wifi (or at least we don't check that right now), it's only available for remote connections.

maniac103:
Maybe I'm missing something, but I can't see the value of that at the moment, as I can't see what the user is going to do with that information...it'll be a 'VPN is down' notification essentially, as myopenhab (hopefully) will rarely be unreachable.

As for local connection detection, we totally should improve that, both by regularly retrying connection and by checking differently (querying rest endpoint instead of just opening socket)

mobile/src/main/res/drawable/ic_none.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="24dp"
+    android:viewportHeight="24.0" android:viewportWidth="24.0"
+    android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:pathData=""/>
```
maniac103:
```
01-31 11:14:19.230  6991  6991 E AndroidRuntime: FATAL EXCEPTION: main
01-31 11:14:19.230  6991  6991 E AndroidRuntime: Process: org.openhab.habdroid.beta, PID: 6991
01-31 11:14:19.230  6991  6991 E AndroidRuntime: android.content.res.Resources$NotFoundException: Drawable org.openhab.habdroid.beta:drawable/ic_none with resource ID #0x7f08009d
01-31 11:14:19.230  6991  6991 E AndroidRuntime: Caused by: android.content.res.Resources$NotFoundException: File res/drawable-anydpi-v21/ic_none.xml from drawable resource ID #0x7f08009d
01-31 11:14:19.230  6991  6991 E AndroidRuntime:        at android.content.res.ResourcesImpl.loadDrawableForCookie(ResourcesImpl.java:725)
[...]
01-31 11:14:19.230  6991  6991 E AndroidRuntime: Caused by: java.lang.IllegalArgumentException: Path string cannot be empty.
01-31 11:14:19.230  6991  6991 E AndroidRuntime:        at android.util.PathParser.nCreatePathDataFromString(Native Method)
01-31 11:14:19.230  6991  6991 E AndroidRuntime:        at android.util.PathParser.-wrap1(PathParser.java)
01-31 11:14:19.230  6991  6991 E AndroidRuntime:        at android.util.PathParser$PathData.<init>(PathParser.java:74)
01-31 11:14:19.230  6991  6991 E AndroidRuntime:        at android.graphics.drawable.VectorDrawable$VFullPath.updateStateFromTypedArray(VectorDrawable.java:1556)
[...]
```

mueller-ma: Fixed by adding an icon. I guess some prefs look better without an icon, so the icons can point the user to the important prefs. Using `android:iconSpaceReserved="true"` didn't work for me, I guess because we don't use PreferenceFragmentCompat. But changing that is out of scope of this PR.

mueller-ma:
> myopenhab (hopefully) will rarely be unreachable.

Yes, it was down a few times in the past and users often request some kind of notification if that happens.

mueller-ma:
> can one get the worker params (and thus input data) for an already scheduled worker?

I don't think it'll be so easy. If the worker fails, we cannot cancel it and get the input data from it (I guess, haven't checked if you really cannot get the input data of failed worker).
We can create a new data holding class "ItemUpdate" containing value and item name. Then it'll be written as json to shared prefs, the worker can read it and remove the successful updates.

mueller-ma:
> We can create a new data holding class "ItemUpdate" containing value and item name. Then it'll be written as json to shared prefs, the worker can read it and remove the successful updates.

I guess a HashMap with index equals item name is better suited. @maniac103 WDYT?

maniac103:
I don't like storing non-pref, non-persistent values in SharedPreferences :-/
I'll have a look at what's possible with workers as soon as I got some free time.

maniac103:
Alternative idea: subscribe for work updates and manage notification from there. Something like this: https://gist.github.com/maniac103/bdc5ba6ea7759c68f11d0e38b1e1a5e0 ... the log line in `onChanged()` looks promising; question is just when to prune old tasks.

mueller-ma:
Your first suggestion sounds good to me.
How do we handle events, where we can get the value only when the event happens, e.g. from an extra from an intent?

maniac103:
With the second suggestion (which is what I'd prefer after playing with both variants) it would just work ;-)

Problem with variant 1 when I played around with it was the retry handling: I found no way to only retry the uploads that failed (as opposed to retrying all of them).

mueller-ma:
I would remove the tags from the set as soon as they succeed.

maniac103:
The thing is, AFAICT we can't remove single work items from the DB. We can only prune all finished (succeeded/failed/canceled) work altogether. Writing some logic to only take the latest item into account is trivial though.

mueller-ma:
Indeed, your second solution is much cleaner than the first one. But I start thinking that we overcomplicate this issue. Can't we just show one failure notification (e.g. "The following actions have failed: Alarm clock") and replace the "Retry" notification action with "Retry all". Also add a second option "Show log", that opens LogActivity with a filtered log, so only worker related entries are shown.

maniac103:
> But I start thinking that we overcomplicate this issue

I have that thought since quite some time ;-)
But I actually think my second suggestion is beneficial, if only to the code structure. I've completed its implementation, but only did some very limited testing so far: https://github.com/maniac103/openhab.android/commit/7c096bda32854b9b5404594c7442346c82e6d887 ... please let me know what you think of it.

mueller-ma:
I tested and added your commit, but I didn't test having two or more background tasks so far.
The only issue I've seen: If a user has only a local connection configured, the item update will fail if the user isn't on his local network.

maniac103:
> If a user has only a local connection configured, the item update will fail if the user isn't on his local network.

That makes sense, but I don't really see a way around that. We could specify we want a WiFi network when scheduling the worker if there's no remote connection configured, but the user still might be on a different network (e.g. work)...might be a bit better nevertheless though?

mueller-ma:
I don't see an easy way to tell the worker that we need wifi (or vpn) network: https://developer.android.com/reference/androidx/work/Constraints.Builder.html#setRequiredNetworkType(androidx.work.NetworkType) and https://developer.android.com/reference/androidx/work/NetworkType.html

maniac103:
Yes, that's right. VPN is impossible; for WiFi the 'unmetered' network type comes close.

mobile/src/main/java/org/openhab/habdroid/background/NotificationUpdateObserver.java
```diff
@@ -111,13 +111,17 @@ public void onChanged(List<WorkInfo> workInfos) {
...
+                        data.getBoolean(ItemUpdateWorker.OUTPUT_DATA_IS_DEMO, false);
                 final int httpStatus =
                         data.getInt(ItemUpdateWorker.OUTPUT_DATA_HTTP_STATUS, 0);
 
                 retryInfos.add(new BackgroundTasksManager.RetryInfo(entry.first, itemName, value));
                 if (hadConnection) {
                     errors.add(mContext.getString(
                             R.string.item_update_http_error, itemName, httpStatus));
+                } else if (isDemo) {
+                    errors.add(mContext.getString(R.string.item_update_connection_is_demo));
```
maniac103: I'm not sure I like this. This feels like it should be handled on preferences side, by
- disabling the upload category in demo mode
- calling `scheduleWorker()` when demo mode toggles and
- checking for demo mode being enabled there and returning early (like for disabled setting) if it is

mueller-ma: I implemented it this way, but if there is an error notification, the user can click retry and it's going to be send to the demo server.

maniac103: Hmm, I guess I don't properly understand the reasoning/use case then. Why would we prevent sending updates to the demo server, when it's (seemingly?) OK to send them? I thought we either shouldn't do it at all or unconditionally allow it.

maniac103: That's a totally fair point ... so why not prevent that upload in demo mode altogether?

mueller-ma: I pushed the implementation you suggested. If a user is connected to a server, there is an error sending some information, the user turns on demo mode and click on retry, the info is send to the demo server.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
```diff
@@ -66,6 +66,10 @@ public void onReceive(Context context, Intent intent) {
 
     public static void scheduleWorker(Context context, String key) {
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+        if (Constants.PREFERENCE_DEMOMODE.equals(key)) {
```
maniac103: That wasn't exactly what I meant. I meant to check `prefs.getBoolean(Constants.PREFERENCE_DEMOMODE, false)` here and call `cancelAllWorkByTag(key)` here if it's true, and additionally...

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
```diff
@@ -288,6 +289,13 @@ protected void updateAndInitPreferences() {
             updateVibrationPreferenceIcon(vibrationPref,
                     prefs.getString(Constants.PREFERENCE_NOTIFICATION_VIBRATION, ""));
 
+            demoModePref.setOnPreferenceChangeListener((preference, newValue) -> {
+                if ((boolean) newValue) {
+                    scheduleWorker(preference.getKey());
```
maniac103: ... call `scheduleWorker(Constants.PREFERENCE_ALARM_CLOCK)` here irregardless of the new pref value.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
mueller-ma: The preference is set to `false` after `setOnPreferenceChangeListener()` returned `true`. So calling `scheduleWorker()` and fetching the pref there returns the old value.

maniac103: No, because `scheduleWorker()` in the prefs fragment schedules the call via a Handler for precisely that reason ;-)

mueller-ma: Done, but the issue still exists.

maniac103: But how? I'd expect this to happen:
- demo mode is enabled
- work is canceled through `scheduleWorker()`
- respective work items have their state change from FAILED to CANCELED
- change is picked up by `NotificationUpdateObserver`
- no item in FAILED state anymore -> notification is removed

What am I missing there?

maniac103: Answer to self: the FAILED -> CANCELED transition doesn't happen, because FAILED is a terminal state (only ENQUEUED, BLOCKED and RUNNING can transition to FAILED)

Fixed by pruning the worker DB in that case: https://github.com/maniac103/openhab.android/commit/8c05ad256db0febc0744ae0e2a4b1ceee680969b

mueller-ma: Thanks. Should be fine to merge now.

PR #1145, opened 2018-12-13
Show hint if a page doesn't contain visible elements

### LINKED ISSUE #1062: Show a hint on a subpage containing no elements
I have a subpage that contains items that are visible under some conditions. It may happen that no item is shown. Whenever this happens my first thought is: "Oh, the app didnt load the page." until I realise that nothing should be shown on this page.
I don't think that I'm the only one who has this thought, so maybe we should add some hint:
1. Just display a hint: "This page contains no elements"
2. Display a hint and a material like image, e.g. from Signal: https://github.com/signalapp/Signal-Android/blob/master/res/drawable-xxxhdpi/conversation_list_empty_state.png
3. Display a funny "Nothing to see here" meme: https://duckduckgo.com/?q=nothing+to+see+here+meme&t=ffab&iar=images&iax=images&ia=images

mueller-ma:
If a page contains no elements at all, openHAB refuses to load the
sitemap. So if a page is "empty" it means the items are not visible.

Closes #1062


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
```diff
@@ -127,10 +131,17 @@ public WidgetAdapter(Context context, Connection connection,
...
+        mRecyclerView = recyclerView;
+        mEmptyPageView = emptyPageView;
     }
 
     public void update(List<Widget> widgets, boolean forceFullUpdate) {
         boolean compatibleUpdate = true;
+        boolean emptyPage = widgets.size() == 0;
+
+        mRecyclerView.setVisibility(emptyPage ? View.GONE : View.VISIBLE);
+        mEmptyPageView.setVisibility(emptyPage ? View.VISIBLE : View.GONE);
```
maniac103: Should do this in the fragment, otherwise you have an inversion of control flow (fragment owns and controls  adapter, not the other way around). `update()` is called from fragment anyway.

mobile/src/main/res/values/strings.xml
```diff
@@ -226,4 +226,5 @@
     <string name="app_intro_skip_button">SKIP</string>
     <!-- Intro "DONE" button -->
     <string name="app_intro_done_button">DONE</string>
+    <string name="empty_page">This page contains widgets, that aren\'t visible due to their visibility configuration</string>
```
maniac103: Comma after widgets should be removed.
Alternative suggestion: `This page does not contain any widgets that are set to be visible`

PR #1152, opened 2018-12-20
Add voice widget w/o app launch shortcut

### LINKED ISSUE #646: [Voice Widget] Widget proposals
While working on the voice widget, I kept asking myself, why there's a 'start openhab' section. Sure, it clearly identifies the voice widget and enables the user beside the voice action to start the app.

For me, I 'm having the voice widget on the same screen as the app itself - so this is redundant.

What do you think of a variation of the voice widget just enabling the user to start the voice action. Per design, maybe just the microphone with an openhab logo in the background.

I'll post a PR the next few days outlining my proposal.

mueller-ma:
Closes #646

Also:
* Use high-res preview images
* Use high-res mic icon (old icon looks pixelated on older devices)
* Open app store to download Google app if it isn't present


maniac103:
Why not nuke the icon altogether? Personally, I never understood why it's there in the first place.

mueller-ma:
Some users might only have the widget and not the app icon on their homescreen to save some space. I personally don't use the widget, so I cannot really tell, but I would keep it.

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/VoiceWidget.java
```diff
@@ -0,0 +1,83 @@
...
+import android.widget.RemoteViews;
+import androidx.annotation.LayoutRes;
+
+import org.openhab.habdroid.R;
+import org.openhab.habdroid.core.VoiceService;
+
+/**
+ * Implementation of App Widget functionality.
+ */
+public class VoiceWidget extends AppWidgetProvider {
```
mueller-ma: @maniac103 Should this be abstract with two subclasses that both override setupOpenhabIcon() and getLayoutRes()?

PR #1157, opened 2018-12-29
[Bugfix] Fix lgtm alerts

frogman544:
Fix the lgtm alerts

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.java
@@ -19,13 +19,11 @@ import androidx.appcompat.app.AppCompatActivity;
 import androidx.appcompat.widget.Toolbar;
 
 import com.google.android.material.floatingactionbutton.FloatingActionButton;
-
 import org.openhab.habdroid.R;
 import org.openhab.habdroid.util.Constants;
 import org.openhab.habdroid.util.Util;
 
 import java.io.BufferedReader;
-import java.io.IOException;
 import java.io.InputStreamReader;
 
 import static org.openhab.habdroid.util.Util.getHostFromUrl;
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/de/duenndns/ssl/MemorizingTrustManager.java
```diff
@@ -418,7 +419,7 @@ private static boolean isPathException(Throwable e) {
     public void checkCertTrusted(X509Certificate[] chain, String authType, boolean isServer)
             throws CertificateException
     {
-        LOGGER.log(Level.FINE, "checkCertTrusted(" + chain + ", " + authType + ", " + isServer + ")");
+        LOGGER.log(Level.FINE, "checkCertTrusted(" + Arrays.toString(chain) + ", " + authType + ", " + isServer + ")");
```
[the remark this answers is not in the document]
mueller-ma: This issue should be fixed upstream: https://github.com/ge0rg/MemorizingTrustManager/issues/57
Using your solution produces a string that it too large to be shown in the log message.

mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.java
```diff
@@ -127,6 +129,7 @@ protected String doInBackground(Boolean... clear) {
                     logBuilder.append(line);
                     logBuilder.append(separator);
                 }
+                bufferedReader.close();
```
mueller-ma: The reader won't be closed if an exception is thrown. Please see https://lgtm.com/rules/10030075/ for the correct way to fix this issue.

frogman544: Thank you for the fix! I implemented as described in the rules.

mueller-ma:
I just learned about try-with-ressources which is way more readable: http://tutorials.jenkov.com/java-exception-handling/try-with-resources.html.
Please make sure your android studio doesn't change the import order.

mueller-ma:
@frogman544 I pushed some minor improvements to your branch. Thanks for your contribution.

PR #1167, opened 2019-01-11
Show shortcut to notification settings for Oreo and above

### LINKED ISSUE #1156: Custom notification ring tone
I see mention in the code that I should be able to select a custom ring tone but that option is not showing up in the preferences. I have enabled myopenhab, and I see successfully registered with gcm i the about, which appears to be a requirement for the custom notification option. I have tried the main openhab app in Google play and the beta version. Notifications show up correctly, I would just like to change the notification sound.

mueller-ma:
Closes #1156

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
@@ -333,13 +333,19 @@ public class PreferencesActivity extends AppCompatActivity {
                 getParent(fullscreenPreference).removePreference(fullscreenPreference);
             }
 
-            if (!CloudMessagingHelper.isSupported()
-                    || Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                Log.d(TAG, "Removing notification prefs");
+            if (CloudMessagingHelper.isSupported()) {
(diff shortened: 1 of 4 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
```diff
@@ -327,6 +339,9 @@ protected void updateAndInitPreferences() {
                 Log.d(TAG, "Removing notification prefs");
```
mueller-ma: #1137 needs to merged first, otherwise this condition is wrong.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
```diff
@@ -308,6 +311,14 @@ protected void updateAndInitPreferences() {
                 return true;
             });
 
+            ringtoneVibrationPref.setOnPreferenceClickListener(preference -> {
+                Intent i = new Intent(android.provider.Settings.ACTION_SETTINGS);
+                i.setAction(Settings.ACTION_APP_NOTIFICATION_SETTINGS);
+                i.putExtra(Settings.EXTRA_APP_PACKAGE, getContext().getApplicationInfo().packageName);
```
maniac103: Can avoid the overly long line by using `getContext().getPackageName()`

PR #1172, opened 2019-01-15
Improve HABPanel viewer

### LINKED ISSUE #1169: HABPanel is reset when device orientation is changed
### Actual behaviour
HABPanel's starting page is shown after the device orientation is changed.

### Expected behaviour
HABPanel should be on the same subpage as before.

Yes

#### Client
* App version : 2.5.6-beta

mueller-ma:
* Don't reset HABPanel on device rotation. Closes #1169
* Set the webview background to transparent. Avoids white flash when
opening HABPanel with a dark app theme.
* Make fragment reusable for other websites

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
@@ -12,6 +12,7 @@ package org.openhab.habdroid.ui.activity;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.graphics.Color;
 import android.graphics.PorterDuff;
 import android.graphics.drawable.Drawable;
 import android.net.wifi.WifiManager;
@@ -231,7 +232,8 @@ public abstract class ContentController implements PageConnectionHolderFragment.
     }
 
     public void showHabpanel() {
-        showTemporaryPage(HabpanelFragment.newInstance());
+        showTemporaryPage(FullScreenWebviewFragment.newInstance(R.string.habpanel_error,
+                "/habpanel/index.html", "/rest/events"));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -668,28 +670,52 @@ public void onClick(View view) {
...
+        static String mUrltoLoad;
+        static String mUrlForError;
 
-        public static HabpanelFragment newInstance() {
-            return new HabpanelFragment();
+        public static FullScreenWebviewFragment newInstance(@StringRes int errorMessage,
+                String urltoLoad, String urlForError) {
+            mErrorMessage = errorMessage;
+            mUrltoLoad = urltoLoad;
+            mUrlForError = urlForError;
```
maniac103: Those 3 should be put into fragment arguments and be non-static.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -737,9 +763,10 @@ public void onReceivedError(WebView view, int errorCode, String description,
                     updateViewVisibility(true, false);
                 }
             });
-            webView.getSettings().setDomStorageEnabled(true);
-            webView.getSettings().setJavaScriptEnabled(true);
-            webView.loadUrl(url);
+            mWebview.getSettings().setDomStorageEnabled(true);
+            mWebview.getSettings().setJavaScriptEnabled(true);
+            mWebview.loadUrl(url);
+            mWebview.setBackgroundColor(Color.TRANSPARENT);
```
maniac103: This should be possible to do in XML as well (`android:background=@null` I think)

mueller-ma:
`onBackPressed()` should be overrriden to navigate back in the HABPanel.

PR #1179, opened 2019-01-22
 Use different ids for each flavor in GCM registration 

mueller-ma:


mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -105,13 +107,15 @@ private void registerGcm(CloudConnection connection) throws IOException {
...
+        String deviceName = getDeviceName()
+                + (BuildConfig.FLAVOR.toLowerCase().contains("beta") ? " (Beta)" : "");
+        deviceName = URLEncoder.encode(deviceName, "UTF-8");
+        String deviceId = Settings.Secure.getString(getContentResolver(),
+                Settings.Secure.ANDROID_ID) + BuildConfig.FLAVOR;
 
         String regUrl = String.format(Locale.US,
                 "addAndroidRegistration?deviceId=%s&deviceModel=%s&regId=%s",
-                deviceId, deviceModel, token);
+                deviceId, deviceName, token);
```
maniac103: Could also place the `URLEncoder.encode(...)` here instead of the reassigning the variable.

mobile/src/full/java/org/openhab/habdroid/core/GcmRegistrationService.java
```diff
@@ -105,13 +107,15 @@ private void registerGcm(CloudConnection connection) throws IOException {
...
         String token = instanceId.getToken(connection.getMessagingSenderId(),
                 GoogleCloudMessaging.INSTANCE_ID_SCOPE, null);
-        String deviceModel = URLEncoder.encode(Build.MODEL, "UTF-8");
-        String deviceId =
-                Settings.Secure.getString(getContentResolver(), Settings.Secure.ANDROID_ID);
+        String deviceName = getDeviceName()
+                + (BuildConfig.FLAVOR.toLowerCase().contains("beta") ? " (Beta)" : "");
+        deviceName = URLEncoder.encode(deviceName, "UTF-8");
+        String deviceId = Settings.Secure.getString(getContentResolver(),
+                Settings.Secure.ANDROID_ID) + BuildConfig.FLAVOR;
```
maniac103: Doing so will invalidate existing registrations, is this intended? IMHO prepending 'beta:' or similar in the beta case should be sufficient.

PR #1191, opened 2019-02-02
Fix webviews to openHAB server with untrusted certificate

mueller-ma:
See #948

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/AnchorWebViewClient.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/AnchorWebViewClient.java
@@ -10,6 +10,7 @@
 
 package org.openhab.habdroid.ui;
 
+import android.content.Context;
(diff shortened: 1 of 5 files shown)


maniac103:
The use of shared prefs bothers me a bit here. Did you try calling isKnown() in mtm, passing in the failing cert?

mobile/src/main/java/org/openhab/habdroid/ui/AnchorWebViewClient.java
```diff
@@ -47,4 +60,40 @@ public void onPageFinished(WebView view, String url) {
             view.loadUrl("javascript:location.hash = '#" + mAnchor + "';");
         }
     }
+
+    public void onReceivedSslError(WebView view, SslErrorHandler handler, SslError error) {
```
maniac103: `@Override`?

PR #1195, opened 2019-02-05
Fix crash with MapView if item has no state

mueller-ma:


maniac103:
Why not pass the ParsedState into ParseLocation and make that one return null if the passed-in state is null:
`String location = state != null ? state.asString() : null;`
?

mueller-ma:
This fixes the crash, but we should think about a better representation for "no location" instead of showing a map without markers.
The follwing icon with the text "No position set"?
map-marker-off

PR #1201, opened 2019-02-07
Fix possible NPE

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
@@ -653,21 +653,22 @@ public class MainActivity extends AppCompatActivity implements
         Drawable defaultIcon = ContextCompat.getDrawable(this, R.drawable.ic_openhab_appicon_24dp);
         item.setIcon(applyDrawerIconTint(defaultIcon));
 
-        if (url != null) {
-            mConnection.getAsyncHttpClient().get(url,
-                    new AsyncHttpClient.BitmapResponseHandler(defaultIcon.getIntrinsicWidth()) {
-                @Override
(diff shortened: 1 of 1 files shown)


mueller-ma:
Stacktrace:
screenshot_20190206-222209

I'm not sure how to reproduce this since `loadSitemapIcon()` isn't called if `mServerProperties` is null. The only case I can imagine is when the device has been rotated and `mServerProperties` is restored, but the connection not fetched.

maniac103:
> I'm not sure how to reproduce this since `loadSitemapIcon()` isn't called if `mServerProperties` is null. The only case I can imagine is when the device has been rotated and `mServerProperties` is restored, but the connection not fetched.

In that case the connection must have changed during the rotation though!?
https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java#L229

I wonder whether this is a race condition where the connection went away during server property fetch?

I wonder whether we shouldn't check mConnection for being null in the `if` check in `loadSitemapIcon` instead? Somehow I feel we had a similar issue in the past, this feels somewhat familar to me...

mueller-ma:
> I wonder whether we shouldn't check mConnection for being null in the `if` check in `loadSitemapIcon` instead?

Like this: https://github.com/openhab/openhab-android/commit/53a8c3bc7f1878bc5bff1517954d77c8e88a4f11?

maniac103:
Yes. IIRC I shot it down back then (right?), but I now really think it's the best solution.

mueller-ma:
No, you didn't, but I thought it's better to have null checks for mConnection and mServerProperties in one place.

mueller-ma:
Stickler issues seem false-positive to me.

PR #1220, opened 2019-02-25
Fix NPE in colorpicker

mueller-ma:


mueller-ma:
> `float[] hsv = mBoundItem.state() != null ? mBoundItem.state().asHsv() : null;`

Well, it only calls .asHsv() only once, which does "a lot" of parsing and therefore is slower.

maniac103:
> Well, it only calls .asHsv() only once, which does "a lot" of parsing and therefore is slower.

I ... don't understand? Why is calling a 'slow' method only once slower?
(Besides that `.asHsv()` isn't really slow ... parsing happens at item construction time)

mueller-ma:
I still had the old getStateAsHsv() implementation in my mind when I wrote the comment :)

PR #1222, opened 2019-02-27
Fix ConnectionFactory.waitForInitialization() behavior.

maniac103:
Unlike stated by its Javadoc, waitForInitialization() didn't wait for
initialization in all cases. In particular, it didn't work when a
connectivity change happened while we were running in background and
some thread started waiting for initialization afterwards.
Ensure the guarantee given by waitForInitialization -
getUsableConnection() returning either a non-null connection or throwing
an exception - actually holds.


mueller-ma:
In #1137

PR #1232, opened 2019-03-03
Wait up to 2 seconds for connection to become available

### LINKED ISSUE #755: Discovery failed error is shown after network reconnect

mueller-ma:
Especially for local connections, like WiFi or maybe VPN, too, a
connection may not be "available" (~ connected to the internet) once it is
"connected". To allow a connection to do whatever is needed to do, wait up
to 2 seconds (in 10 200ms chunks) for the connection to become available.
If it is not available after this timeframe, the connection is returned as
"not reachable".

Fixes #755
Closes #780

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
@@ -104,19 +104,19 @@ public abstract class AbstractConnection implements Connection {
         for (int retries = 0; retries < 10; retries++) {
             try {
                 s.connect(socketAddress, 1000);
+                Log.d(TAG, "Socket connected (attempt  " + retries + ")");
+                return s;
             } catch (SocketTimeoutException e) {
-                Log.d(TAG, "Socket timeout at the " + retries + ". try.", e);
+                Log.d(TAG, "Socket timeout after " + retries + " retries");
                 return null;
             } catch (IOException e) {
-                Log.d(TAG, "Socket connection failed at the " + retries + ". try.", e);
+                Log.d(TAG, "Socket creation failed (attempt  " + retries + ")");
                 try {
                     Thread.sleep(200);
                 } catch (InterruptedException ignored) {
                     // ignored
                 }
-                continue;
             }
-            return s;
         }
         return null;
     }
--- a/mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
@@ -7,8 +7,10 @@ import okhttp3.OkHttpClient;
 import org.openhab.habdroid.util.AsyncHttpClient;
 import org.openhab.habdroid.util.SyncHttpClient;
 
+import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Socket;
+import java.net.SocketTimeoutException;
 import java.net.URL;
 
 public abstract class AbstractConnection implements Connection {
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
```diff
@@ -84,14 +86,38 @@ public boolean checkReachabilityInBackground() {
...
+            Log.d(TAG, e.getMessage());
             return false;
         }
     }
+
+    private Socket createConnectedSocket(InetSocketAddress socketAddress) {
+        Socket s = new Socket();
+        for (int retries = 0; retries < 10; retries++) {
+            try {
+                s.connect(socketAddress, 1000);
```
maniac103: https://github.com/openhab/openhab-android/pull/780/files#r178436704 still applies: could return s here and remove the `continue` / `return s` below.

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
```diff
@@ -84,14 +86,38 @@ public boolean checkReachabilityInBackground() {
...
+            } catch (SocketTimeoutException e) {
+                Log.d(TAG, "Socket timeout at the " + retries + ". try.", e);
```
maniac103: `5. try` is not valid English, as `5.` is a German construct. Could write `Socket timeout after 5 retries` here.

Also, I'm not sure whether printing the stack trace here (and below) is actually useful.

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
```diff
@@ -84,14 +86,38 @@ public boolean checkReachabilityInBackground() {
...
+                return null;
+            } catch (IOException e) {
+                Log.d(TAG, "Socket connection failed at the " + retries + ". try.", e);
```
maniac103: As above, but here one could write `Socket creation failed (attempt 5)`

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
mueller-ma: > Also, I'm not sure whether printing the stack trace here (and below) is actually useful.

No, I don't think so.

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
mueller-ma: Done

maniac103:
@mueller-ma Forgot to push the update? ;-)

mueller-ma:
Yes. Pushed it now.

mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
```diff
@@ -84,14 +86,38 @@ public boolean checkReachabilityInBackground() {
             } else if (url.getProtocol().equals("https") && checkPort == -1) {
                 checkPort = 443;
             }
-            Socket s = new Socket();
-            s.connect(new InetSocketAddress(url.getHost(), checkPort), 1000);
+            Socket s = createConnectedSocket(new InetSocketAddress(url.getHost(), checkPort));
+            if (s == null) {
+                return false;
+            }
             Log.d(TAG, "Socket connected");
```
maniac103: Can drop this message now that `createConnectedSocket` generates virtually the same message.

mueller-ma: Done.

PR #1237, opened 2019-03-05
Fix SSE cancel behavior.

### LINKED ISSUE #1228: Local access mode terribly slow after few pages opened
I am running server snapshot 1544 and Android app 2.5

App is almost unusable in local access mode. Page loading take several seconds and I can then see that icons take time to be loaded. Image cache seems to be no more used ?
I tried to clear the image cache but it does not resolve the problem.

The app is working normally (faster) when I am in remote access mode.

It could be relative to the most recent version of the server I am running. I will try to downgrade to an older version.

maniac103:
Don't flag failure to the respective connection handler if the SSE
onClosed() callback is called due to SSE being canceled instead of retry
exhaustion. Doing so will 'revive' canceled connection handlers, which
will cause all kinds of state tracking confusion.

Closes #1228


mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.java
```diff
@@ -545,7 +545,12 @@ public boolean onRetryError(ServerSentEvent sse,
 
             @Override
             public void onClosed(ServerSentEvent sse) {
-                mFailureCb.handleFailure();
+                // We're only interested in permanent failure here, not in callbacks we caused
+                // ourselves by calling close(), so check the retry count if it actually is
+                // exceeded.
+                if (mRetries >= MAX_RETRIES) {
```
maniac103: Could maybe also check for `mEventStream == sse` here.

PR #1238, opened 2019-03-05
Keep authentication header an implementation detail of HttpClient.

maniac103:
mAuthHeader is only public for unit testing (marked by the
VisibleForTesting annotation), so keep the usage of the member variable
private to the HttpClient class. Doing so also avoids code duplication.


maniac103:
As mentioned in https://github.com/openhab/openhab-android/pull/1231#pullrequestreview-209922653 ... I wasn't aware *why* mAuthHeader was accessible when I wrote this; with that knowledge I more strongly think keeping its usage private is better.

PR #1245, opened 2019-03-06
Clean up public methods needed for SSE in HttpClient.

maniac103:
Previously, there were two public methods in HttpClient that were needed
only by the SSE code. Since they always were used together anyway, we
can as well combine them into one and keep HttpClient public API more
consise.


maniac103:
Sorry for the back-and-forth with those methods, but the option of just creating the SSE class instance in HttpClient didn't occur to me until today :-/

Issue #1247, opened 2019-03-06
Crash due to missing class


mueller-ma:
````
Unable to instantiate service org.openhab.habdroid.core.FcmInstanceIdListenerService: java.lang.ClassNotFoundException: Didn't find class "org.openhab.habdroid.core.FcmInstanceIdListenerService" on path: DexPathList[[zip file "/data/app/org.openhab.habdroid.beta-2/base.apk"],nativeLibraryDirectories=[/data/app/org.openhab.habdroid.beta-2/lib/arm, /vendor/lib, /system/lib]]
````

#### Client
* Android version: Multiple
* Device model: Multiple
* Stock or customized ROM: Multiple
* App version : 2.3.0-beta
* Build flavor : Full
* Device language: Multiple

@maniac103 https://github.com/siacs/Conversations/commit/c2067e3b56364317bc384b55e320b72326fb1de6 wasn't a good idea, I see less info about crashes now.

maniac103:
I was about to ask about the full stacktrace, as this class was removed because it supposedly isn't needed anymore...so we don't have a stacktrave anymore now?

mueller-ma:
It's still linked in the manifest: https://github.com/openhab/openhab-android/blob/master/mobile/src/full/AndroidManifest.xml#L21

maniac103:
This ... explains things *facepalm* I wonder who reviewed this :-P
I'll fix this.

PR #1249, opened 2019-03-07
FCM fixups

maniac103:


maniac103:
@mueller-ma About https://github.com/openhab/openhab-android/issues/1247#issue-417960337 - what do you use for looking at crashes, Play Store or Crashlytics console? If the latter, it surely shouldn't be affected by what we do with Firebase, as it has its own SDK? If the former, can't we kill of Crashlytics SDK in that case?

mueller-ma:
LGTM! Can you also change FcmRegistrationService Line 90 to FCM?
I'm using https://console.firebase.google.com to access crash reports. The first beta version with Crashlytics is 2.6.3-beta. Let's see how it goes now.

maniac103:
> Can you also change FcmRegistrationService Line 90 to FCM?

Done.

PR #1250, opened 2019-03-07
Only send unit in Setpoint/Slider updates if needed.

maniac103:
In particular, only send them for Number:<dimension> items, not for
plain Number ones.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Item.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Item.java
@@ -39,6 +39,7 @@ public abstract class Item implements Parcelable {
         Image,
         Location,
         Number,
+        NumberWithDimension,
         Player,
         Rollershutter,
         StringItem,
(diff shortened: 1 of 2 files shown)


maniac103:
See https://github.com/openhab/openhab-android/issues/1221#issuecomment-469022874

cribskip:
LGTM, @mueller-ma please include this in the next beta for as much users to test.

I've opened a thread in the forums for discussing the desired behavior there:
https://community.openhab.org/t/question-to-the-design-of-uom/69890

mueller-ma:
I don't understand why this is needed. Can `Number` items even have a unit?

maniac103:
Yes, because the unit never exists explicitly. It's parsed from the state.

maniac103:
Looks like the docs have this covered: https://www.openhab.org/docs/concepts/units-of-measurement.html#numberitem-linked-to-quantitytype-channel 
Units can be added to the item either implicitly (if the channel emits them) or explicitly (`Number:XXX`)

PR #1251, opened 2019-03-07
Fix NPE in LogActivity

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.java
@@ -140,16 +140,12 @@ public class LogActivity extends AppCompatActivity {
             String log = logBuilder.toString();
             SharedPreferences sharedPreferences =
                     PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
-            String localUrl =
-                    sharedPreferences.getString(Constants.PREFERENCE_LOCAL_URL, "");
-            String remoteUrl =
-                    sharedPreferences.getString(Constants.PREFERENCE_REMOTE_URL, "");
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.java
```diff
@@ -146,11 +144,11 @@ protected String doInBackground(Boolean... clear) {
...
             String remoteUrl =
                     sharedPreferences.getString(Constants.PREFERENCE_REMOTE_URL, "");
-            if (!TextUtils.isEmpty(localUrl)) {
-                log = log.replaceAll(getHostFromUrl(localUrl), "<openhab-local-address>");
+            if (!TextUtils.isEmpty(localUrl) && Util.getHostFromUrl(localUrl) != null) {
+                log = log.replaceAll(Util.getHostFromUrl(localUrl), "<openhab-local-address>");
             }
-            if (!TextUtils.isEmpty(remoteUrl)) {
-                log = log.replaceAll(getHostFromUrl(remoteUrl), "<openhab-remote-address>");
+            if (!TextUtils.isEmpty(remoteUrl) && Util.getHostFromUrl(remoteUrl) != null) {
```
maniac103: How can host be null if URL is not empty? We validate the URL on input after all.

In any case, I'd avoid the double parsing of the URL. Just make a method:

```java
private String redactHost(String text, String url, String replacement) {
    if (!TextUtils.isEmpty(url)) {
        String host = Util.hostFromUrl(url);
        if (host != null) {
            return text.replaceAll(host, replacement);
        }
    }
    return text;
}
```

mueller-ma: > How can host be null if URL is not empty?

No idea, but the crash happend multiple times.

> Just make a method

Yes, good idea.

PR #1252, opened 2019-03-07
Fix NPE in ContentController#goBack()

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
@@ -398,8 +398,9 @@ public abstract class ContentController implements PageConnectionHolderFragment.
      * @return true if back key was consumed, false otherwise
      */
     public boolean goBack() {
-        if (mTemporaryPage instanceof FullScreenWebviewFragment) {
-            WebView webView = ((FullScreenWebviewFragment) mTemporaryPage).mWebview;
+        WebView webView = mTemporaryPage instanceof FullScreenWebviewFragment ?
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -398,11 +398,14 @@ public boolean canGoBack() {
...
-        if (mTemporaryPage instanceof FullScreenWebviewFragment) {
-            WebView webView = ((FullScreenWebviewFragment) mTemporaryPage).mWebview;
+        WebView webView = mTemporaryPage instanceof FullScreenWebviewFragment
+                ? ((FullScreenWebviewFragment) mTemporaryPage).mWebview : null;
+        if (webView != null) {
             if (webView.canGoBack()) {
                 webView.goBack();
                 return true;
+            } else {
+                webView.loadUrl("about:blank");
```
maniac103: What's the point of this? The WebView is going to be killed off when the fragment is removed anyway.

maniac103:
Why not just add the null check to the existing `if ()` line?

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
mueller-ma: If I go back only once I see the sitemap, but the fragment isn't killed. There are still log messages from HABPanel.

mueller-ma:
Just personal prefference: I didn't want to call `((FullScreenWebviewFragment) mTemporaryPage).mWebview` twice.

PR #1254, opened 2019-03-07
Fix HABPanel HTTP Error 401

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
@@ -30,7 +30,6 @@ import android.webkit.WebChromeClient;
 import android.webkit.WebResourceError;
 import android.webkit.WebResourceRequest;
 import android.webkit.WebView;
-import android.webkit.WebViewDatabase;
 import android.widget.Button;
 import android.widget.ImageView;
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.java
```diff
@@ -791,6 +793,14 @@ public void onReceivedError(WebView view, int errorCode, String description,
...
                 }
             });
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                WebViewDatabase webViewDatabase = WebViewDatabase.getInstance(getContext());
+                webViewDatabase.setHttpAuthUsernamePassword(Util.getHostFromUrl(url), "",
+                        mConnection.getUsername(), mConnection.getPassword());
+            } else {
+                mWebview.setHttpAuthUsernamePassword(Util.getHostFromUrl(url), "",
+                        mConnection.getUsername(), mConnection.getPassword());
+            }
```
maniac103: I wonder whether it would make sense to put this into a Util method: `void applyAuthentication(WebView WebView, Connection connection)`

mueller-ma: Done

PR #1258, opened 2019-03-10
Add "Turn off" action to demo mode snackbar

mueller-ma:


maniac103:
I wonder whether we only should add this if a local or remote URL is set?

mueller-ma:
IMO not, because the discovery service starts if no urls are set.

Issue #1230, opened 2019-03-03
Icon cache apparently cleared when sliding down


lolodomo:
In remote access mode, I can see that the icon cache is correctly handled. When I reopen a page, the icons are quickly displayed, faster than the first time.
Something I don't understand is why the handling is page per page. If I have the same icon in several pages, I would expect that loading is immediate when opening the second page.

I come to the real problem. When I slide down on any page, I get a message telling me that it is useless. Ok but then I can notice that the icon cache is apparently lost for all pages because I can see that the icon loading in any page (the first time) is slower again.
Is there a valid reason to clear the icon cache in this case ? Is there a valid reason to clear it for all pages and not only on the page I asked to refresh ?

mueller-ma:
That's intended, but I agree it's not the best behavior: https://github.com/openhab/openhab-android/pull/929#issuecomment-401471842

lolodomo:
The refresh can be triggered too much easily sliding down, while you just want to go to the top of your sitemap page !
Do you think a preference to disable refresh with slide down could be added ?

maniac103:
I think what we should do instead is only trigger swipe to refresh if the swipe was started at the top of the page. I have done this in the past in another project, just need to dig out that code again.

lolodomo:
Yes, good idea.

PR #1260, opened 2019-03-12
Prevent swipe-to-refresh if swipe isn't started at top of page.

maniac103:
Current it's too easy to accidentally trigger swipe-to-refresh when
swiping to scroll the widget list upwards. Prevent that by only allowing
swipe-to-refresh if the widget list is already at the top when starting
the swipe gesture. This means if one wants to refresh while in the
middle of the list, one needs two swipe gestures.

See #1230

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/widget/RecyclerViewSwipeRefreshLayout.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/widget/RecyclerViewSwipeRefreshLayout.java
@@ -1,0 +1,143 @@
+package org.openhab.habdroid.ui.widget;
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.view.MotionEvent;
+import android.view.View;
+import android.view.ViewConfiguration;
+
+import androidx.core.view.NestedScrollingChildHelper;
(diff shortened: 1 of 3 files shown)


mueller-ma:
Is it possible to add some tolerance to `shouldPreventRefresh()`? Currently if 1 px is hidden by the action bar and you swipe down, the refresh isn't triggered.

maniac103:
I don't think so. One seemingly can only give a direction for checking, not an amount: https://developer.android.com/reference/android/view/View.html#canScrollVertically(int)

mobile/src/main/java/org/openhab/habdroid/ui/widget/RecyclerViewSwipeRefreshLayout.java
```diff
@@ -0,0 +1,143 @@
...
+        return false;
+    }
+
+    @Override
+    public boolean onInterceptTouchEvent(MotionEvent event) {
+        if (event.getAction() != MotionEvent.ACTION_DOWN && shouldPreventRefresh()) {
+            return false;
+        }
+
+        switch (event.getAction()) {
```
mueller-ma: I would add an empty default branch here, but that's personal preference.

PR #1269, opened 2019-03-27
Improve number handling

maniac103:
See https://community.openhab.org/t/setpoint-cant-edit-the-value-with-fractional-digit/70705


mueller-ma:
Using the example from the forum, I can select 0, 1, 1, 2, 2, 3, 3,... if the time has 0 minutes (12:00) and I can select 0.0, 0.5, 1.0,... if the time has 30 minutes (12:30).
As you already said we don't know the mapping values, so this is as best as it can get, right?

maniac103:
I think so, yes. The integer values probably should be dropped from the mapping file to print the fractional values properly. I don't see where we could get a better format from.

PR #1280, opened 2019-03-30
Add security features

mueller-ma:
1. Add setting to prevent screenshots.
2. Add panic button receiver. This seems very specific to me and I'm pretty sure, that most people won't use it. It's possible to create a subscreen only for panic button prefs and open this through the app "Ripple" (https://github.com/guardianproject/Ripple/blob/master/metadata/en-US/phoneScreenshots/screenshot-main.png There is an "edit" button for each app that opens the pref activity (see added intent-filter)), but not link to the subscreen in openHAB. This way it's only visible for people using Ripple.
@maniac103 WDYT?
3.  Add option to remove passwords if the url is changed. An attacker could change the url to a server under his control and get the password by letting the app connect.
4. Add option to enable screen lock.


maniac103:
TBH, I'm not a fan of most of this. Looking at the individual items:

- *Add setting to prevent screenshots.*
  What's the point of this? What is so secret in OH sitemaps that it must not be part of any screenshot? Additionally, what does one gain from having a setting which - when enabled - prevents one from doing things? One would think that the user enabling the option and the user not doing screenshots of secret things has the same precondition - an inforrmed user...
- *Add panic button receiver*
  Besides the very specific usecase (which I wouldn't mind by itself if settings are hidden as long as the app isn't installed, this seems very obscure for another reason: the mentioned 'Ripple' app doesn't seem to have gained any significant user base. Its Play Store link is dead (why?), but from the number of stars on Github it really doesn't seem to be popular. When adding 3rd party app integration, IMHO we should focus on apps with a relevant user base.
- *Add option to remove passwords if the url is changed*
  Why not do this unconditionally? When changing the server URL from one non-empty one to another non-empty one, show a warning ('Note: Changing the server URL will cause username and password to be cleared out for security reasons') and after acknowledging the change just clear them out.
- *Add option to enable screen lock.*
  That one actually is OK with me.

mueller-ma:
> Add setting to prevent screenshots.

It prevents other apps of making screenshots of openHAB and it changes the preview in the latest apps view to a white image.
You can also find this setting in a few security focused apps with a high download count, like Signal or Firefox Klar/Focus.

> Add panic button receiver

I didn't noticed it's not available in the Play Store anymore. However it's available in F-Droid and other apps can provide a panic button as well. How about simplifying the implementation: If screenlock is enabled, close the app when receiving a panic button intent. No settings for this at all.

> Add option to remove passwords if the url is changed

This might annoy the user if he just wants to test something, e.g. temporary disable the remote access. On the other side I don't really like the pref as well.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
```diff
@@ -1208,8 +1211,14 @@ public void promptForDevicePassword() {
...
         }
         KeyguardManager km = (KeyguardManager) this.getSystemService(Context.KEYGUARD_SERVICE);
-        Intent intent = km.createConfirmDeviceCredentialIntent(getString(R.string.app_name), "");
-        startActivityForResult(intent, SCREEN_LOCK_REQUEST_CODE);
+        if ((Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && km.isDeviceSecure())
+                || (Build.VERSION.SDK_INT < Build.VERSION_CODES.M && km.isKeyguardSecure())) {
+            Intent intent = km.createConfirmDeviceCredentialIntent(getString(R.string.app_name), "");
+            startActivityForResult(intent, SCREEN_LOCK_REQUEST_CODE);
+        } else {
+            showSnackbar(R.string.error_no_pin_set); // TODO
```
mueller-ma: This function is called, but the Snackbar never shown.
Also: Add a action here and open the Android settings.

mueller-ma:
I don't like the current lock screen implementation. I split this PR in one for the screen lock and maybe one for the screenshot prevention.
Both should be easier to implement with the new AbstractBaseActivity.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
maniac103: It's not shown because it's both
- replaced by the 'using local/remote/demo connection' snackbar and
- hidden in onAvailableConnectionChanged

mueller-ma:
Well, my proposal has one major flaw: Only the MainActivity is guarded, so an attacker might get access to e.g. settings or openHAB secrets in About.
I have a look at moving the code to AbstractBaseActivity.

maniac103:
In that case we need to manage the authentication state and timestamp in the application object (or some other singleton) though to share it between activities.

mueller-ma:
WIP: https://github.com/mueller-ma/openhab.android/commit/7d454d7ea633ffb9603392c89272819b1af51d47

mueller-ma:
Superseded by #1449

PR #1283, opened 2019-04-02
Support writing NFC tags of Setpoint and Slider

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.java
@@ -149,7 +149,7 @@ public class WidgetListFragment extends Fragment
                 }
             } else if (widget.type() == Widget.Type.Setpoint
                     || widget.type() == Widget.Type.Slider) {
-                if (widget.state() != null) {
+                if (widget.state() != null && widget.state().asNumber() != null) {
                     ParsedState.NumberState state = widget.state().asNumber();
 
                     String currentState = state.toString(Locale.getDefault());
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.java
```diff
@@ -146,6 +147,29 @@ public void onItemLongClicked(final Widget widget) {
                     labels.add(getString(R.string.nfc_action_current_color));
                     commands.add(widget.state().asString());
                 }
+            } else if (widget.type() == Widget.Type.Setpoint
+                    || widget.type() == Widget.Type.Slider) {
+                if (widget.state() != null) {
+                    ParsedState.NumberState state = widget.state().asNumber();
+
+                    String currentState = state.toString(Locale.getDefault());
```
maniac103: Need to null check state here? (e.g. by a ternary: `NumberState state = widget.state() != null ? widget.state().asNumber() : null;`)

Also can omit the locale parameter.

mueller-ma: I check widget.state() for null in line 152 already.

maniac103: `widget.state().asNumber()` can be null even with `widget.state` being != null: in case the state can't be converted to a number.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.java
```diff
@@ -146,6 +147,29 @@ public void onItemLongClicked(final Widget widget) {
                     labels.add(getString(R.string.nfc_action_current_color));
                     commands.add(widget.state().asString());
                 }
+            } else if (widget.type() == Widget.Type.Setpoint
+                    || widget.type() == Widget.Type.Slider) {
+                if (widget.state() != null && widget.state().asNumber() != null) {
+                    ParsedState.NumberState state = widget.state().asNumber();
+
+                    String currentState = state.toString(Locale.getDefault());
```
maniac103: The redundant `Locale.getDefault()` is still there ;-)

Issue #13, opened 2014-08-13
writing nfc tag for a switch crash the app


bogdanioanliviu:
I´d like to add a new issue to the Android app.
 Writing a nfc tag with a toggle switch it force close the app.
 Steps to reproduce: open the app, long hold over a switch, popup appear and select toggle, awaiting for tag, trying to write tag , force close.
 If you try the same with a group of items (from the demo FF_Bedroom items) it is working with out glitch.
Tested with sony xperia tags http://www.sonymobile.com/global-en/products/accessories/xperia-smarttags/
If a logcat is needed please let me know.
Version 1.5.1 
Rom used: CM11
Phone :xperia sp 
Thank you
##

mueller-ma:
#1302 changes the uri for item update tags to `openhabitem://?i=item&s=state`, which is shorter than the current uri.

PR #1302, opened 2019-04-12
Improve NFC item updates

mueller-ma:
Create a new scheme that is only used for item updates. This way we can
have a background responder for NFC item updates. It uses the framework
introduced in #1137.

If an old tag is used, we show a message to the user to renew his tag.

See https://github.com/openhab/openhab-android/issues/538#issuecomment-480483543
See #13 (Shorter parameter names and no sitemap require less bytes)

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -65,21 +65,6 @@
 
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
-            <intent-filter>
-                <action android:name="android.nfc.action.NDEF_DISCOVERED" />
-
-                <category android:name="android.intent.category.DEFAULT" />
-
-                <data android:scheme="openhab" />
-            </intent-filter>
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW" />
-
-                <category android:name="android.intent.category.DEFAULT" />
-                <category android:name="android.intent.category.BROWSABLE" />
-
-                <data android:scheme="openhab" />
-            </intent-filter>
         </activity>
         <activity
             android:name=".ui.WriteTagActivity"
@@ -145,7 +130,7 @@
 
                 <category android:name="android.intent.category.DEFAULT" />
 
-                <data android:scheme="openhabitem" />
+                <data android:scheme="openhab" />
             </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
@@ -153,7 +138,7 @@
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
 
-                <data android:scheme="openhabitem" />
+                <data android:scheme="openhab" />
             </intent-filter>
         </activity>
     </application>
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -23,11 +23,11 @@
         android:allowBackup="false"
         android:icon="@mipmap/icon"
         android:label="@string/app_name"
+        android:networkSecurityConfig="@xml/network_security_config"
         android:resizeableActivity="true"
         android:supportsPictureInPicture="false"
         android:supportsRtl="true"
-        android:theme="@style/HABDroid.Light"
-        android:networkSecurityConfig="@xml/network_security_config">
+        android:theme="@style/HABDroid.Light">
         <activity
             android:name=".ui.PreferencesActivity"
             android:label="@string/app_preferences_name">
@@ -44,7 +44,8 @@
         <activity
             android:name=".ui.AboutActivity"
             android:label="@string/about_title" />
-        <activity android:name=".ui.LogActivity"
+        <activity
+            android:name=".ui.LogActivity"
             android:label="@string/log" />
         <activity
             android:name=".ui.MainActivity"
@@ -107,22 +108,21 @@
                 android:resource="@xml/voice_widget_info" />
         </receiver>
         <receiver
-                android:name=".ui.homescreenwidget.VoiceWidgetWithIcon"
-                android:label="@string/title_voice_widget_icon">
+            android:name=".ui.homescreenwidget.VoiceWidgetWithIcon"
+            android:label="@string/title_voice_widget_icon">
             <intent-filter>
                 <action android:name="android.appwidget.action.APPWIDGET_UPDATE" />
             </intent-filter>
 
             <meta-data
-                    android:name="android.appwidget.provider"
-                    android:resource="@xml/voice_widget_with_icon_info" />
+                android:name="android.appwidget.provider"
+                android:resource="@xml/voice_widget_with_icon_info" />
         </receiver>
         <receiver android:name=".core.OnUpdateBroadcastReceiver">
             <intent-filter>
                 <action android:name="android.intent.action.MY_PACKAGE_REPLACED" />
             </intent-filter>
         </receiver>
-
         <receiver
             android:name=".background.BackgroundTasksManager"
             android:enabled="true"
@@ -131,9 +131,31 @@
                 <action android:name="android.app.action.NEXT_ALARM_CLOCK_CHANGED" />
             </intent-filter>
             <intent-filter>
-                <action android:name="android.intent.action.LOCALE_CHANGED"/>
+                <action android:name="android.intent.action.LOCALE_CHANGED" />
             </intent-filter>
         </receiver>
+
+        <activity
+            android:name=".background.NfcReceiveActivity"
+            android:launchMode="singleInstance"
+            android:theme="@android:style/Theme.NoDisplay"
+            android:noHistory="true">
+            <intent-filter>
+                <action android:name="android.nfc.action.NDEF_DISCOVERED" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+
+                <data android:scheme="openhabitem" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data android:scheme="openhabitem" />
+            </intent-filter>
+        </activity>
     </application>
 
 </manifest>
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
@@ -35,7 +35,7 @@ public class BackgroundTasksManager extends BroadcastReceiver {
     static final String EXTRA_RETRY_INFOS = "retryInfos";
 
     private static final String WORKER_TAG_ITEM_UPLOADS = "itemUploads";
-    public static final String WORKER_TAG_PREFIX_NFC = "nfc-";
+    static final String WORKER_TAG_PREFIX_NFC = "nfc-";
 
     static final List<String> KNOWN_KEYS = Arrays.asList(
         Constants.PREFERENCE_ALARM_CLOCK
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
@@ -9,6 +9,7 @@ import android.os.Build;
 import android.os.Parcel;
 import android.os.Parcelable;
 import android.preference.PreferenceManager;
+import android.text.TextUtils;
 import android.util.Log;
 import android.util.Pair;
 import androidx.annotation.Nullable;
@@ -19,9 +20,11 @@ import androidx.work.OneTimeWorkRequest;
 import androidx.work.WorkInfo;
 import androidx.work.WorkManager;
 
+import org.openhab.habdroid.R;
 import org.openhab.habdroid.model.NfcTag;
 import org.openhab.habdroid.ui.widget.ItemUpdatingPreference;
 import org.openhab.habdroid.util.Constants;
+import org.openhab.habdroid.util.Util;
 
 import java.util.Arrays;
 import java.util.HashMap;
@@ -121,8 +124,15 @@ public class BackgroundTasksManager extends BroadcastReceiver {
         workManager.enqueue(workRequest);
     }
 
-    public static void enqueueNfcUpdateIfNeeded(@Nullable NfcTag tag) {
+    public static void enqueueNfcUpdateIfNeeded(Context context, @Nullable NfcTag tag) {
         if (tag != null && tag.sitemap() == null) {
+            String message;
+            if (TextUtils.isEmpty(tag.label())) {
+                message = context.getString(R.string.nfc_tag_recognized_item, tag.item());
+            } else {
+                message = context.getString(R.string.nfc_tag_recognized_label, tag.label());
+            }
+            Util.showToast(context, message);
             enqueueItemUpload(WORKER_TAG_PREFIX_NFC + tag.item(), tag.item(), tag.state());
         }
     }
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
@@ -36,6 +36,7 @@ public class BackgroundTasksManager extends BroadcastReceiver {
     static final String EXTRA_RETRY_INFOS = "retryInfos";
 
     private static final String WORKER_TAG_ITEM_UPLOADS = "itemUploads";
+    public static final String WORKER_TAG_PREFIX_NFC = "nfc-";
 
     static final List<String> KNOWN_KEYS = Arrays.asList(
         Constants.PREFERENCE_ALARM_CLOCK
@@ -108,7 +109,7 @@ public class BackgroundTasksManager extends BroadcastReceiver {
                 || NfcAdapter.ACTION_NDEF_DISCOVERED.equals(intent.getAction())) {
             Uri nfcData = intent.getData();
             String itemName = nfcData.getQueryParameter(WriteTagActivity.QUERY_PARAMETER_ITEM_NAME);
-            enqueueItemUpload("nfc-" + itemName, itemName,
+            enqueueItemUpload(WORKER_TAG_PREFIX_NFC + itemName, itemName,
                     nfcData.getQueryParameter(WriteTagActivity.QUERY_PARAMETER_STATE));
         }
     }
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
@@ -5,13 +5,14 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.net.Uri;
+import android.nfc.NfcAdapter;
 import android.os.Build;
 import android.os.Parcel;
 import android.os.Parcelable;
 import android.preference.PreferenceManager;
 import android.util.Log;
 import android.util.Pair;
-
 import androidx.lifecycle.LiveData;
 import androidx.work.Constraints;
 import androidx.work.NetworkType;
@@ -19,6 +20,7 @@ import androidx.work.OneTimeWorkRequest;
 import androidx.work.WorkInfo;
 import androidx.work.WorkManager;
 
+import org.openhab.habdroid.ui.WriteTagActivity;
 import org.openhab.habdroid.ui.widget.ItemUpdatingPreference;
 import org.openhab.habdroid.util.Constants;
 
@@ -99,6 +101,16 @@ public class BackgroundTasksManager extends BroadcastReceiver {
 
         String prefix = prefs.getString(Constants.PREFERENCE_SEND_DEVICE_INFO_PREFIX, "");
         enqueueItemUpload(key, prefix + setting.second, getter.getValue(context));
+    }
+
+    public static void enqueueNfcItemUpload(Intent intent) {
+        if (Intent.ACTION_VIEW.equals(intent.getAction())
+                || NfcAdapter.ACTION_NDEF_DISCOVERED.equals(intent.getAction())) {
+            Uri nfcData = intent.getData();
+            String itemName = nfcData.getQueryParameter(WriteTagActivity.QUERY_PARAMETER_ITEM_NAME);
+            enqueueItemUpload("nfc-" + itemName, itemName,
+                    nfcData.getQueryParameter(WriteTagActivity.QUERY_PARAMETER_STATE));
+        }
     }
 
     private static void enqueueItemUpload(String tag, String itemName, String value) {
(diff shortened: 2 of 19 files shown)


mobile/src/main/AndroidManifest.xml
```diff
@@ -23,11 +23,11 @@
         android:allowBackup="false"
         android:icon="@mipmap/icon"
         android:label="@string/app_name"
+        android:networkSecurityConfig="@xml/network_security_config"
```
mueller-ma: These changes have been made by AS.

mobile/src/main/AndroidManifest.xml
```diff
@@ -131,9 +131,31 @@
                 <action android:name="android.app.action.NEXT_ALARM_CLOCK_CHANGED" />
             </intent-filter>
             <intent-filter>
-                <action android:name="android.intent.action.LOCALE_CHANGED"/>
+                <action android:name="android.intent.action.LOCALE_CHANGED" />
             </intent-filter>
         </receiver>
+
+        <activity
```
mueller-ma: I tried to add the intent filter to BackgroundTasksManager, but it seems that the NFC intents can only start activities, no broadcast receivers:

> Intent to start an activity when a tag with NDEF payload is discovered. 

from https://developer.android.com/reference/android/nfc/NfcAdapter.html#ACTION_NDEF_DISCOVERED

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
```diff
@@ -738,30 +739,40 @@ private void launchVoiceRecognitionIfNeeded() {
     }
 
     private void openPendingNfcPageIfNeeded() {
-        if (mPendingNfcData == null || mConnection == null || !mStarted) {
+        if (mPendingNfcIntent == null || mPendingNfcIntent.getData() == null
```
mueller-ma: Not sure if `mPendingNfcIntent.getData()` can be null if `mPendingNfcIntent` isn't.

mobile/src/main/java/org/openhab/habdroid/ui/WriteTagActivity.java
```diff
@@ -147,35 +156,58 @@ public void onNewIntent(Intent intent) {
...
-        Log.d(TAG, "Creating tag object for URI " + uri);
+    private NdefMessage getNdefMessage(String uri) {
+        if (TextUtils.isEmpty(uri)) {
+            return null;
+        }
+        NdefRecord[] longNdefRecords = new NdefRecord[] { NdefRecord.createUri(uri) };
+        return new NdefMessage(longNdefRecords);
+    }
+
+    private void writeTag(Tag tag, String longUri, String shortUri) {
```
mueller-ma: This function calls `writeTagMessage.setText()` a few times, but the text is never changed in the UI. This should be fixed by using a AsyncTask.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
```diff
@@ -101,6 +104,16 @@ private static void scheduleWorker(Context context, String key) {
         enqueueItemUpload(key, prefix + setting.second, getter.getValue(context));
     }
 
+    public static void enqueueNfcItemUpload(Intent intent) {
+        if (Intent.ACTION_VIEW.equals(intent.getAction())
+                || NfcAdapter.ACTION_NDEF_DISCOVERED.equals(intent.getAction())) {
```
maniac103: I think this check should go to the activity and just the URI be passed here.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
```diff
@@ -145,7 +146,7 @@
     private Snackbar mLastSnackbar;
     private Connection mConnection;
 
-    private Uri mPendingNfcData;
+    private Intent mPendingNfcIntent;
```
maniac103: When passing only the URI to BackgroundTasksManager, this change and everything related to it is superfluous.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
```diff
@@ -738,30 +739,40 @@ private void launchVoiceRecognitionIfNeeded() {
...
+                String newPageUrl = String.format(Locale.US, "rest/sitemaps%s", nfcSitemap);
+                mController.openPage(newPageUrl);
+            } else {
+                Log.d(TAG, "This is an old tag for item updates (" + nfcItem + ")");
+                String url = String.format(Locale.US, "rest/items/%s", nfcItem);
+                Util.sendItemCommand(mConnection.getAsyncHttpClient(), url, nfcState);
+                new AlertDialog.Builder(this)
+                        .setMessage(getString(R.string.info_write_tag_again, nfcItem, nfcState))
+                        .setPositiveButton(android.R.string.ok, (dialog, which) -> finish())
+                        .show();
```
maniac103: Maybe just show a snackbar instead of this? Another approach: Route all NFC stuff to the transparent activity and just start MainActivity if the tag is for a sitemap page.

mobile/src/main/java/org/openhab/habdroid/ui/WriteTagActivity.java
```diff
@@ -139,76 +150,126 @@ public void onNewIntent(Intent intent) {
...
-                    if (ndef.isWritable()) {
-                        ndef.writeNdefMessage(message);
+                    sitemapUri = new URI(mSitemapPage);
+                    if (!TextUtils.isEmpty(mItem) && !TextUtils.isEmpty(mState)) {
+                        longUri = "openhabitem://?" + QUERY_PARAMETER_ITEM_NAME + "=" + mItem
+                                + "&" + QUERY_PARAMETER_STATE + "=" + mState
+                                + "&" + QUERY_PARAMETER_MAPPED_STATE + "=" + mMappedState
+                                + "&" + QUERY_PARAMETER_ITEM_LABEL + "=" + mLabel;
+                        shortUri = "openhabitem://?" + QUERY_PARAMETER_ITEM_NAME + "=" + mItem
+                                + "&" + QUERY_PARAMETER_STATE + "=" + mState;
```
maniac103: Wouldn't it be more elegant to use the `android.net.Uri` class methods for constructing those? Could construct `shortUri` first and then `longUri` from it using `Uri.buildUpon()`.
That API should also work for `sitemapUri`.

mobile/src/main/java/org/openhab/habdroid/ui/WriteTagActivity.java
```diff
@@ -139,76 +150,126 @@ public void onNewIntent(Intent intent) {
...
-        Log.d(TAG, "Writing page " + mSitemapPage + " to tag");
+        new AsyncTask<Void, Integer, Boolean>() {
+            @Override
+            protected Boolean doInBackground(Void... voids) {
+                Tag tag = intent.getParcelableExtra(NfcAdapter.EXTRA_TAG);
+                Log.d(TAG, "NFC TAG = " + tag.toString());
+                Log.d(TAG, "Writing page " + mSitemapPage + " to tag");
 
-        TextView writeTagMessage = findViewById(R.id.write_tag_message);
+                publishProgress(R.string.info_write_tag_progress);
```
maniac103: Given this is only called once the progress logic could/should be moved to onPreExecute.

mobile/src/main/java/org/openhab/habdroid/ui/WriteTagActivity.java
```diff
@@ -139,76 +150,126 @@ public void onNewIntent(Intent intent) {
...
+                        ndefFormatable.connect();
+                        try {
+                            ndefFormatable.format(longMessage);
+                        } catch (IOException e) {
+                            if (shortMessage != null) {
+                                Log.d(TAG, "Try with short uri");
+                                ndefFormatable.format(shortMessage);
+                            }
+                        }
+                        ndefFormatable.close();
```
maniac103: Should this go into a finally block?

mobile/src/main/java/org/openhab/habdroid/ui/WriteTagActivity.java
```diff
@@ -139,76 +150,126 @@ public void onNewIntent(Intent intent) {
...
+                                    ndef.writeNdefMessage(longMessage);
+                                } catch (IOException e) {
+                                    if (shortMessage != null) {
+                                        Log.d(TAG, "Try with short uri");
+                                        ndef.writeNdefMessage(shortMessage);
+                                    }
+                                }
+                            }
+                            Log.d(TAG, "Closing");
+                            ndef.close();
```
maniac103: As above.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
mueller-ma: I'll try your second approach.

mueller-ma: Done! Works pretty good and we need no deprecation.

mobile/src/main/java/org/openhab/habdroid/model/NfcTag.java
```diff
@@ -0,0 +1,76 @@
...
+                : uri.getQueryParameter("command");
+        String mappedState = uri.getQueryParameter(WriteTagActivity.QUERY_PARAMETER_MAPPED_STATE);
+
+        return new AutoValue_NfcTag.Builder()
+                .sitemap(sitemap)
+                .item(item)
+                .label(label)
+                .state(state)
+                .mappedState(mappedState)
+                .isOpenHabTag("openhab".equals(uri.getScheme()))
```
maniac103: Isn't that guaranteed to be true anyway?

mobile/src/main/java/org/openhab/habdroid/model/NfcTag.java
```diff
@@ -0,0 +1,76 @@
...
+                .isSitemapTag(!TextUtils.isEmpty(sitemap) && TextUtils.isEmpty(item))
+                .hasLabelAndMappedState(TextUtils.isEmpty(label) || TextUtils.isEmpty(mappedState))
```
maniac103: Implement those two as separate methods instead of data members instead? They're purely derived from other data members, so having them as members creates redundancy.

mobile/src/main/java/org/openhab/habdroid/model/NfcTag.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public boolean mustOpenSitemap() {
+        if (!isOpenHabTag()) {
+            return false;
+        }
+        if (isSitemapTag()) {
+            return true;
+        }
+        BackgroundTasksManager.enqueueItemUpload(WORKER_TAG_PREFIX_NFC + item(), item(), state());
```
maniac103: This doesn't belong here, it's a very surprising side effect. Rather should call this in the else branch in `NfcReceiveActivity`.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
```diff
@@ -354,8 +357,13 @@ private void processIntent(Intent intent) {
...
             case Intent.ACTION_VIEW:
-                mPendingNfcData = intent.getData();
-                openPendingNfcPageIfNeeded();
+                NfcTag tag = NfcTag.fromTagData(intent.getData());
+
+                if (tag.mustOpenSitemap()) {
+                    mPendingOpenSitemapUrl = tag.sitemap();
+                    openPendingSitemapIfNeeded();
+                }
+
```
maniac103: This `case` block - AFAICT - won't be executed anymore since the respective intent filter was removed from the manifest and thus can be removed.

mobile/src/main/java/org/openhab/habdroid/model/NfcTag.java
mueller-ma: Well, I wanted to add a function `handleTag()` that does update the state or open a sitemap, but opening the way how to open the sitemap depends from where it is called.
But the current function name isn't perfect. I was thinking about something like `handleTagAndCheckIfSitemapMustBeOpened()`, but that's quite long.

maniac103: As per my other comment I think this should only be called from NfcReceiveActivity only anyway, so IMHO it should be fine to move the call there...I wonder whether we even still need the NfcTag class in that case.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
mueller-ma: I think it will be used. Don't we tell the nfc adapter to get all tags? I've to check that. https://github.com/openhab/openhab-android/blob/768bc546e43dab81a53828ec70b4ce0dbc58ce7d/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java#L412-L418

kanokwanmaneeon: # [1302](url)

maniac103: Didn't think of that, but can't we route that one towards the NfcReceiveActivity as well?

mueller-ma: I'd like to keep this here as I don't want to start an invisible activity only to re-launch the current one. By using the function `NfcTag#mustOpenSitemap()` there's nearly no duplicate code.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
```diff
@@ -101,7 +101,7 @@ private static void scheduleWorker(Context context, String key) {
         enqueueItemUpload(key, prefix + setting.second, getter.getValue(context));
     }
 
-    private static void enqueueItemUpload(String tag, String itemName, String value) {
+    public static void enqueueItemUpload(String tag, String itemName, String value) {
```
maniac103: Thinking about it, I'd prefer to leave this private (and the prefix constant package private) and introducing a new method `public static void enqueueNfcUpdate(NfcTag tag)` which does the necessary wrapping.

(Reason is that IMHO the public methods should guide the callers in how they're to be used)

mobile/src/main/java/org/openhab/habdroid/model/NfcTag.java
```diff
@@ -0,0 +1,76 @@
...
+        abstract Builder state(String state);
+        abstract Builder mappedState(String mappedState);
+        abstract Builder isOpenHabTag(boolean isOpenHabTag);
+        abstract Builder isSitemapTag(boolean isSitemapTag);
+        abstract Builder hasLabelAndMappedState(boolean hasLabelAndMappedState);
+
+        abstract NfcTag build();
+    }
+
+    public static NfcTag fromTagData(Uri uri) {
```
maniac103: Could just return `null` here if scheme isn't 'openhab'. That way the `isOpenhabTag` method would become superfluous.

mobile/src/main/java/org/openhab/habdroid/model/NfcTag.java
```diff
@@ -0,0 +1,76 @@
...
+        String sitemap = uri.getPath();
+        String item = TextUtils.isEmpty(uri.getQueryParameter("item")) ?
+                uri.getQueryParameter(WriteTagActivity.QUERY_PARAMETER_ITEM_NAME)
+                : uri.getQueryParameter("item");
+        String label = uri.getQueryParameter(WriteTagActivity.QUERY_PARAMETER_ITEM_LABEL);
+        String state = TextUtils.isEmpty(uri.getQueryParameter("command")) ?
+                uri.getQueryParameter(WriteTagActivity.QUERY_PARAMETER_STATE)
+                : uri.getQueryParameter("command");
+        String mappedState = uri.getQueryParameter(WriteTagActivity.QUERY_PARAMETER_MAPPED_STATE);
+
```
maniac103: Could set `sitemap` to null here in case `item` is non-empty, that way the `isSitemapTag` method becomes superfluous.

mobile/src/main/java/org/openhab/habdroid/model/NfcTag.java
maniac103: Fine. My (fairly strong in this case) opinion still is that the item upload doesn't belong into this method though. Actually, with the above suggestions, we can drop this method altogether and just check for `tag != null && !TextUtils.isEmpty(tag.sitemap())`

mobile/src/main/java/org/openhab/habdroid/model/NfcTag.java
mueller-ma: Actually this isn't needed at all.

mobile/src/main/java/org/openhab/habdroid/model/NfcTag.java
mueller-ma: Indeed, that's a better solution.

mueller-ma:
When launching to a subpage via NfcReceiveActivity, the MainActivity shows the root sitemap page. Using the same tag when the MainActivity is open shows the subpage.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.java
```diff
@@ -34,6 +35,7 @@
     static final String EXTRA_RETRY_INFOS = "retryInfos";
 
     private static final String WORKER_TAG_ITEM_UPLOADS = "itemUploads";
+    public static final String WORKER_TAG_PREFIX_NFC = "nfc-";
```
maniac103: Make this package private (by omitting the `public` keyword)?

maniac103:
Looks like there's another issue: In demo mode, I go to the 'ground floor' page, long-press 'kitchen' and write the tag. Reading the tag then takes me to the 'ground floor' page, where I'd expect to be taken to the 'kitchen' page. Am I missing something there?

maniac103:
> When launching to a subpage via NfcReceiveActivity, the MainActivity shows the root sitemap page. Using the same tag when the MainActivity is open shows the subpage.

Since you wait for server properties now [1], you need to call `openPendingSitemapUrlIfNeeded()` when server properties are assigned [2] ... doing so fixes this issue.

[1] https://github.com/openhab/openhab-android/pull/1302/files#diff-c3948f595ec22e3aa2a9a8c75c83afb7R742
[2] https://github.com/openhab/openhab-android/pull/1302/files#diff-c3948f595ec22e3aa2a9a8c75c83afb7R595

mobile/src/main/java/org/openhab/habdroid/ui/WriteTagActivity.java
```diff
@@ -48,14 +51,21 @@
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.Locale;
 
 public class WriteTagActivity extends AppCompatActivity {
     private static final String TAG = WriteTagActivity.class.getSimpleName();
+    public static final String QUERY_PARAMETER_ITEM_NAME = "i";
+    public static final String QUERY_PARAMETER_STATE = "s";
+    public static final String QUERY_PARAMETER_MAPPED_STATE = "m";
```
maniac103: What's the semantic difference of those two? Looks like the same value is passed for both.

maniac103:
See top 2 commits in https://github.com/maniac103/openhab.android/commits/nfc-item-updates ... with those two, IMHO we're good to go.

mueller-ma:
> In demo mode, I go to the 'ground floor' page, long-press 'kitchen' and write the tag. Reading the tag then takes me to the 'ground floor' page, where I'd expect to be taken to the 'kitchen' page. Am I missing something there?

I guess this is intended. @FlorianSW moved the "Write tag for sitemap" action from the action bar to the long-press menu, but nobody realized, that this will change the "context" of this action: Sitemap that is currently open vs sitemap that will be opened via this widget.
With your commit it's not possible any longer to write tags for the root sitemap page, right?

IMO we should
1. For now (in this PR) leave this action as it is currently 
2. Move the action back to the action bar in the overflow menu


> Should probably also use a constant for sitemapPage

I want to add constants for the scheme and the now deprecated query parameters ("item" and "state") and move these consts to NfcTag.

maniac103:
> With your commit it's not possible any longer to write tags for the root sitemap page, right?

That's the only drawback, right, but what's the use case of this action? This would essentially mean to have a tag action to start the app itselt, which, if really needed, can also be achieved by other means.

> Move the action back to the action bar in the overflow menu

That's another way to solve the issue, but I'm currently unsure what the better way is ... less distribution of similar things between different UI elements is preferable IMHO.
What other reasons do you want to reintroduce the overflow menu for?

> and the now deprecated query parameters ("item" and "state")

IMHO not worth introducing consts for those, as they're used only once.

mueller-ma:
> use case of this action?

Opening a sitemap that isn't the default one.

> What other reasons do you want to reintroduce the overflow menu for?

I want to use the overflow menu for #1322 which is quite similar to the nfc sitemap tag.

> IMHO not worth introducing consts for those, as they're used only once.

IMO it's clearer why they are used "DEPRECATED_QUERY_PARAMETER_ITEM_NAME".

> Only thing I was wondering was whether we should show a toast when enqueuing an NFC item update to make it obvious that the tag was recognized.

Good idea. What should the tag say? Just "Tag recognized" or be more precise about what item was changed.

FlorianSW:
> > In demo mode, I go to the 'ground floor' page, long-press 'kitchen' and write the tag. Reading the tag then takes me to the 'ground floor' page, where I'd expect to be taken to the 'kitchen' page. Am I missing something there?
> 
> I guess this is intended. @FlorianSW moved the "Write tag for sitemap" action from the action bar to the long-press menu, but nobody realized, that this will change the "context" of this action: Sitemap that is currently open vs sitemap that will be opened via this widget.
> With your commit it's not possible any longer to write tags for the root sitemap page, right?

That's IMHO the expected behaviour, it takes you to the sitemap page which you're currently on. Maybe the wording of the menu item is a bit misleading though :/ However, I can still navigate to all pages in a sitemap, even the root one, or am I misunderstanding something? (At least it works on the latest version of the app) :)

maniac103:
>> use case of this action?

> Opening a sitemap that isn't the default one.

Let me get this straight: You'd program a tag to open the root sitemap, then open the drawer and select a different sitemap? Still, where's the advantage over simply opening the app? I thought the tags are supposed to be a shortcut?

> I want to use the overflow menu for #1322 which is quite similar to the nfc sitemap tag.

Yes, and it has a similar issue: think e.g. the two-pane view: which pane does the overflow menu relate to?

> Good idea. What should the tag say? Just "Tag recognized" or be more precise about what item was changed.

My suggestion would be a wording like `NFC tag for Item %s recognized`.

> That's IMHO the expected behaviour, it takes you to the sitemap page which you're currently on. Maybe the wording of the menu item is a bit misleading though :/

That, and it's weird in the context ... all other items in that long-press menu relate to the widget itself, just that one item relates to the sitemap page the widget belongs to. This is why I propose to only show this item for widgets that point to a page and make the NFC tag link to the page the widget points to. As mentioned, the only lost capability with that approach is the ability to write a tag pointing to the sitemap root (as no widget can link to it), but as mentioned, I question the use case for that scenario; IOW, I don't think it's a real loss.

> However, I can still navigate to all pages in a sitemap, even the root one, or am I misunderstanding something? (At least it works on the latest version of the app) :)

That's the case currently, and I propose changing that behavior as outlined above.

mueller-ma:
With your commit, it's more intuitive now.
There was one missing openPendingSitemapIfNeeded() call, but after adding this, I see the "sitemap contains no widgets" error, when I want to open a sitemap while the app is closed.

maniac103:
> but after adding this, I see the "sitemap contains no widgets" error, when I want to open a sitemap while the app is closed.

This should fix it: https://github.com/maniac103/openhab.android/commit/57c5c09220f10ea49bf804de5a8908715143b161

mueller-ma:
After fixing the ui test, this rtm from my side.

mueller-ma:
:ship:

PR #1312, opened 2019-04-21
 Show all activities in fullscreen if enabled

### LINKED ISSUE #530: Full screen mode is exited when opening settings/about

mueller-ma:
Fixes #530


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.java
```diff
@@ -236,11 +236,10 @@ protected void onCreate(Bundle savedInstanceState) {
 
         processIntent(getIntent());
 
-        if (isFullscreenEnabled()) {
+        if (Util.isFullscreenEnabled(this)) {
             IntentFilter filter = new IntentFilter(Intent.ACTION_DREAMING_STARTED);
             filter.addAction(Intent.ACTION_DREAMING_STOPPED);
             registerReceiver(mDreamReceiver, filter);
-            checkFullscreen();
         }
```
maniac103: IMHO we should check whether this receiver is actually (still) needed. I'd expect `onResume()` being called when coming back from daydream.
@digitaldan Do you remember details of why this is/was needed?

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -185,6 +187,43 @@ public static void setActivityTheme(@NonNull final Activity activity, String the
...
+        if (isEnabled) {
+            uiOptions = uiOptions
+                    | View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
+                    | View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY
+                    | View.SYSTEM_UI_FLAG_FULLSCREEN;
+        } else {
+            uiOptions = uiOptions
+                    & View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
+                    & View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY
+                    & View.SYSTEM_UI_FLAG_FULLSCREEN;
```
maniac103: This seems off. It should rather be `uiOptions &= ~(...flags...)`

Suggestion:
```
final int flags = View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
        | View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY
        | View.SYSTEM_UI_FLAG_FULLSCREEN;
if (isEnabled) {
    uiOptions |= flags;
} else {
    uiOptions &= ~flags;
}
```

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -185,6 +187,43 @@ public static void setActivityTheme(@NonNull final Activity activity, String the
...
+        }
+        activity.getWindow().getDecorView().setSystemUiVisibility(uiOptions);
+    }
+
+    /**
+     * If we are 4.4 we can use fullscreen mode and Daydream features
+     */
```
maniac103: Comment seems off here. Maybe use `// The system UI flags for fullscreen mode only work on 4.4+`

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -185,6 +187,43 @@ public static void setActivityTheme(@NonNull final Activity activity, String the
         return R.style.HABDroid_Light;
     }
 
+    /**
+     * If fullscreen is enabled and we are on at least android 4.4 set
+     * the system visibility to fullscreen + immersive + noNav
+     *
+     * @author Dan Cunningham
+     */
```
maniac103: Comment really needed?

mueller-ma:
> I'm wondering whether having a common activity base class would make sense for this.

Makes sense, since we have to override two functions now: onCreate() and onResume().

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -185,6 +167,32 @@ public static void setActivityTheme(@NonNull final Activity activity, String the
         return R.style.HABDroid_Light;
     }
 
+    public static void checkFullscreen(Activity activity) {
+        checkFullscreen(activity, isFullscreenEnabled(activity));
+    }
+
+    public static void checkFullscreen(Activity activity, boolean isEnabled) {
```
mueller-ma: Should probably remove this Checkstyle check: "Missing a Javadoc comment."

mobile/src/main/java/org/openhab/habdroid/util/BaseActivity.java
```diff
@@ -0,0 +1,39 @@
+package org.openhab.habdroid.util;
```
maniac103: This one should live in the `ui` package.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -185,6 +167,32 @@ public static void setActivityTheme(@NonNull final Activity activity, String the
         return R.style.HABDroid_Light;
     }
 
+    public static void checkFullscreen(Activity activity) {
```
maniac103: Why not move those as protected methods to `BaseActivity`?

mobile/src/main/java/org/openhab/habdroid/util/BaseActivity.java
mueller-ma: But isn't it just a utility?

maniac103: In a sense, yes. But it's also the base class to all activities, which clearly is UI. What I'd like to avoid is the util package becoming a dumping ground for stuff which 'doesn't have a better place', and as an activity (even an abstract one) clearly is UI, I think it should live there.

mobile/src/main/java/org/openhab/habdroid/util/BaseActivity.java
```diff
@@ -0,0 +1,39 @@
...
+import android.graphics.BitmapFactory;
+import android.os.Build;
+import android.os.Bundle;
+import android.util.TypedValue;
+import androidx.annotation.CallSuper;
+import androidx.appcompat.app.AppCompatActivity;
+
+import org.openhab.habdroid.R;
+
+public class BaseActivity extends AppCompatActivity {
```
maniac103: I wonder whether this shouldn't be made `abstract`?

mobile/src/main/java/org/openhab/habdroid/util/Util.java
mueller-ma: Done

PR #1317, opened 2019-04-21
Fetch about screen information async

mueller-ma:


maniac103:
What problem does this change solve? The about list is populated in a background thread already, so AFAICT this would only help with showing 'something' fast for very (VERY) slow servers.

mueller-ma:
Some months ago myopenhab had some performance issues and was pretty slow. When I opened the about screen on a hardware-slow device, it took a few seconds until it was displayed. With this patch, the list is shown instantly and some fields display "Loading" until the values are fetched.

PR #1318, opened 2019-04-21
Improve connection check reachablility

mueller-ma:
Don't fail on a timeout instantly, but increase the retry count.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
+++ b/mobile/src/main/java/org/openhab/habdroid/core/connection/AbstractConnection.java
@@ -100,7 +100,7 @@ public abstract class AbstractConnection implements Connection {
 
     private Socket createConnectedSocket(InetSocketAddress socketAddress) {
         Socket s = new Socket();
-        for (int retries = 0; retries < 20; retries++) {
+        for (int retries = 0; retries < 10; retries++) {
             try {
                 s.connect(socketAddress, 1000);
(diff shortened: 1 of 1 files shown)


maniac103:
What's the use case for this ... IOW, in which cases does that help?

mueller-ma:
On one device I get SocketTimeoutExceptions quite often, but not always, when I open the app. Currently if the first connection attempt throws a timeout exception, we wait 1 second in total. Ten retries in total need 2 seconds. Instead of failing after the first timeout, increase `retries` by 5 (5*200 == 1000). This way the function won't run longer than before.

maniac103:
It will (potentially) run a bit longer than before: imagine a timeout happening every time; it will run 4 seconds (1 second timeout x 4 runs [with retries being 0, 5, 10 and 15]). Probably not that much of a big deal (and can be fixed by increasing by 10), I was just wondering _why_ that would happen.

mueller-ma:
> I was just wondering why that would happen

I guess this is an issue with my Nexus 9 running Lineage 14, maybe due to some too aggressive power-saving WiFi disconnects. I don't have this issue with other devices on the same WiFi and openHAB server.

maniac103:
OK, fair enough. Is doubling the check time in case of actual failure OK though? Why not keep `retries` at 10 and increase by 3 or 4 in case of timeout?

mueller-ma:
Ok, I reverted the max retries back to 10.

Issue #1295, opened 2019-04-09
Preparation for Android P and Q


mueller-ma:
- [x] Wifi cannot be enabled by apps anymore (https://developer.android.com/preview/privacy/camera-connectivity#restriction-enable-disable-wifi). Instead Google recommends using settings panels (https://developer.android.com/preview/features#settings-panels). Since they announced to release a backport of settings panels in AndroidX and Android Q isn't released yet, we can wait for the backport to be released.  #1458

- [x] Use settings panel to enable NFC in WriteTagActivity. #1516

- [x] android.preference library is now deprecated (https://developer.android.com/preview/behavior-changes-all#preferences)
- [x] The prompt for device pin or fingerprint is deprecated. Instead use https://developer.android.com/preview/features#improved-biometric-auth.
- [ ] Notification channel groups can be blocked. openHAB shouldn't post notifications on these channels (https://developer.android.com/about/versions/pie/android-9.0#channel_settings_broadcasts_and_do_not_disturb).
- [x] This shouldn't be an issue, because WebViews are only called from MainActivity: https://developer.android.com/about/versions/pie/android-9.0-changes-28#web-data-dirs

PR #1325, opened 2019-04-25
Fix PreferencesActivity crash on API < 17

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.java
@@ -92,8 +92,8 @@ public class PreferencesActivity extends AppCompatActivity {
 
     @Override
     public boolean onOptionsItemSelected(MenuItem item) {
-        if (isFinishing() ||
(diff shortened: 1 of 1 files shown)


maniac103:
I wonder whether simply calling just `isFinishing()` wouldn't be sufficient here?

mueller-ma:
Good question. Since I cannot reproduce this crash, I've looked up on the Internet and found this fix.

maniac103:
Probably could/should check `FragmentManager.isStateSaved()` instead, and move the check into the respective case statement.

mueller-ma:
`FragmentManager.isStateSaved()` isn't available on all API levels we support. This would require us changing the whole PreferenceActivity to use the AppCompat version of PreferenceFragment.

maniac103:
Wouldn't that be a good either either way anyway? ;-)

mueller-ma:
Definitely, but that's a larger task and I want to focus on the pending PRs and #257 for now. It's tracked in https://github.com/openhab/openhab-android/issues/1295.
I removed the `isDestroyed()` call.

PR #1327, opened 2019-04-27
Remove removed features from Readme

mueller-ma:


mueller-ma:
Added `Signed-off-by: mueller-ma <mueller-ma@users.noreply.github.com>` when merging.

PR #1331, opened 2019-04-30
Implement row click action for section switch

### LINKED ISSUE #1328: Add row click for section switch if it has two mappings
Add row click for section switch if it has two mappings and currently the state is equal to one of the two mappings.

mueller-ma:
Send state if there's one mapping. Toggle state if there are two
mappings and the current state is equal to one of the mappings.

Fixes #1328

@maniac103 Is there a reason why you have chosen to hide spare views instead of removing them from the radio group?


maniac103:
> Is there a reason why you have chosen to hide spare views instead of removing them from the radio group?

Yes. I want to avoid view reinflation when the view is reused when scrolling the widget list and/or reused in another page (due to the shared view pool). View reinflation is an expensive operation, while view hiding is not.

mueller-ma:
Ok, so `getVisibleViewCount()` is needed.

PR #1336, opened 2019-05-03
Update screenshots in docs

mueller-ma:
Pixel 2 frame, generated with https://developer.android.com/distribute/marketing-tools/device-art-generator


mueller-ma:
@Confectrician Is there a way to preview this change? I want to display the screenshots in a row.

Confectrician:
Nope or at least there is no known to me.
We could check if the vuepress style is documented and has some grid system which we could use through a html class attribute.

mueller-ma:
Ok, we'll see...

Confectrician:
I have rescheduled the jenkins job that will fetch the external ressources, so we dont have to wait too much.

Confectrician:
https://github.com/vuejs/vuepress/issues/578

working with html is basically the correct way, when we want to use no special extension that add resizing to the supported markdown features.

Confectrician:
Looks good 👍 

https://www.openhab.org/docs/apps/android.html

mueller-ma:
Nice :+1:

PR #1338, opened 2019-05-04
Add Tasker/Locale action plugin

mueller-ma:
This plugin works with
https://play.google.com/store/apps/details?id=com.twofortyfouram.locale
and
https://play.google.com/store/apps/details?id=net.dinglisch.android.taskerm.

See #257

TODO:

- [x] Make list searchable
- [x] Hide read-only items
- [x] Add more predefined states to pick from
- [x] Add feature to docs and app store description?


mueller-ma:
Here's the apk: [mobile-foss-beta-debug.zip](https://github.com/openhab/openhab-android/files/3144155/mobile-foss-beta-debug.zip)

- Rename it to .apk
- Install my apk

Confectrician:
Works already very well.
Would it be possible to also GET an item stat instead of only setting one?

mueller-ma:
>  Would it be possible to also GET an item state instead of only setting one?

That would be a condition plugin. Yes, it's possible, but not in this PR.

Confectrician:
>That would be a condition plugin. Yes, it's possible, but not in this PR.

Ah sorry, i think we have discussed this already somewhere.
I remember now.
Thanks for clarifiaction.

mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerActivity.java
```diff
@@ -0,0 +1,272 @@
...
+    @Override
+    public void onItemClicked(Item item) {
+        if (item == null) {
+            return;
+        }
+
+        ArrayList<String> labels = new ArrayList<>();
+        ArrayList<String> commands = new ArrayList<>();
+
+        if (item.isOfTypeOrGroupType(Item.Type.Switch)) {
```
mueller-ma: This is basically the same mapping (item to possible/common) state as in WidgetListFragment, except that WidgetListFragment also looks at the widget.
@maniac103 Do you think it makes sense to add the functions `addCommandsForItem(Item item, ArrayList<String> labels,ArrayList<String> commands)` and `addCommandsForWidget(Widget widget, ArrayList<String> labels,ArrayList<String> commands)`, use the forItem function here. And to use both in WidgetListFragment and remove duplicate commands from the ArrayLists?

mueller-ma:
Maybe I should add a pref to enable this plugin. Otherwise any app can send item updates.

mobile/src/main/java/org/openhab/habdroid/model/Item.java
```diff
@@ -47,6 +47,8 @@
 
     public abstract String name();
     public abstract String label();
+    @Nullable
+    public abstract String icon();
```
maniac103: Why not name this `category` to follow the JSON field?

mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerActivity.java
```diff
@@ -0,0 +1,277 @@
...
+                .setTitle(R.string.item_picker_dialog_title)
+                .setItems(labelArray, (dialog, which) -> {
+                    if (which < labelArray.length - 1) {
+                        finish(true, item.name(), commands.get(which));
+                    } else {
+                        final EditText input = new EditText(this);
+                        LinearLayout.LayoutParams lp = new LinearLayout.LayoutParams(
+                                LinearLayout.LayoutParams.MATCH_PARENT,
+                                LinearLayout.LayoutParams.MATCH_PARENT);
+                        input.setLayoutParams(lp);
```
maniac103: Is this really needed? IOW: what happens when omitting this?

mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerAdapter.java
```diff
@@ -0,0 +1,174 @@
...
+import org.openhab.habdroid.ui.widget.WidgetImageView;
+import org.openhab.habdroid.util.Constants;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Locale;
+
+public class ItemPickerAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder>
```
maniac103: The generic type should be `ItemViewHolder` (you only have one type of item)

mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerAdapter.java
```diff
@@ -0,0 +1,174 @@
...
+            mItemNameView.setText(item.name());
+            mItemLabelView.setText(item.label());
+            mItemTypeView.setText(item.type().toString());
+
+            Connection connection;
+            try {
+                connection = ConnectionFactory.getUsableConnection();
+            } catch (ConnectionException e) {
+                connection = null;
+            }
```
maniac103: We should probably introduce a `getUsableConnectionOrNull` method to `ConnectionFactory` to avoid the throwing overhead when we're not interested in the actual exception. I'll do this in the Kotlin conversion.

mobile/src/main/java/org/openhab/habdroid/util/SuggestCommandsFactory.java
```diff
@@ -0,0 +1,149 @@
...
+
+import java.util.List;
+
+public class SuggestCommandsFactory {
+    Context mContext;
+    List<String> mCommands;
+    List<String> mLabels;
+    boolean mShowUndef;
+
+    public SuggestCommandsFactory(Context context, List<String> commands, List<String> labels,
```
maniac103: I'd prefer to make the lists getters instead of passing them into the constructor. Basically this
```
private final List<String mCommands = new ArrayList<>();
private final List<String mLabels = new ArrayList<>();
[...]
public SuggestCommandsFactory(Context context, boolean showUndef) {
    [...]
}

public List<String> getCommands() {
    return mCommands;
}
public List<String> getLabels() {
    return mLabels;
}
```

(Could maybe even use a single `List<Pair<String, String>> mCommandsAndLabels`)

That way it becomes much clearer from the usage of the class what the class actually does.

mobile/src/main/java/org/openhab/habdroid/util/SuggestCommandsFactory.java
```diff
@@ -0,0 +1,149 @@
...
+        } else if (item.isOfTypeOrGroupType(Item.Type.Contact)) {
+            // Contact items cannot receive commands
+        } else if (item.isOfTypeOrGroupType(Item.Type.Dimmer)) {
+            addOnOffCommands();
+            addIncreaseDecreaseCommands();
+            addCommonPercentCommands();
+        } else if (item.isOfTypeOrGroupType(Item.Type.Number)) {
+            // Don't suggest numbers that might be totally out of context if there's already
+            // at least one command
+            if (mCommands.size() == 0) {
```
maniac103: `isEmpty()`

mobile/src/main/java/org/openhab/habdroid/util/SuggestCommandsFactory.java
```diff
@@ -0,0 +1,149 @@
...
+        } else {
+            add("UNDEF", R.string.nfc_action_undefined);
+        }
+    }
+    
+    private void add(String commandAndlabel) {
+        if (!mCommands.contains(commandAndlabel)) {
+            mCommands.add(commandAndlabel);
+            mLabels.add(commandAndlabel);
+        }
```
maniac103: Can shorten to `add(commandAndLabel, commandAndLabel);`

mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerAdapter.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.List;
+import java.util.Locale;
+
+public class ItemPickerAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder>
+        implements View.OnClickListener {
+    public interface ItemClickListener {
+        void onItemClicked(Item item);
+    }
+
+    private ArrayList<Item> mItems = new ArrayList<>();
```
maniac103: `mFilteredItems`?

mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerAdapter.java
```diff
@@ -0,0 +1,174 @@
...
+                .getString(Constants.PREFERENCE_ICON_FORMAT, "");
+        mInflater = LayoutInflater.from(context);
+        mItemClickListener = itemClickListener;
+    }
+
+    public void addLoadedItems(List<Item> items) {
+        mItems.clear();
+        mItems.addAll(items);
+        Collections.sort(mItems, new ItemNameComparator());
+        mAllItems = new ArrayList<>(mItems);
```
maniac103: Why not `clear()` + `addAll()`?

mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerAdapter.java
```diff
@@ -0,0 +1,174 @@
...
+
+    public ItemPickerAdapter(Context context, ItemClickListener itemClickListener) {
+        super();
+        mIconFormat = PreferenceManager.getDefaultSharedPreferences(context)
+                .getString(Constants.PREFERENCE_ICON_FORMAT, "");
+        mInflater = LayoutInflater.from(context);
+        mItemClickListener = itemClickListener;
+    }
+
+    public void addLoadedItems(List<Item> items) {
```
maniac103: `setItems()`? After all, you're replacing the list, not amending to it.

mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerAdapter.java
mueller-ma: Because I already call clear(), addAll() and sort() on mItems. After that IMO copying is smarter than calling the three functions again.

maniac103: I didn't say you should re-sort, but do `addAll(mItems)` ;-) I think clear + addAll is better because it avoids the memory reallocation.

mueller-ma: I thought you mean `mAllItems.addAll(items)`.

mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerActivity.java
mueller-ma: Actually it does nothing, because `setLayoutParams()` must be called after `AlertDialog#show()`. And it should be `FrameLayout.LayoutParams`, otherwise the app crashes. I removed it.

mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerActivity.java
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        SuggestCommandsFactory suggestCommandsFactory =
+                new SuggestCommandsFactory(this, true);
+        suggestCommandsFactory.fill(item);
+
+        List<String> labels = suggestCommandsFactory.getLabels();
+        List<String> commands = suggestCommandsFactory.getCommands();
+
+        labels.add(getString(R.string.item_picker_custom));
```
mueller-ma: This shouldn't be shown if makes no sense, because `commands` already contains all possible states, e.g. for Switches.

mueller-ma:
TODO:

- [ ] Auto open/hide keyboard on clicking "Custom"
- [ ] Fix UoM: Units aren't sent to server for Setpoints with units

mueller-ma:
> Maybe I should add a pref to enable this plugin. Otherwise any app can send item updates.

@maniac103 WDYT?

maniac103:
@mueller-ma I agree, making this opt-in sounds like a good idea. Pref naming suggestion: 'Tasker integration' with summary 'Enable Tasker action plugin which allows updating Items via Tasker.' ... and only show it if either the pref is enabled at page enter time or Tasker is installed.

mueller-ma:
Should something happen if an app wants to update an item without permission, e.g. show a toast?

maniac103:
I personally wouldn't do it. If the plugin is disabled, just treat it as such and keep it as silent as possible.

mueller-ma:
RTM from my side.

Issue #1342, opened 2019-05-05
Inconsistent web widget behavior


Sonic-Amiga:
I am running BasicUI on wall panel. I added Webview, displaying weather informer from one of local providers. When using this, i notice two problems:
1. The provider insists on having links to itself in the widget, it doesn't work if they are removed. These links have target="_blank" attribute, so they would open in a new browser window, but instead the WebWidget jumps to the new URL itself, breaking the UI.
2. The widget has a fixed width, which is smaller than panel display width. This leaves colored rectangle around the content, which looks ugly. Setting transparent background via CSS results in white color, apparently the WebView itself is not transparent

Sonic-Amiga:
Here is the proposed fix: https://github.com/openhab/openhab-android/pull/1343
In the long run perhaps it would be nice to have preference item to enable or disable opening these links in external browser. Sometimes it can be handy, for instance my weather widget would open 2-week forecast. But even simple disable is better than current behavior.

mueller-ma:
> _blank: Load the URL into a new browsing context. This is usually a tab, but users can configure browsers to use new windows instead.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

Links with `target="_blank"` should always be opened in a browser, no setting needed here.

Sonic-Amiga:
I imagined a dedicated setup where HABDroid replaces the launcher. In this case perhaps you would not want the user to escape at all.
 OK, makes sense too. Let's do this first, then add preference item if really needed or requested.

mueller-ma:
Even if a preference is requested, I still disagree. The website explicitly requests to be opened in a new tab/window.

Sonic-Amiga:
I've tested it myself, and you know what... This isn't bad. Now i can tap on the informer and see 2 weeks forecast in the browser. This is convenient.

Sonic-Amiga:
Since the patch has landed, i am closing this. There's a separate issue reported against core regarding transparency

PR #1343, opened 2019-05-05
[#1342] Improve web widget in BasicUI

Sonic-Amiga:
Set supportsMultipleWindows for the WebView. This prevents the WebView from
following links with target="_blank", which previously were opened in the same
WebView, corrupting the UI. The same adjustment is also made for HABPanel, can
be good there too.

Also set WebWidget background to transparent; this enables to use transparent
HTML background. Useful for e.g. weather informers.

Signed-off-by: Pavel Fedin <pavel_fedin@mail.ru>

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
@@ -1089,11 +1089,9 @@ public class WidgetAdapter extends RecyclerView.Adapter<WidgetAdapter.ViewHolder
             }
 
             String url = mConnection.getAsyncHttpClient().buildUrl(widget.url()).toString();
-            Util.applyAuthentication(mWebView, mConnection, url);
+            Util.initWebView(mWebView, mConnection, url);
             mWebView.setWebViewClient(new AnchorWebViewClient(url,
                     mConnection.getUsername(), mConnection.getPassword()));
-            mWebView.getSettings().setDomStorageEnabled(true);
-            mWebView.getSettings().setJavaScriptEnabled(true);
             mWebView.loadUrl(url);
         }
     }
--- a/mobile/src/main/java/org/openhab/habdroid/util/Util.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Util.java
@@ -27,7 +27,6 @@ import android.util.TypedValue;
 import android.webkit.WebChromeClient;
 import android.webkit.WebView;
 import android.webkit.WebViewDatabase;
-import android.widget.Toast;
 import androidx.annotation.AttrRes;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -303,21 +302,22 @@ public class Util {
                 Message href = view.getHandler().obtainMessage();
                 view.requestFocusNodeHref(href);
                 String url = href.getData().getString("url");
-                openInBrowser(url, view.getContext());
+                openInBrowser(view.getContext(), url);
                 return false;
             }
         });
     }
 
-    public static void openInBrowser(@Nullable String url, Context context) {
-        if (!TextUtils.isEmpty(url)) {
-            Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
-            try {
-                context.startActivity(intent);
-            } catch (ActivityNotFoundException e) {
-                Toasty.error(context, R.string.error_no_browser_found,
-                        Toast.LENGTH_LONG, true).show();
-            }
+    public static void openInBrowser(Context context, @Nullable String url) {
+        if (TextUtils.isEmpty(url)) {
+            Log.e(TAG, "Got empty url");
+            return;
+        }
+        Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+        try {
+            context.startActivity(intent);
+        } catch (ActivityNotFoundException e) {
+            Toasty.error(context, R.string.error_no_browser_found, Toasty.LENGTH_LONG).show();
         }
     }
 
--- a/mobile/src/main/java/org/openhab/habdroid/util/Util.java
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Util.java
@@ -396,7 +396,7 @@ public class Util {
      */
     public static void showToast(Context context, CharSequence message) {
         new Handler(Looper.getMainLooper()).post(() -> Toasty.custom(context, message,
-                R.drawable.ic_openhab_appicon_24dp, R.color.openhab_orange, Toast.LENGTH_SHORT,
+                R.drawable.ic_openhab_appicon_24dp, R.color.openhab_orange, Toasty.LENGTH_SHORT,
                 true, true).show());
     }
 }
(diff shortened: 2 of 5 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
```diff
@@ -1069,6 +1069,8 @@ public void bind(Widget widget) {
                     mConnection.getUsername(), mConnection.getPassword()));
             mWebView.getSettings().setDomStorageEnabled(true);
             mWebView.getSettings().setJavaScriptEnabled(true);
+            mWebView.getSettings().setSupportMultipleWindows(true);
```
mueller-ma: I guess we should move the duplicate (here and in WebViewFragment) call to a helper function. Maybe to `Util.applyAuthentication()` and rename it to `Util.initWebview()`.

> If set to true, WebChromeClient#onCreateWindow must be implemented by the host application. The default is false.

https://developer.android.com/reference/android/webkit/WebSettings.html#setSupportMultipleWindows(boolean)

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
```diff
@@ -1069,6 +1069,8 @@ public void bind(Widget widget) {
...
+            mWebView.setBackgroundColor(Color.TRANSPARENT);
```
mueller-ma: This breaks websites that rely on having a white background and don't set one in CSS, e.g. default Apache HTTP error sites.
Maybe there's a way to set the webview background to transparent only if the website background is set to transparent.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
Sonic-Amiga: And if you look at ChromeWebClient, the default implementation of onCreateWindow() always returns false, so those links get blocked.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
Sonic-Amiga: Errr... After all it's a widget on a control panel, it's not a generic browser. Aren't we supposed to put only specially designed content there?
In other words: is there any real-world case which would break?
 I'll have a look if there's a cheap way to dig into loaded DOM.

Sonic-Amiga:
Please feel free to find the updated PR.
I've left out transparency, let's discuss first in the issue.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
Sonic-Amiga: I've looked at this from different perspectives.
1. Indeed. If a web site doesn't specify a background, it will be taken from what we set for the WebView.
2. Plain text is still readable using "dark" theme. So the user will see 404.
3. There's no reliable way to inspect the whole DOM inside the WebView. You could only get HTML as a string, but this then means you would need to parse the whole thing, perhaps loading external CSS... Ending up in reimplementing your own browser. Using some magic tag/comment would be a hack and again you can think of a case where this woudn't work (e. g. the widget contents is hosted completely elsewhere).
4. If you still think it's an issue, a cheap way to solve this would be to add a setting.
5. Perhaps the most user-friendly and flexible way would be to allow to specify backround RGBA value in the sitemap for a particular widget. But i guess this would be hard to implement since it requires changes to OpenHAB core too.
 So what's your opinion?
 BTW, HABPanel does use transparent background and apparently nobody suffers.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -281,6 +284,26 @@ public static void applyAuthentication(WebView webView, Connection connection, S
...
         }
+
+        webView.getSettings().setDomStorageEnabled(true);
+        webView.getSettings().setJavaScriptEnabled(true);
+        webView.getSettings().setSupportMultipleWindows(true);
+
+        webView.setWebChromeClient(new WebChromeClient() {
+            @Override
+            public boolean onCreateWindow(WebView view, boolean dialog, boolean userGesture,
+                    android.os.Message resultMsg) {
```
maniac103: As you import it above you don't need the fully qualified name here.

Sonic-Amiga: Fixed

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
mueller-ma: > But i guess this would be hard to implement since it requires changes to OpenHAB core too.

If something _has to be_ changed in openHAB core, you can open an issue there. Something like `transparentBackground=true` to set the background transparent. Defaults to false. Let's discuss it there as it also affects Basic UI.

>  BTW, HABPanel does use transparent background and apparently nobody suffers.

I guess no one suffers, because HABPanel has it's own background color.

Sonic-Amiga: Well, BasicUI has it too.

Sonic-Amiga: https://github.com/openhab/openhab-core/issues/809

Sonic-Amiga:
Rebased and pushed again, now all tests have passed. I don't know why there was a failure, i didn't change anything.

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -281,6 +284,26 @@ public static void applyAuthentication(WebView webView, Connection connection, S
...
+            @Override
+            public boolean onCreateWindow(WebView view, boolean dialog, boolean userGesture,
+                    Message resultMsg) {
+                Message href = view.getHandler().obtainMessage();
+                view.requestFocusNodeHref(href);
+                String url = href.getData().getString("url");
+                if (url != null) {
+                    Context context = view.getContext();
+                    Intent browserIntent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+                    context.startActivity(browserIntent);
```
mueller-ma: Should catch having no browser installed: https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/ui/AboutActivity.java#L334-L339

Sonic-Amiga:
Done, with some more code factoring out

mobile/src/main/java/org/openhab/habdroid/util/Util.java
```diff
@@ -281,6 +287,34 @@ public static void applyAuthentication(WebView webView, Connection connection, S
...
+                if (url != null) {
+                    openInBrowser(url, view.getContext());
+                }
+                return false;
+            }
+        });
+    }
+
+    public static void openInBrowser(String url, Context context) {
+        Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
```
mueller-ma: IMO you should check here for `!TextUtils.isEmpty(url)` and for url being a valid url. Then add `@Nullable String url` in the method args and remove the null check in `onCreateWindow()`.

Sonic-Amiga:
Have fun

PR #1345, opened 2019-05-06
Improve location parsing

maniac103:
Move it to the correct place, and make it more lenient to also cover the 'latitude,longitude,elevation' case.

Fixes issue reported in https://community.openhab.org/t/mapview-in-android-app-does-not-show-the-location/73379

### DIFF
--- a/mobile/src/full/java/org/openhab/habdroid/ui/MapViewHelper.java
+++ b/mobile/src/full/java/org/openhab/habdroid/ui/MapViewHelper.java
@@ -189,7 +189,7 @@ public class MapViewHelper {
 
         private static LatLng toLatLng(ParsedState state) {
             Location location = state != null ? state.asLocation() : null;
-            return state != null
+            return location != null
                     ? new LatLng(location.getLatitude(), location.getLongitude())
                     : null;
         }
--- a/mobile/src/full/java/org/openhab/habdroid/ui/MapViewHelper.java
+++ b/mobile/src/full/java/org/openhab/habdroid/ui/MapViewHelper.java
@@ -2,6 +2,7 @@ package org.openhab.habdroid.ui;
 
 import android.app.AlertDialog;
 import android.content.res.Resources;
+import android.location.Location;
 import android.view.LayoutInflater;
 import android.view.ViewGroup;
 
@@ -151,7 +152,7 @@ public class MapViewHelper {
             if (!mBoundItem.members().isEmpty()) {
                 ArrayList<LatLng> positions = new ArrayList<>();
                 for (Item item : mBoundItem.members()) {
-                    LatLng position = parseLocation(item.state());
+                    LatLng position = toLatLng(item.state());
                     if (position != null) {
                         setMarker(map, position, item, item.label(), canDragMarker);
                         positions.add(position);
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/model/ParsedState.java
```diff
@@ -212,6 +218,28 @@ private static NumberState parseAsNumber(String state, String format) {
...
+    private static Location parseAsLocation(String state) {
+        String[] splitState = state.split(",");
+        // Valid states are either "latitude,longitude" or "latitude,longitude,elevation",
+        // (we ignore elevation in the latter case)
+        if (splitState.length == 2 || splitState.length == 3) {
+            try {
+                Location l = new Location("openhab");
+                l.setLatitude(Float.valueOf(splitState[0]));
+                l.setLongitude(Float.valueOf(splitState[1]));
+                l.setTime(System.currentTimeMillis());
```
mueller-ma: The time when the location was changed in openHAB isn't the time when the user starts the app. I suggest to not set the time here.

mueller-ma:
Can you build and upload a beta apk?

mobile/src/main/java/org/openhab/habdroid/model/ParsedState.java
maniac103: I basically only set the time to make the debug output look nicer, as `Location.toString()` prints some question marks if it isn't set ;-)

mueller-ma:
The binding that causes this issue might always set the third parameter to "0", but other bindings might use it. I prefer to pass it to the MapView.

maniac103:
> I prefer to pass it to the MapView

How do you want to do that? MapView only accepts LatLng instance, which are latitude/longitude pairs ;-)

mueller-ma:
https://developer.android.com/reference/android/location/Location.html#setAltitude(double)
I haven't looked at the Maps API.

maniac103:
Oops, missed that. I'll set it; but maps API indeed doesn't use it.

mobile/src/full/java/org/openhab/habdroid/ui/MapViewHelper.java
```diff
@@ -186,17 +187,11 @@ private static void setMarker(GoogleMap map, LatLng position, Item item,
...
-                try {
-                    return new LatLng(Float.valueOf(splitState[0]), Float.valueOf(splitState[1]));
-                } catch (NumberFormatException e) {
-                    // ignored
-                }
-            }
-            return null;
+        private static LatLng toLatLng(ParsedState state) {
+            Location location = state != null ? state.asLocation() : null;
+            return state != null
```
mueller-ma: Need to check for location == null here.

maniac103: Oops. It's `!=` actually, but thanks, will fix this.

mueller-ma:
Do we want to add a common (full and foss) no-position handling? Like hiding the map and display an "empty view", like in notification list?

maniac103:
> Do we want to add a common (full and foss) no-position handling? Like hiding the map and display an "empty view", like in notification list?

For foss I surely don't ... what's the point in showing a 'no map support' view for every MapView? For full we could do that, but I think that's outside of scope of this PR.

PR #1357, opened 2019-05-14
Add ability to create sitemap shortcuts

### LINKED ISSUE #1322: Ability to create launcher shorts for sitemaps and sitemap subpages
Add the ability to create launcher shorts for sitemaps and sitemap subpages, like it's possible to write nfc tags for these. Suggested UI: Long click any widget => Pin to home screen
I want to improve how sitemaps are opened in #1302 via nfc tags, so I'll make it easy to implement this request.

### LINKED ISSUE #1337: PNG widget icons might be shown too small
### Actual behavior
On devices with high dpi (e.g. Pixel 2), the widget and notification icons (png) are shown too small. The parameter `mSize` isn't honored here: https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/util/AsyncHttpClient.java#L74

### Expected behavior
The PNG icons should be scaled to be shown with the same size than the svg ones.

Yes

#### Client
* Android version: P
* Device model: Pixel 2
* Stock or customized ROM: Stock

mueller-ma:
* Fixes #1322
* Fixes #1337

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -10,7 +10,6 @@
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.NFC" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
-    <uses-permission android:name="android.permission.INSTALL_SHORTCUT"/>
     <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT" />
 
     <uses-feature
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -10,6 +10,8 @@
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.NFC" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <uses-permission android:name="android.permission.INSTALL_SHORTCUT"/>
+    <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT" />
 
     <uses-feature
         android:name="android.hardware.location.gps"
--- a/mobile/src/main/java/org/openhab/habdroid/model/LinkedPage.java
+++ b/mobile/src/main/java/org/openhab/habdroid/model/LinkedPage.java
@@ -25,6 +25,7 @@ public abstract class LinkedPage implements Parcelable {
     public abstract String id();
     public abstract String title();
     public abstract String icon();
+    public abstract String iconPath();
     public abstract String link();
 
     @AutoValue.Builder
@@ -32,6 +33,7 @@ public abstract class LinkedPage implements Parcelable {
         public abstract Builder id(String id);
         public abstract Builder title(String title);
         public abstract Builder icon(String icon);
+        public abstract Builder iconPath(String iconPath);
         public abstract Builder link(String link);
 
         public LinkedPage build() {
@@ -67,6 +69,7 @@ public abstract class LinkedPage implements Parcelable {
                 .id(id)
                 .title(title)
                 .icon(icon)
+                .iconPath(String.format("images/%s.png", icon))
                 .link(link)
                 .build();
     }
@@ -75,10 +78,12 @@ public abstract class LinkedPage implements Parcelable {
         if (jsonObject == null) {
             return null;
         }
+        String icon = jsonObject.optString("icon", null);
         return new AutoValue_LinkedPage.Builder()
                 .id(jsonObject.optString("id", null))
                 .title(jsonObject.optString("title", null))
-                .icon(jsonObject.optString("icon", null))
+                .icon(icon)
+                .iconPath(String.format("icon/%s", icon))
                 .link(jsonObject.optString("link", null))
                 .build();
     }
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.java
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.java
@@ -22,7 +22,6 @@ import android.os.AsyncTask;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.text.TextUtils;
-import android.util.DisplayMetrics;
 import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.View;
@@ -203,7 +202,9 @@ public class WidgetListFragment extends Fragment
 
         if (widget.linkedPage() != null) {
             labels.add(getString(R.string.nfc_action_to_sitemap_page));
-            labels.add(getString(R.string.home_shortcut_pin_to_home));
+            if (ShortcutManagerCompat.isRequestPinShortcutSupported(getContext())) {
+                labels.add(getString(R.string.home_shortcut_pin_to_home));
+            }
         }
 
         if (!labels.isEmpty()) {
@@ -212,12 +213,12 @@ public class WidgetListFragment extends Fragment
                     .setTitle(R.string.nfc_dialog_title)
                     .setItems(labelArray, (dialog, which) -> {
                         if (which < commands.size()) {
-                            startActivityForResult(WriteTagActivity.createItemUpdateIntent(
+                            startActivity(WriteTagActivity.createItemUpdateIntent(
                                     getActivity(), widget.item().name(), commands.get(which),
-                                    labels.get(which), widget.item().label()), 0);
+                                    labels.get(which), widget.item().label()));
                         } else if (which == commands.size()) {
-                            startActivityForResult(WriteTagActivity.createSitemapNavigationIntent(
-                                    getActivity(), widget.linkedPage().link()), 0);
+                            startActivity(WriteTagActivity.createSitemapNavigationIntent(
+                                    getActivity(), widget.linkedPage().link()));
                         } else {
                             createShortcut(widget.linkedPage());
                         }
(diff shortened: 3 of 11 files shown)


mobile/src/main/java/org/openhab/habdroid/util/HttpClient.java
```diff
@@ -111,4 +121,76 @@ protected Call prepareCall(String url, String method, Map<String, String> additi
         }
         return builder;
     }
+
+    protected static Bitmap getBitmapFromSvgInputstream(ResponseBody body, int size)
```
maniac103: This method is misnamed; the `Svg` in there should be dropped.

mobile/src/main/java/org/openhab/habdroid/util/SyncHttpClient.java
```diff
@@ -61,6 +63,10 @@ public HttpTextResult asText() {
             return new HttpTextResult(this);
         }
 
+        public HttpBitmapResult asBitmap(int sizeInPixel) {
```
maniac103: `sizeInPixels`

mobile/src/main/java/org/openhab/habdroid/model/LinkedPage.java
```diff
@@ -75,10 +78,12 @@ public static LinkedPage fromJson(JSONObject jsonObject) {
         if (jsonObject == null) {
             return null;
         }
+        String icon = jsonObject.optString("icon", null);
         return new AutoValue_LinkedPage.Builder()
                 .id(jsonObject.optString("id", null))
                 .title(jsonObject.optString("title", null))
-                .icon(jsonObject.optString("icon", null))
+                .icon(icon)
+                .iconPath(String.format("icon/%s", icon))
```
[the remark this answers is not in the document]
mueller-ma: The format differs for openHAB 1 and 2. This is why I put it here.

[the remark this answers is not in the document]
maniac103: Makes sense ... however, looking at the format, it seems like the cloud notification code uses the OH1 format ('images/%s.png') as well ... I wonder whether we need to do something there; and if we do, I wonder *what* to do there (no ServerProperties at that point).

mobile/src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,8 @@
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.NFC" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <uses-permission android:name="android.permission.INSTALL_SHORTCUT"/>
+    <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT" />
```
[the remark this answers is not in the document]
mueller-ma: I've seen this in several tutorials about creating home screen shortcuts. I haven't seen it in the official docs though.

mobile/src/main/java/org/openhab/habdroid/model/LinkedPage.java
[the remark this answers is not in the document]
mueller-ma: TBH I don't really like that api at all. The app must create a socket to the server in order to download the icon and display it in the notification. This is discouraged by Google and I'm afraid a slow connection will cause the app to be killed before the notification is shown.
I haven't found time so far to write an issue, but what about putting the icon inside the FCM message, even if that means that we might have to drop png support for those icons (due to size restrictions)?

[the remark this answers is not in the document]
maniac103: > This is discouraged by Google and I'm afraid a slow connection will cause the app to be killed before the notification is shown.

Why do you think that's the case? `FcmMessageListenerService` ultimately derives from `IntentService`, which executes in a background thread.

Do you mean that quote?
```
Because a small portion of the Android mobile population are on high latency networks, avoid opening a connection to your servers before displaying a notification. Calling back to the server before the end of the allowed processing time may be risky for users on high latency networks. 
```
If so, to comply to that, we could look into posting the notification without icon at first and then, after downloading the icon, update it to include the icon.

> but what about putting the icon inside the FCM message, even if that means that we might have to drop png support for those icons (due to size restrictions)?

Not sure whether openhab-cloud people will be fond of storing the icons in their database...and if they don't do that, the very same question applies to showing the list of (previous) notifications.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.java
```diff
@@ -184,26 +203,30 @@ public void onItemLongClicked(final Widget widget) {
...
             final String[] labelArray = labels.toArray(new String[0]);
             new AlertDialog.Builder(getActivity())
                     .setTitle(R.string.nfc_dialog_title)
                     .setItems(labelArray, (dialog, which) -> {
-                        final Intent writeTagIntent;
                         if (which < commands.size()) {
-                            writeTagIntent = WriteTagActivity.createItemUpdateIntent(getActivity(),
-                                    widget.item().name(), commands.get(which),
-                                    labels.get(which), widget.item().label());
+                            startActivityForResult(WriteTagActivity.createItemUpdateIntent(
```
maniac103: While you're at it, you can make this call `startActivity` instead of `startActivityForResult` here and below. There's nothing listening for a result (no `onActivityResult` implementation) anyway.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.java
```diff
@@ -251,6 +274,102 @@ public void onPause() {
...
+                    // ignored
+                }
+
+                if (context == null || connection == null) {
+                    return null;
+                }
+
+                if (url != null) {
+                    /** Icon size is defined in {@link AdaptiveIconDrawable}. **/
+                    int foregroundSize = (int) Util.convertDpToPixel(46, context);
```
maniac103: I don't seem to be able to find those magic numbers (46/31) in `AdaptiveIconDrawable`'s sources...instead, it says this:
```
 * <li>Both foreground and background layers should be sized at 108 x 108 dp.</li>
 * <li>The inner 72 x 72 dp  of the icon appears within the masked viewport.</li>
 * <li>The outer 18 dp on each of the 4 sides of the layers is reserved for use by the system UI
 * surfaces to create interesting visual effects, such as parallax or pulsing.</li>
 * </ul>
```
So foregroundSize should be 72dp and border size should be 18dp ... what am I missing?

mobile/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
maniac103: Looks like it's needed for legacy (API <= 25) support.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.java
```diff
@@ -184,26 +203,30 @@ public void onItemLongClicked(final Widget widget) {
 
         if (widget.linkedPage() != null) {
             labels.add(getString(R.string.nfc_action_to_sitemap_page));
+            labels.add(getString(R.string.home_shortcut_pin_to_home));
```
maniac103: Should check https://developer.android.com/reference/androidx/core/content/pm/ShortcutManagerCompat.html#isRequestPinShortcutSupported(android.content.Context) before adding this.

mobile/src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,8 @@
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.NFC" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <uses-permission android:name="android.permission.INSTALL_SHORTCUT"/>
```
maniac103: This one doesn't seem to be a thing though.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.java
mueller-ma: First I tried it with these values, but then realized, that "masked viewport" and foreground size are two different things: If I set foregroundSize to 72dp and have a square homescreen icon, the icon from openHAB touches the border of the homescreen icon. So I made the foregroundSize smaller and the border bigger: 46 + (2 x 31) = 108 (the border is added on each site of the icon).
I should add that to the comment.

mobile/src/main/AndroidManifest.xml
mueller-ma: If I'm searching for it, I find a few messengers that use this permission:
https://github.com/signalapp/Signal-Android/blob/master/AndroidManifest.xml#L72
https://github.com/DrKLO/Telegram/blob/master/TMessagesProj/src/main/AndroidManifest.xml#L43
and maybe WhatsApp as well: https://duckduckgo.com/?q=%22android.permission.INSTALL_SHORTCUT%22&t=ffcm&ia=web

maniac103: *shrug* I grepped the Android framework code (which includes support libs) for that string, and while I could find references to the com.android.launcher one, I couldn't find this one.

maniac103: I'm also not convinced e.g. Signal actually needs this permission either. The commit that introduced it seems rather ... random: https://github.com/signalapp/Signal-Android/commit/df674719597226485e1d3ecbd5b320e02ce1f797

mueller-ma: I tested this with API 21 and 25 emulators and both didn't need the `android.permission.INSTALL_SHORTCUT` permission. I'll remove it.

Issue #574, opened 2017-12-29
Add support for Location items (through Mapview widget)


kaikreuzer:
The Android client is lacking one of the standard widgets of openHAB: The [Mapview widget](http://docs.openhab.org/configuration/sitemaps.html#element-type-mapview), which can be used to display a geo location for a "Location" item.

In Basic UI (and Paper UI), we use OSM maps as these are freely available and the best fit for open source usage that does not want to confront its users with any special terms&conditions.

I'd assume that for Android, it might be more natural to use Google maps, though, which should be fine because Android users anyhow already have accepted all kinds of T&Cs from Google.

I leave the decision on how to do it best to the experts here. With all the recent activity on the app, I just wanted to take the opportunity to report this missing feature in the hope that it can be added :-)

FTR: I'd also like to extend the "Mapview" support to refer to a group item. In such a case, all members that are locations should be rendered on a single map. But this is probably something for which we might also want to enhance the sitemap information in the REST API.

mueller-ma:
There are some people running android without play services ;)

kaikreuzer:
Maybe one could offer both options (and use Google maps as a default, if the play services are installed)?

Btw, a suitable location pin can be found [here](https://github.com/htreu/openhab-core/blob/8768ec4396d6a2aa043f380084949b5a0160c72d/bundles/org.openhab.ui.paperui/patch/web/img/logo_pointer.png) in case it is helpful.

maniac103:
Using OSM for the FOSS variant and Google Maps for the Play Store version indeed sounds like a good idea.

maniac103:
I implemented support for Google Maps on top of PR #650: https://github.com/maniac103/openhab.android/commit/cb519d5e78a10c01a224c898c168ad9c3100d89e

It looks like this:
![screenshot_1516882414](https://user-images.githubusercontent.com/781373/35387894-e1e3134c-01d1-11e8-8f1e-60e1895a2705.png)
![screenshot_1516882426](https://user-images.githubusercontent.com/781373/35387895-e1fed3b6-01d1-11e8-856d-a4edbea44036.png)

Thoughts about that UI?

kaikreuzer:
Looks great! But why is this done as a modal window and not directly embedded in the page (as it is done for the web UIs, e.g. Basic UI)?

maniac103:
> Looks great! But why is this done as a modal window and not directly embedded in the page (as it is done for the web UIs, e.g. Basic UI)?

[...]

PR #1358, opened 2019-05-15
Support MapView in foss flavor

mueller-ma:
See #574

### DIFF
--- a/mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.java
+++ b/mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.java
@@ -1,13 +1,255 @@
 package org.openhab.habdroid.ui;
 
+import android.app.AlertDialog;
+import android.content.Context;
+import android.content.res.Resources;
+import android.location.Location;
+import android.os.Handler;
+import android.preference.PreferenceManager;
+import android.util.DisplayMetrics;
+import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.ViewGroup;
 
+import androidx.core.content.ContextCompat;
+
+import org.openhab.habdroid.R;
 import org.openhab.habdroid.core.connection.Connection;
+import org.openhab.habdroid.model.Item;
+import org.openhab.habdroid.model.ParsedState;
+import org.openhab.habdroid.model.Widget;
+import org.openhab.habdroid.util.Util;
+import org.osmdroid.api.IMapController;
+import org.osmdroid.config.Configuration;
+import org.osmdroid.events.MapEventsReceiver;
+import org.osmdroid.tileprovider.tilesource.TileSourceFactory;
+import org.osmdroid.util.BoundingBox;
+import org.osmdroid.util.GeoPoint;
+import org.osmdroid.views.MapView;
+import org.osmdroid.views.overlay.CopyrightOverlay;
+import org.osmdroid.views.overlay.MapEventsOverlay;
+import org.osmdroid.views.overlay.Marker;
+
+import java.util.ArrayList;
+import java.util.Locale;
 
 public class MapViewHelper {
+    private static final String TAG = MapViewHelper.class.getSimpleName();
+
     public static WidgetAdapter.ViewHolder createViewHolder(LayoutInflater inflater,
             ViewGroup parent, Connection connection, WidgetAdapter.ColorMapper colorMapper) {
-        return new WidgetAdapter.GenericViewHolder(inflater, parent, connection, colorMapper);
+        Context context = inflater.getContext();
+        Configuration.getInstance().load(context,
+                PreferenceManager.getDefaultSharedPreferences(context));
+        return new OsmViewHolder(inflater, parent, connection, colorMapper);
+    }
+
+    private static class OsmViewHolder extends WidgetAdapter.LabeledItemBaseViewHolder
+            implements Marker.OnMarkerDragListener {
+        private final MapView mMapView;
+        private final int mRowHeightPixels;
+        private Item mBoundItem;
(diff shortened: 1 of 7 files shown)


mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.java
```diff
@@ -1,13 +1,255 @@
...
+        }
+
+        /**
+         * This method converts dp unit to equivalent pixels, depending on device density.
+         *
+         * @param dp A value in dp (density independent pixels) unit. Which we need to convert into pixels
+         * @param context Context to get resources and device specific display metrics
+         * @return A float value to represent px equivalent to dp depending on device density
+         */
+        private static float convertDpToPixel(float dp, Context context){
```
mueller-ma: I'll use the copy introduced in https://github.com/openhab/openhab-android/pull/1357/files#diff-6a62feee90062d42673ee06ee0626d43R384.

mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.java
```diff
@@ -1,13 +1,255 @@
...
+                    return true;
+                }
+
+                @Override
+                public boolean longPressHelper(GeoPoint p) {
+                    return false;
+                }
+            }));
+
+            new Handler().post(() -> applyPositionAndLabel(mMapView, 15.0f, false, false));
```
maniac103: Is the handler really needed? If so, should probably make it a member variable considering it's used multiple times.

mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.java
```diff
@@ -1,13 +1,255 @@
...
+
+            AlertDialog dialog = new AlertDialog.Builder(itemView.getContext())
+                    .setView(mapView)
+                    .setCancelable(true)
+                    .setNegativeButton(R.string.close, null)
+                    .create();
+
+            dialog.setOnDismissListener(dialogInterface -> {
+                mapView.onPause();
+            });
```
maniac103: Can shorten this to one line.

mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.java
mueller-ma: If I don't use the Handler here, the markers are applied before the MapView is inflated, leading to runtime crashes.

maniac103: I can not reproduce that using the emulator. How does the crash trace look like for you?

mueller-ma:
```
05-10 22:53:23.356 org.openhab.habdroid.beta D/MapViewHelper: North 52.516998, south 52.506340, west 10.388490, east 15.388490
05-10 22:53:23.357 org.openhab.habdroid.beta D/AndroidRuntime: Shutting down VM
05-10 22:53:23.361 org.openhab.habdroid.beta E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.openhab.habdroid.beta, PID: 7685
    java.lang.IllegalArgumentException: north must be in [-85.05112877980658,85.05112877980658]
        at org.osmdroid.util.BoundingBox.set(BoundingBox.java:67)
        at org.osmdroid.views.Projection.refresh(Projection.java:680)
        at org.osmdroid.views.Projection.<init>(Projection.java:99)
        at org.osmdroid.views.Projection.<init>(Projection.java:110)
        at org.osmdroid.views.MapView.zoomToBoundingBox(MapView.java:556)
        at org.osmdroid.views.MapView.zoomToBoundingBox(MapView.java:582)
        at org.openhab.habdroid.ui.MapViewHelper$OsmViewHolder.applyPositionAndLabel(MapViewHelper.java:207)
```

Opened an issue here: https://github.com/osmdroid/osmdroid/issues/1339#issuecomment-491487934

This only occurs if you have more than one markers and remove all the Handlers and the try/catch below.

maniac103: Ah, OK, only tried with a single item. Thanks for that link giving context; let's see what they'll say. The `Handler` seems like a workaround for an issue on their side ... and if we want to keep it, we might rather attach a `ViewTreeObserver.OnGlobalLayoutListener` to see whether the map is already laid out.

maniac103: (And BTW, you're using incorrect terminology there: the view is already `inflated` (XML description was used to create a view hierarchy), it's just not laid out yet)

mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.java
```diff
@@ -1,13 +1,259 @@
...
+            mapView.setMultiTouchControls(true);
+            mapView.setVerticalMapRepetitionEnabled(false);
+            mapView.getOverlays().add(new CopyrightOverlay(itemView.getContext()));
+            mapView.onResume();
+            applyPositionAndLabelWhenReady(mapView, 16.0f, true, true);
+        }
+
+        private void applyPositionAndLabelWhenReady(MapView mapView, float zoomLevel,
+                boolean allowDrag, boolean allowScroll) {
+            mHandler.post(() -> applyPositionAndLabel(mapView, zoomLevel, allowDrag, allowScroll));
```
maniac103: Still not too happy with this, but I realize it's an upstream issue. Should probably merge this now and revisit when they fix this.

PR #1361, opened 2019-05-16
Fix crash in SectionSwitchViewHolder#handleRowClick()

mueller-ma:


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.java
```diff
@@ -785,7 +785,11 @@ protected void handleRowClick() {
             if (visibleChildCount == 1) {
                 onClick(mRadioGroup.getChildAt(0));
             } else if (visibleChildCount == 2) {
-                String state = mBoundItem.state().asString();
+                ParsedState parsedState = mBoundItem.state();
+                if (parsedState == null) {
+                    return;
+                }
+                String state = parsedState.toString();
```
maniac103: `asString`, not `toString` (the latter - IIRC - prints all members)

PR #1390, opened 2019-06-08
Fix Setpoint widget

### LINKED ISSUE #1389: Setpoint with step 0.5 not working after it has been changed in Basic UI
### Actual behaviour
Item:
````
Number ASetpoint "Target Temperature [%.1f °C]"
````
Sitemap:
````
Setpoint item=ASetpoint minValue=23 maxValue=26 step=0.5
````
If I set the value to 24.5 via Basic UI and open the setpoint in the app, everything is working. If I change it via Basic UI to 24, the selectable values in the app are 23, 24, 24, 25, 25, 26, 26.
It's important to use Basic UI to change the item values. If I only use the app, everything is working.

Yes: Widget Overview => Temperature (http://demo.openhab.org:8080/basicui/app?w=0202&sitemap=demo)

### LINKED ISSUE #1397: Setpoint not workig with empty number item
### Actual behaviour
A setpoint entry for a number item that has no value (new item / not persisted) correctely shows "-" as the value, but does not allow to change the value through the up/down buttons or the normal selection pop-up. After assigning a value to the item (e.g. slider element in app; rule or BasicUI), the setpoint entry works as it should, again.
This behaviour completely prevents setting new number items to a value through the app, if there is no other input element for this item (e.g. slider).

### Expected behaviour
The setpoint should allow to change (set) the value of the number item the same way as it does when the item already has a value.

 
### Steps to reproduce
1. Create a new number item in Openhab.
2. Add a setpoint option for this item anywhere on a sitemap.
3. Try to use the up/down/set elements

Not tested.

#### Client
* Android version: 6.0.1
* Device model: Samsung Galaxy S5 (SM-G900F)
* Stock or customized ROM: stock
* App version: 2.7.0
* Build flavor: full 
* Device language: german

#### Server
* Server version: 2.4. ...? (Latest official as of July 2019

#### openHAB Cloud
* Self hosted cloud or myopenhab.org: myopenhab.org, but also happens locally

#### App log
Unavailable as I have no new empty items at the moment

mueller-ma:
* Parse state as Float if step size is a Float. Fixes #1389
* Fix not showing min value, but one step over max value.
* Fix Setpoint if state is null. Fixes #1397

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
@@ -61,18 +61,14 @@ data class ParsedState internal constructor(
 
         internal fun parseAsNumber(state: String, format: String?): NumberState? {
             return when (state) {
-                "ON" -> NumberState(100)
-                "OFF" -> NumberState(0)
+                "ON" -> NumberState(100F)
+                "OFF" -> NumberState(0F)
                 else -> {
                     val spacePos = state.indexOf(' ')
                     val number = if (spacePos >= 0) state.substring(0, spacePos) else state
                     val unit = if (spacePos >= 0) state.substring(spacePos + 1) else null
                     return try {
-                        if (number.indexOf('.') > 0) {
-                            NumberState(number.toFloat(), unit, format)
-                        } else {
-                            NumberState(number.toInt(), unit, format)
-                        }
+                        NumberState(number.toFloat(), unit, format)
                     } catch (e: NumberFormatException) {
                         null
                     }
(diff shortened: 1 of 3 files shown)


maniac103:
The first item, should be handled differently, I think. Continuing my comment https://github.com/openhab/openhab-android/issues/1389#issuecomment-500399772 I dug up what core foes for formatting: https://github.com/openhab/openhab-core/blob/10acf5cf4675ba908ba69361df951bf36e9744a6/bundles/org.openhab.core.compat1x/src/main/java/org/openhab/core/library/types/DecimalType.java#L66

I think we should do the same, handle the value as double internally and format using the rounded value if required by the format string. This should both fix the issue and simplify the code.

maniac103:
Untested implementation of the approach outlined above: https://github.com/maniac103/openhab.android/commit/4985a0f176f6deaf2428c06d5c56fd30788c855a

mueller-ma:
Yes, both issues are fixed.

PR #1401, opened 2019-06-22
Retry background item updates more often

### LINKED ISSUE #1370: Behaviour speed regression of NFC tag when connection not yet established
#### Current behaviour
I am reaching the front door of my house and I tap successfully the NFC tag. A toast message is displayed. The mobile will eventually connect to the WiFi. **_After a while, longer,_** the NFC action will trigger the switch correctly.

#### Previous behaviour
I am reaching the front door of my house and I tap successfully the NFC tag. As the connection to the OpenHAB on local network is not yet established, the activity page with the big "no signal" icon is displayed. The mobile eventually connects to the WiFi. **_Instantly_** as the WiFi is connected, the NFC action will trigger the switch correctly, the activity page with the "no signal" icon disappears immediately.
This behaviour was preferred for speedy performance on WiFi established connection.

To display a Toast as the NFC action is recognised is fine.
In the case the connection to OpenHAB are initially not successful, it would be ideal if, say for the first 60s/30s, the further attempts to connect to OpenHAB are done faster.

- NFC action is triggered. If connection to OpenHAB is successful, display a Toast. If connection to OpenHAB is not initially successful, display the "no signal" icon activity; that is fallback to previous behaviour.
- A settings on the mobile app to choose between new Toast based for NFC action (new behaviour) or bring up Activity (old behaviour)

### Additional context
The use case is approaching the front door of the house and use the NFC tap to trigger some OpenHAB actions of Welcome: this used to be very responsive with the previous behaviour (Activity page with "no signal" icon) because really immediately after WiFi did connect, the action on OpenHAB was performed; with the current Toast-based behaviour, there is still a response, but this happening way delayed, I didn't clocked but definitely at least about 30s later.

Can you kindly consider either of the 3 solutions above, or additional, to resolve this regression in behaviour speed, please?

btw Thank You, OpenHAB is really awesome!! 🙏 :)

mueller-ma:
The first retry is now after 10 seconds, the second one after 20, ...

Fixes #1370

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
@@ -1,5 +1,6 @@
 package org.openhab.habdroid.background
 
+import android.annotation.SuppressLint
 import android.app.AlarmManager
 import android.content.BroadcastReceiver
 import android.content.Context
@@ -96,6 +97,7 @@ class BackgroundTasksManager : BroadcastReceiver() {
 
         // need to keep a ref for this to avoid it being GC'ed
         // (SharedPreferences only keeps a WeakReference)
+        @SuppressLint("StaticFieldLeak")
         private lateinit var prefsListener: PrefsListener
 
         fun initialize(context: Context) {
@@ -114,7 +116,7 @@ class BackgroundTasksManager : BroadcastReceiver() {
                 else
                     context.getString(R.string.nfc_tag_recognized_item, tag.item)
                 Util.showToast(context, message)
-                enqueueItemUpload(WORKER_TAG_PREFIX_NFC + tag.item, tag.item, tag.state)
+                enqueueItemUpload(WORKER_TAG_PREFIX_NFC + tag.item, tag.item, tag.state, BackoffPolicy.LINEAR)
             }
         }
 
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -145,6 +143,8 @@ class BackgroundTasksManager : BroadcastReceiver() {
                 .build()
             val workRequest = OneTimeWorkRequest.Builder(ItemUpdateWorker::class.java)
                 .setConstraints(constraints)
+                .setBackoffCriteria(BackoffPolicy.EXPONENTIAL,
+                    WorkRequest.MIN_BACKOFF_MILLIS, TimeUnit.MILLISECONDS)
```
maniac103: Isn't this the default anyway? https://developer.android.com/reference/kotlin/androidx/work/WorkRequest.Builder.html#setBackoffCriteria(androidx.work.BackoffPolicy,%20java.time.Duration)

I wonder whether there is an actual use case for exponential backoff in our case? Shouldn't linear backoff not work just fine?

mueller-ma: The default value is 30 seconds, `MIN_BACKOFF_MILLIS` is 10.

> I wonder whether there is an actual use case for exponential backoff in our case? Shouldn't linear backoff not work just fine?

I want to avoid spamming users that don't have a cloud instance set up. Imagine the alarm clock changes and they're not at home: After 10 * 10s = 100s they get an error notification.
With `EXPONENTIAL` the error is shown much later, but at the same time the first few retries are quite fast.

maniac103: Ah, right. Maybe use different criteria then? Linear for NFC uploads, exponential for background ones.

mueller-ma: Makes sense.

mueller-ma: Done

PR #1403, opened 2019-06-23
Add ability to pin HABPanel shortcut to home

mueller-ma:
Users want to access HABPanel directly from home, but don't have an
Android version supporting App Shortcuts.


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -990,6 +1004,25 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         }
     }
 
+    private fun pinHabpanelShortcut(context: Context) = GlobalScope.launch {
+        val intent = Intent(context, MainActivity::class.java)
+            .setAction(ACTION_HABPANEL_SELECTED)
+        val shortcutInfo = ShortcutInfoCompat.Builder(context, "habpanel-" + System.currentTimeMillis())
```
maniac103: Is this really required to be made unique by adding the timestamp?

mueller-ma: I've to investigate this. On one of my devices I cannot pin this shortcut more than once, even with this timestamp. Maybe launchers check for the same action, name, icon or a combination of this.

mueller-ma:
> for the HABPanel fragment, pre-create the shortcut info and put it into arguments; control menu item visibility by presence of that argument

Make the action (`ACTION_HABPANEL_SELECTED`) an argument of `WebViewFragment`? If null, the shortcut action is hidden.

maniac103:
> Make the action (ACTION_HABPANEL_SELECTED) an argument of WebViewFragment? If null, the shortcut action is hidden.

`WebViewFragment` is generic though, while the action name is not. That's why I suggested to either keep the class implementation generic or to make the class name point towards it being HABPanel specific.

mueller-ma:
I put the shortcut info in the args of WebViewFragment, but `java.lang.IllegalArgumentException: Illegal value type androidx.core.content.pm.ShortcutInfoCompat for key "shortcut_info"`.

maniac103:
Damn, `ShortcutInfo` is parcelable, `ShortcutInfoCompat` is not ... why, Google, why?
There's two options then: pass in the data needed for the shortcut in a `Bundle`, or ask the activity to create the shortcut when required from some identification token passed to arguments.

mueller-ma:
I'll look into it why the menu isn't shown tomorrow.

mobile/src/main/java/org/openhab/habdroid/ui/activity/WebViewFragment.kt
```diff
@@ -1,44 +1,68 @@
...
         @StringRes get() = arguments!!.getInt(KEY_PAGE_TITLE)
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
-        return inflater.inflate(R.layout.fragment_fullscreenwebview, container, false)
+        return inflater.inflate(R.layout.fragment_webview, container, false)
     }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         val args = arguments!!
+        val context = context!!
```
maniac103: Could avoid the `!!` operator by using `view.context`

mobile/src/main/java/org/openhab/habdroid/ui/activity/WebViewFragment.kt
```diff
@@ -77,6 +101,31 @@ class WebViewFragment : Fragment(), ConnectionFactory.UpdateListener {
...
     }
 
+    override fun onCreateOptionsMenu(menu: Menu?, inflater: MenuInflater?) {
+        shortcutInfo.let { inflater?.inflate(R.menu.webview_menu, menu) }
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        return when (item.itemId) {
+            R.id.webview_add_shortcut -> {
+                pinShortcut(context!!)
```
maniac103: Can omit the parameter here...

mobile/src/main/java/org/openhab/habdroid/ui/activity/WebViewFragment.kt
```diff
@@ -77,6 +101,31 @@ class WebViewFragment : Fragment(), ConnectionFactory.UpdateListener {
...
+                true
+            }
+            else -> super.onOptionsItemSelected(item)
+        }
+    }
+
+    private fun pinShortcut(context: Context) = GlobalScope.launch {
+        val success = ShortcutManagerCompat.requestPinShortcut(context, shortcutInfo!!, null)
```
maniac103: ...and do this here:
```
val context = context ?: return
val info = shortcutInfo ?: return
```

Doing so avoids the `!!` operators usage.

mobile/src/main/java/org/openhab/habdroid/ui/activity/WebViewFragment.kt
```diff
@@ -77,6 +101,31 @@ class WebViewFragment : Fragment(), ConnectionFactory.UpdateListener {
         }
     }
 
+    override fun onCreateOptionsMenu(menu: Menu?, inflater: MenuInflater?) {
+        shortcutInfo.let { inflater?.inflate(R.menu.webview_menu, menu) }
```
maniac103: I wonder whether we only should do this pre-N? In N, one can use the launcher icon to pin the shortcut.

mueller-ma: Yes, users can pin the shortcut, but
1. it's not so visible to the user
2. maybe not every launcher supports this
3. if we add more WebViewFragments, we cannot add a shortcut for them, because only 4 to 5 are supported.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
mueller-ma: No, the timestamp isn't needed.

PR #1405, opened 2019-06-23
Improve command suggestions

mueller-ma:
* Don't suggest commands for Charts
* Suggest current value for Numbers and Strings


mobile/src/main/java/org/openhab/habdroid/util/SuggestedCommandsFactory.kt
```diff
@@ -71,6 +71,7 @@ class SuggestedCommandsFactory(private val context: Context, private val showUnd
             if (suggestedCommands.commands.isEmpty()) {
                 addCommonNumberCommands(suggestedCommands)
             }
+            add(suggestedCommands, item.state?.asString ?: "UNDEF")
```
maniac103: Maybe rather `item.state?.asString.let { value -> add(suggestedCommands, value) }`?

mueller-ma: Done.

PR #1409, opened 2019-06-24
Improve SSE failure retry logic.

### LINKED ISSUE #1282: Widget update still not fully reliable
Even now that SSE is implemented in app, unfortunately, it looks like the refresh of widget is still not yet fully reliable. Sometimes, I need to reopen a page to have the correct widget states.

Last aap store version.

maniac103:
Don't waste time with retries on closed connection, and retry SSE
instead of falling back to long polling.

May fix #1282

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.kt
@@ -411,11 +411,16 @@ class PageConnectionHolderFragment : Fragment(), CoroutineScope {
             }
         }
 
-        internal fun handleSseSubscriptionFailure() {
-            Log.w(TAG, "SSE processing failed for $url, using long polling")
-            eventHelper = null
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.kt
```diff
@@ -411,11 +411,16 @@ class PageConnectionHolderFragment : Fragment(), CoroutineScope {
...
         }
 
-        internal fun handleSseSubscriptionFailure() {
-            Log.w(TAG, "SSE processing failed for $url, using long polling")
-            eventHelper = null
-            if (longPolling) {
-                load()
+        internal fun handleSseSubscriptionFailure(sseUnsupported: Boolean) {
+            if (sseUnsupported) {
+                Log.w(TAG, "SSE processing failed for $url, using long polling")
```
mueller-ma: "SSE unsupported for ..."?

Issue #1326, opened 2019-04-27
Render switch with mappings as selection


mueller-ma:
### Is your feature request related to a problem? Please describe.
Sometimes a switch with a lot of mappings cannot be displayed in one line. The user could shorten the labels, but we cannot expect the user to create multiple versions of the same sitemap for a small phone, a tablet and BasicUI on PC.

### Describe the solution you'd like
Check if the section switch is wider than the screen. If so, render it as selection widget instead.

PR #1410, opened 2019-06-24
Always use item options instead of mappings if available.

### LINKED ISSUE #1281: Switch widget with dynamic options
The app is not handling correctly the switch widget with dynamic options.

It handles static options with mappings defined in the sitemap by displaying a button for each mapping.

In case of dynamic options, it should display a button for each option too. Of course, it should do it only if the number of options is low. In Basic UI and Classic UI, the limit is set to 4 options (4 buttons). If there are 5 options or more, the widget is rendered as a simple switch.

Current behavior: dynamic options are ignored and a simple switch is displayed.

maniac103:
Closes #1281


maniac103:
We may want to rework the segmented switch button layout before doing this, though ... it doesn't deal well at all with the options being larger than the available space.

mueller-ma:
> We may want to rework the segmented switch button layout before doing this, though ... it doesn't deal well at all with the options being larger than the available space.

#1326?

Which binding do you use to generate the dynamic options?

maniac103:
Yeah, that. I'm using the Squeezebox binding.

mueller-ma:
@lolodomo Do you want to review this?
@maniac103 Can you rebase this?

maniac103:
@mueller-ma Done.

mueller-ma:
@maniac103 RTM? I didn't find a binding that has this feature and requires no hardware.

maniac103:
@mueller-ma Taking aside the switch button layout, which has #1326, yes, IMHO it is.

PR #1413, opened 2019-06-24
Migrate preference code to AndroidX

maniac103:
Gets rid of deprecation warnings, and ensures consistent UI (e.g. for dialogs) across Android versions.


mueller-ma:
In xml files there are warnings, that our custom prefs aren't allowed there.
You can remove the icon `ic_none` now. I added this as workaround to have to the items w/o icon indented.

maniac103:
Both done.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -1,14 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
-<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.preference.PreferenceScreen
```
mueller-ma: For consistency I'd like to see this in all pref xml files.

maniac103: Good point, I'll update that.

PR #1414, opened 2019-06-24
Remove daydream receiver from MainActivity.

maniac103:
Since we already re-check fullscreen in onResume(), there's no reason to
handle daydream separately since we go to a pause/resume cycle anyway
when daydreaming starts.

@digitaldan Do you remember what the reason for introducing this code was? I can't see any reason for doing this.


maniac103:
@digitaldan Any news here?

mueller-ma:
@maniac103 Can you rebase this? IMO this is ready for beta.

maniac103:
Done.

mueller-ma:
I just released the beta version with this change. We'll see, if it's still an issue.

PR #1417, opened 2019-06-24
Fix race conditions in ConnectionFactory state updating.

maniac103:
Make sure the various dynamic state variables (available connection,
failure reason, cloud connection, init done tracking) are set
atomically.


mueller-ma:
Build failed, but probably unrelated to this PR:

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

mueller-ma:
Same failure reason again.

maniac103:
Huh, I can't see tests failing, only the stalled build? What do you mean?

mueller-ma:
Me neither, but something makes the build stall.

maniac103:
Okay, now we have something to work with. I'll look into it.

maniac103:
At least one aspect of the problem is that ConnectionFactoryTest updates the ConnectionFactory from its own thread instead of main thread. The initialization isn't thread safe though, just `waitForInitialization()` can be called from non-main threads. I'll update the test.

maniac103:
I think I got it now.

PR #1418, opened 2019-06-24
Handle row click for MapView items

### LINKED ISSUE #1367: Handle row click for MapView items
After #1356

mueller-ma:
Fixes #1367


maniac103:
Hmm, is the base class really useful for the 3-line method?

mueller-ma:
In the first revision I added `handleRowClick()` to both MapViewHolders, but that felt kind of wrong to me.

maniac103:
I don't see that much of a problem, considering the amount of constructor boilerplate needed otherwise. If we use a base class, I think we need to add more stuff to it, e.g. the popup itself. But for now, duplicating the handleRowClick method seems better to me.

PR #1420, opened 2019-06-24
Require pressing back twice to exit app in fullscreen

### LINKED ISSUE #877: Unable to get exit from the app.
### Actual behaviour
When i opened the openHAB opensource android app and then i started using the app and after that when i tried to leave the app and i tried to get back to homescreen by clicking the back button twice. But unfortunately I was not able to get exit from the app.

### Steps to reproduce
* First you have to download the app from fdroid.

* Then install the app.

* Open the app.

* Then try to get back to the homescreen.

* Click the back button twice. But unfortunately you will not be able to get back to the homescreen.

https://youtu.be/3y25frtgWtA

* Android version: 7.1.2
* Device model: Redmi 4
* Stock or customized ROM: Android
* App version:2.1.5-beta
* Device language: English

mueller-ma:
Closes #877

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -362,9 +362,9 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         Log.d(TAG, "onBackPressed()")
         if (controller.canGoBack()) {
             controller.goBack()
-        } else if (!isFullscreenEnabled) {
-            // Only handle back action in non-fullscreen mode, as we don't want to exit
-            // the app via back button in fullscreen mode
+        } else if (isFullscreenEnabled && lastSnackbar?.isShown != true) {
+            showSnackbar(R.string.press_back_to_exit)
+        } else {
             super.onBackPressed()
         }
     }
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -52,7 +52,9 @@ import androidx.drawerlayout.widget.DrawerLayout
 import androidx.recyclerview.widget.RecyclerView
 import com.google.android.material.navigation.NavigationView
 import com.google.android.material.snackbar.Snackbar
-import kotlinx.coroutines.*
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
 import okhttp3.Request
 import org.openhab.habdroid.R
 import org.openhab.habdroid.background.BackgroundTasksManager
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -369,9 +369,9 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         Log.d(TAG, "onBackPressed()")
         if (controller.canGoBack()) {
             controller.goBack()
-        } else if (!isFullscreenEnabled) {
-            // Only handle back action in non-fullscreen mode, as we don't want to exit
-            // the app via back button in fullscreen mode
+        } else if (isFullscreenEnabled && lastSnackbar?.isShown != true) {
+            showSnackbar(R.string.press_back_to_exit)
```
maniac103: Can one attach tags to SnackBars? Would be a good application here, as right now, when pressing back in the wrong moment (e.g. connection change), the app will exit immediately.

mueller-ma: I'm not aware of that. I also tried to get and compare the snackbar text, no success.

maniac103: Maybe via `lastSnackbar.getView()`? Needs investigation, I'm not sure whether that returns the container or the actual view.

mueller-ma: One can attach a tag to `lastSnackbar.view`.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -360,12 +362,15 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
...
-        } else if (!isFullscreenEnabled) {
-            // Only handle back action in non-fullscreen mode, as we don't want to exit
-            // the app via back button in fullscreen mode
-            super.onBackPressed()
+        when {
+            controller.canGoBack() -> controller.goBack()
+            isFullscreenEnabled -> when {
+                lastSnackbar?.isShown != true ->
+                    showSnackbar(R.string.press_back_to_exit, tag = TAG_SNACKBAR_PRESS_AGAIN_EXIT)
+                lastSnackbar?.view?.tag?.equals(TAG_SNACKBAR_PRESS_AGAIN_EXIT) == true -> super.onBackPressed()
```
maniac103: Can use `==` instead of `equals` I think, and then drop the `== true` part.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -877,13 +882,15 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
...
         @StringRes actionResId: Int = 0,
+        tag: String? = null,
         onClickListener: (() -> Unit)? = null
     ) {
         hideSnackbar()
         val snackbar = Snackbar.make(findViewById(android.R.id.content), messageResId, Snackbar.LENGTH_LONG)
         if (actionResId != 0 && onClickListener != null) {
             snackbar.setAction(actionResId) { onClickListener() }
         }
+        snackbar.view.tag = tag.toString()
```
maniac103: `toString()` seems useless on a string ;-)

mueller-ma: Yes, but not on a `String?`.

maniac103: But you don't need to convert to the string 'null' anyway. If a null tag is passed, why shouldn't we assign it as-is? I think we could even make the `tag` parameter to be of type `Any?`.

mueller-ma: I want to keep tag a `String?` as long we don't need an `Any?` tag.

maniac103: That's fine, just the toString call is useless ;-)

PR #1424, opened 2019-06-30
Fix typos

mueller-ma:


mueller-ma:
@maniac103 I noticed that AS saves user-specific stuff in workspace.xml, so we have to exclude it from git.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -46,8 +46,8 @@ class BackgroundTasksManager : BroadcastReceiver() {
                 NotificationUpdateObserver.createNotificationChannels(context)
             }
             ACTION_RETRY_UPLOAD -> {
-                val retryInfos = intent.getParcelableArrayListExtra<RetryInfo>(EXTRA_RETRY_INFOS)
-                for (info in retryInfos) {
+                val retryInfo = intent.getParcelableArrayListExtra<RetryInfo>(EXTRA_RETRY_INFO)
+                for (info in retryInfo) {
```
maniac103: Old spelling seems more correct (it's a list, hence plural makes more sense)

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -97,7 +97,7 @@ class BackgroundTasksManager : BroadcastReceiver() {
         private val TAG = BackgroundTasksManager::class.java.simpleName
 
         internal const val ACTION_RETRY_UPLOAD = "org.openhab.habdroid.background.action.RETRY_UPLOAD"
-        internal const val EXTRA_RETRY_INFOS = "retryInfos"
+        internal const val EXTRA_RETRY_INFO = "retryInfo"
```
maniac103: As above

mobile/src/main/java/org/openhab/habdroid/background/NotificationUpdateObserver.kt
```diff
@@ -75,30 +75,30 @@ internal class NotificationUpdateObserver(context: Context) : Observer<List<Work
         // - failed
         val hasEnqueuedWork = latestInfoByTag.any { (_, info) -> info.state == WorkInfo.State.ENQUEUED }
         val hasRunningWork = latestInfoByTag.any { (_, info) -> info.state == WorkInfo.State.FAILED }
-        val failedInfos = latestInfoByTag.filter { (_, info) -> info.state == WorkInfo.State.FAILED }
+        val failedInfo = latestInfoByTag.filter { (_, info) -> info.state == WorkInfo.State.FAILED }
```
maniac103: As above

mobile/src/main/java/org/openhab/habdroid/background/NotificationUpdateObserver.kt
```diff
@@ -75,30 +75,30 @@ internal class NotificationUpdateObserver(context: Context) : Observer<List<Work
...
         val nm = context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
 
-        if (failedInfos.isNotEmpty()) {
+        if (failedInfo.isNotEmpty()) {
             // show error notification
             val errors = ArrayList<CharSequence>()
-            val retryInfos = ArrayList<BackgroundTasksManager.RetryInfo>()
-            for ((tag, info) in failedInfos) {
+            val retryInfo = ArrayList<BackgroundTasksManager.RetryInfo>()
```
maniac103: As above

mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
```diff
@@ -148,9 +148,9 @@ class WidgetImageView constructor(context: Context, attrs: AttributeSet?) : AppC
         scope = null
     }
 
-    fun setRefreshRate(msec: Int) {
+    fun setRefreshRate(sec: Int) {
```
maniac103: This seems off, as the argument is given in milliseconds ... maybe name it `ms`?

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
mueller-ma: Yes, plural is better here, but the plural of information is information.

maniac103: OK, so how about either `retryInfoList` or simply
```
for (info in intent.getParcelableArrayListExtra<RetryInfo>(EXTRA_RETRY_INFO_LIST))
```
(same line of thought applies below)?

I really don't want variable names to be misleading, even if the misleading name is linguistically more correct.

mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
mueller-ma: `ms` is better, I didn't know what `msec` means.

mueller-ma:
I also removed the changes in all `values-*/strings.xml` files.

mobile/src/main/java/org/openhab/habdroid/ui/activity/WebViewFragment.kt
```diff
@@ -106,7 +106,7 @@ class WebViewFragment : Fragment(), ConnectionFactory.UpdateListener {
         return false
     }
 
-    private fun loadWebsite(urlToLoad: String = urltoLoad) {
+    private fun loadWebsite(urlToLoad: String = this.urlToLoad) {
```
maniac103: `this.` needed?

mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
```diff
@@ -148,9 +148,9 @@ class WidgetImageView constructor(context: Context, attrs: AttributeSet?) : AppC
         scope = null
     }
 
-    fun setRefreshRate(msec: Int) {
+    fun setRefreshRate(ms: Int) {
```
maniac103: Thinking about it, maybe we could/should name this `refreshDelayMs` or something?

mobile/src/main/java/org/openhab/habdroid/ui/activity/WebViewFragment.kt
mueller-ma: Yes, previously it was `urlToLoad` (fun) vs `urltoLoad` (class).

PR #1440, opened 2019-07-09
Fix widget visibility tracking algorithm

### LINKED ISSUE #1437: Frame visibility in sitemap ignored
PR #1174 looks to be included in 2.7.10-beta, so it may have caused this, or it isn't included and may fix this issue.

### Actual behaviour
Sitemaps with items that should be hidden are always visible. Basic UI displays them properly.

### Expected behaviour
Sitemaps with items that should be hidden should be hidden.

### Steps to reproduce
1. Add Item1 to a sitemap inside a frame with visibility based on Item2.
2. Toggle Item2 and notice visibility of Item1 does not change.

Here is a snip from the sitemap where I see this occurring...
```
    Frame visibility=[Mode=="Morning"] {
        Switch item=Wakeup_Alarm
        Switch item=Wakeup_Alarm_Snooze
    }
```

#### Client
* Android version: 2.7.10-beta

#### Server
* Server version: S1627

maniac103:
Closes #1437

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -63,8 +63,9 @@ class WidgetAdapter(
     private val connection: Connection,
     private val itemClickListener: ItemClickListener
 ) : RecyclerView.Adapter<WidgetAdapter.ViewHolder>(), View.OnClickListener, View.OnLongClickListener {
-    private val items = ArrayList<Widget>()
+    private val items = mutableListOf<Widget>()
     val itemList: List<Widget> get() = items
+    private val widgetsById = mutableMapOf<String, Widget>()
 
     private val inflater = LayoutInflater.from(context)
     private val chartTheme: CharSequence
@@ -90,13 +91,14 @@ class WidgetAdapter(
         if (compatibleUpdate) {
             widgets.forEachIndexed { index, widget ->
                 if (items[index] != widget) {
-                    items[index] = widget
-                    notifyItemChanged(index)
+                    updateWidgetAtPosition(index, widget)
                 }
             }
         } else {
             items.clear()
             items.addAll(widgets)
+            widgetsById.clear()
+            widgets.forEach { w -> widgetsById[w.id] = w }
             notifyDataSetChanged()
         }
     }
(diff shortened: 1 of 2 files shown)


openhab-5iver:
Thank you!

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -202,8 +203,29 @@ class WidgetAdapter(
...
+        // If visibility of a container with at least one child changes, refresh the whole list to make sure
+        // the child visibility is also updated. Otherwise it's sufficient to update the single widget only.
+        if (oldWidget.visibility != widget.visibility && items.any { w -> w.parentId == widget.id }) {
+            notifyDataSetChanged()
+        } else {
+            notifyItemChanged(position)
+        }
+    }
+
+    private tailrec fun isWidgetIncludingAllParentsVisible(widget : Widget): Boolean {
```
mueller-ma: AS claims `tailrec` isn't needed here.

maniac103: Where do you see this? I don't see any notice like that in the code editor, during build or in the 'analyze code' output.

mueller-ma: This is a lint inspection that is run during editing code.

maniac103: OK. Looks like one would need to help the compiler here:

```diff
diff --git a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
index 1a5609c1..ce681cda 100644
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -220,8 +220,8 @@ class WidgetAdapter(
         if (!widget.visibility) {
             return false
         }
-        val parent = widget.parentId?.let { id -> widgetsById[id] }
-        return parent == null || isWidgetIncludingAllParentsVisible(parent)
+        val parent = widget.parentId?.let { id -> widgetsById[id] } ?: return true
+        return isWidgetIncludingAllParentsVisible(parent)
     }
 
     private fun getItemViewType(widget: Widget): Int {
```

I've checked in the bytecode output that it indeed currently doesn't do tail recursion optimization.

PR #1448, opened 2019-07-10
Fix lint issues in PreferencesActivity

mueller-ma:
* Only set on prefs click listener if ringtone and vibration prefs are shown. This way the constants Settings.ACTION_APP_NOTIFICATION_SETTINGS and Settings.EXTRA_APP_PACKAGE are guarded by an API check.
* Use elvis operator
* Explicitly set type for `prefs`

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
@@ -14,7 +14,6 @@
 package org.openhab.habdroid.ui
 
 import android.content.Intent
-import android.content.SharedPreferences
 import android.content.pm.PackageManager
 import android.media.RingtoneManager
 import android.net.Uri
--- a/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
@@ -14,6 +14,7 @@
 package org.openhab.habdroid.ui
 
 import android.content.Intent
+import android.content.SharedPreferences
 import android.content.pm.PackageManager
 import android.media.RingtoneManager
 import android.net.Uri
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -109,7 +110,7 @@ class PreferencesActivity : AbstractBaseActivity() {
         protected abstract val titleResId: Int
 
         protected val parentActivity get() = activity as PreferencesActivity
-        protected val prefs get() = preferenceScreen.sharedPreferences
+        protected val prefs get() = preferenceScreen.sharedPreferences as SharedPreferences
```
maniac103: Is this for getting rid of nullability? Can shortcut with `!!` instead of cast in that case.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -363,7 +364,7 @@ class PreferencesActivity : AbstractBaseActivity() {
                 val ringtonePref = findPreference(Constants.PREFERENCE_TONE)
                 updateRingtonePreferenceSummary(ringtonePref, ringtoneUri)
                 prefs.edit {
-                    putString(Constants.PREFERENCE_TONE, if (ringtoneUri != null) ringtoneUri.toString() else "")
+                    putString(Constants.PREFERENCE_TONE, ringtoneUri?.toString() ?: "")
```
maniac103: Can also use `.orEmpty()`, it's probably more elegant than Elvis.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
mueller-ma: Kotlin throws a lint issue here, because it doesn't know for sure if it receives `SharedPreferences` or `SharedPreferences?`.

maniac103: That's what I guessed. As I said, can shortcut using the `!!` operator in that case.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
maniac103: Nevermind, missed the `toString()`.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
mueller-ma: Done.

PR #1449, opened 2019-07-11
Add lock screen

### LINKED ISSUE #1216: adding authentication with PIN and/or fingerprint for app
**Is your feature request related to a problem? Please describe.**
Hey, this is not a problem but pure feature request. Smart home in my case (but I conclude from forums that others' too) contains many sensitive settings for example heating unit control or gate/garage gate control. In this case anyone having access to the phone can play with it and accidentally change important setting or even get compromised by malicious behaviour of stranger. This is quite common for people with kids who take parents' phone to play

**Describe the solution you'd like**
Application should open only after entering PIN or fingerprint (depends which solution is chosen by developer).  This is known behavior from any bank apps

**Describe alternatives you've considered**
Alternative is to delete password from the app for remote and local links after using the app and entering it again for any new access but this is very inconvenient and one day will get forgotten for sure.

mueller-ma:
Users can set lock screen to:
* off (default)
* on
* Kiosk mode: MainActivity can be accessed without screen lock, but
prefs, about, etc. are locked

Fixes #1216

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
@@ -30,8 +30,4 @@ class OpenHabApplication : MultiDexApplication() {
         super.onTerminate()
         ConnectionFactory.shutdown()
     }
-
-    companion object {
-        var lastAuthenticationTimestamp = 0L
-    }
 }
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
@@ -30,4 +30,8 @@ class OpenHabApplication : MultiDexApplication() {
         super.onTerminate()
         ConnectionFactory.shutdown()
     }
+
+    companion object {
+        var lastAuthenticationTimestamp = 0L
+    }
 }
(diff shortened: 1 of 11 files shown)


mueller-ma:
I've found two issues:
* On my Amazon tablet I always have to enter the code twice, on Android emulators it works fine.
* If I cancel the screen lock by clicking back, the screen lock is "duplicated". I have to enter the code twice to enter the app.

maniac103:
Logic simplification suggestion: https://github.com/openhab/openhab-android/commit/b6fda7123042390bb2c8ba6e787125249f1d8149

I'll look into the latter issue.

maniac103:
Latter issue fixed. Commit + some more fixes can be found here: https://github.com/maniac103/openhab.android/commits/1216-lockscreen

mueller-ma:
I guess `lastAuthenticationTimestamp` should be set to `SystemClock.elapsedRealtime()` in `onPause()`. Otherwise a user has to re-authenticate after 2 minutes, when entering a different activity, even if the app wasn't closed.

maniac103:
That's right, but I don't really see this as a problem. Users will stay in MainActivity 99% of the time anyway.
What we'd need is a way to pass information on whether the activity invocation was internal or external, and skip auth in the former case. Don't have a good, non-fakeable idea for that yet though.

mueller-ma:
RTM from my side, but I want to wait until the next stable version is released.

PR #1454, opened 2019-07-24
Make sure that intro is shown again if app is stopped

mueller-ma:
Previous if the app got stopped during the intro, the intro won't be
shown again on next start.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -193,15 +193,13 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         //  Create a new boolean and preference and set it to true
         val isFirstStart = prefs.getBoolean(Constants.PREFERENCE_FIRST_START, true)
 
-        val prefsEditor = prefs.edit()
         //  If the activity has never started before...
         if (isFirstStart) {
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -193,15 +193,13 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
...
         if (isFirstStart) {
             //  Launch app intro
             val i = Intent(this@MainActivity, IntroActivity::class.java)
             startActivityForResult(i, INTRO_REQUEST_CODE)
-
-            prefsEditor.putBoolean(Constants.PREFERENCE_FIRST_START, false)
         }
+        val prefsEditor = prefs.edit()
         OnUpdateBroadcastReceiver.updateComparableVersion(prefsEditor)
         prefsEditor.apply()
```
maniac103: After this change, we can probably pass `prefs` into the version update and only edit there if needed, using the android-ktx syntax.

mueller-ma: `updateComparableVersion` is also called from `OnUpdateBroadcastReceiver` inside a `prefs.edit` block. Now the change is applied twice when called from `OnUpdateBroadcastReceiver`.

PR #1455, opened 2019-07-24
Make ItemPickerActivity abstract

mueller-ma:
Renaming ItemPickerActivity to TaskerItemPickerActivity breaks
configured actions in Tasker. I want to rename it, before the Tasker
plugin goes live in the stable version.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
@@ -26,7 +26,8 @@ import androidx.work.*
 import kotlinx.android.parcel.Parcelize
 import org.openhab.habdroid.R
 import org.openhab.habdroid.model.NfcTag
-import org.openhab.habdroid.ui.ItemPickerActivity
+import org.openhab.habdroid.ui.AbstractItemPickerActivity
+import org.openhab.habdroid.ui.TaskerItemPickerActivity
 import org.openhab.habdroid.ui.preference.toItemUpdatePrefValue
 import org.openhab.habdroid.util.*
 import java.util.*
@@ -56,8 +57,8 @@ class BackgroundTasksManager : BroadcastReceiver() {
                     return
                 }
                 val bundle = intent.getBundleExtra(TaskerIntent.EXTRA_BUNDLE) ?: return
-                val itemName = bundle.getString(ItemPickerActivity.EXTRA_ITEM_NAME)
-                val state = bundle.getString(ItemPickerActivity.EXTRA_ITEM_STATE)
+                val itemName = bundle.getString(AbstractItemPickerActivity.EXTRA_ITEM_NAME)
+                val state = bundle.getString(AbstractItemPickerActivity.EXTRA_ITEM_STATE)
                 if (itemName.isNullOrEmpty() || state.isNullOrEmpty()) {
                     return
                 }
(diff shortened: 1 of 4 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
```diff
@@ -253,22 +231,24 @@ class ItemPickerActivity : AbstractBaseActivity(), SwipeRefreshLayout.OnRefreshL
...
             loadError -> R.string.item_picker_list_error
-            isDisabled -> R.string.settings_tasker_plugin_summary
+            isDisabled -> getDisabledMessage()
             else -> R.string.item_picker_list_empty
         })
         retryButton.setText(if (isDisabled) R.string.turn_on else R.string.try_again_button)
         retryButton.isVisible = loadError || isDisabled
     }
 
+    protected abstract fun getDisabledMessage() : Int
```
maniac103: This can become a property I think.

mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
```diff
@@ -25,43 +24,45 @@ import android.widget.EditText
...
 import org.openhab.habdroid.model.toItem
 import org.openhab.habdroid.ui.widget.DividerItemDecoration
-import org.openhab.habdroid.util.*
-
-class ItemPickerActivity : AbstractBaseActivity(), SwipeRefreshLayout.OnRefreshListener,
+import org.openhab.habdroid.util.HttpClient
+import org.openhab.habdroid.util.SuggestedCommandsFactory
+import org.openhab.habdroid.util.TaskerIntent
+import org.openhab.habdroid.util.getPrefs
+import org.openhab.habdroid.util.isTaskerPluginEnabled
```
maniac103: The Tasker specific stuff at the end of onCreate should be moved to the derived class as well, shouldn't it?

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -56,8 +56,8 @@ class BackgroundTasksManager : BroadcastReceiver() {
                     return
                 }
                 val bundle = intent.getBundleExtra(TaskerIntent.EXTRA_BUNDLE) ?: return
-                val itemName = bundle.getString(ItemPickerActivity.EXTRA_ITEM_NAME)
-                val state = bundle.getString(ItemPickerActivity.EXTRA_ITEM_STATE)
+                val itemName = bundle.getString(AbstractItemPickerActivity.EXTRA_ITEM_NAME)
+                val state = bundle.getString(AbstractItemPickerActivity.EXTRA_ITEM_STATE)
```
maniac103: Since the bundle name references `TaskerIntent`, I think it's better to use `TaskerItemPickerActivity` here.

mueller-ma: IMO the child classes of `AbstractItemPickerActivity` should always send back a bundle containing item name and state, so we always need these two constants.

mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
mueller-ma: Yes, forgot to remove them.

PR #1461, opened 2019-08-01
Fix host formatting in prefs

mueller-ma:


mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -439,7 +444,7 @@ class PreferencesActivity : AbstractBaseActivity() {
 
             @VisibleForTesting fun beautifyUrl(url: String?): String {
                 val host = url?.toUri()?.host.orEmpty()
-                return if (host.contains("myopenhab.org")) "myopenHAB" else url.orEmpty()
+                return if (host.contains("myopenhab.org")) "myopenHAB" else host
```
mueller-ma: I want to replace `contains()` with a regex for `(home.)myopenhab.org`.

PR #1462, opened 2019-08-02
Fix crash in FCM registration

mueller-ma:
````
Fatal Exception: java.lang.RuntimeException: Unable to start receiver org.openhab.habdroid.core.FcmRegistrationService$ProxyReceiver: java.lang.IllegalArgumentException: work must not be null
        [...]
Caused by java.lang.IllegalArgumentException: work must not be null
       at androidx.core.app.JobIntentService.enqueueWork + 518(JobIntentService.java:518)
       at androidx.core.app.JobIntentService.enqueueWork + 501(JobIntentService.java:501)
       at org.openhab.habdroid.core.FcmRegistrationService$ProxyReceiver.onReceive + 116(FcmRegistrationService.kt:116)
````

### DIFF
--- a/mobile/src/full/java/org/openhab/habdroid/core/FcmRegistrationService.kt
+++ b/mobile/src/full/java/org/openhab/habdroid/core/FcmRegistrationService.kt
@@ -112,7 +112,7 @@ class FcmRegistrationService : JobIntentService() {
 
     class ProxyReceiver : BroadcastReceiver() {
         override fun onReceive(context: Context, intent: Intent) {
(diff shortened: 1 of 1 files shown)


maniac103:
Adding the check seems harmless enough, but how did the receiver receive a broadcast without this extra? I wonder whether somebody toyed with calling it externally ... how many such crashes did you see?

mueller-ma:
> I wonder whether somebody toyed with calling it externally

Possible, I only see one crash. Nevertheless I'd like to merge this as sooner or later Google will annotate the parameter with `@NonNull` or even rewritten in it Kotlin.

PR #1464, opened 2019-08-03
Implement widgets

### LINKED ISSUE #6: Widgets for homescreen
This was requested a couple of times on the mailinglist already, but nobody seemed to have cared to open an issue for it.
It would be great to be able to put a customizable widget directly on the Android homescreen - e.g. for an "All off" functionality or similar.
##

mueller-ma:
Fixes #6

The widgets are configured with a config activity. It initializes the widget, but from time to time `ItemUpdateWidget#onUpdate()` is called which resets the widgets to default. I haven't found a practical way to persist the config data yet.


mueller-ma:
I used prefs to save information about the widgets, including item, state, label and mapped state. I want to save the input stream from the icon get request and if it's svg or not. The icon can be re-rendered whenever the widget size is changed.
@maniac103 Should I go on with this approach?

maniac103:
@mueller-ma Sounds fine, except I probably would use separate shared pref files and logic for the widgets. You can have one shared pref instance per widget by passing e.g. `"widget-$widgetId"` as name to `Context.getSharedPreferences()`.

mueller-ma:
I guess it's better to save the downloaded icon to directly to disk. This way it's easier to remove all icons if the icon cache is cleared (remove all files in the icon dir). If it's svg or not can be read from the prefs, no need to save it again. If the icon format is changed, the icon cache should be cleared.

mueller-ma:
Ready to review

mueller-ma:
The widget should switch to the small layout if the text cannot be displayed completely, but I don't know how to check that.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundIntentReceiveActivity.kt
```diff
@@ -42,8 +41,14 @@ class NfcReceiveActivity : Activity() {
                 }
                 startActivity(startMainIntent)
             }
+        } else if (intent.action == BackgroundTasksManager.ACTION_UPDATE_WIDGET && intent.extras != null) {
+            BackgroundTasksManager.enqueueWidgetItemUpdateIfNeeded(intent.extras!!)
```
maniac103: I don't like the `!!` operator usage. How about this:

```
} else if (intent.action == BackgroundTasksManager.ACTION_UPDATE_WIDGET) {
    intent.extras?.let { extras -> BackgroundTasksManager.enqueueWidgetItemUpdateIfNeeded(extras) }
}

File: mobile/src/main/java/org/openhab/habdroid/ui/ItemUpdateWidgetItemPickerActivity.kt [hunk 2]
```diff
@@ -0,0 +1,36 @@
...
+        ) ?: AppWidgetManager.INVALID_APPWIDGET_ID
+    }
+
+    override fun finish(item: Item, state: String, mappedState: String) {
+        val appWidgetManager: AppWidgetManager = AppWidgetManager.getInstance(this)
+        val label = if (item.label.isNullOrEmpty()) item.name else item.label
+        val data = ItemUpdateWidget.ItemUpdateWidgetData(item.name, state, label, mappedState, item.category.orEmpty())
+        ItemUpdateWidget.setupWidget(this, data, appWidgetId, appWidgetManager)
+        ItemUpdateWidget.saveInfoForWidget(this, data, appWidgetId)
+        val resultValue = Intent().apply {
```
maniac103: `apply` seems useless given there's only one line in the block ... can call Intent().putExtra() directly.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -348,15 +350,46 @@ class PreferencesActivity : AbstractBaseActivity() {
...
+            } else {
+                iconFormatPreference.setOnPreferenceChangeListener { pref, _ ->
+                    val context = pref.context
+                    clearImageCache(context)
+                    val intent = Intent(context, ItemUpdateWidget::class.java)
+                    intent.action = AppWidgetManager.ACTION_APPWIDGET_UPDATE
+                    val ids = AppWidgetManager.getInstance(context)
+                        .getAppWidgetIds(ComponentName(context, ItemUpdateWidget::class.java))
+                    intent.putExtra(AppWidgetManager.EXTRA_APPWIDGET_IDS, ids)
+                    context.sendBroadcast(intent)
```
maniac103:
```
val widgetComponent = ComponentName(context, ItemUpdateWidget::class.java)
val ids = AppWidgetManager.getInstance(context).getAppWidgetIds(widgetComponent)
val intent = Intent(widgetComponent)
    .setAction(AppWidgetManager.ACTION_APPWIDGET_UPDATE)
    .putExtra(AppWidgetManager.EXTRA_APPWIDGET_IDS, ids)
context.setBroadcast(intent)
```

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -348,15 +350,46 @@ class PreferencesActivity : AbstractBaseActivity() {
                 ?: preferenceScreen.sharedPreferences.getInt(Constants.PREV_SERVER_FLAGS, 0)
 
             if (flags and ServerProperties.SERVER_FLAG_ICON_FORMAT_SUPPORT == 0) {
-                val iconFormatPreference = preferenceScreen.findPreference(Constants.PREFERENCE_ICON_FORMAT)
                 preferenceScreen.removePreferenceFromHierarchy(iconFormatPreference)
+            } else {
+                iconFormatPreference.setOnPreferenceChangeListener { pref, _ ->
+                    val context = pref.context
+                    clearImageCache(context)
```
maniac103: Clearing the cache once should be sufficient, shouldn't it? ;-)

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -348,15 +350,46 @@ class PreferencesActivity : AbstractBaseActivity() {
...
 
+        private fun clearImageCache(context: Context) {
+            // Clear widget icons
+            val files = context.fileList()
+            files.forEach {
+                if (it.matches("^widget-\\d+$".toRegex())) {
+                    Log.d(TAG, "Deleting file $it")
+                    context.deleteFile(it)
+                }
+            }
```
maniac103: I wouldn't use the files directory for storing the icons. I'd suggest to use the cache directory instead, so the cached icons appear as 'cache data' instead of 'app data' in the app listing.
Deleting the cache then essentially becomes something among the lines of `File(context.cacheDir, "widgeticons").listFiles().forEach { f -> f.delete() }`

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
```diff
@@ -114,8 +114,13 @@ fun ResponseBody.toBitmap(targetSize: Int, enforceSize: Boolean = false): Bitmap
         }
     }
 
+    return bitmapToSvg(string(), targetSize)
+}
+
+@Throws(IOException::class)
+fun bitmapToSvg(svgString: String, targetSize: Int): Bitmap {
```
maniac103: The naming is off. The method name implies a bitmap-to-SVG conversion, but it's the other way around.
Besides that, I'd suggest making this `fun InputStream.asSvgToBitmap(targetSize: Int): Bitmap` ... the current method isn't an extension function, which makes it a bit mis-placed in that file ;-) and looking at how the method is used, we have an `InputStream` readily available, so why do the extra string conversion?

mobile/src/main/res/xml/item_update_widget.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<appwidget-provider xmlns:android="http://schemas.android.com/apk/res/android"
+                    android:minWidth="40dp"
```
maniac103: Nit: Indent by 4

mobile/src/main/res/xml/item_update_widget.xml
```diff
@@ -0,0 +1,11 @@
...
+                    android:minHeight="40dp"
+                    android:updatePeriodMillis="0"
+                    android:previewImage="@drawable/item_update_widget_preview"
+                    android:initialLayout="@layout/widget_item_update"
+                    android:widgetCategory="home_screen|keyguard"
```
maniac103: I know homescreen widgets aren't used anymore, but for the Android versions that *have* homescreen widgets, this one doesn't seem suitable to place there.

mobile/src/main/java/org/openhab/habdroid/ui/ItemUpdateWidgetItemPickerActivity.kt
```diff
@@ -0,0 +1,36 @@
...
+            AppWidgetManager.EXTRA_APPWIDGET_ID,
+            AppWidgetManager.INVALID_APPWIDGET_ID
+        ) ?: AppWidgetManager.INVALID_APPWIDGET_ID
+    }
+
+    override fun finish(item: Item, state: String, mappedState: String) {
+        val appWidgetManager: AppWidgetManager = AppWidgetManager.getInstance(this)
+        val label = if (item.label.isNullOrEmpty()) item.name else item.label
+        val data = ItemUpdateWidget.ItemUpdateWidgetData(item.name, state, label, mappedState, item.category.orEmpty())
+        ItemUpdateWidget.setupWidget(this, data, appWidgetId, appWidgetManager)
```
maniac103: Wouldn't it be more elegant to just send the broadcast to the widget provider and let that one handle the update?

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
```diff
@@ -0,0 +1,237 @@
...
+open class ItemUpdateWidget : AppWidgetProvider() {
+    override fun onUpdate(context: Context, appWidgetManager: AppWidgetManager, appWidgetIds: IntArray) {
+        Log.d(TAG, "onUpdate()")
+        appWidgetIds.forEach {
+            Log.d(TAG, "id: $it")
+            val data = getInfoForWidget(context, it)
+            if (data != null) {
+                setupWidget(context, data, it, appWidgetManager)
+            }
+        }
```
maniac103: I don't know if it's overkill, but I'd find this more elegant:
```
appWidgetIds.map { id -> Pair(id, getInfoForWidget(context, id)) }
    .filter { id, data -> data != null }
    .forEach { id, data -> setupWidget(context, data, id, appWidgetManager }

File: mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt [hunk 11]
```diff
@@ -0,0 +1,237 @@
...
+    ) {
+        Log.d(TAG, "onAppWidgetOptionsChanged()")
+        val data = getInfoForWidget(context ?: return, appWidgetId)
+        setupWidget(context, data ?: return, appWidgetId, appWidgetManager ?: return)
+    }
+
+    override fun onReceive(context: Context?, intent: Intent?) {
+        Log.d(TAG, "onReceive() $intent")
+        super.onReceive(context, intent)
+    }
```
maniac103: Method implementation seems useless.

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
```diff
@@ -0,0 +1,237 @@
...
+        if (context == null || appWidgetIds == null) {
+            return
+        }
+        appWidgetIds.forEach {
+            Log.d(TAG, "Deleting data for id $it")
+            context.deleteFile(getFileNameForWidget(it))
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                context.deleteSharedPreferences(getFileNameForWidget(it))
+            } else {
+                saveInfoForWidget(context, ItemUpdateWidgetData("", "", "", "", ""), it)
```
maniac103: How about
```
context.getSharedPreferences(getFileNameForWidget(it), MODE_PRIVATE).edit {
    clear()
}
```

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
```diff
@@ -0,0 +1,237 @@
...
+        private val TAG = ItemUpdateWidget::class.java.simpleName
+
+        fun setupWidget(
+            context: Context,
+            data: ItemUpdateWidgetData,
+            appWidgetId: Int,
+            appWidgetManager: AppWidgetManager
+        ) {
+            val widgetOptions = appWidgetManager.getAppWidgetOptions(appWidgetId)
+            val smallWidget = widgetOptions.getInt(AppWidgetManager.OPTION_APPWIDGET_MAX_HEIGHT) < 200
```
maniac103: Maybe use a dimension value instead of the magic number '200'? `R.dimen.small_widget_threshold` or something?

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
```diff
@@ -0,0 +1,237 @@
...
+                val encodedIcon = Uri.encode(data.icon)
+                val iconFormat = context.getPrefs().getIconFormat()
+                val iconUrl = "icon/$encodedIcon?state=${data.state}&format=$iconFormat"
+
+                try {
+                    if (context.fileList().contains(getFileNameForWidget(appWidgetId))) {
+                        Log.d(TAG, "Icon exits")
+                    } else {
+                        Log.d(TAG, "Download icon")
+                        val bitmap = connection.httpClient.get(iconUrl).response.bytes()
```
maniac103: `val content = ...` ... it's not a bitmap after all.

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
```diff
@@ -0,0 +1,237 @@
...
+            return ItemUpdateWidgetData(item, state, label, mappedState, icon)
+        }
+
+        fun saveInfoForWidget(
+            context: Context,
+            data: ItemUpdateWidgetData,
+            id: Int
+        ) {
+            val prefs = context.getSharedPreferences(getFileNameForWidget(id), MODE_PRIVATE)
+            prefs.edit()
```
maniac103:
```
context.getSharedPreferences(getFileNameForWidget(id), MODE_PRIVATE).edit {
   putString(...)
   ...
}
```

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
```diff
@@ -0,0 +1,237 @@
...
+        fun getInfoForWidget(context: Context, id: Int): ItemUpdateWidgetData? {
+            val prefs = context.getSharedPreferences(getFileNameForWidget(id), MODE_PRIVATE)
+            val item = prefs.getString(PreferencesActivity.ITEM_UPDATE_WIDGET_ITEM)
+            val state = prefs.getString(PreferencesActivity.ITEM_UPDATE_WIDGET_STATE)
+            val label = prefs.getString(PreferencesActivity.ITEM_UPDATE_WIDGET_LABEL)
+            val mappedState = prefs.getString(PreferencesActivity.ITEM_UPDATE_WIDGET_MAPPED_STATE)
+            val icon = prefs.getString(PreferencesActivity.ITEM_UPDATE_WIDGET_ICON)
+            if (item.isEmpty() || state.isEmpty() || label.isEmpty() || mappedState.isEmpty()) {
+                return null
+            }
```
maniac103: This check can be removed, as you won't get broadcasts for non-existing widgets.

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
```diff
@@ -0,0 +1,237 @@
...
+                saveInfoForWidget(context, ItemUpdateWidgetData("", "", "", "", ""), it)
+            }
+        }
+        super.onDeleted(context, appWidgetIds)
+    }
+
+    companion object {
+        private val TAG = ItemUpdateWidget::class.java.simpleName
+
+        fun setupWidget(
```
maniac103: With the suggestion above (invoke onUpdate on creation via broadcast) all of those methods except `saveInfoForWidget` can be moved out of companion object and be made private.

mobile/src/main/java/org/openhab/habdroid/ui/ItemUpdateWidgetItemPickerActivity.kt
mueller-ma: Actually this is copy and paste from Google (https://developer.android.com/guide/topics/appwidgets/#UpdatingFromTheConfiguration). I'll change it.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
mueller-ma: `ComponentName` cannot be passed to `Intent()`.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
mueller-ma: I thought about using the cache directory, but I won't to avoid that any "optimization" app clears the cache and forces us to re-download the icon again, which only wastes data.

mobile/src/main/res/xml/item_update_widget.xml
mueller-ma: Do you mean keyguard? IMO it's practical to quickly turn on lights, for example.

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
mueller-ma: It's for debugging. I want to keep this for some time in beta.

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
mueller-ma: By making the return type of `getInfoForWidget()` non-nullable, this can be simplified to:
````kotlin
        appWidgetIds.forEach {
            Log.d(TAG, "id: $it")
            setupWidget(context, getInfoForWidget(context, it), it, appWidgetManager)
        }
````

**Comment by mueller-ma** [comment]
If multiple widgets are created, they all send the item update of the first widget.

File: mobile/src/main/java/org/openhab/habdroid/ui/ItemUpdateWidgetItemPickerActivity.kt (hunk 9 above)
Comment (mueller-ma) [reply mobile/src/main/java/org/openhab/habdroid/ui/ItemUpdateWidgetItemPickerActivity.kt]: The docs say, that the configuration activity should handle the setup of the widget.
I can remove `data`  and `appWidgetManager` from the function and let `setupWidget()` fetch them.

File: mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt (hunk 3 above)
Comment (mueller-ma) [reply mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt]: I've optimized the `setAction` and `putExtra` calls.

File: mobile/src/main/res/xml/item_update_widget.xml [hunk 18]
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<appwidget-provider xmlns:android="http://schemas.android.com/apk/res/android"
+    android:minWidth="40dp"
+    android:minHeight="40dp"
+    android:updatePeriodMillis="0"
```
mueller-ma: I set this to 0, but `onUpdate()` is called quite often. I should try setting a pretty high number here.

mueller-ma:
`SuggestedCommandsFactory#addCommonPercentCommands()` should use a non-breaking space between the number and the percent sign.

mobile/src/main/java/org/openhab/habdroid/ui/ItemUpdateWidgetItemPickerActivity.kt
maniac103: If the configuration activity manually triggers the widget receiver, it also handles the widget setup ;-) It's just a matter of code structure.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
maniac103: Looks good now. I was totally sure `Intent` has a constructor taking a `ComponentName`, because the `Intent(Context, Class<?>)` one also creates a ComponentName internally, but obviously I was wrong.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
maniac103: This argument doesn't make too much sense to me, TBH. It can be equally applied to a web browser's cache, where redownloading web pages wastes data, but still, a web browser's cache lives in the cache directory.

Thinking more about this argument, the whole additional caching layer seems unneeded/redundant. The OkHttp caching layer also caches on disk, so if one passes `FORCE_CACHE_IF_POSSIBLE` as `cachingMode` to `HttpClient.get()`, one achieves exactly the same effect as with the additional layer.

mobile/src/main/java/org/openhab/habdroid/ui/ItemUpdateWidgetItemPickerActivity.kt
mueller-ma: I tried to send an intent to `ItemUpdateWidget` from `finish()`, but the widget is created, after `finish()`. So the widget with the `appWidgetId` isn't present at the time I send the broadcast.

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
```diff
@@ -0,0 +1,234 @@
...
+            data: ItemUpdateWidgetData,
+            appWidgetId: Int,
+            appWidgetManager: AppWidgetManager
+        ) {
+            val widgetOptions = appWidgetManager.getAppWidgetOptions(appWidgetId)
+            val smallWidget = widgetOptions.getInt(AppWidgetManager.OPTION_APPWIDGET_MAX_HEIGHT) <
+                context.resources.getDimension(R.dimen.small_widget_threshold)
+            val views = RemoteViews(context.packageName,
+                if (smallWidget) R.layout.widget_item_update_small else R.layout.widget_item_update)
+            val intent = Intent(context, BackgroundIntentReceiveActivity::class.java).apply {
```
mueller-ma: Widgets should just send their id to `ItemUpdateWidget` and `onReceive()` should handle the intent.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
mueller-ma: How can I get a usable `HttpClient` if `connection` is `null`?

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
mueller-ma: Now it's used.

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
```diff
@@ -0,0 +1,242 @@
...
+            setIcon(context, views, smallWidget, appWidgetId, appWidgetManager)
+        }
+
+        private fun setIcon(
+            context: Context,
+            views: RemoteViews,
+            smallWidget: Boolean,
+            appWidgetId: Int,
+            appWidgetManager: AppWidgetManager
+        ) = GlobalScope.launch {
```
maniac103: Could also pass the InputStream (be it the file one or the HTTP response one) directly into this one, so an icon just downloaded doesn't need to be reopened from disk.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
maniac103: That's a very good point. I therefore withdraw the redundancy argument ;-) But please put the cache management code to CacheManager, so we have a single place managing all cache related stuff. I could see two methods there:
- `fun getWidgetIconStream(widgetId: Int): Pair<InputStream, Boolean>?`, where the latter denotes whether the icon is PNG or SVG (think the anyFormat proposal) ... null would be returned if nothing can be found
- `fun saveWidgetIcon(widgetId: Int, iconData: InputStream, isSvg: Boolean)` saves the data to disk.

As for cache vs. files, it's probably fine either way, since I guess it would feel strange if widget icons are gone after clearing the cache directory. We have to make sure the icon for a given widget is cleared when that widget is removed, though. When using the files directory, I'd also prefer if we'd use a subdirectory for the widget images.

mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
```diff
@@ -0,0 +1,242 @@
...
+            val bitmap = try {
+                context.openFileInput(getFileNameForWidget(appWidgetId)).use { fileInputStream ->
+                    val inputStream = fileInputStream.readBytes().inputStream()
```
maniac103: Why are you converting an InputStream to a different InputStream here? The file stream should be totally fine by itself, shouldn't it?

maniac103:
Implemented the review suggestions here: https://github.com/maniac103/openhab.android/tree/6-widgets ... please let me know about your thoughts.

What also would be good were some kind of feedback the action is going to be executed ... a ripple background probably would be sufficient for that, I think.

mueller-ma:
I added some comments to your commits, otherwise I'm fine with them.

> What also would be good were some kind of feedback the action is going to be executed ... a ripple background probably would be sufficient for that, I think.

Maybe show a toast, like it's done for nfc item updates.

mobile/src/main/res/xml/item_update_widget.xml
mueller-ma: @maniac103 This is still unresolved.

maniac103: Yes, I meant keyguard. I still think only information widgets should be placed there, but I don't have a strong opinion on this one, so it's fine.

mueller-ma:
> Maybe show a toast, like it's done for nfc item updates.

A toast is shown when the item is updated.
RTM from my side.

mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
```diff
@@ -42,18 +41,20 @@ class ItemUpdateWorker(context: Context, params: WorkerParameters) : Worker(cont
...
         val value = inputData.getString(INPUT_DATA_VALUE) as String
+        val successToastMessage = inputData.getString(INPUT_DATA_SUCCESS_TOAST_MESSAGE)
 
         return runBlocking {
             try {
                 val result = connection.httpClient
                     .post("rest/items/$item", value, "text/plain;charset=UTF-8")
                     .asStatus()
                 Log.d(TAG, "Item '$item' successfully updated to value $value")
+                successToastMessage?.let { Util.showToast(applicationContext, it) }
```
maniac103: I think I'd prefer an `if` check instead of `let` for clarity.

(I usually only use `let` for mutable variables)

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -359,15 +356,45 @@ class PreferencesActivity : AbstractBaseActivity() {
...
```
maniac103: This is a leftover that should be removed.

PR #1465, opened 2019-08-03
Improve long press menu

mueller-ma:
* Show a menu with the possible actions: write item update tag, write
sitemap tag or create shortcut
* Show "custom" state for item update tags

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
@@ -20,9 +20,13 @@ import android.graphics.Bitmap
 import android.graphics.Canvas
 import android.graphics.Color
 import android.net.Uri
+import android.nfc.NfcAdapter
 import android.os.Bundle
 import android.util.Log
+import android.view.ContextMenu
 import android.view.LayoutInflater
+import android.view.Menu
+import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
 import android.view.WindowManager
@@ -47,6 +51,7 @@ import org.openhab.habdroid.R
 import org.openhab.habdroid.core.connection.ConnectionFactory
 import org.openhab.habdroid.model.LinkedPage
 import org.openhab.habdroid.model.Widget
+import org.openhab.habdroid.ui.widget.ContextMenuAwareRecyclerView
 import org.openhab.habdroid.ui.widget.RecyclerViewSwipeRefreshLayout
 import org.openhab.habdroid.util.CacheManager
 import org.openhab.habdroid.util.HttpClient
@@ -54,7 +59,6 @@ import org.openhab.habdroid.util.SuggestedCommandsFactory
 import org.openhab.habdroid.util.dpToPixel
 import org.openhab.habdroid.util.getIconFormat
 import org.openhab.habdroid.util.getPrefs
-import java.util.ArrayList
 
 /**
  * This class is apps' main fragment which displays list of openHAB
@@ -112,6 +116,7 @@ class WidgetListFragment : Fragment(), WidgetAdapter.ItemClickListener {
         recyclerView.layoutManager = layoutManager
         recyclerView.adapter = adapter
         (recyclerView.itemAnimator as SimpleItemAnimator).supportsChangeAnimations = false
+        registerForContextMenu(recyclerView)
 
         refreshLayout = view.findViewById(R.id.swiperefresh)
         refreshLayout.applyColors(R.attr.colorPrimary, R.attr.colorAccent)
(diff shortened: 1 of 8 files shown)


maniac103:
I really don't like two dialogs opening subsequently. I think instead of the dialogs we should use (nested) context menus. RecyclerView doesn't support those out of the box, but I have coded this in the past: https://github.com/slapperwan/gh4a/blob/master/app/src/main/java/com/gh4a/widget/ContextMenuAwareRecyclerView.java ... the sub menu will probably be a bit tricky (see e.g. http://www.blackmoonit.com/2010/09/using-submenu-in-listactivitys-context/) but I think it will provide much better UX.

mueller-ma:
I downloaded OctoDroid. Can you tell me, where this dialogs are used?

maniac103:
E.g. in the news feed (default start view if you're logged in) ... on long press, options for visiting the respective repo, user or issue are offered.

mueller-ma:
This menu doesn't seem to have any submenus. What about the menu when clicking on the three dots right to a comment and than "Add reaction"?

maniac103:
That's what I meant by 'tricky' ... that part doesn't have a ready-to-use example yet. Inflation of a sub menu should work the same way in context menus and popup menus though. The code handling the menu you mentioned can be found here: https://github.com/slapperwan/gh4a/blob/master/app/src/main/java/com/gh4a/adapter/CommitNoteAdapter.java#L272-L277

mueller-ma:
> but I think it will provide much better UX.

IMO the UX might be a bit better, but I don't see why two menus are bad UI.

maniac103:
No one said it is. It's fine in that it isn't confusing and is usable, but it just feels unpolished and clunky to mme

maniac103:
PoC implementation of context menu instead of AlertDialogs: https://github.com/openhab/openhab-android/commit/7e8f9216dd40623e65fcc2837eb92065c6801abd

mueller-ma:
I agree that it's looking more polished, but the context menu is much smaller. The title of the NFC state selection menu is cut and most of the time that state menu must be scrolled to view all possible states.
Removing the two NFC actions if NFC isn't available is better UX, but it makes it impossible for UI tests to check for this menu. Emulators don't have NFC support. Maybe we can add some debug switch.

maniac103:
> I agree that it's looking more polished, but the context menu is much smaller. The title of the NFC state selection menu is cut and most of the time that state menu must be scrolled to view all possible states.

I don't see the scrolling as a big issue. The title in the NFC menu can be shortened, as it's clear what the menu is about due to the first level item.

> Removing the two NFC actions if NFC isn't available is better UX, but it makes it impossible for UI tests to check for this menu. Emulators don't have NFC support. Maybe we can add some debug switch.

Yeah, we can force it by checking for emulator, or maybe we can leverage https://developer.android.com/reference/kotlin/android/app/ActivityManager?hl=en#isuseramonkey ... if neither works nicely, we can still use a resource boolean or test-only code overlay or something. All of this shouldn't be a blocker.

mueller-ma:
> The title in the NFC menu can be shortened, as it's clear what the menu is about due to the first level item.

The first level isn't present if the widget doesn't have a linked page.

maniac103:
> The first level isn't present if the widget doesn't have a linked page. And even if we shorten the English string, translations might be longer.

Then we also can shorten translations. TBH, the menu title always felt *way* too long and verbose for me, especially given there's an activity triggered by the menu items which doesn't do anything by itself.
We could make the title e.g. 'Write to NFC tag' and show a text in the NFC activity among the lines of 'Touch NFC tag to make it send the command XY to Item YZ' or 'Touch NFC tag to make it open the app with page XY'.

mueller-ma:
I shortened the labels, should be RTM now.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
```diff
@@ -140,41 +154,118 @@ class WidgetListFragment : Fragment(), WidgetAdapter.ItemClickListener {
...
-                    } else if (linkToHandle != null) {
-                        startActivity(WriteTagActivity.createSitemapNavigationIntent(context, linkToHandle))
+    private fun populateNfcStatesMenu(
+        menu: Menu,
+        context: Context,
+        suggestedCommands: SuggestedCommandsFactory.SuggestedCommands,
+        widget: Widget
+    ) {
+        val name = widget.item?.name ?: return
+        val listener = object: MenuItem.OnMenuItemClickListener {
```
mueller-ma: False positive?

maniac103: You mean the Stickler error? Doesn't look like a false positive to me, as there's a rule to put a space before and after the colon in the `DerivedClassName : BaseClassName` syntax.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
```diff
@@ -140,41 +154,118 @@ class WidgetListFragment : Fragment(), WidgetAdapter.ItemClickListener {
...
+    private fun isEmulator(): Boolean {
+        return Build.FINGERPRINT.startsWith("generic") ||
+            Build.FINGERPRINT.startsWith("unknown") ||
+            Build.MODEL.contains("google_sdk") ||
+            Build.MODEL.contains("Emulator") ||
+            Build.MODEL.contains("Android SDK built for x86") ||
+            Build.MANUFACTURER.contains("Genymotion") ||
+            (Build.BRAND.startsWith("generic") && Build.DEVICE.startsWith("generic")) ||
+            "google_sdk" == Build.PRODUCT
+    }
```
maniac103: Maybe put this to `Util`?

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
```diff
@@ -140,41 +154,118 @@ class WidgetListFragment : Fragment(), WidgetAdapter.ItemClickListener {
...
+                val id = item?.itemId ?: return false
+                if (id == CONTEXT_MENU_ID_WRITE_CUSTOM_TAG) {
+                    val input = EditText(context)
+                    input.inputType = suggestedCommands.inputTypeFlags
+                    val customDialog = AlertDialog.Builder(context)
+                        .setTitle(getString(R.string.item_picker_custom))
+                        .setView(input)
+                        .setPositiveButton(android.R.string.ok) { _, _ ->
+                            startActivity(WriteTagActivity.createItemUpdateIntent(context, name, input.text.toString(),
+                                input.text.toString(), widget.item.label.orEmpty())) }
```
maniac103: Shouldn't the closing `}` be on a new line?

mueller-ma:
The tests pass on my local machine with an api 22 emulator.

PR #1477, opened 2019-08-22
Support autofill for username and password prefs

### LINKED ISSUE #1294: Check if autofill is supported
Autofill should be explicitly disabled for all text fields except the username and password settings dialogs.

https://developer.android.com/guide/topics/text/autofill-optimize

mueller-ma:
Closes #1294


mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
```diff
@@ -25,33 +26,52 @@ import androidx.preference.EditTextPreferenceDialogFragmentCompat
...
             inputType = getInt(0, 0)
             recycle()
         }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            context.obtainStyledAttributes(attrs, intArrayOf(android.R.attr.autofillHints)).apply {
+                autofillHints = getString(0)
+                recycle()
+            }
+        }
```
maniac103: Can't we unify those two blocks into one? The `attr` definition should be present no matter what the actual API level is.
I'd also do the parsing (string -> string array conversion) here and pass around the array, as the string format is specific to the attribute.

Also, `android:autoFillHints` is comma separated, not pipe separated: https://developer.android.com/reference/android/view/View.html#attr_android:autofillHints

mobile/src/main/res/xml/local_connection_preferences.xml
```diff
@@ -14,11 +14,13 @@
         android:inputType="textNoSuggestions"
         android:key="default_openhab_username"
         android:title="@string/settings_openhab_username"
+        android:autofillHints="username|emailAddress"
```
maniac103: `android:autoFillHints="username,emailAddress"`

mobile/src/main/res/xml/remote_connection_preferences.xml
```diff
@@ -14,11 +14,13 @@
         android:inputType="textNoSuggestions"
         android:key="default_openhab_remote_username"
         android:title="@string/settings_openhab_username"
+        android:autofillHints="username|emailAddress"
```
maniac103: As above.

mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
mueller-ma: Done, but can you give it a try on api 22 emulators? The app freezes there if I open the settings activity.

maniac103: Doesn't seem to happen for me (?) I created a new Emulator for Nexus 7, using API 22 + Google APIs as emulator image. Then I started a fossBeta build in that emulator, and settings activity seems just fine for me.

mueller-ma: It seems that there was an issue with the emulator itself, not the build.

PR #1486, opened 2019-08-27
Fix build warnings

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Sitemap.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Sitemap.kt
@@ -28,7 +28,6 @@ import org.w3c.dom.Node
 data class Sitemap internal constructor(
     val name: String,
     val label: String,
-    val link: String,
     val icon: String?,
     val iconPath: String,
     val homepageLink: String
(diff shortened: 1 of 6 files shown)


mobile/src/main/java/org/openhab/habdroid/model/Sitemap.kt
```diff
@@ -38,14 +38,12 @@ fun Node.toSitemap(): Sitemap? {
     var label: String? = null
     var name: String? = null
     var icon: String? = null
-    var link: String? = null
```
maniac103: It looks like the actual issue is that the 3rd parameter to the Sitemap construction below should be `link`, not `homepageLink`.

mueller-ma: The link property of `Sitemap` is only used in `SitemapTest#testGetLink()`. Should I remove `link` at all?

mueller-ma:
There's also a warning, that `actualValue` (`ParsedState` Line 162) cannot be written to parcel, but `NumberState` is annotated with `@Parcelize`. I haven't found the usage of any parcel related functions of `ParsedState`, `Item` or `Widget`. Can `@Parcelize` be removed from these classes?

mobile/src/main/java/org/openhab/habdroid/model/Sitemap.kt
maniac103: I don't have a strong opinion there ... do whatever you prefer there :-)

maniac103:
> There's also a warning, that actualValue (ParsedState Line 162) cannot be written to parcel, but NumberState is annotated with @Parcelize. 

Probably need to implement `actualValue` as method instead of property then.

> I haven't found the usage of any parcel related functions of ParsedState, Item or Widget. Can @Parcelize be removed from these classes?

Well, I made them parcelable so they can be passed around as intent extras. It seems we're currently using this capability only for `ServerProperties`, but I'd keep the capability nevertheless.

PR #1487, opened 2019-08-27
Always render Switch items without mappings as switches.

maniac103:
The notable difference to the old behavior is that we now ignore item
options for Switch items.

See [1] for discussion and [2] for equivalent implementation in BasicUI.

[1] https://github.com/openhab/openhab-core/issues/952
[2] https://github.com/openhab/openhab-webui/pull/94


mueller-ma:
I don't have add-on that creates such mappings. Can you upload a debug apk and link it in the discussion?

maniac103:
@lolodomo Can you please try?

lolodomo:
I have no such binding too. In fact, the Z-Wave binding is the main one but you need to own the appropriate Z-Wave devices.

maniac103:
Hmm, dunno who could test this then ... I don't have Z-Wave devices either ¯\_(ツ)_/¯
I'm reasonablity certain the logic change is correct (in: matches what was done for the web UIs) though.

lolodomo:
Maybe I own an appropriate device to test but I don't know which channel/devices have these switch options.

lolodomo:
In fact, my Fibaro motion sensors are good examples. With the 2.8 version of Android App, I get Ok and Alarm buttons (while a simple switch in Basic UI).
So I can test the change but I need a test apk.

maniac103:
Here you are ... just rename zip to apk:
[mobile-foss-beta-release-unsigned.zip](https://github.com/openhab/openhab-android/files/3555347/mobile-foss-beta-release-unsigned.zip)

mueller-ma:
@lolodomo Did you review this?

lolodomo:
Unfortunately; the apk fails to install on my tablet.

mueller-ma:
Did you uninstall the previous installation before?

lolodomo:
Yes and I enabled the unknown source option.

maniac103:
Sorry, I missed signing the apk. This one should be better:
[mobile-foss-beta-debug.zip](https://github.com/openhab/openhab-android/files/3573321/mobile-foss-beta-debug.zip)

lolodomo:
Tested, it works as expected. The Z-wave channel is now rendered as a switch rather than two buttons, and I have another switch widget with mappings which is still rendered with buttons.

mueller-ma:
Thanks!

PR #1495, opened 2019-08-30
Improve UrlInputPreference

mueller-ma:
* Change input type from text to url
* Remove whitespaces before saving instead of showing an error
* Make sure ok button is disabled if invalid url is entered

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/preference/UrlInputPreference.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/preference/UrlInputPreference.kt
@@ -13,14 +13,15 @@
 
 package org.openhab.habdroid.ui.preference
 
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/preference/UrlInputPreference.kt
```diff
@@ -43,6 +48,7 @@ class UrlInputPreference constructor(context: Context, attrs: AttributeSet) : Ed
             if (view != null) {
                 editor = view.findViewById(android.R.id.edit)
                 editor.addTextChangedListener(this)
+                editor.inputType = InputType.TYPE_TEXT_VARIATION_URI
```
mueller-ma: Setting this is xml didn't work.

PR #1496, opened 2019-08-30
Fix build warnings

mueller-ma:
TODO:

- [x] `JSONObject#optString()` doesn't accept `null` as fallback anymore. IMO `null` is better than an empty string.
- [ ] ~~`KeyguardManager.createConfirmDeviceCredentialIntent()`, see #1295~~ => new PR

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
@@ -58,7 +58,8 @@ class BackgroundTasksManager : BroadcastReceiver() {
                 NotificationUpdateObserver.createNotificationChannels(context)
             }
             ACTION_RETRY_UPLOAD -> {
-                for (info in intent.getParcelableArrayListExtra<RetryInfo>(EXTRA_RETRY_INFO_LIST)) {
+                val retryInfoList = intent.getParcelableArrayListExtra<RetryInfo>(EXTRA_RETRY_INFO_LIST) ?: return
+                for (info in retryInfoList) {
                     enqueueItemUpload(context, info.tag, info.itemName, info.value)
                 }
             }
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
@@ -58,8 +58,7 @@ class BackgroundTasksManager : BroadcastReceiver() {
                 NotificationUpdateObserver.createNotificationChannels(context)
             }
             ACTION_RETRY_UPLOAD -> {
-                val retryInfoList = intent.getParcelableArrayListExtra<RetryInfo>(EXTRA_RETRY_INFO_LIST) ?: return
-                for (info in retryInfoList) {
+                intent.getParcelableArrayListExtra<RetryInfo>(EXTRA_RETRY_INFO_LIST)?.forEach { info ->
                     enqueueItemUpload(context, info.tag, info.itemName, info.value)
                 }
             }
--- a/mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt
@@ -101,17 +101,17 @@ interface ConnectionManagerHelper {
     @TargetApi(26)
     private open class ChangeCallbackHelperApi26 constructor(context: Context) : ConnectivityManager.NetworkCallback() {
         var changeCallback: ConnectionChangedCallback? = null
-        private val connectivityManager = context.getSystemService(ConnectivityManager::class.java)
+        private val connectivityManager = context.getSystemService(ConnectivityManager::class.java)!!
         private val lastKnownCaps = HashMap<Network, NetworkCapabilities>()
         private var callbackJob: Job? = null
 
         init {
-            connectivityManager?.registerDefaultNetworkCallback(this)
+            connectivityManager.registerDefaultNetworkCallback(this)
         }
 
         fun shutdown() {
             callbackJob?.cancel()
-            connectivityManager?.unregisterNetworkCallback(this)
+            connectivityManager.unregisterNetworkCallback(this)
         }
 
         override fun onCapabilitiesChanged(network: Network, networkCapabilities: NetworkCapabilities) {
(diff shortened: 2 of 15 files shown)


mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -57,7 +57,8 @@ class BackgroundTasksManager : BroadcastReceiver() {
                 NotificationUpdateObserver.createNotificationChannels(context)
             }
             ACTION_RETRY_UPLOAD -> {
-                for (info in intent.getParcelableArrayListExtra<RetryInfo>(EXTRA_RETRY_INFO_LIST)) {
+                val retryInfoList = intent.getParcelableArrayListExtra<RetryInfo>(EXTRA_RETRY_INFO_LIST) ?: return
+                for (info in retryInfoList) {
```
mueller-ma: Is there something like `for (info in? retryInfoList)`?

maniac103: Maybe `retryInfoList?.forEach { info ->`?

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt
```diff
@@ -100,39 +106,35 @@ interface ConnectionManagerHelper {
...
-                        scheduleCallback()
-                    }
-                    lastKnownCaps[network] = networkCapabilities
+        override fun onCapabilitiesChanged(network: Network, networkCapabilities: NetworkCapabilities) {
+            val knownCaps = lastKnownCaps[network]
+            if (knownCaps != null) {
+                if (knownCaps.isUsable() != networkCapabilities.isUsable()) {
+                    scheduleCallback()
                 }
+                lastKnownCaps[network] = networkCapabilities
```
maniac103: Can pull this out of the if and collapse the two ifs into one.

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt
```diff
@@ -174,10 +176,13 @@ interface ConnectionManagerHelper {
...
         private val connectivityManager = context.getSystemService(ConnectivityManager::class.java)
         val currentConnection: ConnectionType get() {
+            if (connectivityManager == null) {
+                return ConnectionType.None
+            }
             val activeConnectionTypes = connectivityManager.allNetworks
                     .map { network -> connectivityManager.getNetworkInfo(network) }
-                    .filter { info -> info.isConnected }
-                    .map { info -> info.type }
+                    .filter { info -> info?.isConnected ?: false}
```
maniac103: Checking for `== true` would be clearer than Elvis IMHO.

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt
```diff
@@ -174,10 +176,13 @@ interface ConnectionManagerHelper {
     private class NetworkTypeHelperApi23 constructor(context: Context) {
         private val connectivityManager = context.getSystemService(ConnectivityManager::class.java)
         val currentConnection: ConnectionType get() {
+            if (connectivityManager == null) {
```
maniac103: As much as I dislike the `!!` operator, I wonder whether the `getSystemService` call wouldn't be a good exception.

mobile/src/main/java/org/openhab/habdroid/model/CloudNotification.kt
```diff
@@ -35,10 +36,10 @@ data class CloudNotification internal constructor(
 fun JSONObject.toCloudNotification(): CloudNotification {
     var created: Long = 0
     if (has("created")) {
-        val format = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.S'Z'")
+        val format = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.S'Z'", Locale.US)
         format.timeZone = TimeZone.getTimeZone("UTC")
         try {
-            created = format.parse(getString("created")).time
+            created = format.parse(getString("created"))?.time  ?: 0
```
maniac103: Excess space character.

Besides that, maybe instead of the has() call above, do getString() earlier and skip if result is null?

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt
mueller-ma: However the method got annotated with `@Nullable` for a reason, so I'd check for null.

maniac103: The reason is 'not every platform has every feature' (e.g. device has no Bluetooth -> `getSystemService(BLUETOOTH_SERVICE)` returns null). I don't think that will happen with connectivity manager.

mobile/src/main/java/org/openhab/habdroid/model/CloudNotification.kt
mueller-ma: `parse()` may return null, so I need this elvis even if I move `JSONObject.getString()` above.

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
```diff
@@ -177,6 +177,11 @@ inline fun <T> JSONArray.map(transform: (JSONObject) -> T): List<T> {
     return (0 until length()).map { index -> transform(getJSONObject(index)) }
 }
 
+fun JSONObject.getStringOrNull(key: String): String? {
```
mueller-ma: While this fixes all the warnings, I really don't like the solution. Why must be the fallback value in optString non-null? :|

maniac103: We can't change that anyway, so thinking about that rationale is pointless ;-)
I actually like the solution, but the method needs to be called `optStringOrNull`, since the get* methods are for fetching *mandatory* fields (JsonException is thrown if they aren't present).

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
```diff
@@ -177,6 +177,11 @@ inline fun <T> JSONArray.map(transform: (JSONObject) -> T): List<T> {
...
+    val value = optString(key, "")
```
maniac103: Should probably implement this as `return if (has(key)) getString(key) else null` since otherwise a field which is present but empty is treated like a missing field.

mueller-ma: Done.

Issue #1351, opened 2019-05-08
Frames without label are not rendered


mueller-ma:
### Actual behaviour
Frames without label are not rendered.

PR #1502, opened 2019-09-02
Fix handling of frames without label.

### LINKED ISSUE #1437: Frame visibility in sitemap ignored

maniac103:
We shouldn't simply filter them, as doing so
- produces wrong visual output (no spacer)
- creates issues when visibility of that frame is changed

Instead, just hide the label and leave the spacer in place.

Fixes #1437

### DIFF
--- a/mobile/src/androidTest/java/org/openhab/habdroid/ui/BasicWidgetTest.kt
+++ b/mobile/src/androidTest/java/org/openhab/habdroid/ui/BasicWidgetTest.kt
@@ -45,25 +45,25 @@ class BasicWidgetTest : TestWithoutIntro() {
         val recyclerView = onView(withId(R.id.recyclerview))
 
         recyclerView
-                .perform(scrollToPosition<RecyclerView.ViewHolder>(0))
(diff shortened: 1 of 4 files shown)


openhab-5iver:
Thank you!

mueller-ma:
Besides the failing test case, LGTM!

maniac103:
Fixed.

mueller-ma:
Also fixes #1351.

PR #1503, opened 2019-09-03
Use new androidx.media2 video playback library

maniac103:
This improves VideoView sizing behavior and adds video playback controls. This library has a minSdkVersion of 19 (Android 4.4, Kitkat) though.


maniac103:
@mueller-ma Can you check whether we have a significant number of users on 4.1, 4.2 or 4.3?

mueller-ma:
Android version 4.1, 4.2, 4.3 and together are less than 1% of the current users. 4.4 is about 2%.
IMO using the library is worth dropping support for pre 4.4 devices.

I tested this PR with one mp4 video and it works good. The video didn't play at all with the old video player.
After fixing the Stickler issues and remove any code path that are only for pre 4.4 devices (if there are any), LGTM!

mueller-ma:
Did you check if we can drop `MjpegVideoViewHolder` now?

maniac103:
I didn't, but I'd be somewhat surprised if that were the case...MJPEG is somewhat unusual.

PR #1504, opened 2019-09-03
Offer opening MapViews in Browser or native Maps app

mueller-ma:

### DIFF
--- a/mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.kt
+++ b/mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.kt
@@ -262,6 +262,6 @@ fun Location.toGeoPoint(): GeoPoint {
     return GeoPoint(this)
 }
 
-fun getUrlForMaps(location: Location): String {
-    return "https://www.openstreetmap.org/#map=16/${location.latitude}/${location.longitude}"
+fun Location.toMapsUrl(): String? {
+    return "https://www.openstreetmap.org/#map=16/$latitude/$longitude"
 }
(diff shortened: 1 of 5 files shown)


mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.kt
```diff
@@ -260,3 +261,7 @@ fun MapView.setMarker(
 fun Location.toGeoPoint(): GeoPoint {
     return GeoPoint(this)
 }
+
+fun getUrlForMaps(location: Location): String {
+    return "https://www.openstreetmap.org/#map=16/${location.latitude}/${location.longitude}"
+}
```
maniac103: Suggestion: `fun Location?.toMapsUrl(): String?`

mobile/src/full/java/org/openhab/habdroid/ui/MapViewHelper.kt
```diff
@@ -193,3 +193,7 @@ fun GoogleMap.setMarker(position: LatLng, item: Item, label: CharSequence?, canD
 fun Location.toLatLng(): LatLng {
     return LatLng(latitude, longitude)
 }
+
+fun getUrlForMaps(location: Location): String {
+    return "https://www.google.de/maps/@${location.latitude},${location.longitude},16z"
+}
```
maniac103: As above

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
```diff
@@ -183,6 +184,10 @@ class WidgetListFragment : Fragment(), WidgetAdapter.ItemClickListener {
                     widget.linkedPage?.let { createShortcut(context, it) }
                     return true
                 }
+                CONTEXT_MENU_ID_OPEN_IN_MAPS -> {
+                    getUrlForMaps(widget.item?.state?.asLocation ?: return true).toUri().openInBrowser(context)
```
maniac103: With the suggestion above, this becomes `widget.item?.state?.asLocation?.toMapsUrl()?.toUri()?.openInBrowser(context)`

Still slightly weird with the large number of question marks, but IMHO better than the return in the middle.

mueller-ma: Imagine doing that in Java xD

PR #1511, opened 2019-09-04
Improve slider thumb

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -433,7 +433,7 @@ class WidgetAdapter(
         init {
             seekBar.setOnSeekBarChangeListener(this)
             val now = Calendar.getInstance()
-            if (now.get(Calendar.DAY_OF_MONTH) == 31 && now.get(Calendar.MONTH) == 10) {
+            if (now.get(Calendar.DAY_OF_MONTH) == 31 && now.get(Calendar.MONTH) == Calendar.OCTOBER) {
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -399,14 +421,19 @@ class WidgetAdapter(
         inflater: LayoutInflater,
         parent: ViewGroup,
         private val connection: Connection,
-        colorMapper: ColorMapper
+        colorMapper: ColorMapper,
+        context: Context
```
maniac103: This is redundant, can use `itemView.context`.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -399,14 +421,19 @@ class WidgetAdapter(
...
     ) : LabeledItemBaseViewHolder(inflater, parent, R.layout.widgetlist_slideritem, connection, colorMapper),
         SeekBar.OnSeekBarChangeListener {
         private val seekBar: SeekBar = itemView.findViewById(R.id.seekbar)
         private var boundWidget: Widget? = null
 
         init {
             seekBar.setOnSeekBarChangeListener(this)
+            val date = SimpleDateFormat("MM-dd", Locale.US).format(Date())
+            if (date == "10-31") {
```
maniac103:
```
val now = Date()
if (now.dayOfMonth == 31 && now.month == Month.OCTOBER) {
```

mueller-ma: These functions are deprecated, but I the recommended alternative.

mueller-ma:
> I assume this is going to be temporary? ;-)

I'd like to keep this.

PR #1515, opened 2019-09-05
Send call state to server

### LINKED ISSUE #1512: Update an Item with the state of phone calls
Like the alarm clock, it would be great to know when one of our phones was on a cellular call so that the audio notifications in our home could be disabled. Bonus if VoIP and video calls were included, like when on a Skype call. When on a call, update the Item to ON. When the call ends, update the Item to OFF.

Tasker could probably be used for this.

mueller-ma:
Fixes #1512


mobile/src/main/res/xml/preferences.xml
```diff
@@ -91,6 +91,14 @@
             android:title="@string/settings_alarm_clock"
             android:dependency="default_openhab_demomode"
             android:icon="@drawable/ic_alarm_grey_24dp" />
+        <org.openhab.habdroid.ui.preference.ItemUpdatingPreference
+            android:defaultValue="false|CallState"
+            android:key="phoneState"
+            app:helpHint="@string/settings_alarm_clock_howto_summary"
```
maniac103: item_update_pref_howto_summary? (Same above)

mueller-ma:
I want to add a phone-off icon, when it's available as outline: https://github.com/Templarian/MaterialDesign/issues/4376

PeterShaggyNoble:
We're currently wrapping up the next release of MDI (v4.4.95); `phone-off-outline` will be available in the release after that (v4.5.95) but we don't yet have a timeframe for it.

mueller-ma:
@PeterShaggyNoble Thanks for the icon!

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -592,6 +599,13 @@ fun Preference?.getPrefValue(defaultValue: String? = null): String? {
     return sharedPreferences?.getString(key, defaultValue)
 }
 
+fun Preference?.setPrefValue(value: String? = null) {
+    if (this == null || value == null) {
+        return
+    }
+    sharedPreferences?.edit()?.putString(key, value)?.apply()
+}
```
maniac103: This seems unused and can be removed.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -371,6 +374,20 @@ class PreferencesActivity : AbstractBaseActivity() {
             }
         }
 
+        override fun onRequestPermissionsResult(requestCode: Int, permissions: Array<String>, grantResults: IntArray) {
+            when (requestCode) {
+                PERMISSIONS_REQUEST_READ_PHONE_STATE -> {
+                    if (grantResults.isEmpty() || grantResults[0] != PackageManager.PERMISSION_GRANTED) {
+                        context!!.showToast(R.string.settings_phone_state_permission_denied)
```
maniac103: Maybe rather `context?.showToast()`? In the unlikely case of the fragment being detached, it seems fine to not show a toast, better than crashing at least.

mobile/src/main/java/org/openhab/habdroid/util/PrefExtensions.kt
```diff
@@ -120,3 +121,8 @@ fun SharedPreferences.Editor.updateDefaultSitemap(sitemap: Sitemap?) {
         putString(Constants.PREFERENCE_SITEMAP_LABEL, sitemap.label)
     }
 }
+
+fun SharedPreferences.Editor.disableItemUpdatePref(context: Context, key: String) {
+    val item = context.getPrefs().getString(key).toItemUpdatePrefValue().second
+    putString(key, "false|$item")
+}
```
maniac103: The definition extension function seems a little unclear/misleading to me. You call it on a specific prefs editor, but fetch the used value from `getPrefs()`, which the editor may or may not belong to. I think it's better to not make this an extension function and inline this code into the (one) call place.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -321,25 +324,25 @@ class PreferencesActivity : AbstractBaseActivity() {
...
+
+                true
             }
 
             sendDeviceInfoPrefixPref.setOnPreferenceChangeListener { _, newValue ->
-                updateAlarmClockPreferenceSummary(alarmClockPref, newValue as String,
-                    alarmClockPref.getPrefValue().toItemUpdatePrefValue())
+                val prefix = newValue as String
+                (alarmClockPref as ItemUpdatingPreference).updateSummaryAndIcon(prefix)
+                (phoneStatePref as ItemUpdatingPreference).updateSummaryAndIcon(prefix)
```
maniac103: Maybe rather cast the result of `findPreference`? Would be a bit more readable, I guess.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -321,25 +324,25 @@ class PreferencesActivity : AbstractBaseActivity() {
...
-                    updateAlarmClockPreferenceIcon(preference, newValue)
-                    updateAlarmClockPreferenceSummary(preference, prefix, value)
-                    true
+            }
+
+            phoneStatePref.setOnPreferenceChangeListener { preference, newValue ->
+                @Suppress("UNCHECKED_CAST")
+                val value = newValue as Pair<Boolean, String>
+                if (value.first && ContextCompat.checkSelfPermission(preference.context,
+                        Manifest.permission.READ_PHONE_STATE) != PackageManager.PERMISSION_GRANTED) {
```
maniac103: Suggestion: Make an extension function: `fun Context.hasPermission(String permission): Boolean` for this.

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
mueller-ma: I cast the result of `findPreference()` to `Preference` in #1517. After merging that PR, I can rebase this one and adjust the cast.

openhab-5iver:
Tested this on two phones and everything is working great! This is a great feature... thank you!

mueller-ma:
@PeterShaggyNoble Thanks! The icon was available before on the website http://materialdesignicons.com/ where I downloaded it.

PR #1516, opened 2019-09-05
Use Settings Panel to enable NFC in WriteTagActivity

mueller-ma:
Also listen for NFC adapter state changes.
See #1295


mobile/src/main/java/org/openhab/habdroid/ui/WriteTagActivity.kt
```diff
@@ -259,8 +271,14 @@ class WriteTagActivity : AbstractBaseActivity(), CoroutineScope {
 
             val nfcActivate = view?.findViewById<TextView>(R.id.nfc_activate)
             nfcActivate?.isVisible = true
+
+            val action = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+                Settings.Panel.ACTION_NFC
```
mueller-ma: I haven't tested the setting panel as I don't have a device with NFC and Android Q.

PR #1523, opened 2019-09-06
Allow writing empty states on NFC tags

mueller-ma:


maniac103:
What's the rationale behind this? On a first look, this doesn't seem to make sense to me.

mueller-ma:
For String items it can make sense to set them to an empty string.
I discovered that it's possible to select "Custom", enter nothing and click on. In this case the app crashes. I could also disable the "OK" button if the edittext is empty, but this seems the better solution to me.

PR #1526, opened 2019-09-06
Enable backups and encrypt username and password

### LINKED ISSUE #1508: Enable backups
[Backups have been disable](https://github.com/openhab/openhab-android/blob/master/mobile/src/main/AndroidManifest.xml#L24) to make sure that usernames and passwords aren't saved in clear text to Google Drive or any other backup service.
Usernames and passwords should be moved to a second shared prefs file, that will be encrypted (https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences) and excluded from backups.

mueller-ma:
Closes #1508

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
@@ -22,33 +22,23 @@ import androidx.core.os.bundleOf
 import androidx.fragment.app.DialogFragment
 import androidx.preference.EditTextPreference
 import androidx.preference.EditTextPreferenceDialogFragmentCompat
-import org.openhab.habdroid.R
-import org.openhab.habdroid.util.Constants
-import org.openhab.habdroid.util.getPrefs
-import org.openhab.habdroid.util.getSecretPrefs
-import org.openhab.habdroid.util.getString
 
 class CustomInputTypePreference constructor(context: Context, attrs: AttributeSet) :
     EditTextPreference(context, attrs) {
     private val inputType: Int
     private var autofillHints: Array<String>? = null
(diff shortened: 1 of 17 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
```diff
@@ -57,16 +70,24 @@ class CustomInputTypePreference constructor(context: Context, attrs: AttributeSe
                 }
             }
             super.onBindDialogView(view)
+            if (arguments?.getBoolean(KEY_IS_PREF_SECRET) == true) {
+                editor?.setText(context?.getSecretPrefs()?.getString(this.preference.key))
```
maniac103: Where/how is the value written to the secret prefs?

I think a better approach would be writing a (trivial) `PreferenceDataStore` around the secret prefs and attaching that to the respective prefs from within `PreferenceActivity`.

maniac103: Implementation to above suggestion: https://github.com/maniac103/openhab.android/commits/backup ... IMHO it simplifies logic quite a bit.

mueller-ma: Indeed, that's better.

mueller-ma:
I'd add some hint to the user that he has to re-enter the credentials after restoring. Is a simple toast enough?

maniac103:
Maybe a Snackbar with button to open settings?

mueller-ma:
I'd rather go for a notification since the snackbars on startup are easily dismissed by other snackbars, e.g. those for showing if connecting to local or remote server.

PR #1532, opened 2019-09-09
Add checksum-dependency-plugin for verification of plugin/dependency checksums

vlsi:
`checksum-dependency-plugin` is a superset of `gradle-witness`, and it enables to increase the level of security.

See https://github.com/vlsi/vlsi-release-plugins#checksum-dependency-plugin
See https://medium.com/@vladimirsitniko/dependency-verification-checksum-vs-pgp-582e76207019?sk=7485298b76eaf9f935b899b002f4c3b5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ android:
 install:
   - echo "mapsApiKey=AIzaSyA2t2893e9wccs9EnlnauJRgtqFNYQSp-o" >> ~/.gradle/gradle.properties
   - echo yes | sdkmanager "platforms;android-27"
-  - ./gradlew --version
+  - ./gradlew --version -PchecksumPrint
   - sdkmanager --list || true
 
 stages:
--- a/checksum.xml
+++ b/checksum.xml
@@ -1,0 +1,526 @@
+<?xml version='1.0' encoding='utf-8'?>
+<dependency-verification version='1'>
+  <trust-requirement pgp='GROUP' checksum='NONE' />
+  <ignored-keys />
+  <trusted-keys>
+    <trusted-key id='c4c8cb73b1435348' group='com.android.tools.build' />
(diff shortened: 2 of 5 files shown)


mueller-ma:
Thanks for your PR. What needs to be changed if a dependency is bumped?

vlsi:
A) Travis would print new `checksum.xml` (I added `-PchecksumPrint` there)
B) If dependency added/updated, the plugin would print the ways to proceed. For instance, if I comment some checksums in checksum.xml I get the following output:

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'openhab-android'.
> Checksum/PGP violations detected on resolving configuration :classpath
    No trusted PGP keys are configured for group com.squareup:
      com.squareup:javapoet:1.8.0 (pgp=[8e3f0de7ae354651], sha512=[computation skipped])
      com.squareup:javawriter:2.5.0 (pgp=[80c08b1c29100955], sha512=[computation skipped])

  You might want to add -PchecksumFailOn=build_finish if you are brave enough
  It will collect all the violations, however untrusted code might be executed (e.g. from a plugin)
  You can find updated checksum.xml file at .../openhab-android/build/checksum/checksum.xml.
  You might add -PchecksumUpdate to update root checksum.xml file.
```

Does the above answer your question?

So you can edit `checksum.xml` manually (e.g. to cleanup for unused entries), and you can use `-PchecksumUpdate` to automatically add new ones.

mueller-ma:
We're using Dependabot for automated version bumps, so that would stop working. Thus I'm closing the PR.

vlsi:
Well, that depends.

1) I'm not sure, however if Dependabot can add `-PchecksumUpdate`, then the same bot would automatically update `checksum.xml`

2) Then dependencies are often published by the same release managers, so PGP keys would be intact, and often you don't have to update `checksum.xml` file

PR #1534, opened 2019-09-10
Fix ktlint issues automatically

mueller-ma:
Let's see how good it works.


maniac103:
Hmm, I'm not a fan of automatic code formatting ... I guess we'll see.

PR #1541, opened 2019-09-14
Try to fix IllegalStateException in ContentController

mueller-ma:
````
Fatal Exception: java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
       at androidx.fragment.app.FragmentManagerImpl.checkStateLoss + 1536(FragmentManagerImpl.java:1536)
       at androidx.fragment.app.FragmentManagerImpl.enqueueAction + 1558(FragmentManagerImpl.java:1558)
       at androidx.fragment.app.BackStackRecord.commitInternal + 317(BackStackRecord.java:317)
       at androidx.fragment.app.BackStackRecord.commit + 282(BackStackRecord.java:282)
       at org.openhab.habdroid.ui.activity.ContentControllerOnePane.executeStateUpdate$mobile_fullBetaRelease + 50(ContentControllerOnePane.kt:50)
       at org.openhab.habdroid.ui.activity.ContentController.updateFragmentState + 450(ContentController.kt:450)
       at org.openhab.habdroid.ui.activity.ContentController.onLoadFailure + 438(ContentController.kt:438)
       at org.openhab.habdroid.ui.activity.PageConnectionHolderFragment$ConnectionHandler$load$1.invokeSuspend + 298(PageConnectionHolderFragment.kt:298)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith + 33(ContinuationImpl.kt:33)
````

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -235,7 +235,6 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
     override fun onStart() {
         Log.d(TAG, "onStart()")
         super.onStart()
-        isStarted = true
 
         ConnectionFactory.addListener(this)
 
@@ -251,7 +250,6 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
 
     public override fun onStop() {
         Log.d(TAG, "onStop()")
-        isStarted = false
         super.onStop()
         ConnectionFactory.removeListener(this)
         serviceResolveJob?.cancel()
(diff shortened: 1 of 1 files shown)


maniac103:
Hmm, I don't think this fix is correct. The connections are normally shut down on onStop, which means this normally can't happen. I wonder whether this is a case of load failure during activity recreation (after rotating phone), which kinda would explain this.

How many instances of this crash did you see?

mueller-ma:
> How many instances of this crash did you see?

This crash happened ~100 times for ~90 users in 2.8.0.

I don't know how to reproduce this crash, but my guess was, that setting `isStarted` to true in `onStart()` is too early. Maybe it should be set to true in `onRestoreInstanceState()`, since it's set to false in `onSaveInstanceState()`.

maniac103:
I think I once traced that exception being thrown in the fragment code, and it's bound to a flag which is set in onSaveInstanceState and reset in onStart.

PR #1551, opened 2019-09-26
Ignore alarm clocks set by Tasker

### LINKED ISSUE #1341: AlarmClock<->Tasker Issue
### Actual behaviour
Describe us what happens:
The time of the alarmclock is not sent if you are using tasker at the same time. So you have to deactivate the tasker app and the Item AlarmClock is updated correctly.

Using the latest beta app 2.6.21-beta

mueller-ma:
Tasker mis-uses this api, so we ignore alarm clocks set by Tasker.

Closes #1341

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
@@ -124,7 +124,10 @@ class BackgroundTasksManager : BroadcastReceiver() {
             Constants.PREFERENCE_ALARM_CLOCK,
             Constants.PREFERENCE_PHONE_STATE
         )
-        private val VALUE_GETTER_MAP = HashMap<String, (Context) -> String>()
+        private val IGNORED_PACKAGES_FOR_ALARM = listOf(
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -124,7 +124,10 @@ class BackgroundTasksManager : BroadcastReceiver() {
             Constants.PREFERENCE_ALARM_CLOCK,
             Constants.PREFERENCE_PHONE_STATE
         )
-        private val VALUE_GETTER_MAP = HashMap<String, (Context) -> String>()
+        private val IGNORED_PACKAGES_FOR_ALARM = listOf(
+            "net.dinglisch.android.taskerm"
+        )
+        private val VALUE_GETTER_MAP = HashMap<String, (Context) -> String?>()
```
maniac103: Probably should make those (up to and including `KNOWN_KEYS`) const

mueller-ma: `const` does only work with primitives or `String`s.

PR #1552, opened 2019-09-26
Explain how to speed up chart rendering

mueller-ma:
Closes https://github.com/openhab/openhab-docs/issues/666


maniac103:
`The charts are loaded to slow` -> `Chart loading is too slow`

PR #1554, opened 2019-09-26
Hide Sitemaps in drawer

### LINKED ISSUE #1305: Hide Sitemaps
> Possibility to make certain sitemaps hidden for the family (some are interesting for me but confusing for family members)

from a Play Store review.

mueller-ma:
The Sitemaps are hidden in the drawer by default, but there's a setting to change that.

See #392
Closes #1305

### DIFF
--- a/mobile/src/main/res/xml/preferences.xml
+++ b/mobile/src/main/res/xml/preferences.xml
@@ -31,6 +31,10 @@
             android:clickable="true"
             android:key="default_openhab_clear_default_sitemap"
             android:title="@string/settings_clear_default_sitemap" />
+        <SwitchPreference
+            android:defaultValue="false"
+            android:key="show_sitemaps"
+            android:title="@string/settings_show_sitemaps_in_drawer" />
     </PreferenceCategory>
     <PreferenceCategory android:title="@string/settings_display_title">
         <ListPreference
--- a/mobile/src/main/res/xml/preferences.xml
+++ b/mobile/src/main/res/xml/preferences.xml
@@ -27,6 +27,8 @@
             android:summary="@string/settings_openhab_none"
             android:title="@string/settings_openhab_sslclientcert"
             android:dependency="default_openhab_demomode" />
+    </PreferenceCategory>
+    <PreferenceCategory android:title="@string/settings_sitemap_title">
         <Preference
             android:clickable="true"
             android:key="default_openhab_clear_default_sitemap"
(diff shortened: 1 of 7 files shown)


maniac103:
I don't really like having a separate setting for this ... how about the following:
- If only one sitemap is present, unconditionally hide sitemap list
- If more than one is present, enhance the default sitemap selection dialog so that it allows both selecting a default sitemap and the set of sitemaps to be shown
- If this set is empty, don't show sitemaps in drawer
- Probably also need to store last sitemap list and re-show dialog if set is non-empty and sitemap list changed

mueller-ma:
> If only one sitemap is present, unconditionally hide sitemap list

:+1: 

> If more than one is present, enhance the default sitemap selection dialog so that it allows both selecting a default sitemap and the set of sitemaps to be shown

This sounds a bit overkill to me. Sitemaps are not intended to be a navigation element. The only real reason for having multiple sitemaps I see is having different sitemaps for different users or a sitemap for debugging purpose. Thus IMO it's enough to show all sitemaps or none.

> If this set is empty, don't show sitemaps in drawer

:+1:

mueller-ma:
What about adding a simple switch ("Show other sitemaps in drawer") to the selection dialog of the default sitemap?

maniac103:
> What about adding a simple switch ("Show other sitemaps in drawer") to the selection dialog of the default sitemap?

Sounds like a good compromise to me, I'm just unsure of the wording ... does the average user know the term 'drawer'? How about `Show all sitemaps in side menu`?

mueller-ma:
> What about adding a simple switch ("Show other sitemaps in drawer") to the selection dialog of the default sitemap?

Actually I don't like the idea anymore. If a user updates the app and still wants to see the sitemaps, going to settings and resetting the default sitemap sounds undiscoverable to me.

mobile/src/main/res/xml/preferences.xml
```diff
@@ -31,6 +31,10 @@
             android:clickable="true"
             android:key="default_openhab_clear_default_sitemap"
             android:title="@string/settings_clear_default_sitemap" />
+        <SwitchPreference
```
maniac103: I wonder whether we shouldn't introduce a new category here, since both sitemap prefs aren't connection related...can't think of a good header string at the moment though :-(

mueller-ma:
I added the default sitemap to the drawer. IMO the drawer should always contain the "currently active" item.

PR #1556, opened 2019-09-26
Don't treat # as anchor if followed by a slash

### LINKED ISSUE #1549: (Bug) Android App Webview rendering issue
The Android app has a rendering issue with below URL in Webview:
https://mapviewer-deredactie.be-mobile.biz/#/map?lang=nl_BE&@=50.98955680038983,4.395904541015624,9

In Android it gives a line of text displaying only
`/#/map?lang=nl_BE&@=50.98955680038983,4.395904541015624,9`

However, it does render correct (as supposed to be) when using the Sitemap in Safari or Firefox on a Mac.

mueller-ma:
Fixes #1549

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/AnchorWebViewClient.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/AnchorWebViewClient.kt
@@ -39,7 +39,7 @@ open class AnchorWebViewClient(
 
     init {
         val pos = url.lastIndexOf("#") + 1
-        if (pos != 0 && pos < url.length - 1) {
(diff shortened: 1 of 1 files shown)


maniac103:
I wonder whether we shouldn't do `url.toUri().fragment` to get the fragment ... the framework parsing code is well tested, I'd assume.

mueller-ma:
Nevermind, the anchor handling doesn't work for wikipedia either. I'll check it.

> I wonder whether we shouldn't do url.toUri().fragment to get the fragment ... the framework parsing code is well tested, I'd assume.

:+1:

PR #1557, opened 2019-09-26
Prune work when disabling item update pref

### LINKED ISSUE #1289: Alarm clock item update notification keeps showing up
### Actual behaviour
From time to time (mostly after app updates or after the app has been killed by the os and restarted by the user) the notification "1 Item could not be updated" shows up.

### Expected behaviour
It shouldn't do so, because I have disabled it in the settings.

 
### Steps to reproduce
1. Enable sending alarm clock
2. Don't create the item on the server
3. Get the notification
4. Disable sending alarm clock
5. Force close app
6. Open app
7. Get the notification

#### Client
* Android version: 7.1.2
* Device model: Oneplus X
* Stock or customized ROM: Lineage 14.2
* App version : 2.6.13

mueller-ma:
Closes #1289

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
@@ -177,7 +177,10 @@ class BackgroundTasksManager : BroadcastReceiver() {
             }
 
             if (!setting.first) {
-                WorkManager.getInstance(context).cancelAllWorkByTag(key)
+                with(WorkManager.getInstance(context)) {
+                    cancelAllWorkByTag(key)
+                    pruneWork()
+                }
                 return
             }
 


mobile/src/main/java/org/openhab/habdroid/ui/preference/ItemUpdatingPreference.kt
```diff
@@ -206,6 +206,11 @@ fun String?.toItemUpdatePrefValue(): Pair<Boolean, String> {
     return Pair(this!!.substring(0, pos).toBoolean(), substring(pos + 1))
 }
 
+fun Any?.toItemUpdatePrefValue(): Pair<Boolean, String> {
+    @Suppress("UNCHECKED_CAST")
+    return this as Pair<Boolean, String>
+}
```
maniac103: How's this related to the issue in question?

(Besides that, I'm not a fan of this extension function: it's only used once and only used as a warning suppressor. Supressing the warning were only OK for me if the method ensures the cast is safe)

mueller-ma: This change is a left over from a previous attempt to fix this issue. Since it's only used in one place now, I'll remove it.

PR #1563, opened 2019-09-30
Improve preference activity startup time

maniac103:
Creating the shared prefs instance is an operation that isn't trivial
(takes ~50ms on a OnePlus 6), so avoid doing it over and over again.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
@@ -13,13 +13,32 @@
 
 package org.openhab.habdroid.core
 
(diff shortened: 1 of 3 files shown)


maniac103:
I widened the scope to include other performance enhancement fixes for the pref activity.

mueller-ma:
> Creating the shared prefs instance is an operation that isn't trivial

Same applies to `Context.getPrefs()`, right?

maniac103:
> Same applies to Context.getPrefs(), right?

No, as the result of `Context.getSharedPreferences` is cached: https://android.googlesource.com/platform/frameworks/base/+/refs/heads/android10-release/core/java/android/app/ContextImpl.java#451

PR #1564, opened 2019-09-30
Update OkHttp to latest version.

maniac103:
Update to latest version in 3.12.x series, as we don't want to bump min
required SDK version at this point.


maniac103:
Dependabot likely only tracks the newest version in the main branch.

PR #1565, opened 2019-09-30
Fix HTTP response body leaks.

### LINKED ISSUE #1560: HTTP response body leak
Latest git master as of today yields the following in logcat:
```
09-28 21:15:35.086 12666 22297 W System  : A resource failed to call response.body().close(). 
09-28 21:15:35.086 12666 22297 I chatty  : uid=10211(org.openhab.habdroid) OkHttp Connecti identical 6 lines
09-28 21:15:35.086 12666 22297 W System  : A resource failed to call response.body().close(). 
```

I'll investigate this...

maniac103:
Fixes #1560

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/HttpClient.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/util/HttpClient.kt
@@ -128,7 +128,7 @@ class HttpClient constructor(private val client: OkHttpClient, baseUrl: String?,
 
                 when {
                     !response.isSuccessful -> {
-                        body?.let { it.close() }
+                        body?.close()
                         cont.resumeWithException(
                             HttpException(call.request(), url, response.message(), response.code()))
                     }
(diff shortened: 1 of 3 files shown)


maniac103:
Need to additionally update OkSse here: https://github.com/heremaps/oksse/issues/17

mobile/src/main/java/org/openhab/habdroid/util/HttpClient.kt
```diff
@@ -128,6 +128,7 @@ class HttpClient constructor(private val client: OkHttpClient, baseUrl: String?,
 
                 when {
                     !response.isSuccessful -> {
+                        body?.let { it.close() }
```
mueller-ma:
```suggestion
                        body?.close()
```

maniac103: Fixed, not sure what I was thinking.

Issue #1558, opened 2019-09-27
User name and password settings are reset when opening


mueller-ma:
### Actual behaviour
When opening the user name or password prefs, an empty EditPrefs is shown.

### Expected behaviour
When opening the user name or password prefs, the EditPrefs should be filled with the current values.

### Steps to reproduce
1. Set user name
2. Close app
3. Open settings
4. The correct value is shown as summary
5. When clicking on the pref, an empty EditPrefs is shown.

#### Client
* App version :  2.8.26-beta

PR #1566, opened 2019-10-01
Fix CustomInputType issue #1558

VMadalin:
### Description
https://github.com/openhab/openhab-android/issues/1558

### Overview
By default the `getDefaultSharedPreferences` is used for set-up correctly the stored values like on `sendDeviceInfoPrefix` preference. But the problem starts when you want to work with different `SharedPreferences` in this case `getSecretPrefs`, and it's require to set it manually 

untitled

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
@@ -22,6 +22,7 @@ import androidx.core.os.bundleOf
 import androidx.fragment.app.DialogFragment
 import androidx.preference.EditTextPreference
 import androidx.preference.EditTextPreferenceDialogFragmentCompat
+import org.openhab.habdroid.R
 
 class CustomInputTypePreference constructor(context: Context, attrs: AttributeSet) :
     EditTextPreference(context, attrs) {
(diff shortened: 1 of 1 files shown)


maniac103:
Thanks for this patch. I want to dig into this a bit more though, as it feels like a big shortcoming in the preference framework if this really is required, so @mueller-ma please hold off merging this for now.

maniac103:
I'd expect the text to be loaded via [1], whose implementation is [2], which in turn is populated via [3], which is populated from [4]. The latter loads from data store (instead of shared prefs) if present, so IMHO we must be missing something here.

[1] https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-master-dev/preference/preference/src/main/java/androidx/preference/EditTextPreferenceDialogFragmentCompat.java#49
[2] https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-master-dev/preference/preference/src/main/java/androidx/preference/EditTextPreference.java#95
[3] https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-master-dev/preference/preference/src/main/java/androidx/preference/EditTextPreference.java#106
[4] https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-master-dev/preference/preference/src/main/java/androidx/preference/Preference.java#1676

maniac103:
https://github.com/openhab/openhab-android/pull/1568 is my take on fixing this.

VMadalin:
@maniac103 Absolutely agree with you, the approach is quite better and more clean. So strange behaviour and thank you for report issue to google. I close this PR

PR #1568, opened 2019-10-02
Correctly initialize editor prefs with custom data store.

### LINKED ISSUE #1558: User name and password settings are reset when opening

maniac103:
When using editor prefs with a custom data store, the initial data load
happens before assigning the data store and thus too early. Re-do the
initialization when assigning the data store to fetch the content from
the correct place.

Fixes #1558


maniac103:
Created a bug report for that: https://issuetracker.google.com/issues/141918360

PR #1569, opened 2019-10-02
Fix lock prompt on API 28+

### LINKED ISSUE #1525: Lock screen not working on Android Q
The current logic behind the lock screen is:
Open the Android lock screen in `onStart()` if needed. The lock screen is full screen, so nothing of the openHAB activity is shown. If canceled, Android call openHAB activity's `onStart()` again and the lock screen is shown again.

However this is an issue with Android Q's biometric prompt, that replaces the lock screen. It isn't full screen, so parts of the openHAB activity are shown, and it's dismissable without calling `onStart()`. Thus the app can be accessed without authenticating.

Example image:

I have a WIP branch using `androidx.biometric:biometric:1.0.0-beta01` to replace `KeyguardManager#createConfirmDeviceCredentialIntent()`. It's possible to allow device credentials instead of biometrics, but if no biometrics are on a device, the prompt fails instantly. I don't know if that's intented or not, but it seems that we need to keep `KeyguardManager#createConfirmDeviceCredentialIntent()` as fallback.

maniac103:
Make the logic deal with the lock prompt no longer being fullscreen
anymore.

Fixes #1525


maniac103:
As mentioned in https://github.com/openhab/openhab-android/issues/1525#issuecomment-537398627

mueller-ma:
```
org.openhab.habdroid.beta E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.openhab.habdroid.beta, PID: 7478
        at androidx.fragment.app.FragmentManagerImpl.checkStateLoss(FragmentManagerImpl.java:1536)
        at androidx.fragment.app.FragmentManagerImpl.enqueueAction(FragmentManagerImpl.java:1558)
        at androidx.fragment.app.BackStackRecord.commitInternal(BackStackRecord.java:317)
        at androidx.fragment.app.BackStackRecord.commit(BackStackRecord.java:282)
        at androidx.biometric.BiometricPrompt.authenticateInternal(BiometricPrompt.java:728)
        at androidx.biometric.BiometricPrompt.authenticate(BiometricPrompt.java:662)
        at org.openhab.habdroid.ui.AbstractBaseActivity$AuthPrompt.authenticate(AbstractBaseActivity.kt:171)
        at org.openhab.habdroid.ui.AbstractBaseActivity.promptForDevicePassword(AbstractBaseActivity.kt:114)
        at org.openhab.habdroid.ui.AbstractBaseActivity.promptForDevicePasswordIfRequired(AbstractBaseActivity.kt:131)
        at org.openhab.habdroid.ui.AbstractBaseActivity.onStart(AbstractBaseActivity.kt:80)
        at org.openhab.habdroid.ui.MainActivity.onStart(MainActivity.kt:237)
```

mueller-ma:
With fingerprint on Android 10 I have to re-authenticate whenever I open a new activity.

maniac103:
Both fixed (the latter definitely, the former hopefully).

mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
```diff
@@ -173,24 +154,32 @@ abstract class AbstractBaseActivity : AppCompatActivity(), CoroutineScope {
...
             authPrompt = null
         }
 
         override fun onAuthenticationFailed() {
             super.onAuthenticationFailed()
-            finish()
+            finishActivity()
+        }
+
+        private fun finishActivity() {
```
mueller-ma: You can make this `protected` and remove it [here](https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/background/NfcReceiveActivity.kt#L49).

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
```diff
@@ -226,3 +228,12 @@ fun Context.showToast(@StringRes message: Int) {
 fun Context.hasPermission(permission: String): Boolean {
     return ContextCompat.checkSelfPermission(this, permission) != PackageManager.PERMISSION_GRANTED
 }
+
+fun Activity.finishAndRemoveTaskIfPossible() {
```
mueller-ma: Why as extension function? A `protected` fun in `AbstractBaseActivity` should be fine.

mueller-ma:
What about `WriteTagActivity`?

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
maniac103: Except `NfcReceiveActivity` doesn't derive from that and also doesn't need to when doing this as extension function ;-)

maniac103:
> What about WriteTagActivity?

You mean whether we should lock it? Doesn't it already derive from `AbstractBaseActivity`?

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
mueller-ma: Ah, I didn't see that.

mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
```diff
@@ -134,25 +140,40 @@ abstract class AbstractBaseActivity : AppCompatActivity(), CoroutineScope {
...
-                }
-            }
+    private inner class AuthPrompt : BiometricPrompt.AuthenticationCallback() {
+        private val contentView = findViewById<View>(R.id.activity_content)
+        private val prompt = BiometricPrompt(this@AbstractBaseActivity, Dispatchers.Main.asExecutor(), this)
+
+        fun authenticate() {
+            val info = BiometricPrompt.PromptInfo.Builder()
+                .setTitle(getString(R.string.app_name))
+                .setDescription(getString(R.string.screen_lock_unlock_screen_description))
```
mueller-ma: The description could be changed to "Unlock to access settings" if kiosk mode is enabled.

mueller-ma:
```diff
diff --git a/mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt b/mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
index 6b5d6843..457f8ed5 100644
--- a/mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
@@ -145,9 +145,15 @@ abstract class AbstractBaseActivity : AppCompatActivity(), CoroutineScope {
         private val prompt = BiometricPrompt(this@AbstractBaseActivity, Dispatchers.Main.asExecutor(), this)
 
         fun authenticate() {
+            val description = if (getPrefs().getScreenLockMode(this@AbstractBaseActivity) ==
+                ScreenLockMode.KioskMode) {
+                getString(R.string.screen_lock_unlock_preferences_description)
+            } else {
+                getString(R.string.screen_lock_unlock_screen_description)
+            }
             val info = BiometricPrompt.PromptInfo.Builder()
                 .setTitle(getString(R.string.app_name))
-                .setDescription(getString(R.string.screen_lock_unlock_screen_description))
+                .setDescription(description)
                 .setDeviceCredentialAllowed(true)
                 .build()
             contentView.isInvisible = true
diff --git a/mobile/src/main/res/values/strings.xml b/mobile/src/main/res/values/strings.xml
index 92e04d98..4fe95f50 100644
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -78,6 +78,7 @@
     <string name="settings_screen_lock">Screen lock</string>
     <string name="settings_screen_lock_summary">Lock openHAB access with Android screen lock or fingerprint. Voice widgets can always be used.</string>
     <string name="screen_lock_unlock_screen_description">Unlock to access openHAB</string>
+    <string name="screen_lock_unlock_preferences_description">Unlock to access settings</string>
     <string name="settings_screen_lock_off">Off</string>
     <string name="settings_screen_lock_kiosk">Kiosk mode</string>
     <string name="settings_screen_lock_on">On</string>

```

mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
```diff
@@ -134,25 +140,40 @@ abstract class AbstractBaseActivity : AppCompatActivity(), CoroutineScope {
...
+                .setDeviceCredentialAllowed(true)
+                .build()
+            contentView.isInvisible = true
+            prompt.authenticate(info)
+        }
+
+        override fun onAuthenticationError(errorCode: Int, errString: CharSequence) {
+            super.onAuthenticationError(errorCode, errString)
+            finishAndRemoveTaskIfPossible()
```
mueller-ma: I don't know _why_ I used `finishAndRemoveTask()` previously and I think just finishing here is better.

mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
```diff
@@ -105,9 +109,8 @@ abstract class AbstractBaseActivity : AppCompatActivity(), CoroutineScope {
         val km = getSystemService(KEYGUARD_SERVICE) as KeyguardManager
         val locked = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) km.isDeviceSecure else km.isKeyguardSecure
         if (locked) {
```
mueller-ma: `else` show a Snackbar with "No screen lock set" and a button that takes the user to the Android settings? This Snackbar should be shown as well if the user sets up the screen lock in the openHAB settings.

maniac103: This seems a bit excessive to me, at least the proposed else branch. How about disabling the screen lock option in the settings activity if Android lock is disabled?

mueller-ma: If the pref is going to be disabled, the summary should be something like "No Android screen lock oder fingerprint set".

maniac103: Yeah, was thinking about something among those lines ... maybe a bit more verbose: `You need to enable a screen lock for your device for using this feature.`

mobile/src/main/res/values/strings.xml
```diff
@@ -76,8 +76,12 @@
     <string name="settings_insecure_connection_summary">Insecurely connected to %s</string>
     <string name="beta">Beta</string>
     <string name="settings_screen_lock">Screen lock</string>
-    <string name="settings_screen_lock_summary">Lock openHAB access with Android screen lock or fingerprint. Widgets can always be used.</string>
+    <string name="settings_screen_lock_nolock_summary">You need to enable a screen lock for your device for enabling app access locking.</string>
```
mueller-ma:
```suggestion
    <string name="settings_screen_lock_nolock_summary">You need to enable a screen lock for your device for enabling app access locking</string>
```

mobile/src/main/res/values/strings.xml
```diff
@@ -76,8 +76,12 @@
...
+    <string name="settings_screen_lock_off_summary">Access to the app is not locked.</string>
```
mueller-ma:
```suggestion
    <string name="settings_screen_lock_off_summary">Access to the app is not locked</string>
```

mobile/src/main/res/values/strings.xml
```diff
@@ -76,8 +76,12 @@
...
+    <string name="settings_screen_lock_kiosk_summary">Access to app settings is locked with Android screen lock or fingerprint.</string>
```
mueller-ma:
```suggestion
    <string name="settings_screen_lock_kiosk_summary">Access to app settings is locked with Android screen lock or fingerprint</string>
```

PR #1582, opened 2019-10-03
Beautify success messages

mueller-ma:
"Light set to On" doesn't sound nice. This PR makes success toasts for a
few common states translatable. The success messages are not used for
NFC tags, because they are too long to be saved to the tag.


maniac103:
I think at some point we wanted to do the TOGGLE action in-app by fetching the Item before applying the new state (which then depends on current state). Maybe we should hold off on this beautification until we have made that refactoring? Advantage of that is that we have Item type etc. handily available when we show the toast, thus we can generate the texts at that time.

mueller-ma:
I'm having a fix locally for #1576. After that fix has been merged, I can rebase this PR. Then I can easily pass the label to the toast, not the item name.

PR #1583, opened 2019-10-04
Clears statusbar notifications on opening a fragment

### LINKED ISSUE #1348: Status bar notification not cleared when visiting notifications page
### Actual behaviour
Status bar notification isn't cleared when visiting the notification page in the app

### Expected behaviour
Status bar notification should be cleared

 
### Steps to reproduce
1. Get a notification from the server
2. Open app from launcher
3. Go to notification page via drawer

No (no notifications in demo mode)

#### Client
* Android version: 9
* Device model: OnePlus 6
* Stock or customized ROM: LineageOS 16.0
* App version: 2.6.19 beta
* Build flavor : full
* Device language: DE

#### Server
* Server version: 2.4
* Reverse Proxy: no
* Authentication method : none

#### openHAB Cloud
* openHAB Cloud version: myopenhab
* Self hosted cloud or myopenhab.org: myopenhab
* Authentication method : user/pw

Kshitij09:
App notifications are cleared on opening the notifications page in the app. Currently clearing all the notifications.
Fixes #1348

Signed-off by: Kshitij Patil <kshitijpatil98@gmail.com> (github:Kshitij09)

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/CloudNotificationListFragment.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/CloudNotificationListFragment.kt
@@ -13,6 +13,8 @@
 
 package org.openhab.habdroid.ui
 
+import android.app.NotificationManager
+import android.content.Context.NOTIFICATION_SERVICE
 import android.os.Bundle
 import android.util.Log
 import android.view.LayoutInflater
@@ -111,6 +113,7 @@ class CloudNotificationListFragment : Fragment(), View.OnClickListener, SwipeRef
     }
 
     private fun loadNotifications(clearExisting: Boolean) {
+        clearStatusbarNotifications()
         val activity = activity as AbstractBaseActivity? ?: return
         val conn = ConnectionFactory.cloudConnection
         if (conn == null) {
@@ -145,6 +148,14 @@ class CloudNotificationListFragment : Fragment(), View.OnClickListener, SwipeRef
                 Log.e(TAG, "Notifications request failure", e)
             }
         }
+    }
+
+    // Statusbar notifications should be cleared on visiting notifications page
+    // in the app. Currently clearing all the notifications, but it could be
+    // modified further to remove specific only.
+    private fun clearStatusbarNotifications() {
+        val notificationManager: NotificationManager = activity?.getSystemService(NOTIFICATION_SERVICE) as NotificationManager
+        notificationManager.cancelAll()
     }
 
     private fun handleInitialHighlight() {


mobile/src/main/java/org/openhab/habdroid/ui/CloudNotificationListFragment.kt
```diff
@@ -147,6 +150,14 @@ class CloudNotificationListFragment : Fragment(), View.OnClickListener, SwipeRef
         }
     }
 
+    // Statusbar notifications should be cleared on visiting notifications page
+    // in the app. Currently clearing all the notifications, but it could be
+    // modified further to remove specific only.
```
maniac103: Doing that is necessary, as we also have other notifications (e.g. for background work). You'd need to add a new method to `CloudMessagingHelper` (e.g. named `clearAllNotifications`) which then probably should call into a static method in `FcmMessageListenerService`, clearing the summary notification + all GCM notifications (see `getGcmNotificationCount` for how to detect this) there.

Kshitij09:
I accidently deleted the local and fork repo of this Pull Request. I'll follow up with new one considering all the suggestions given here.

mobile/src/main/java/org/openhab/habdroid/ui/CloudNotificationListFragment.kt
Kshitij09: So should I make provision to delete individual notification (by notificationId) as well ?

maniac103: What do you mean exactly? We certainly don't need a way to locally remove individual notifications, but pushing out 'hideNotification' messages for the IDs we clear locally during clear-all seems like a good idea to me. In fact, doing so is already supported by passing `FcmMessageListenerService.SUMMARY_NOTIFICATION_ID` to the notification hiding methods in `FcmRegistrationService`.

Kshitij09: I've implemented a method in CloudMessagingHelper which clears all the notifications using NotificationManager. I could've reused the code from FirebaseMessagingService for the same, but my implementation adds up only two lines of code. So I suppose this would be fine.

However, one strange thing I noticed is, I'm getting 0 notifications in response to http request made in NotificationListFragment to get all notifications. Is this because I'm using dummy UUID and secret to send notifications from server ? or is this some kind of bug ?

Kshitij09: How can I get non-null context for calling FcmRegistrationService's static methods ?

mueller-ma: It depends. If you're in an activitiy, you can use `this` as context. Otherwise you can enter context and Android Studio should suggest a context getter function.
If not, you must pass `context` as argument to the function where you need it.

mueller-ma: > However, one strange thing I noticed is, I'm getting 0 notifications in response to http request made in NotificationListFragment to get all notifications. Is this because I'm using dummy UUID and secret to send notifications from server ? or is this some kind of bug ?

The notifications send via the web interface are no "real" notifications, thus you don't see them if you enter the Notifications site in the app.

Kshitij09: If I've to send hideNotificationRequest() using FcmRegistrationService, what would be the senderId ?

maniac103: See https://github.com/openhab/openhab-android/blob/master/mobile/src/full/java/org/openhab/habdroid/core/FcmRegistrationService.kt#L107 and its caller.

PR #1585, opened 2019-10-04
Implement TOGGLE command client-side only.

maniac103:
As discussed in https://github.com/eclipse/smarthome/pull/4638, particularly https://github.com/eclipse/smarthome/pull/4638#issuecomment-347509218

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
@@ -59,9 +59,7 @@ class ItemUpdateWorker(context: Context, params: WorkerParameters) : Worker(cont
         return runBlocking {
             try {
                 val item = loadItem(connection, itemName)
-                if (item == null) {
-                    return@runBlocking Result.failure(buildOutputData(true, 500))
-                }
+                    ?: return@runBlocking Result.failure(buildOutputData(true, 500))
                 if (value == "TOGGLE") {
                     value = determineOppositeState(item)
                 }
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
```diff
@@ -41,27 +52,79 @@ class ItemUpdateWorker(context: Context, params: WorkerParameters) : Worker(cont
...
         }
     }
 
+    private suspend fun loadItem(connection: Connection, itemName: String): Item? {
+        val response = connection.httpClient.get("rest/items/$itemName")
+        val contentType = response.response.contentType()
+        val content = response.asText().response
+
+
+        if (contentType?.type() == "application" && contentType?.subtype() == "json") {
```
mueller-ma:
```suggestion
        if (contentType?.type() == "application" && contentType.subtype() == "json") {
```

mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
```diff
@@ -41,27 +52,79 @@ class ItemUpdateWorker(context: Context, params: WorkerParameters) : Worker(cont
...
+
+    private fun determineOppositeState(item: Item): String {
+        if (item.isOfTypeOrGroupType(Item.Type.Rollershutter) || item.isOfTypeOrGroupType(Item.Type.Dimmer)) {
+            // If shutter is (partially) closed, open it, else close it
+            return if (item.state?.asNumber?.value == 0F) "100" else "0"
+        } else if (item.state?.asBoolean == true) {
+            return "OFF"
+        } else {
+            return "ON"
+        }
```
mueller-ma: You can lift out the return: `return if (item....`

mueller-ma:
See https://github.com/openhab/openhab-android/commit/b619fa9b45a7b63213fc8b5be83a417b465d629e for my suggestions.

Issue #1587, opened 2019-10-04
Add docs for Settings> Item name prefix


openhab-5iver:
The [documentation](https://www.openhab.org/docs/apps/android.html) does not mention what this setting is used for. I assume it is used as a filter in a REST API call, which could provide a performance improvement (we have 1684 Items)? Also, the value is not displayed once it is set, like is done for the CallState (I have it currently set to "Phone_")...

![image](https://user-images.githubusercontent.com/21967853/66219267-da190100-e698-11e9-9760-76a91ead9f21.png)

maniac103:
It's not used as a filter, but as a prefix ;-) Suppose you have the alarm clock item bane set to 'AlarmClock' and the prefix is set to 'Phone_', the alarm is uploaded to item 'Phone_AlarmClock'.

openhab-5iver:
In that case, I change my request to remove it entirely, since the functionality is simplified without it! :smile:

mueller-ma:
> In that case, I change my request to remove it entirely, since the functionality is simplified without it! 😄

No, it's not: If you have only one phone, you can ignore the prefix at all. If you have two or more phones, you just enter the phone name as prefix and don't have to change the item names for alarm clock and call state.

mueller-ma:
> Also, the value is not displayed once it is set, like is done for the CallState (I have it currently set to "Phone_")...

In #1589

openhab-5iver:
> No, it's not: If you have only one phone, you can ignore the prefix at all. If you have two or more phones, you just enter the phone name as prefix and don't have to change the item names for alarm clock and call state.

@mueller-ma, I find it offensive and rude that you would close this issue without discussion and before an appropriate resolution. I took time out of my day to point out multiple gaps in the app that _you_ are maintaining. You've introduced unneeded functionality without documentation that makes the app needlessly complex and confusing, and in a best case will save the user from typing a few characters!

maniac103:
From a complexity point of view, @openhab-5iver certainly has a point. The usefulness of the option probably depends on how many additional item upload features we'll add in the future...however, I can't think of that many additional options at the moment, so while I don't have a strong opinion on the topic, I'd be fine with removing the option.

mueller-ma:
> I took time out of my day to point out multiple gaps in the app that you are maintaining.

Thanks for your feedback. The prefix not being displayed in the summary will be fixed in #1589.

> You've introduced unneeded functionality

[...]

PR #1589, opened 2019-10-04
Show prefix in summary

mueller-ma:
See #1587

### DIFF
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -157,7 +157,8 @@
     <!-- Background tasks -->
     <string name="send_device_info_to_server">Send device information to server</string>
     <string name="send_device_info_item_prefix">Item name prefix</string>
-    <string name="send_device_info_item_prefix_summary">The Item names of the following features have this prefix: %s</string>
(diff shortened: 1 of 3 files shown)


mobile/src/main/res/values/strings.xml
```diff
@@ -157,7 +157,7 @@
     <!-- Background tasks -->
     <string name="send_device_info_to_server">Send device information to server</string>
     <string name="send_device_info_item_prefix">Item name prefix</string>
-    <string name="send_device_info_item_prefix_summary">The Item names of the following features have this prefix</string>
+    <string name="send_device_info_item_prefix_summary">The Item names of the following features have this prefix: %s</string>
```
maniac103: '... have this prefix: %s' -> 'are prefixed with %s'?

mueller-ma: The first option sounds better if no prefix is set.
Should I add a different string if no prefix is set? "The Item names of the following features have no prefix"

maniac103: Yeah, a separate string makes sense to me. How about 'No prefix is set for the Item names of the following features.'?

PR #1594, opened 2019-10-04
Clears statusbar notifications on opening a fragment

### LINKED ISSUE #1348: Status bar notification not cleared when visiting notifications page

Kshitij09:
App notifications are cleared after opening notifications page in the app.
Currently using NotificationManager to perform the task. IMO it is unnecessary to send fcm message for 'hideNotification' unless it isn't able to clear all the notifications (including Summary and GCM notifications).
fixes #1348

Signed-off-by: Kshitij Patil <kshitijpatil98@gmail.com> (github: Kshitij09)


mobile/src/full/java/org/openhab/habdroid/core/CloudMessagingHelper.kt
```diff
@@ -70,4 +72,12 @@ object CloudMessagingHelper {
         }
         else -> context.getString(R.string.info_openhab_gcm_connected)
     }
+
+    // Used to clear all notifications from statusbar
+    fun clearAllNotifications(context: Context?) {
+        val nm = context?.getSystemService(NOTIFICATION_SERVICE) as NotificationManager
+        nm.cancelAll()
```
maniac103: As mentioned before, this doesn't cut it. We have more notifications than just GCM ones, which must not be hidden by this.
Also need to update the foss variant of CloudMessagingHelper.

maniac103:
> IMO it is unnecessary to send fcm message for 'hideNotification'

It is necessary ... if you have two devices using the same account, hiding notifications on one device shall also hide them from the other device...that's actually the sole reason for the existence of the hideNotification GCM message.

Kshitij09:
Okay! I got that now. But I tried calling scheduleHideNotification with notificationId=0, it didn't work

mobile/src/full/java/org/openhab/habdroid/core/CloudMessagingHelper.kt
Kshitij09: fossStableDebug right ?

maniac103:
You need to do both ... clear the local notifications via NotificationManager and the remote ones via GCM message.

mobile/src/full/java/org/openhab/habdroid/core/CloudMessagingHelper.kt
maniac103: Yeah, any foss build will work.

mueller-ma: I can do the foss part if you like.

Kshitij09:
Will the sending FCM message as @maniac103  mentioned clear any notifications that I could check? Currently NotificationManager is clearing all the notifications, but there's nothing noticeable if I send FCM message.

maniac103:
It really isn't that complicated: NotificationManager clears the notifications on *your* device, and the FCM message will be received by the app on *other* devices, where the app will use NotificationManager to clear notifications on *those* devices.

Kshitij09:
Indeed. But I'm asking whether the FCM message sent via one device will be received by the same as well? or it will affect the others only?

maniac103:
The openhab-cloud instance forwards the GCM message only to the other devices, the sending device doesn't receive it by itself.

Kshitij09:
My method needs a class Specific to full variant. How should I incorporate the same in foss variant (as the required class is absent) ?

maniac103:
What does that class do? AFAICT you'd only need a new method in CloudNotificationHelper, which receives an implementation for full variant and is kept as stub in the foss variant.

mobile/src/full/java/org/openhab/habdroid/core/CloudMessagingHelper.kt
```diff
@@ -70,4 +72,15 @@ object CloudMessagingHelper {
         }
         else -> context.getString(R.string.info_openhab_gcm_connected)
     }
+
+    // Used to clear all notifications from statusbar
+    fun clearAllNotifications(context: Context) {
+        // to clear the notifications on current device
+        val notificationManager: NotificationManager =
+            context.getSystemService(NOTIFICATION_SERVICE) as NotificationManager
+        notificationManager.cancelAll()
```
maniac103: As mentioned before: This is wrong. We have non-FCM notifications as well, and we must not clear those.

mobile/src/full/java/org/openhab/habdroid/core/FcmRegistrationService.kt
```diff
@@ -105,6 +105,7 @@ class FcmRegistrationService : JobIntentService() {
     }
 
     private fun sendHideNotificationRequest(notificationId: Int, senderId: String) {
+
```
maniac103: Please avoid random whitespace changes.

mobile/src/main/java/org/openhab/habdroid/ui/CloudNotificationListFragment.kt
```diff
@@ -84,6 +85,9 @@ class CloudNotificationListFragment : Fragment(), View.OnClickListener, SwipeRef
         recyclerView.layoutManager = layoutManager
         recyclerView.addItemDecoration(DividerItemDecoration(view.context))
         recyclerView.adapter = adapter
+
+        // Clear notifications from statusbar
+        CloudMessagingHelper.clearAllNotifications(context!!)
```
maniac103: Can use `view.context` here to avoid using the `!!` operator.

mobile/src/full/java/org/openhab/habdroid/core/CloudMessagingHelper.kt
Kshitij09: So should I check whether it's a FCM notification and then clear?

Kshitij09: I came to we can use `n.groupKey?.endsWith("gcm")` to identify fcm notifications after API 23. How can I do the same for prior APIs ?

Kshitij09:
I understood why @maniac103  said creating static method in `FcmMessageListenerService` would be a good idea. But I'm not sure what sending `SUMMARY_NOTIFCATION_ID` to `nm.hide()` method does. So I've pushed the fcm message for both Summary Notifications and all the fcm notifications cleared on current device.

mobile/src/full/java/org/openhab/habdroid/core/CloudMessagingHelper.kt
maniac103: AFAICT there is no way to implement this on API < 23.

mobile/src/full/java/org/openhab/habdroid/core/FcmMessageListenerService.kt
```diff
@@ -202,10 +203,24 @@ class FcmMessageListenerService : FirebaseMessagingService() {
...
+        fun clearFcmNotifications(context: Context) {
+            val nm = context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+            nm.cancel(SUMMARY_NOTIFICATION_ID)
+            if (HAS_GROUPING_SUPPORT) {
+                val activeNotifications = nm.activeNotifications
+                activeNotifications
+                    .filter { it.groupKey.endsWith("gcm") }
+                    .forEach {
+                        nm.cancel(it.id)
+                        FcmRegistrationService.scheduleHideNotification(context, it.id)
```
maniac103: This is unneeded, as the subnotifications are already cleared. See the handling of the incoming FCM message here: https://github.com/openhab/openhab-android/pull/1594/files#diff-120bc65ca3879a5d1faad0e4bb6e4f65R93

.idea/codeStyles/Project.xml
```diff
@@ -1,8 +1,5 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
-    <AndroidXmlCodeStyleSettings>
-      <option name="ARRANGEMENT_SETTINGS_MIGRATED_TO_191" value="true" />
-    </AndroidXmlCodeStyleSettings>
```
maniac103: Please remove the project file changes from this PR. We need to figure out how to avoid the unwanted changes to checked in files in a separate PR.

mobile/src/full/java/org/openhab/habdroid/core/FcmMessageListenerService.kt
Kshitij09: Exactly which part of it is unneeded ? `scheduleHideNotification()` for each notification or the entire forEach block ? I'm calling nm.cancel(it.id) to clear locally and scheduling the same id to clear remotely. Which part is redundant ?

maniac103: `scheduleHideNotification` for each notification is redundant, since the FCM hideNotification message with notificationId == SUMMARY_NOTIFICATION_ID will take care of it.

Kshitij09: I kind of had doubt that just clearing the summary notifications would clear out whole group, so no need to iterate manually. Is that what you talking about ?

.idea/codeStyles/Project.xml
Kshitij09: Could you help me how can I do this ?

maniac103: See suggestions here: https://stackoverflow.com/questions/12481639/remove-files-from-git-commit

Kshitij09: Just want to confirm with you before making any changes -
commit hash when first time this file was added: 254916d6e23c1247bae77620995d7261a3fba1ea

1. `git reset HEAD^ 254916d6e23c1247bae77620995d7261a3fba1ea -- .idea/codeStyles/Project.xml`
1. `git commit --amend --no-edit`

or

1.  `git checkout 254916d6e23c1247bae77620995d7261a3fba1ea -- .idea/codeStyles/Project.xml`
1. `git commit ...`

which one should I do ?

mueller-ma: Can you update Android Studio to the latest version? This should add the code back.

Kshitij09: So I've to keep the changes made by IDE itself? that would be sufficient?

Kshitij09: My Android Studio is up-to-date.

Android Studio 3.5
Build #AI-191.8026.42.35.5791312, built on August 9, 2019
JRE: 1.8.0_202-release-1483-b03 amd64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Windows 10 10.0

mueller-ma: > So I've to keep the changes made by IDE itself? that would be sufficient?

My Android Studio added this after updating to the latest version. It must not be changed in this PR.

Kshitij09: But What am I supposed to do here? If you need these lines of code, I could simply copy paste and commit the same.

maniac103: Just do not change them in this PR. Whether you want to remove them from your existing commits and force push the PR branch, or whether you add a commit that restores them, is up to you, it's fine either way.

Kshitij09: Actually it was changed in one of the first commits of this PR and I've squashed the changes in single commit before opening this PR. So its a bit difficult to identify that particular commit so that I could revert this file to that version.

mobile/src/main/java/org/openhab/habdroid/ui/CloudNotificationListFragment.kt
```diff
@@ -84,6 +85,9 @@ class CloudNotificationListFragment : Fragment(), View.OnClickListener, SwipeRef
         recyclerView.layoutManager = layoutManager
         recyclerView.addItemDecoration(DividerItemDecoration(view.context))
         recyclerView.adapter = adapter
+
+        // Clear notifications from statusbar
+        CloudMessagingHelper.clearAllNotifications(view.context)
```
mueller-ma: I'd move this call to `onResume()` as the notification list is loaded there as well. You can also drop the comment, the function name is self-explanatory.

Kshitij09: Then it would send the FCM `hideNotification()` mesage every time onResume() is called. Is there any way to check if there are any active notifications, prior to Android N ?

Kshitij09: @mueller-ma my method needs non-null context, is it safe to pass `context!!` if I move the method call to `onResume()` ?

maniac103: Yes, it is (even though I don't like using the `!!` operator, there's no other choice there).

Kshitij09:
Is there anything I could do to get this merged ?

mueller-ma:
This works for me on one device (clearing the local notifications), but I don't have a second device with api >= 24 and play services.

Kshitij09:
Should I perform any testing on my devices and report you back? I suppose there's no error in this PR now. Consider removing _bug_ tag.

mueller-ma:
> Should I perform any testing on my devices and report you back?

Not needed. Currently I don't have much time, but I'll review this soon.

> Consider removing bug tag.

It's tagged with "bug", because it's a PR that fixes a bug.

Kshitij09:
Okay. Thanks :-)

mueller-ma:
Opening notifications does clear them on the local device, but not on others.

Kshitij09:
> Opening notifications does clear them on the local device, but not on others.

Is there any way I can reproduce the results ? So that I'd be able to tackle the problem.

maniac103:
> Is there any way I can reproduce the results ? So that I'd be able to tackle the problem.

- Have two devices
- Install app on both, connect to myopenhab on both
- Enter notifications page on one of them
- Check whether notification is gone on the other one

That's what I'd guess at least.

@mueller-ma Does your second test device have API level >= 24?

mueller-ma:
Yes, both devices have a higher level than 24.

PR #1598, opened 2019-10-08
Use MaterialButtonToggleGroup for segmented switch.

### LINKED ISSUE #1597: Convert segmented button to MaterialButtonToggleGroup
Now that we use Material Components, we can drop our custom segmented button implementation and instead use [MaterialButtonToggleGroup](https://material.io/develop/android/components/material-button-toggle-group/)

maniac103:
Closes #1597


maniac103:
This is how it looks:
Screenshot_1570518725

We could also use borderless buttons:
Screenshot_1570517853

... but with the outlined one the segmentation probably is a bit more obvious?

mueller-ma:
> the outlined one the segmentation probably is a bit more obvious

:+1:

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -609,41 +610,56 @@ class WidgetAdapter(
 
             val mappings = widget.mappingsOrItemOptions
             // inflate missing views
-            for (i in radioGroup.childCount until mappings.size) {
-                val view = inflater.inflate(R.layout.widgetlist_sectionswitchitem_button, radioGroup, false)
+            while (!spareViews.isEmpty() && group.childCount < mappings.size) {
```
mueller-ma:
```suggestion
            while (spareViews.isNotEmpty() && group.childCount < mappings.size) {
```

PR #1600, opened 2019-10-08
Switch to OkHttp-internal SSE implementation.

maniac103:
This one doesn't require us to use random git revisions.

Tested only lightly so far.


mueller-ma:
I haven't found any issue when testing this. Do you think it's RTM?

(retriggering Github Actions)

PR #1632, opened 2019-10-25
Grey dark theme: Increase contrast between buttons and background

mueller-ma:


maniac103:
Got a before/after screenshot?

mueller-ma:
# Dark
## Before
before dark
## After
after dark

# Light
## Before
before light
## After
after light

maniac103:
Dark version makes sense to me, but I'm not sure about the light one. What about using a color resource initialized to blue_grey_700 and overridden to white for -night?

Issue #1626, opened 2019-10-22
Use discrete sliders if only a few states are possible


mueller-ma:
### Is your feature request related to a problem? Please describe.
By setting minValue, maxValue and/or step it's possible to create sliders, that can only have a few states. It's not visible to the user where the slider will accept the change of the state.
The slider might reject state changes if the new value isn't close to a valid state. The slider will than jump back unexpectedly.

### Describe the solution you'd like
Slider with less than, let's say 10, possible states, should be discrete ones: https://material.io/components/sliders/#discrete-slider

mueller-ma:
This might be useful: https://stackoverflow.com/questions/35370671/how-to-show-tick-marks-for-discrete-slider

maniac103:
We also already have a discrete slider in preferences.

PR #1637, opened 2019-10-26
Show discrete sliders when only few states are possible

vjgarciag96:
As requested per #1626, in this PR you'll see the changes needed to have a discrete slider when there are less than 10 possible states for it. Otherwise, we will have the continuous slider as usual.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -445,6 +445,13 @@ class WidgetAdapter(
             boundWidget = widget
 
             val stepCount = (widget.maxValue - widget.minValue) / widget.step
+
+            if (stepCount <= DISCRETE_SLIDER_STATES_THRESHOLD) {
+                seekBar.context.setTheme(R.style.Widget_AppCompat_SeekBar_Discrete)
+            } else {
+                seekBar.context.setTheme(R.style.Widget_AppCompat_SeekBar)
+            }
+
             seekBar.max = Math.ceil(stepCount.toDouble()).toInt()
             seekBar.progress = 0
 
@@ -488,6 +495,10 @@ class WidgetAdapter(
             val item = widget?.item ?: return
             val newValue = widget.minValue + widget.step * progress
             connection.httpClient.sendItemUpdate(item, item.state?.asNumber.withValue(newValue))
+        }
+
+        private companion object {
+            const val DISCRETE_SLIDER_STATES_THRESHOLD = 10
         }
     }
 


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -445,6 +445,13 @@ class WidgetAdapter(
             boundWidget = widget
 
             val stepCount = (widget.maxValue - widget.minValue) / widget.step
+
+            if (stepCount <= DISCRETE_SLIDER_STATES_THRESHOLD) {
+                seekBar.context.setTheme(R.style.Widget_AppCompat_SeekBar_Discrete)
+            } else {
+                seekBar.context.setTheme(R.style.Widget_AppCompat_SeekBar)
```
maniac103: Changing the theme for the whole activity (that's what the view context is) isn't a viable solution, especially given a widget style (a collection of widget specific attributes) isn't the same thing as a theme (a collection of activity specific attributes).
Are you sure this even works? I fail to see how it could...

vjgarciag96: Yes, you're right, changing activity theme is a bad option. It's possible to apply a style as a theme and it works but I agree with you that we mustn't do that. 

The ideal approach IMO would be constructing the seek bar programmatically passing the correspondent style in the constructor depending on the number of states. However, I didn't implemented that option because inside `SliderViewHolder` the seek bar is used before knowing the number of states it's going to have (there is code to set listeners on the `init` method of the view holder and it's also used on `handleRowClick`).

I could try to implement the programatical approach, but I would like to know your opinion on how should this be implemented before going for it.

maniac103: If reconfiguring the SeekBar between discrete and continuous at runtime doesn't work (maybe checking the Discrete style would give a clue what needs to be done there?), I'd likely go with two different view types and two layouts and distinguishing between them via SliderViewHolder constructor parameter.

mueller-ma: > If reconfiguring the SeekBar between discrete and continuous at runtime doesn't work

I tried to add the tick marks manually, but this doesn't work for Android < N: https://github.com/mueller-ma/openhab.android/commit/7606d828e1890e271e6170ad359034b6b8265f75

maniac103: What's the issue on < N?

mueller-ma:
````
rg.openhab.habdroid.beta E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.openhab.habdroid.beta, PID: 8949
    java.lang.NoSuchMethodError: No virtual method setTickMark(Landroid/graphics/drawable/Drawable;)V in class Landroidx/appcompat/widget/AppCompatSeekBar; or its super classes (declaration of 'androidx.appcompat.widget.AppCompatSeekBar' appears in /data/app/org.openhab.habdroid.beta-2/base.apk)
        at org.openhab.habdroid.ui.WidgetAdapter$SliderViewHolder.bind(WidgetAdapter.kt:469)
        at org.openhab.habdroid.ui.WidgetAdapter.onBindViewHolder(WidgetAdapter.kt:187)
        at org.openhab.habdroid.ui.WidgetAdapter.onBindViewHolder(WidgetAdapter.kt:88)
        at androidx.recyclerview.widget.RecyclerView$Adapter.onBindViewHolder(RecyclerView.java:7059)
        at androidx.recyclerview.widget.RecyclerView$Adapter.bindViewHolder(RecyclerView.java:7101)
````

mueller-ma:
The slider from material components should be used instead: https://github.com/openhab/openhab-android/pull/1677#issuecomment-555463226

PR #1643, opened 2019-10-28
Work around appcompat forcing font scale to 1.0

### LINKED ISSUE #1613: Font size too small with high-res tablet since version 2.9
After updating to version 2.9, I discovered that the font size is very small on my old but very high-res tablet (Google Nexus 10).
On my very recent smartphone, the font size is normal.
Is it possible to restore a bigger font for high-res tablet ? Or to add a setting to select the font size ?

maniac103:
Fixes #1613

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -232,6 +232,14 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         drawerToggle.onConfigurationChanged(newConfig)
     }
 
+    override fun applyOverrideConfiguration(overrideConfiguration: Configuration?) {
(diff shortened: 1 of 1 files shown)


maniac103:
This commit can be reverted after updating appcompat to a version which includes https://android.googlesource.com/platform/frameworks/support/+/1795e4ebdd262598f3c90b79cd3d75d79a5a2264

lolodomo:
I am available in case you need my help to test a pre-version.
By the way, thanks for analyzing an fixing my problem, I really appreciate.

mueller-ma:
Can you please test the following apk?

- Rename it to .apk
- Install my apk

[mobile-foss-beta-debug.zip](https://github.com/openhab/openhab-android/files/3780262/mobile-foss-beta-debug.zip)

lolodomo:
I can confirm that the bug is fixed. Bravo.

lolodomo:
Please note that the settings and about pages are still in a smaller (unscalled ?) font size.

maniac103:
Damn, should have put that in the activity base class. Will submit a new PR.

PR #1648, opened 2019-10-29
Disabled http protocol for remote URL

jalalkhademi:


jalalkhademi:
I added that attribute to xml file and parse it in UrlInputPreference class.

maniac103:
Please use an attribute resource for that. `ItemUpdatingPreference` can serve as an example here.
Additionally, the existing error message should be either kept unchanged or the protocol be inserted via positional argument, as http is still valid in local case.

jalalkhademi:
@maniac103 Many thanks for your fast and helpful review. I fixed the issues.

mobile/src/main/java/org/openhab/habdroid/ui/preference/UrlInputPreference.kt
```diff
@@ -116,8 +130,8 @@
         companion object {
             private const val KEY_TITLE = "title"
 
-            fun newInstance(key: String, title: CharSequence): PrefFragment {
-                val f = PrefFragment()
+            fun newInstance(key: String, title: CharSequence, isHttpEnabled: Boolean): PrefFragment {
+                val f = PrefFragment(isHttpEnabled)
```
maniac103: Fragments must not have constructor parameters. Just put the boolean into the arguments bundle as well.

mobile/src/main/java/org/openhab/habdroid/ui/preference/UrlInputPreference.kt
```diff
@@ -89,15 +100,18 @@
                         val url = URL(value)
                         urlIsValid = true
                         when (url.protocol) {
-                            "http" -> portSeemsInvalid = url.port == 443 || url.port == 8443
                             "https" -> portSeemsInvalid = url.port == 80 || url.port == 8080
+                            "http" -> when
+                                    {
```
maniac103: Curly brace should be on old line. Also maybe could use an if here?

mobile/src/main/java/org/openhab/habdroid/ui/preference/UrlInputPreference.kt
```diff
@@ -78,6 +88,7 @@
 
         override fun afterTextChanged(editable: Editable) {
             var portSeemsInvalid = false
+            val res = editor.resources
```
maniac103: This variable movement seems unneeded?

mobile/src/main/java/org/openhab/habdroid/ui/preference/UrlInputPreference.kt
```diff
@@ -89,15 +100,18 @@
...
+                                    }
                         }
                     } catch (e: MalformedURLException) {
                         urlIsValid = false
                     }
                 }
             }
-            val res = editor.resources
             wrapper.error = when {
                 !urlIsValid -> res.getString(R.string.error_invalid_url)
```
maniac103: Need to distinguish strings between 'http allowed' and 'http not allowed' here.

mueller-ma:
Currently two required checks are failing. You need to configure your git cli name and mail to match those in GitHub.
`git config --local user.name jalalkhademi && git config --local user.email jalal.khademi66@gmail.com`
After that you must change the author of the commits. My preferred way is:
`git reset --soft HEAD~4` (`4` because of 4 commits in this PR). `git add . && git commit -s -m "Disabled http protocol for remote URL"`. `git push --force-with-lease`.

jalalkhademi:
@maniac103 fixed all mentioned points. Just I have no idea what is the process for updating translated strings.

mobile/src/main/java/org/openhab/habdroid/ui/preference/UrlInputPreference.kt
```diff
@@ -33,8 +33,17 @@ import java.net.URL
 class UrlInputPreference constructor(context: Context, attrs: AttributeSet) :
     CustomInputTypePreference(context, attrs) {
 
+    var isHttpEnabled = false
```
maniac103: Oops, sorry, I missed this earlier: this should be a val, I think.

mobile/src/main/java/org/openhab/habdroid/ui/preference/UrlInputPreference.kt
```diff
@@ -45,6 +54,7 @@ class UrlInputPreference constructor(context: Context, attrs: AttributeSet) :
         private lateinit var wrapper: TextInputLayout
         private lateinit var editor: EditText
         private var urlIsValid: Boolean = false
+        private var isHttpEnabled = false
```
maniac103: No need for a member for this, just use a local variable.

jalalkhademi:
Those two points are also resolved now

mobile/src/main/java/org/openhab/habdroid/ui/preference/UrlInputPreference.kt
```diff
@@ -33,8 +33,11 @@ import java.net.URL
 class UrlInputPreference constructor(context: Context, attrs: AttributeSet) :
     CustomInputTypePreference(context, attrs) {
 
+    private val isHttpEnabled = context.obtainStyledAttributes(attrs, R.styleable.UrlInputPreference)
+        .getBoolean(R.styleable.UrlInputPreference_isHttpEnabled, false)
```
maniac103: This must stay in the init block for the recycle() call.

jalalkhademi:
Done

jalalkhademi:
@mueller-ma Done.

mueller-ma:
Thanks. Will be merged tomorrow (https://github.com/openhab/openhab-android/issues/549#issuecomment-547916527).

PR #1673, opened 2019-11-14
Use Action for WIP

mueller-ma:


mueller-ma:
Needs much longer than the current WIP check.

PR #1677, opened 2019-11-19
Show discrete sliders when only few states are possible

### LINKED ISSUE #1626: Use discrete sliders if only a few states are possible

mueller-ma:
Fixes #1626
Closes #1637

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -20,6 +20,7 @@ import android.content.res.ColorStateList
 import android.graphics.BitmapFactory
 import android.graphics.Color
 import android.net.Uri
+import android.os.Build
 import android.os.Handler
 import android.os.Message
 import android.util.Base64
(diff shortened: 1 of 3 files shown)


maniac103:
As mentioned [here](https://github.com/openhab/openhab-android/pull/1623#issuecomment-547110358) I think it would be better to defer this until Material Components 1.2.0 is released, and at that point implement it by switching to their `Slider` class.

mueller-ma:
The new Material Slider uses a new `OnChangeListener` (https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/slider/Slider.java#L205) that has only one callback. The old Slider has a callback that is called, after the slider has been changed. The new Slider callback is called, while dragging the slider. This causes the slider to jump around.

maniac103:
That doesn't seem to be much of a problem though. Just schedule a timeout on change and only run it when things have settled (IOW, reschedule it on every change).

mueller-ma:
> As mentioned here I think it would be better to defer this until Material Components 1.2.0 is released, and at that point implement it by switching to their Slider class.

What about this as an interim solution?

maniac103:
*shrug* Personally I don't see the need for an interim solution, since this is just a cosmetic change.

PR #1705, opened 2019-11-27
Item Picker: Wait for connection

### LINKED ISSUE #1696: AbstractItemPickerActivity should call ConnectionFactory.waitForInitialization()
https://community.openhab.org/t/solved-openhab-app-widget-loading-items-failed/86136/3

The widget item picker and tasker item picker can be opened when the app isn't started.

Also: `forceNonFullscreen` shouldn't be true in the nfc item picker as it's called from within the app.

mueller-ma:
Fixes #1696

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
@@ -222,16 +222,18 @@ abstract class AbstractItemPickerActivity : AbstractBaseActivity(), SwipeRefresh
             return
         }
 
-        val connection = ConnectionFactory.usableConnectionOrNull
-        if (connection == null) {
-            updateViewVisibility(loading = false, loadError = true, showHint = false)
-            return
-        }
-
-        itemPickerAdapter.clear()
         updateViewVisibility(loading = true, loadError = false, showHint = false)
+        itemPickerAdapter.clear()
 
         requestJob = launch {
+            ConnectionFactory.waitForInitialization()
+
+            val connection = ConnectionFactory.usableConnectionOrNull
+            if (connection == null) {
+                updateViewVisibility(loading = false, loadError = true, showHint = false)
+                return@launch
+            }
+
             try {
                 val result = connection.httpClient.get("rest/items").asText()
                 var items = JSONArray(result.response)


mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
```diff
@@ -222,6 +222,11 @@ abstract class AbstractItemPickerActivity : AbstractBaseActivity(), SwipeRefresh
             return
         }
 
+        updateViewVisibility(loading = true, loadError = false, showHint = false)
+        launch {
+            ConnectionFactory.waitForInitialization()
```
maniac103: This does not work. The `launch` block spawns a new thread, which then waits for init, but the code below does not block for that wait to finish. You need to run the code below in main thread context inside this launch block.

maniac103: Something like this will probably do (untested):

```diff
diff --git a/mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt b/mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
index 20d795be..a2268d9c 100644
--- a/mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
@@ -222,16 +222,18 @@ abstract class AbstractItemPickerActivity : AbstractBaseActivity(), SwipeRefresh
             return
         }
 
-        val connection = ConnectionFactory.usableConnectionOrNull
-        if (connection == null) {
-            updateViewVisibility(loading = false, loadError = true, showHint = false)
-            return
-        }
-
-        itemPickerAdapter.clear()
         updateViewVisibility(loading = true, loadError = false, showHint = false)
+        itemPickerAdapter.clear()
 
         requestJob = launch {
+            ConnectionFactory.waitForInitialization()
+
+            val connection = ConnectionFactory.usableConnectionOrNull
+            if (connection == null) {
+                updateViewVisibility(loading = false, loadError = true, showHint = false)
+                return@launch
+            }
+
             try {
                 val result = connection.httpClient.get("rest/items").asText()
                 var items = JSONArray(result.response)
```

mueller-ma: Thanks!

maniac103:
(I assume you - unlike me - actually tested this, right? ;-) )

mueller-ma:
I did, but I'm not sure if it's just me or the loading times of the activity did increase.

mueller-ma:
> I did, but I'm not sure if it's just me or the loading times of the activity did increase.

It's just me :)

PR #1707, opened 2019-11-30
Support DateTime items as Alarm clock receiver

### LINKED ISSUE #1618: Support DateTime items as Alarm clock receiver
See https://community.openhab.org/t/android-app-sends-alarm-time-as-long-not-datetime/83758/

mueller-ma:
Fixes #1618

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
@@ -305,11 +305,7 @@ class BackgroundTasksManager : BroadcastReceiver() {
                         putBoolean(Constants.PREFERENCE_ALARM_CLOCK_LAST_VALUE_WAS_ZERO, time == "0" || time == null)
                     }
 
-                    if (time == null) {
-                        null
-                    } else {
-                        ItemUpdateWorker.ValueWithInfo(time, type = ItemUpdateWorker.ValueType.Timestamp)
-                    }
+                    time?.let { ItemUpdateWorker.ValueWithInfo(it, type = ItemUpdateWorker.ValueType.Timestamp) }
                 }
             }
             VALUE_GETTER_MAP[Constants.PREFERENCE_PHONE_STATE] = { context ->
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
@@ -72,7 +72,6 @@ class BackgroundTasksManager : BroadcastReceiver() {
                         info.itemName,
                         info.label,
                         info.value,
-                        info.mappedValue,
                         BackoffPolicy.EXPONENTIAL,
                         info.showToast,
                         info.taskerIntent,
@@ -110,8 +109,7 @@ class BackgroundTasksManager : BroadcastReceiver() {
                     WORKER_TAG_PREFIX_TASKER + itemName,
                     itemName,
                     label,
-                    state,
-                    mappedState,
+                    ItemUpdateWorker.ValueWithInfo(state, mappedState),
                     BackoffPolicy.EXPONENTIAL,
                     false,
                     intent.getStringExtra(TaskerPlugin.Setting.EXTRA_PLUGIN_COMPLETION_INTENT),
@@ -129,8 +127,7 @@ class BackgroundTasksManager : BroadcastReceiver() {
         val tag: String,
         val itemName: String,
         val label: String?,
-        val value: String,
-        val mappedValue: String?,
+        val value: ItemUpdateWorker.ValueWithInfo,
         val showToast: Boolean,
         val taskerIntent: String?,
         val asCommand: Boolean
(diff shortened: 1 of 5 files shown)


mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
```diff
@@ -91,34 +90,64 @@
...
-                Log.e(TAG, "Error updating item '$itemName' to value $value. Got HTTP error ${e.statusCode}", e)
+                Log.e(TAG, "Error updating item '$itemName'. Got HTTP error ${e.statusCode}", e)
                 sendTaskerSignalIfNeeded(taskerIntent, true, e.statusCode, e.localizedMessage)
                 Result.failure(buildOutputData(true, e.statusCode))
             }
         }
     }
 
+    private fun mapValueAccordingToItemTypeAndValue(value: ValueWithInfo, item: Item) = when {
+        value.value == "TOGGLE" && item.isOfTypeOrGroupType(Item.Type.Switch) -> determineOppositeState(item)
```
maniac103: We need to extend this check to also cover Dimmer and Rollershutter (and maybe also Contact?)

mueller-ma: Why not:

```suggestion
        value.value == "TOGGLE" -> determineOppositeState(item)
```
See https://github.com/openhab/openhab-android/commit/b619fa9b45a7b63213fc8b5be83a417b465d629e#r35360730

mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
```diff
@@ -91,34 +90,63 @@ class ItemUpdateWorker(context: Context, params: WorkerParameters) : Worker(cont
                     )
                     return@runBlocking Result.failure(buildOutputData(true, 500))
                 }
-                if (value == "TOGGLE") {
-                    value = determineOppositeState(item)
-                    mappedValue = value
+
+                val value = inputData.getValueWithInfo(INPUT_DATA_VALUE)!!
```
mueller-ma: Why move this inside the blocking block? If it's outside `valueToBeSent` could be logged in a failure.

maniac103: Because `valueToBeSent` needs `item` to be determined.

mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
maniac103: I don't want to fall back to `ON` for people who want to sent the string `TOGGLE` to e.g. a `StringItem`.

mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
mueller-ma: Makes sense, but I like to log the value in case of an error. Especially for cases where the value is in an inappropriate format for the item, e.g. a badly formatted DateTime string.
I added the logging.

mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
mueller-ma: I adjusted this check so all items are toggled where we suggest the `TOGGLE` value. Currently Contact items cannot be toggled, but this may be a good addition.

mobile/src/main/java/org/openhab/habdroid/model/Item.kt
```diff
@@ -159,3 +158,10 @@ fun String?.toItemType(): Item.Type {
         Item.Type.None
     }
 }
+
+fun Item.canBeToggled(): Boolean {
```
maniac103: This does not make much sense as extension function, as it can be added to the actual class.

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -312,17 +305,22 @@ class BackgroundTasksManager : BroadcastReceiver() {
                         putBoolean(Constants.PREFERENCE_ALARM_CLOCK_LAST_VALUE_WAS_ZERO, time == "0" || time == null)
                     }
 
-                    time
+                    if (time == null) {
+                        null
+                    } else {
+                        ItemUpdateWorker.ValueWithInfo(time, type = ItemUpdateWorker.ValueType.Timestamp)
+                    }
```
maniac103: I know this is my code ;-) but using `time?.let { ItemUpdateWorker.ValueWithInfo(it, type = ...) }` would be more elegant here I think.

mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
```diff
@@ -71,13 +78,9 @@ class ItemUpdateWorker(context: Context, params: WorkerParameters) : Worker(cont
...
         val itemName = inputData.getString(INPUT_DATA_ITEM_NAME)!!
-        var value = inputData.getString(INPUT_DATA_VALUE)!!
 
-        var label = inputData.getString(INPUT_DATA_LABEL)
-        if (label.isNullOrEmpty()) label = itemName
-
-        var mappedValue = inputData.getString(INPUT_DATA_MAPPED_VALUE)
-        if (mappedValue.isNullOrEmpty()) mappedValue = value
+        var valueForLog: String? = null
+        var valueToBeSentForLog: String? = null
```
maniac103: I find those additional variables slightly cumbersome. How about logging `value.value` and `valueToBeSent` before the actual HTTP call? Doing so avoids the weird variables (at the expense of having two log lines, which IMHO is OK).

PR #1709, opened 2019-12-02
Set alarm clock to 0 if alarm gets canceled

### LINKED ISSUE #1617: Set alarm clock to 0 if alarm gets canceled
See https://community.openhab.org/t/android-app-beta/41382/152

mueller-ma:
Instead of ignoring the alarm clock set by Tasker, send "0" to the
server. Also cache if the last sent alarm clock was 0 and don't send
again if so.

Fixes #1617


maniac103:
> Also cache if the last sent alarm clock was 0 and don't send
> again if so.

Why is this needed?

mueller-ma:
I've seen alarm clocks by tasker every few minutes. I don't want to waste battery by sending them all.

PR #1713, opened 2019-12-06
Bump 'com.android.tools.build:gradle' to 3.5.3

mueller-ma:


mueller-ma:
#1715

PR #1717, opened 2019-12-10
Refresh Image and Chart items after resuming the app

### LINKED ISSUE #1712: Image refresh timer: Use elapsedRealtime()
See https://community.openhab.org/t/image-refresh-not-working/86631/4

mueller-ma:
Fixes #1712


maniac103:
What's the advantage of this over using `elapsedRealtime()`? If e.g. refresh is set to 60000 and the app is paused for 5 seconds, there's no need to refresh on resume.

mueller-ma:
That depends on what the intended behavior should be: If the app resumes, should the image items be updated? I think so.
Imagine you have an image without auto refresh or a really long refresh time. You open the app and see the latest image. You "close" the app by pressing the home button and click on the app icon in the launcher a few hours later: You don't see the latest image, but the one from hours ago.

PR #1730, opened 2019-12-18
Override Uri.Builder#appendQueryParameter() to allow bool values

mueller-ma:


mueller-ma:
With https://github.com/openhab/openhab-android/pull/1731/commits/e35a513c4f8048de76d5a865e91717e04cd0f188 the function is used at only one place, so I don't think we need this anymore.

PR #1731, opened 2019-12-18
Use anyFormat for all icons

### LINKED ISSUE #1728: Some icons don't use anyFormat
I forgot to update some icon request in the full flavor, e.g. https://github.com/openhab/openhab-android/blob/master/mobile/src/full/java/org/openhab/habdroid/core/FcmMessageListenerService.kt#L129

_Originally posted by @mueller-ma in https://github.com/openhab/openhab-android/pull/1667#issuecomment-567001808_

mueller-ma:
Fixes #1728

### DIFF
--- a/mobile/src/full/java/org/openhab/habdroid/core/FcmMessageListenerService.kt
+++ b/mobile/src/full/java/org/openhab/habdroid/core/FcmMessageListenerService.kt
@@ -20,6 +20,7 @@ import android.app.NotificationManager
 import android.app.PendingIntent
 import android.content.Intent
 import android.graphics.Bitmap
+import android.net.Uri
 import android.os.Build
 import android.service.notification.StatusBarNotification
 import androidx.core.app.NotificationCompat
@@ -31,6 +32,8 @@ import org.openhab.habdroid.R
 import org.openhab.habdroid.core.connection.ConnectionFactory
 import org.openhab.habdroid.ui.MainActivity
 import org.openhab.habdroid.util.HttpClient
+import org.openhab.habdroid.util.IconFormat
+import org.openhab.habdroid.util.getIconFormat
 import org.openhab.habdroid.util.getNotificationTone
 import org.openhab.habdroid.util.getNotificationVibrationPattern
 import org.openhab.habdroid.util.getPrefs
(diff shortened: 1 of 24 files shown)


mobile/src/full/java/org/openhab/habdroid/core/FcmMessageListenerService.kt
```diff
@@ -123,10 +126,15 @@ class FcmMessageListenerService : FirebaseMessagingService() {
...
+            val encodedIcon = Uri.encode(icon)
+            val suffix = when (getPrefs().getIconFormat()) {
+                IconFormat.Png -> "png"
+                IconFormat.Svg -> "svg"
+            }
             if (connection != null) {
                 try {
                     iconBitmap = connection.httpClient
-                            .get("images/$icon.png", timeoutMillis = 1000)
+                            .get("icon/$encodedIcon?&format=$suffix&anyFormat=true", timeoutMillis = 1000)
```
maniac103: `&format` -> `format` (`&` isn't needed right after the question mark)

mobile/src/main/java/org/openhab/habdroid/ui/CloudNotificationAdapter.kt
```diff
@@ -110,7 +113,11 @@ class CloudNotificationAdapter(context: Context, private val loadMoreListener: (
             val conn = ConnectionFactory.cloudConnection
             if (notification.icon != null && conn != null) {
                 val encodedIcon = Uri.encode(notification.icon)
-                iconView.setImageUrl(conn, "images/$encodedIcon.png",
+                val suffix = when (itemView.context.getPrefs().getIconFormat()) {
+                    IconFormat.Png -> "png"
+                    IconFormat.Svg -> "svg"
+                }
+                iconView.setImageUrl(conn, "icon/$encodedIcon?&format=$suffix&anyFormat=true",
```
maniac103: As above.

mueller-ma:
> I wonder whether we should centralize this logic

I was thinking about this as well, but couldn't find a good way to pass both icon name and/or icon path (used by LinkedPage) to the new function.

mobile/src/full/java/org/openhab/habdroid/core/FcmMessageListenerService.kt
mueller-ma: Thanks!

maniac103:
@mueller-ma What do you think about https://github.com/maniac103/openhab.android/commit/6e6bc147d397756baeccc6f578f8577fc6390bdb ? I thought about this a bit more and think the problem in finding a good extension function is the inconsistent icon path handling in the app ... hence this is what I came up with. (NB: it's only compile time tested so far)

mueller-ma:
TODO:

- [ ] Fix missing icon if widget is created from within the app
- [ ] Add fallback or migrate existing item update widget data

maniac103:
> Add fallback or migrate existing item update widget data

https://github.com/maniac103/openhab.android/commit/dc6600dcaef63a123052fc41e5e9a6a04900af80

mueller-ma:
Should be RTM now.

PR #1741, opened 2019-12-31
Do SSL Client auth in webview using the certificate configured for the

rcatolino:
This commit uses the certificate configured for the openhab server in webview widgets as well.

Ideally the user should be able to choose another certificate if needed, but I this fixes the case where the webview URL is behind the same reverse proxy as the openhab server and the certificate needed is the same.

Signed-off-by: rca <raphael.catolino@gmail.com>

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/ConnectionWebViewClient.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/ConnectionWebViewClient.kt
@@ -15,19 +15,28 @@ package org.openhab.habdroid.ui
 
 import android.net.http.SslCertificate
 import android.net.http.SslError
+import android.security.KeyChain
+import android.security.KeyChainException
 import android.util.Base64
 import android.util.Log
+import android.webkit.ClientCertRequest
 import android.webkit.HttpAuthHandler
 import android.webkit.SslErrorHandler
 import android.webkit.WebView
 import android.webkit.WebViewClient
 import de.duenndns.ssl.MemorizingTrustManager
-import org.openhab.habdroid.R
-import org.openhab.habdroid.core.connection.Connection
 import java.io.ByteArrayInputStream
 import java.security.cert.Certificate
 import java.security.cert.CertificateException
 import java.security.cert.CertificateFactory
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import org.openhab.habdroid.core.connection.Connection
+import org.openhab.habdroid.R
+import org.openhab.habdroid.util.Constants
+import org.openhab.habdroid.util.getPrefs
+import org.openhab.habdroid.util.isDemoModeEnabled
 
 open class ConnectionWebViewClient(
     private val connection: Connection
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/ConnectionWebViewClient.kt
```diff
@@ -62,6 +69,41 @@ open class ConnectionWebViewClient(
...
+        if (prefs.isDemoModeEnabled()) {
+            request.cancel()
+        }
+
+        val alias = prefs.getString(Constants.PREFERENCE_SSL_CLIENT_CERT, null)
+        Log.d(TAG, "Using alias " + alias)
+        if (alias == null) {
+            request.cancel()
+        } else {
+            ForwardRequestTask(request, view).execute(alias)
```
maniac103: This should be implemented using a coroutine via `launch` block instead.

mobile/src/main/java/org/openhab/habdroid/ui/ConnectionWebViewClient.kt
```diff
@@ -62,6 +70,34 @@ open class ConnectionWebViewClient(
...
+        Log.d(TAG, "SSL Cient Cert required")
+        val prefs = view.context.getPrefs()
+        if (prefs.isDemoModeEnabled()) {
+            request.cancel()
+        }
+
+        val alias = prefs.getString(Constants.PREFERENCE_SSL_CLIENT_CERT, null)
+        Log.d(TAG, "Using alias " + alias)
+        if (alias == null) {
+            request.cancel()
```
mueller-ma: You can return here and remove the `else`.

mobile/src/main/java/org/openhab/habdroid/ui/ConnectionWebViewClient.kt
```diff
@@ -62,6 +71,36 @@ open class ConnectionWebViewClient(
...
+    override fun onReceivedClientCertRequest(view: WebView, request: ClientCertRequest) {
+        Log.d(TAG, "SSL Cient Cert required")
+        val prefs = view.context.getPrefs()
+        if (prefs.isDemoModeEnabled()) {
+            request.cancel()
+            return
+        }
+
+        val alias = prefs.getString(Constants.PREFERENCE_SSL_CLIENT_CERT, null)
+        Log.d(TAG, "Using alias " + alias)
```
mueller-ma:
```suggestion
        Log.d(TAG, "Using alias $alias")
```

mobile/src/main/java/org/openhab/habdroid/ui/ConnectionWebViewClient.kt
```diff
@@ -62,6 +71,36 @@ open class ConnectionWebViewClient(
...
+            return
+        }
+
+        GlobalScope.launch(Dispatchers.IO) {
+            try {
+                val chain = KeyChain.getCertificateChain(view.context, alias)
+                val privateKey = KeyChain.getPrivateKey(view.context, alias)
+                request.proceed(privateKey, chain)
+            } catch (e: KeyChainException) {
+                Log.d(TAG, "KeyChain Exception", e)
```
mueller-ma:
```suggestion
                Log.d(TAG, "Error getting chain or private key", e)
```

The exception itself is also passed the log line, so the string doesn't need to contain the exception name.

mobile/src/main/java/org/openhab/habdroid/ui/ConnectionWebViewClient.kt
```diff
@@ -62,6 +71,36 @@ open class ConnectionWebViewClient(
...
+                request.ignore()
+            } catch (e: InterruptedException) {
+                Log.d(TAG, "Interrupted Exception", e)
```
mueller-ma: Same as above.

mobile/src/main/java/org/openhab/habdroid/ui/ConnectionWebViewClient.kt
```diff
@@ -62,6 +71,36 @@ open class ConnectionWebViewClient(
         }
     }
 
+    override fun onReceivedClientCertRequest(view: WebView, request: ClientCertRequest) {
+        Log.d(TAG, "SSL Cient Cert required")
```
mueller-ma:
```suggestion
        Log.d(TAG, "SSL Client Cert required")
```

mobile/src/main/java/org/openhab/habdroid/ui/ConnectionWebViewClient.kt
```diff
@@ -62,6 +71,36 @@ open class ConnectionWebViewClient(
...
+        if (alias == null) {
+            request.cancel()
+            return
+        }
+
+        GlobalScope.launch(Dispatchers.IO) {
+            try {
+                val chain = KeyChain.getCertificateChain(view.context, alias)
+                val privateKey = KeyChain.getPrivateKey(view.context, alias)
+                request.proceed(privateKey, chain)
```
maniac103: I wonder whether it's OK to do the callback on a different thread, or whether we should do it in the main thread.

rcatolino: FWIW I haven't had any issues using it so far so it doesn't seem to cause any problems.
I'm actually wondering the opposite : the whole class is only used for networking stuff, shouldn't all that be happening on the IO thread ?
(Note that I don't know the first thing about kotlin, and I haven't worked on Android for ages, so I don't really know what I'm talking about.)

maniac103: I looked into the docs for this, and the current behavior seems fine, so NVM.

PR #1756, opened 2020-01-05
Add ability to create shortcut with dark background

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
@@ -22,7 +22,6 @@ import android.content.Intent
 import android.graphics.Bitmap
 import android.graphics.Canvas
 import android.graphics.Color
-import android.net.Uri
 import android.nfc.NfcAdapter
 import android.os.Build
 import android.os.Bundle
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
```diff
@@ -240,7 +244,9 @@ class WidgetListFragment : Fragment(), WidgetAdapter.ItemClickListener {
                 }
             }
             if (ShortcutManagerCompat.isRequestPinShortcutSupported(context)) {
-                menu.add(Menu.NONE, CONTEXT_MENU_ID_PIN_HOME, Menu.NONE, R.string.home_shortcut_pin_to_home)
+                menu.add(Menu.NONE, CONTEXT_MENU_ID_PIN_HOME_WHITE, Menu.NONE,
+                    R.string.home_shortcut_pin_to_home_bright)
+                menu.add(Menu.NONE, CONTEXT_MENU_ID_PIN_HOME_BLACK, Menu.NONE, R.string.home_shortcut_pin_to_home_dark)
```
maniac103: Having the color options just besides each other seems somewhat random menu design to me. How about either
- making this a preference or
- using a submenu?

mueller-ma: > using a submenu?

:+1: 

I added a submenu, but now I cannot get the menuInfo of the submenu entries.

mueller-ma: I found a solution.

PR #1757, opened 2020-01-05
Add comments to strings with multiple substitutions

mueller-ma:


mueller-ma:
I wasn't aware of xliff tags. I opened https://github.com/openhab/openhab-android/pull/1759 and updated a few strings. Let's see how Crowdin handles them before updating all strings.

mueller-ma:
It seems that Crowdin doesn't support this very well:
grafik

maniac103:
Damn. Let's go back to the other solution then. I just remembered xliff tags are used throughout AOSP code base, but didn't re-check how those look in Crowdin.

PR #1761, opened 2020-01-07
Fix error message for SSL errors

mueller-ma:


mueller-ma:
`Throwable?.hasCause()` needs to be updated to check for parent classes as well.
Example: If an `SSLHandshakeException` occurs and `hasCause(SSLException::class.java)` is called, it returns false. However `SSLHandshakeException` is a child class of `SSLException`.

mueller-ma:
As a workaround for the problem I described above, I added handling for `SSLHandshakeException`.

PR #1765, opened 2020-01-09
Use notification channels in MTM

mueller-ma:
Otherwise there's no error message if an SSL error occurs in background tasks.


mobile/src/main/java/de/duenndns/ssl/MemorizingTrustManager.java
```diff
@@ -619,37 +619,26 @@ private static void setLatestEventInfoReflective(Notification notification,
...
+            .setContentTitle(mtmNotification)
+            .setContentText(certName)
+            .setTicker(certName)
+            .setSmallIcon(android.R.drawable.ic_lock_lock)
+            .setWhen(currentMillis)
+            .setContentIntent(call)
+            .setAutoCancel(true);
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            notification.setChannelId(NotificationUpdateObserver.CHANNEL_ID_BACKGROUND_ERROR);
```
mueller-ma: Need to create the channel before.

PR #1766, opened 2020-01-09
Add 'clear errors' action to background error notification

mueller-ma:
Otherwise it's not possible to get rid of 404 errors if an item has been removed from the server, but the client still has a corresponding widget.

Also add a 404 error string for notifications.

### DIFF
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -154,6 +154,7 @@
     <string name="error_short_certificate_wrong_host">Certificate not valid for %s</string>
     <string name="error_short_http_code_401">Authentication failed (HTTP code %d)</string>
     <string name="error_short_http_code_403" translatable="false">@string/error_short_http_code_401</string>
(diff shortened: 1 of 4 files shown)


mobile/src/main/res/values/strings.xml
```diff
@@ -220,6 +221,7 @@
     <string name="item_update_connection_error_label">%1$s update failed (No connection)</string>
     <string name="error_sending_alarm_clock_item_empty">Please enter a valid Item name</string>
     <string name="retry">Retry</string>
+    <string name="clear_errors">Clear errors</string>
```
mueller-ma: Maybe a better wording would be "Ignore" or "Cancel".

maniac103: Ignore sounds good.

mueller-ma: Right, "Cancel" sounds like it would stop sending further alarm clocks.

PR #1771, opened 2020-01-15
Close dialogs when opening a new page

### LINKED ISSUE #1755: Close color picker popup when opening new subpage
If I open the color picker popup, minimize the app and use a sitemap shortcut on my home screen, the sitemap is opened, but the popup is still open.
This might be the same for mapview and setpoint popups.

mueller-ma:
Fixes #1755

### DIFF
--- a/mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.kt
+++ b/mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.kt
@@ -67,6 +67,7 @@ object MapViewHelper {
         private val mapView: MapView = itemView.findViewById(R.id.mapview)
         private val handler: Handler = Handler()
         private var started: Boolean = false
+        override val dialogManager = WidgetAdapter.DialogManager()
 
         init {
             with(mapView) {
@@ -144,6 +145,7 @@ object MapViewHelper {
                     .setNegativeButton(R.string.close, null)
                     .create()
 
+            dialogManager.manage(dialog)
             with(dialog) {
                 setOnDismissListener { mapView.onPause() }
                 setCanceledOnTouchOutside(true)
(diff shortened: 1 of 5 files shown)


mueller-ma:
If I close the dialogs in `onStop()` in the ViewHolder, the dialog gets closed if the app goes to background even for a very short time.
I'd like to keep the dialog open as long as possible.

maniac103:
Then the base ViewHolder class needs a new callback for that (or a getter that returns a `DialogInterface`) ... the static var really is not a good solution.

maniac103:
My suggestion would be something like this (only compile time tested): https://github.com/maniac103/openhab.android/commit/2c66ce21d231191606618da27a264f5f2d430b18

mueller-ma:
In demo mode: Open the color picker in "Widget Overview". Minimize the app and open the shortcut to "Outside temperature". Color picker is still open.
`collectWidgetFragments()` gets two fragments: "Main Menu" and "Outside temperature".

maniac103:
Damn, right. Does this work?

```diff
diff --git a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
index 5a8bacaa..90023bb5 100644
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
@@ -136,6 +136,11 @@ class WidgetListFragment : Fragment(), WidgetAdapter.ItemClickListener {
         emptyPageView = view.findViewById(android.R.id.empty)
     }
 
+    override fun onDetach() {
+        closeAllDialogs()
+        super.onDetach()
+    }
+
     override fun onStart() {
         Log.d(TAG, "onStart() $displayPageUrl")
         super.onStart()
```

mueller-ma:
Yes, that works. Should be RTM now.

PR #1775, opened 2020-01-19
Fix "IllegalArgumentException: Malformed \uxxxx encoding."

mueller-ma:
````
Fatal Exception: java.lang.IllegalArgumentException: Malformed \uxxxx encoding.
       at java.util.Properties.loadConvert(Properties.java:573)
       at java.util.Properties.load0(Properties.java:388)
       at java.util.Properties.load(Properties.java:336)
       at org.openhab.habdroid.util.MjpegInputStream.parseContentLength(MjpegInputStream.kt:54)
       at org.openhab.habdroid.util.MjpegInputStream.readMjpegFrame(MjpegInputStream.kt:72)
       at org.openhab.habdroid.util.MjpegStreamer$doStream$1.invokeSuspend(MjpegStreamer.kt:50)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
````

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/MjpegInputStream.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/util/MjpegInputStream.kt
@@ -21,6 +21,7 @@ import java.io.ByteArrayInputStream
 import java.io.DataInputStream
 import java.io.IOException
 import java.io.InputStream
+import java.lang.IllegalArgumentException
 import java.util.Properties
 
 class MjpegInputStream(stream: InputStream) : DataInputStream(BufferedInputStream(stream, FRAME_MAX_LENGTH)) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: mobile/src/main/java/org/openhab/habdroid/util/MjpegInputStream.kt [hunk 1]
```diff
@@ -21,6 +21,7 @@ import java.io.ByteArrayInputStream
 import java.io.DataInputStream
 import java.io.IOException
 import java.io.InputStream
+import java.lang.IllegalArgumentException
```
maniac103: This import seems unneeded to me.

mueller-ma: Why? `IllegalArgumentException` is catched below and all other exceptions here are also imported.

maniac103: Because `java.lang` imports happen implicitly anyway. You don't import `java.lang.String` either ;-)

Issue #1767, opened 2020-01-12
after long time open it tries to load, but no sucess. menu still works


ThomasRausU:
### Actual behaviour
I have an wall tablet, which has only openhab app running. After some time the sitemap is not displayed. Just showing the loading circle. App itself is not hanging, menu changes are possible. Only changing the server url back and forth it's working correct for a while.

### Expected behaviour
app should not lose it's connection over time. Reconnection shouldn't be any problem

 
### Steps to reproduce
1. wait

### Can you reproduce the issue in demo mode?

-> don't know yet, trying. But since its connection related I bet it's not reproduceably in demo mode

#### Client
* Android version: 8.1.0
* Device model: TERRA_PAD_1005P0KO
* App version : 2.11.5-beta 8.1.2020  (tried it which default non beta version, same problem)
* Build flavor : Full
* Device language: de

#### Server
* Server version: 2.5.1~20200108044053-1 (and non beta same problem)
* Reverse Proxy: 
* Authentication method : None

#### openHAB Cloud
* Self hosted cloud or myopenhab.org: 
* openHAB Cloud version : 
* Authentication method : 

#### App log
<details>
  <summary>Click to expand</summary>

```
Please add the app log if the issue is not a pure UI issue or it cannot be reproduced in demo mode.
Open the app, go to `Settings` => `View log` and insert the log here.
```

</details>

#### openHAB Server log
no entry

ThomasRausU:
-----------------------
Device information
Fingerprint: TERRA/TERRA_PAD_1005POKO/TERRA_PAD_1005POKO:8.1.0/O11019/vID100103:user/release-keys
Model: TERRA_PAD_1005POKO
Manufacturer: WORTMANN AG
Brand: TERRA
Device: TERRA_PAD_1005POKO
Product: TERRA_PAD_1005POKO
OS: 8.1.0
-----------------------

--------- beginning of events
01-12 19:17:14.010 29935 29935 I am_on_resume_called: [0,org.openhab.habdroid.ui.MainActivity,RESUME_ACTIVITY]
01-12 19:19:19.159 29935 29935 I am_on_paused_called: [0,org.openhab.habdroid.ui.MainActivity,handlePauseActivity]
01-12 19:19:19.198 29935 29935 I am_on_stop_called: [0,org.openhab.habdroid.ui.MainActivity,sleeping]
[...]

PR #1781, opened 2020-01-21
Auto retry after connection failed

mueller-ma:
Automatically retry after 30 seconds if there's no connection or
connection error.

See https://github.com/openhab/openhab-android/issues/1767#issuecomment-576285773

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -67,6 +67,7 @@ import androidx.drawerlayout.widget.DrawerLayout
 import androidx.recyclerview.widget.RecyclerView
 import com.google.android.material.navigation.NavigationView
 import com.google.android.material.snackbar.Snackbar
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.Job
@@ -459,7 +460,7 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
 
     private fun scheduleRetry(runAfterDelay: () -> Unit) {
         retryJob?.cancel(CancellationException("scheduleRetry() was called"))
-        retryJob = GlobalScope.launch(Dispatchers.Main) {
+        retryJob = CoroutineScope(Dispatchers.Main + Job()).launch {
             delay(30 * 1000)
             Log.d(TAG, "runAfterDelay()")
             runAfterDelay()
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -67,8 +67,10 @@ import androidx.drawerlayout.widget.DrawerLayout
 import androidx.recyclerview.widget.RecyclerView
 import com.google.android.material.navigation.NavigationView
 import com.google.android.material.snackbar.Snackbar
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -445,6 +456,15 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         updateTitle()
     }
 
+    private fun scheduleRetry(runAfterDelay: () -> Unit) {
+        retryJob?.cancel(CancellationException("scheduleRetry() was called"))
+        retryJob = GlobalScope.launch(Dispatchers.Main) {
```
maniac103: The scope should be an activity specific scope, not the global scope.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -491,12 +511,16 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
...
+        retryJob?.cancel(CancellationException("queryServerProperties() was called"))
         val successCb: (ServerProperties) -> Unit = { props ->
             serverProperties = props
             updateSitemapAndHabPanelDrawerItems()
             if (props.sitemaps.isEmpty()) {
                 Log.e(TAG, "openHAB returned empty Sitemap list")
                 controller.indicateServerCommunicationFailure(getString(R.string.error_empty_sitemap_list))
+                scheduleRetry {
+                    retryServerPropertyQuery()
+                }
```
maniac103: Does a retry really help here?

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -965,6 +989,9 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         }
 
         controller.indicateServerCommunicationFailure(message)
+        scheduleRetry {
+            retryServerPropertyQuery()
+        }
```
maniac103: I would suggest to only selectively retry here. I don't think retrying on HTTP 4xx return codes provides any benefit.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -436,6 +443,10 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
             }
             else -> {
                 controller.indicateNoNetwork(getString(R.string.error_network_not_available), false)
+                scheduleRetry {
+                    ConnectionFactory.restartNetworkCheck()
```
maniac103: If this helps, our connection tracking is broken. Do we really want to mask such cases?

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
mueller-ma: I don't think this error is shown at all, because the server always returns a sitemap based on things.
But let's say the app connects to the server in the same moment the server reloads the sitemap definitions (and the default sitemap is not available). In this case a retry helps.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
mueller-ma: I just copied the code from https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt#L537

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
mueller-ma: IMO a retry doesn't hurt either and there are some 4XX errors that can be "fixed" by waiting, e.g. 429 (Too many requests), 423 (Locked) or 408 (Request Timeout). (Not that I have ever seen these codes in the wild).

PR #1784, opened 2020-01-22
Use Slider from "Material Components for Android"

### LINKED ISSUE #1626: Use discrete sliders if only a few states are possible

mueller-ma:
Fixes #1626

The sliders are always discrete if a step size is specified.

TODO:
- [x] Find a way to set the slider to integer only without setting a
step size
- [x] Make the halloween easteregg working or remove it
- [x] Think about a better fix for "The stepSize must be 0, or a factor
of the valueFrom-valueTo range"
- [x] ~~Wait until the lib is in beta~~ The alpha version is already in the app.


mueller-ma:
I replaced the slider in the ChartScalingPref as well, but I'm not satisfied with the result. The discrete dots are hardly visible and the new value can already be seen next to the slider.

mueller-ma:
RTM from my side.

mueller-ma:
Sliders should use `colorSecondary`.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -517,69 +516,75 @@ class WidgetAdapter(
...
                 }
             } else {
                 val number = state.asNumber
                 if (number != null) {
-                    val progress = (number.value - widget.minValue) / widget.step
-                    seekBar.progress = Math.round(progress)
+                    var progress = number.value
+                    progress = min(progress, seekBar.valueTo)
+                    progress = max(progress, seekBar.valueFrom)
+                    seekBar.value = progress
```
maniac103, line 553: Can use this Kotlin method: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/coerce-in.html

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -517,69 +516,75 @@ class WidgetAdapter(
...
         }
 
-        override fun onStartTrackingTouch(seekBar: SeekBar) {
-            Log.d(TAG, "onStartTrackingTouch position = ${seekBar.progress}")
+        override fun onStopTrackingTouch(slider: Slider) {
+            val value = slider.value
+            Log.e(TAG, "onValueChange value = $value")
+            val item = boundWidget?.item ?: return
+            if (item.isOfTypeOrGroupType(Item.Type.Color)) {
+                connection.httpClient.sendItemCommand(item, value.toString())
```
maniac103, line 575: This is now a float string, instead of the integer it was before. Can the server deal with that?

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -517,69 +516,75 @@ class WidgetAdapter(
...
-            Log.d(TAG, "onStopTrackingTouch position = $progress")
-            val widget = boundWidget
-            val item = widget?.item ?: return
-            if (item.isOfTypeOrGroupType(Item.Type.Color)) {
-                connection.httpClient.sendItemCommand(item, progress.toString())
+        override fun getFormattedValue(value: Float): String {
+            val item = boundWidget?.item ?: return ""
+            val realValue = if (boundWidget?.step == 1F) value.toInt().toFloat() else value
+            return if (item.isOfTypeOrGroupType(Item.Type.Color)) {
+                realValue.toString()
```
maniac103, line 585: Would it be better to convert to percent here?

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -517,69 +516,75 @@ class WidgetAdapter(
...
             super.bind(widget)
             boundWidget = widget
 
-            val stepCount = (widget.maxValue - widget.minValue) / widget.step
-            seekBar.max = Math.ceil(stepCount.toDouble()).toInt()
-            seekBar.progress = 0
+            // Fix "The stepSize must be 0, or a factor of the valueFrom-valueTo range" exception
+            seekBar.valueTo = widget.maxValue - (widget.maxValue - widget.minValue).rem(widget.step)
+            seekBar.valueFrom = widget.minValue
+            seekBar.stepSize = if (widget.step == 1F) 0F else widget.step
```
maniac103, line 535: Why special case 1 here?

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -517,69 +516,75 @@ class WidgetAdapter(
         private val connection: Connection,
         colorMapper: ColorMapper
     ) : LabeledItemBaseViewHolder(inflater, parent, R.layout.widgetlist_slideritem, connection, colorMapper),
-        SeekBar.OnSeekBarChangeListener {
-        private val seekBar: SeekBar = itemView.findViewById(R.id.seekbar)
+        Slider.OnSliderTouchListener, Slider.LabelFormatter {
+        private val seekBar: Slider = itemView.findViewById(R.id.seekbar)
```
maniac103, line 520: Name this `slider`?

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: If `stepSize > 0` the slider is set to discrete. The step size sent by the server is 1 if no step size is defined.
IMO if the user defines no step size the slider shouldn't be discrete. Checking for `step == 1` here is not the perfect solution, though, as it may be requested by the user.

mueller-ma: Thinking about this, the better solution would be to just set the step size for now and see what feedback we get in https://github.com/material-components/material-components-android/issues/1233.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: I just tested with a color bulb and it works fine.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: I just tested it with a color bulb and it works fine.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: Yes, makes sense.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: I added a function to remove trailing `.0` from floats.

PR #1794, opened 2020-01-25
Don't set Sitemap from demo mode as default

### LINKED ISSUE #1083: Default sitemap is set to "default" when enabling demo mode.
### Actual behaviour
The default sitemap is "default" after using demo mode.

### Expected behaviour
The default sitemap for demo mode and for regular connections should be different.

 
### Steps to reproduce
1. Have a server with at least two sitemaps. One must be "default", the other one e.g. "foo".
2. Connect to your server
3. Select "foo" as default sitemap
4. Enable demo mode
5. The demo server contains only one sitemap, so this one is saved as default
6. Disable demo mode
7. Your default sitemap isn't "foo" anymore

mueller-ma:
Fixes #1083

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -640,11 +640,13 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
                     handled = true
                 }
                 R.id.default_sitemap -> {
-                    val sitemap = serverProperties?.sitemaps?.firstOrNull { s -> s.name == prefs.getDefaultSitemap() }
+                    val sitemap = serverProperties?.sitemaps?.firstOrNull {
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -640,11 +640,13 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
                     handled = true
                 }
                 R.id.default_sitemap -> {
-                    val sitemap = serverProperties?.sitemaps?.firstOrNull { s -> s.name == prefs.getDefaultSitemap() }
+                    val sitemap = serverProperties?.sitemaps?.firstOrNull {
+                            s -> s.name == prefs.getDefaultSitemap(connection)
```
maniac103: `s ->`  should be on old line + indent by 4 spaces

(here and below)

mueller-ma: Done

PR #1796, opened 2020-01-28
Implement data saver

### LINKED ISSUE #660: Check if device runs in data saver mode
Data saver restricts background data. I haven't checked if apps know if data saver is on, but if that's the case, we could use this information to reduce data. Some ideas:
1. Hide images, videos, webviews, etc until the user taps on their placeholder
2. No auto refresh
3. Show default icon (e.g. temperature) if no icon for a specific state (e.g. temperature-40) is cached

mueller-ma:
Respect device wide data saver and add setting to enable data saver for
openHAB only.

If enabled, ...
* widgets with high data usage are replaced by a button that shows them
* automatic image refresh is disabled
* icon paths don't contain the state => reduces icon downloads
* chart resolution is lowered

Fixes #660

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
@@ -33,12 +33,12 @@ class IconResource internal constructor(
     internal val isOh2: Boolean,
     internal val customState: String?
 ) : Parcelable {
-    fun toUrl(context: Context): String {
-        return toUrl(context.getPrefs().getIconFormat())
+    fun toUrl(context: Context, includeState: Boolean): String {
+        return toUrl(includeState, context.getPrefs().getIconFormat())
     }
 
     @VisibleForTesting
-    fun toUrl(iconFormat: IconFormat): String {
+    fun toUrl(includeState: Boolean, iconFormat: IconFormat): String {
         if (!isOh2) {
             return "images/$icon.png"
         }
@@ -54,7 +54,7 @@ class IconResource internal constructor(
             .appendQueryParameter("format", suffix)
             .appendQueryParameter("anyFormat", true)
 
-        if (!customState.isNullOrEmpty()) {
+        if (!customState.isNullOrEmpty() && includeState) {
             builder.appendQueryParameter("state", customState)
         }
 
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -34,12 +34,14 @@ import android.view.ViewGroup
 import android.view.WindowManager
 import android.webkit.WebView
 import android.widget.ArrayAdapter
+import android.widget.Button
 import android.widget.ImageView
 import android.widget.LinearLayout
 import android.widget.NumberPicker
 import android.widget.SeekBar
 import android.widget.TextView
 import androidx.annotation.LayoutRes
+import androidx.annotation.StringRes
 import androidx.annotation.VisibleForTesting
 import androidx.appcompat.app.AlertDialog
 import androidx.core.content.ContextCompat
@@ -70,9 +72,12 @@ import org.openhab.habdroid.ui.widget.ContextMenuAwareRecyclerView
 import org.openhab.habdroid.ui.widget.DividerItemDecoration
 import org.openhab.habdroid.ui.widget.ExtendedSpinner
 import org.openhab.habdroid.ui.widget.WidgetImageView
+import org.openhab.habdroid.util.CacheManager
 import org.openhab.habdroid.util.HttpClient
 import org.openhab.habdroid.util.MjpegStreamer
 import org.openhab.habdroid.util.getPrefs
+import org.openhab.habdroid.util.isDataSaverActive
+import org.openhab.habdroid.util.orDefaultIfEmpty
 import java.util.Calendar
 import java.util.HashMap
 import java.util.Locale
@@ -169,7 +174,7 @@ class WidgetAdapter(
             TYPE_ROLLERSHUTTER -> RollerShutterViewHolder(inflater, parent, connection, colorMapper)
             TYPE_SETPOINT -> SetpointViewHolder(inflater, parent, connection, colorMapper)
             TYPE_CHART -> ChartViewHolder(inflater, parent, chartTheme, connection)
-            TYPE_VIDEO -> VideoViewHolder(inflater, parent)
+            TYPE_VIDEO -> VideoViewHolder(inflater, parent, connection)
             TYPE_WEB -> WebViewHolder(inflater, parent, connection)
             TYPE_COLOR -> ColorViewHolder(inflater, parent, connection, colorMapper)
             TYPE_VIDEO_MJPEG -> MjpegVideoViewHolder(inflater, parent, connection)
@@ -324,6 +329,75 @@ class WidgetAdapter(
             }
             iconView.loadWidgetIcon(connection, widget, colorMapper)
         }
+    }
+
+    abstract class HeavyDataViewHolder internal constructor(
+        inflater: LayoutInflater,
+        parent: ViewGroup,
+        @LayoutRes layoutResId: Int,
+        protected val connection: Connection
+    ) : ViewHolder(inflater, parent, layoutResId) {
+        private var boundWidget: Widget? = null
+        protected val widgetContentView: View = itemView.findViewById(R.id.widget_content)
+        private val dataSaverView: View = itemView.findViewById(R.id.data_saver)
+        private val dataSaverButton: Button = itemView.findViewById(R.id.data_saver_button)
+        private val dataSaverHint: TextView = itemView.findViewById(R.id.data_saver_hint)
+
+        override fun bind(widget: Widget) {
+            boundWidget = widget
+
+            val overrideDataSaver = when {
+                // We do only cache images. Maps may be also cached, but we cannot check their cache.
+                widget.type != Widget.Type.Image -> false
+                // If image is cached, display it.
+                widget.url != null -> CacheManager.getInstance(itemView.context)
+                    .isBitmapCached(connection.httpClient.buildUrl(widget.url))
+                // The image item is either undefined or the image is included as state and already loaded.
+                else -> true
+            }
+
+            handleDataSaver(overrideDataSaver)
+        }
+
+        private fun handleDataSaver(overrideDataSaver: Boolean) {
+            val widget = boundWidget ?: return
+            val dataSaverActive = itemView.context.isDataSaverActive() && !overrideDataSaver
+
+            dataSaverView.isVisible = dataSaverActive
+            widgetContentView.isVisible = !dataSaverView.isVisible
+
+            if (dataSaverActive) {
+                dataSaverButton.setOnClickListener {
+                    loadWidget(widget)
+                    handleDataSaver(true)
+                }
+
+                @StringRes val typeResId = when (widget.type) {
+                    Widget.Type.Image -> R.string.widget_type_image
+                    Widget.Type.Webview -> R.string.widget_type_webview
+                    Widget.Type.Video -> R.string.widget_type_video
+                    Widget.Type.Chart -> R.string.widget_type_chart
+                    else -> 0
+                }
+
+                dataSaverHint.text = itemView.context.getString(R.string.data_saver_hint,
+                    widget.label.orDefaultIfEmpty(itemView.context.getString(typeResId)))
+            } else {
+                dataSaverButton.setOnClickListener(null)
+                loadWidget(widget)
+            }
+        }
+
+        fun handleDataSaverChange(turnedOn: Boolean) {
+            if (turnedOn && (this is ImageViewHolder || this is ChartViewHolder)) {
+                // For Images and Charts continue showing the old data, but stop the auto refresh
+                stop()
+            } else {
+                bind(boundWidget ?: return)
+            }
+        }
+
+        abstract fun loadWidget(widget: Widget)
     }
 
     class GenericViewHolder internal constructor(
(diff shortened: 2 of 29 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -328,6 +330,38 @@ class WidgetAdapter(
         }
     }
 
+    abstract class HeavyDataViewHolder internal constructor(
+        inflater: LayoutInflater,
+        parent: ViewGroup,
+        @LayoutRes layoutResId: Int
+    ) : ViewHolder(inflater, parent, layoutResId) {
+        protected val widgetContentView: View = itemView.findViewById(R.id.widget_content)
```
mueller-ma: I must use `View` here, but this is generic to be used by the child view holders.

mobile/src/main/res/values/strings.xml
```diff
@@ -416,4 +416,10 @@
     <string name="item_update_widget_auto_generate_widget_label">Auto generate widget label</string>
     <string name="item_update_widget_widget_label">Widget label</string>
     <string name="item_update_widget_default_label"><![CDATA[<Item label>: <Action>]]></string>
+    <string name="data_saver_title">Data saver</string>
+    <string name="data_saver_on">Reduce data usage on mobile data</string>
```
mueller-ma: Actually the data usage is reduced on metered networks, but I'm not sure if every user knows what that means.

mobile/src/main/res/values/strings.xml
```diff
@@ -416,4 +416,10 @@
...
+    <string name="data_saver_off">Respect device setting</string>
+    <string name="data_saver_webview">WebView Items are not rendered if data saver is active</string>
```
mueller-ma: Note to self: String unused

mueller-ma:
@maniac103 How do you think about this approach to hide the items? I implemented it for Image items as a draft for the other items.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -328,6 +330,38 @@ class WidgetAdapter(
...
+        private val dataSaverView: View = itemView.findViewById(R.id.data_saver)
+        private val dataSaverButton: Button = itemView.findViewById(R.id.data_saver_button)
+        private val dataSaverHint: TextView = itemView.findViewById(R.id.data_saver_hint)
+        private var overrideDataSaver: Boolean = false
```
maniac103: Having a plain member variable for this doesn't work (as intended). ViewHolders can be re-bound to different widgets at any time, which needs special handling. IMHO, the flow should be something like
- on rebind, check whether the request can be satisfied easily (e.g. image is already cached)
- if that isn't the case, set flag to false, show button etc.
- if it is the case, just show the image

(To test this, just try having e.g. 10 image items with alternating URLs -> item 1 with URL 1, item 2 with URL 2, item 3 with URL 1, etc.)

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
```diff
@@ -307,6 +309,15 @@ fun Context.getHumanReadableErrorMessage(url: String, httpCode: Int, error: Thro
     }
 }
 
+fun Context.isDataSaverActive(): Boolean {
+    val dataSaverPref = getPrefs().getBoolean(Constants.PREFERENCE_DATA_SAVER, false)
+    val cm = getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+    if ((dataSaverPref && cm.isActiveNetworkMetered) || Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
```
maniac103: I think we should take the 'is metered' aspect from the active `Connection`, since we may bind the data traffic to a specific network. We can extract that information from the `NetworkInfo` or `NetworkCapabilities` objects and transfer it to `Connection` from there.
One example for this is 'Wifi without internet' + 'mobile network': active network will report being metered, but we won't transfer data over that interface.
On the other hand, I'm not sure whether we would want data saver if the user uses an unmetered VPN over a metered mobile connection. :man_shrugging:

mobile/src/main/res/values/strings.xml
```diff
@@ -416,4 +416,10 @@
...
+    <string name="data_saver_load">Load anyway</string>
+    <string name="data_saver_hint">%s is not loaded to save data</string>
```
maniac103: Should have a meaningful fallback for when the widget has no label ... and probably is -> was.

mobile/src/main/res/layout/widgetlist_data_saver.xml
```diff
@@ -0,0 +1,39 @@
...
+
+    <TextView
+        android:id="@+id/data_saver_hint"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center_horizontal"
+        android:padding="16dp"
+        android:textAppearance="?android:attr/textAppearanceMedium"
+        android:textColor="@color/empty_list_text_color"
+        tools:text="Webcam is not loaded to save data" />
```
maniac103: This looks good in general, I'd probably make it a tad smaller though. Maybe 48dp for the image size, 8dp vertical padding and `?attr/textAppearanceBody2` as text appearance?

mobile/src/main/res/drawable/ic_data_usage_black_64dp.xml
```diff
@@ -0,0 +1,9 @@
+<vector
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:height="64dp"
+    android:width="64dp"
+    android:viewportHeight="24.0"
+    android:viewportWidth="24.0"
+    android:tint="#242424">
```
maniac103: Can nuke tint here?

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: I thought about having a global list with ids of whitelisted widgets. It's totally fine to save this list in ram only, so the after the app is stopped, all widgets with heavy data usage are removed from the whitelist.

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
mueller-ma: > We can extract that information from the NetworkInfo or NetworkCapabilities objects and transfer it to Connection from there.

:+1: 

> On the other hand, I'm not sure whether we would want data saver if the user uses an unmetered VPN over a metered mobile connection.

The question is if the VPN counts as unmetered if the "carrier" connection is metered.

mobile/src/main/res/values/strings.xml
mueller-ma: `widget.label` always returns non-null. For Image items without label is returns something like "Image Url [-]".
`%s` can be replaced by the widget type if no label is present: "Image is not loaded to save data".

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
maniac103: > The question is if the VPN counts as unmetered if the "carrier" connection is metered.

Good point. I just checked, indeed that's the case. VPN over Wifi has the NOT_METERED capability set, VPN over cellular does not.

mobile/src/main/res/values/strings.xml

mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.kt
```diff
@@ -197,14 +199,15 @@ class LogActivity : AbstractBaseActivity(), SwipeRefreshLayout.OnRefreshListener
         log
     }
 
-    private fun getDeviceInfo(): String {
+    private fun getDeviceInfo(context: Context): String {
```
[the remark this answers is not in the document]
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
```diff
@@ -32,12 +33,12 @@ class IconResource internal constructor(
     internal val isOh2: Boolean,
     internal val customState: String?
 ) : Parcelable {
-    fun toUrl(context: Context): String {
-        return toUrl(context.getPrefs().getIconFormat())
+    fun toUrl(context: Context, obeyDataSaver: Boolean = true): String {
```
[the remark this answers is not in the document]
mueller-ma: This requires that the caller has to call `Context#isDataSaverActive()` itself. IMO using `obeyDataSaver` simplifies the call.

mobile/src/main/res/drawable/ic_data_usage_black_64dp.xml
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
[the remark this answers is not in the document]
maniac103: Yes, it's shorter to write it like that, but it mixes concerns: data saver state is nothing a model class should care of (IMHO, that is) ... my point was that the API design would be clearer that way.

mobile/src/main/res/layout/widgetlist_data_saver.xml
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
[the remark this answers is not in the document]
mueller-ma: Makes sense, done.

mobile/src/main/res/values/strings.xml
mueller-ma: The server always supplies a label for charts, even if the user hasn't set one. I'll check if the server does so for images, videos and mapviews as well.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: > on rebind, check whether the request can be satisfied easily

How can I do that for videos? They aren't cached.

maniac103: You probably can't. Just unset the flag in that case.

mueller-ma:
I think I have to change the pref from a simple on/off switch to something with three states:

- Off
- Moderate (Don't load videos on metered connections, but otherwise the same as Off)
- On

With Moderate as default and as behavior if the app is whitelisted.

> The user has enabled Data Saver but the app is whitelisted. Apps should still make an effort to limit foreground and background data usage. 

from https://developer.android.com/training/basics/network-ops/data-saver

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
```diff
@@ -307,6 +310,15 @@ fun Context.getHumanReadableErrorMessage(url: String, httpCode: Int, error: Thro
     }
 }
 
+fun Context.isDataSaverActive(connection: Connection): Boolean {
+    val dataSaverPref = getPrefs().getBoolean(Constants.PREFERENCE_DATA_SAVER, false)
+    if ((dataSaverPref && connection.isMetered) || Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
+        return dataSaverPref
```
mueller-ma:
```suggestion
        return dataSaverPref && connection.isMetered
```

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: This is why I prefer a different way to whitelist widgets. Can you have a look at https://github.com/openhab/openhab-android/pull/1796/commits/7a13ecf31a36d60bce11ddaac1f4f057d4b688f9?

maniac103: I'm not sure this approach is meaningful. Assume data saver is on, and you chose to play a video nevertheless. If you scroll down to another widget and scroll up again, do you really want to auto-load the video again, using up your data for a *second* time?

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
mueller-ma: Done

mobile/src/main/res/values/strings.xml
mueller-ma: The widget type is now shown instead of the label.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: Fair enough. I implemented your suggestion: https://github.com/openhab/openhab-android/pull/1796/commits/eceea36e1194124286b80c6de38c4c1f383c9d91
However there's a side effect: Swipe to refresh now clears the data saver state of any image.

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
mueller-ma: > We can extract that information from the NetworkInfo or NetworkCapabilities objects and transfer it to Connection from there.

I started with this, but I don't know where exactly the information should be added to the connection.

mueller-ma: @maniac103 Do you have an idea of how to pass the `isMetered` to the connection? If not, I'd remove the whole `isMetered` and

1. activate data saver on every connection if enabled in settings
2. or remove the pref and just respect the device wide data saver.

mueller-ma: I implemented option 1. The pref turned off is exactly like option 2.

mueller-ma:
Triggering Travis...

Issue #320, opened 2017-06-19
FR Tap on charts to see them in fullscreen


mueller-ma:

mueller-ma:
I have added this functionality for charts and images [here](https://github.com/openhab/openhab-android/commit/ba7874a0c9dc1842ee552c1125016d6dd01da5e9) (based on #850). I'm not sure if it makes sense to add this for image items as they are already displayed in max width.

maniac103:
I wonder whether it wouldn't make sense to open the popup as separate activity which overrides (force enables) rotation and shows landscape images in actual landscape orientation? Ideally with zooming support? :-)

PR #1803, opened 2020-02-01
Add fullscreen activity for charts

mueller-ma:
See #320

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
@@ -16,7 +16,6 @@ package org.openhab.habdroid.model
 import android.content.SharedPreferences
 import android.net.Uri
 import android.os.Parcelable
-import android.view.ViewGroup
 import kotlinx.android.parcel.Parcelize
 import org.json.JSONException
 import org.json.JSONObject
--- a/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
@@ -90,13 +90,15 @@ data class Widget(
         density: Int,
         forcedPeriod: String = period,
         forcedLegend: Boolean? = legend
-    ): String {
+    ): String? {
+        item ?: return null
+
         val actualDensity = density.toFloat() / prefs.getChartScalingFactor()
         val resDivider = if (prefs.shouldRequestHighResChart()) 1 else 2
 
         val chartUrl = Uri.Builder()
             .path("chart")
-            .appendQueryParameter(if (item!!.type === Item.Type.Group) "groups" else "items", item!!.name)
+            .appendQueryParameter(if (item.type === Item.Type.Group) "groups" else "items", item.name)
             .appendQueryParameter("dpi", actualDensity.toInt() / resDivider)
             .appendQueryParameter("period", forcedPeriod)
             .appendQueryParameter("random", random.nextInt())
--- a/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
@@ -98,7 +98,7 @@ data class Widget(
         val chartUrl = Uri.Builder()
             .path("chart")
             .appendQueryParameter(if (item!!.type === Item.Type.Group) "groups" else "items", item!!.name)
-            .appendQueryParameter("dpi", (actualDensity.toInt() / resDivider))
+            .appendQueryParameter("dpi", actualDensity.toInt() / resDivider)
             .appendQueryParameter("period", forcedPeriod)
             .appendQueryParameter("random", random.nextInt())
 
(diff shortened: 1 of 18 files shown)


mueller-ma:
Chart activity should override `doesLockModeRequirePrompt()`, so it's not locked if the app is in kiosk mode.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -75,6 +81,43 @@ data class Widget(
...
+        density: Int,
+        forcedPeriod: String = period,
+        forcedLegend: Boolean? = legend
+    ): String {
+        val actualDensity = density.toFloat() / prefs.getChartScalingFactor()
+        val resDivider = if (prefs.shouldRequestHighResChart()) 1 else 2
+
+        val chartUrl = Uri.Builder()
+            .path("chart")
+            .appendQueryParameter(if (item!!.type === Item.Type.Group) "groups" else "items", item!!.name)
```
maniac103: I don't really like the `!!` operator usage here. Can we fall back to a sane default (null?) if item is null?

mobile/src/main/java/org/openhab/habdroid/ui/ChartActivity.kt
```diff
@@ -0,0 +1,200 @@
...
+    private var showLegend: Boolean = true
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        setContentView(R.layout.activity_chart)
+
+        widget = intent.getParcelableExtra(WIDGET)!!
+        period = widget.period
+        showLegend = widget.legend ?: true
```
maniac103: How about doing `!= false` instead of elvis? Or, even more clear, `widget != null && widget.legend`?

mobile/src/main/java/org/openhab/habdroid/ui/ChartActivity.kt
```diff
@@ -0,0 +1,200 @@
...
+            R.id.period -> true
+            else -> {
+                updatePeriod(item.itemId)
+                onRefresh()
+                true
+            }
+        }
+    }
+
+    private fun updatePeriod(itemId: Int) {
```
maniac103: Probably more elegant like this:
```
private fun periodForMenuItem(itemId: Int) = when (itemId) {
   ...
}

...

period = periodForMenuItem(item.itemId)
```

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -877,6 +853,13 @@ class WidgetAdapter(
         override fun stop() {
             chart.cancelRefresh()
         }
+
+        override fun onClick(v: View?) {
+            v ?: return
```
maniac103: `val context = v?.context ?: return` (and use that below) would be more elegant, I think.

mobile/src/main/res/values/strings.xml
```diff
@@ -254,6 +254,25 @@
...
+    <string name="chart_activity_period_12_hour">12 h</string>
+    <string name="chart_activity_period_day">1 D</string>
+    <string name="chart_activity_period_2_day">2 D</string>
+    <string name="chart_activity_period_3_day">3 D</string>
+    <string name="chart_activity_period_week">1 W</string>
+    <string name="chart_activity_period_2_week">2 W</string>
+    <string name="chart_activity_period_month">1 M</string>
+    <string name="chart_activity_period_2_month">2 M</string>
+    <string name="chart_activity_period_4_month">4 M</string>
+    <string name="chart_activity_period_year">1 Y</string>
```
maniac103: I think the labels should be spelled out: `1 day`, `3 days`, `1 week` etc.

mobile/src/main/java/org/openhab/habdroid/ui/ChartActivity.kt
mueller-ma: `legend` itself can be `null`, so checking for `widget != null` doesn't help here. I find the elvis operator more clear in this case.
And IMO it's ok to crash if the widget is null.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
mueller-ma: Currently `toChartUrl()` isn't called if `widget.item` is null, so a NPE won't occur here. If I return `null` here, I have to add two more null checks (one to each of the caller code).

maniac103: `Currently` being the keyword here. IMHO the additional null checks are better than this implicit assumption.

mueller-ma: Done!

mobile/src/main/res/values/strings.xml

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/ui/ChartActivity.kt
mueller-ma: Done

mueller-ma:
I downloaded the latest ktlint and ran `./ktlint --android applyToIDEA` again, but the Stickler error doesn't disappear.

PR #1814, opened 2020-02-08
Implement new color picker

mueller-ma:
The color picker from the package `com.jaredrummler:colorpicker` has no callback when the color is selected, but when the dialog is closed. It also has no brightness slider.
IMO both is enough to implement a color picker on our own, even when it has no possibility to freely select a color.

The new color picker has a brightness slider. If the user select a color and brightness is 0, it is set to 100. Otherwise the current brightness is kept.

Also see https://community.openhab.org/t/updating-the-color-picker/92356 for user feedback.

I want to have some feedback before continuing. TODO:

- [ ] Remove old code
- [ ] Show indicator which color is currently selected
- [ ] Add an item if the current color is not on of the presets
- [ ] Show ripple if the user clicks on a color
- [ ] Make the padding of the color items clickable
- [ ] Add more white tones


mobile/src/main/java/org/openhab/habdroid/util/ColorPicker.kt
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        this.listener = listener
+
+        val colorGridAdapter = ColorPickerAdapter(context, R.layout.color_picker_item, gridItems, listener)
+        val gridView: GridView = alertLayout.findViewById(R.id.grid)
+        gridView.adapter = colorGridAdapter
+        val resources = context.resources
+        val columnCount = resources.pixelToDp(resources.displayMetrics.widthPixels).toInt() / 100
+        gridView.numColumns = max(columnCount, 2)
```
mueller-ma: This is a workaround. If I use `AUTO_FIT`, the gridview only has two columns, which is too less.

mueller-ma:
Closing for now

PR #1817, opened 2020-02-14
New colorpicker

mueller-ma:
https://community.openhab.org/t/updating-the-color-picker/92356/8


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -989,41 +996,86 @@ class WidgetAdapter(
             return false
         }
 
+        override fun onColorSelected(selectedColor: Int) {
+            Log.e(TAG, "onColorSelected($selectedColor)")
```
mueller-ma: Will be changed to `Log.d()` before merging.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -989,41 +996,86 @@
...
+
+        override fun onStartTrackingTouch(slider: Slider) {
+            // no-op
+        }
+
+        override fun onStopTrackingTouch(slider: Slider) {
+            handleChange(delay = 0)
+        }
+
+        private fun handleChange(newColor: Int = 0, delay: Long = 1000) {
```
maniac103: 1000 seems a bit excessive. Why not keep the previous value of 100?

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -989,41 +996,86 @@
...
+
+                addOnColorChangedListener(this@ColorViewHolder)
+                addOnColorSelectedListener(this@ColorViewHolder)
             }
 
-            val initialColor = boundItem?.state?.asHsv?.toColor()
-            if (initialColor != null) {
-                picker.color = initialColor
+            slider = contentView.findViewById<Slider>(R.id.brightness_slider).apply {
+                boundItem?.state?.asBrightness?.let { it -> value = it.toFloat() }
```
maniac103: `it ->` is redundant I think.

mobile/src/main/res/layout/color_picker_dialog.xml
```diff
@@ -1,28 +1,28 @@
...
     android:orientation="vertical">
 
-    <com.larswerkman.holocolorpicker.ColorPicker
+    <com.flask.colorpicker.ColorPickerView
         android:id="@+id/picker"
         android:layout_width="240dp"
-        android:layout_height="240dp" />
+        android:layout_height="240dp"
+        app:alphaSlider="true"
+        app:density="8"
```
maniac103: From your screenshot, I think the wheel could be a bit larger, and density be increased as well. I'll play with that.

What's the meaning of the slider attributes?

mueller-ma: > What's the meaning of the slider attributes?

I think they can be removed.

maniac103: Suggestion:
- density 8 -> 12
- size 240dp -> match_parent

maniac103:
The only issue I have with this library is that the current color selection during touch is hidden below the finger :-/

mobile/src/main/res/layout/color_picker_dialog.xml
mueller-ma: > density 8 -> 12

IMO 12 is too much. What about 10?

> match_parent

Then the slider disappears. Do you have a working patch?

maniac103: The color wheel needs height=0dp + weight=1. I can post a patch for that tomorrow. With that + density 12 it looks like this:

https://www.dropbox.com/s/6vgsst3895shh4e/Screenshot_20200217-212830_openHAB.png?dl=0

maniac103: Layout diff for that screenshot:
```diff
diff --git a/mobile/src/main/res/layout/color_picker_dialog.xml b/mobile/src/main/res/layout/color_picker_dialog.xml
index 0fb1d862..ef4c3e6c 100644
--- a/mobile/src/main/res/layout/color_picker_dialog.xml
+++ b/mobile/src/main/res/layout/color_picker_dialog.xml
@@ -2,8 +2,8 @@
 <LinearLayout
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
-    android:layout_width="wrap_content"
-    android:layout_height="wrap_content"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
     android:gravity="center_horizontal"
     android:paddingStart="16dp"
     android:paddingEnd="16dp"
@@ -12,15 +12,17 @@
 
     <com.flask.colorpicker.ColorPickerView
         android:id="@+id/picker"
-        android:layout_width="240dp"
-        android:layout_height="240dp"
-        app:density="8"
+        android:layout_width="match_parent"
+        android:layout_height="0dp"
+        android:layout_weight="1"
+        app:density="12"
         app:wheelType="CIRCLE" />
 
     <com.google.android.material.slider.Slider
         android:id="@+id/brightness_slider"
-        android:layout_width="240dp"
+        android:layout_width="match_parent"
         android:layout_height="wrap_content"
+        android:layout_marginTop="8dp"
         android:valueFrom="0"
         android:valueTo="100" />
 </LinearLayout>
```

mueller-ma: Picked

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: I had some issues with the item updates sending too often. I added a commit where the previous item update is now canceled, if a new updates is queued.

mobile/build.gradle
```diff
@@ -142,7 +142,7 @@ dependencies {
     implementation "androidx.constraintlayout:constraintlayout:1.1.3"
     implementation "androidx.preference:preference:1.1.0"
     implementation "androidx.biometric:biometric:1.0.1"
-    implementation "com.google.android.material:material:1.1.0"
+    implementation "com.google.android.material:material:1.2.0-alpha04"
```
mueller-ma: This PR uses `material:1.2.0-alpha04`, so maybe I should wait with merging this.

mobile/src/main/res/layout/color_picker_dialog.xml
mueller-ma: I changed the density to 10. Let's look at the feedback when the code goes live.

maniac103: You really think we'll get feedback for that, looking at the responses in the forum thread? ;-)

If 12 is too dense for you and 10 is too loose for me, we could settle at a compromise value of 11.

mueller-ma: I think we get feedback if the new color picker is in the beta version.

> If 12 is too dense for you and 10 is too loose for me, we could settle at a compromise value of 11.

Isn't 10 already a compromise between 8 and 12? ;)
I really have to see and test this when it's in beta, so I'd keep it at 10 for now.

maniac103: *sigh* OK then :)

PR #1828, opened 2020-02-25
Support bluetooth tethering

### LINKED ISSUE #1800: Support Bluetooth tethering
Currently bluetooth tethering connections are ignored.

Support them and treat them like metered wifi, as we don't know, if the hotspot device is in wifi or not.

### Additional context
https://developer.android.com/reference/kotlin/android/net/NetworkCapabilities#transport_bluetooth

mueller-ma:
Fixes #1800

Tested with two Android 7 devices.


mueller-ma:
Besides that maybe we should try the remote connection if the connection type is unknown.

Issue #1829, opened 2020-02-26
Video widget doesn't work with basic auth


mueller-ma:
### Actual behaviour
The video widget doesn't play videos if the reverse proxy has basic authentication enabled.

### Expected behaviour
It should play the video.

### Can you reproduce the issue in demo mode?

No

#### Client
* Android version: 7
* App version : 2.11.0

#### Server
* Reverse Proxy: nginx
* Authentication method :  Basic auth

### Fix

I tried the following patch, but the media player doesn't send the header I've set to the server.

````
diff --git a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
index 2a632d0f..3846ee7e 100644
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -169,7 +169,7 @@ class WidgetAdapter(
             TYPE_ROLLERSHUTTER -> RollerShutterViewHolder(inflater, parent, connection, colorMapper)
             TYPE_SETPOINT -> SetpointViewHolder(inflater, parent, connection, colorMapper)
             TYPE_CHART -> ChartViewHolder(inflater, parent, chartTheme, connection)
-            TYPE_VIDEO -> VideoViewHolder(inflater, parent)
+            TYPE_VIDEO -> VideoViewHolder(inflater, parent, connection)
             TYPE_WEB -> WebViewHolder(inflater, parent, connection)
             TYPE_COLOR -> ColorViewHolder(inflater, parent, connection, colorMapper)
             TYPE_VIDEO_MJPEG -> MjpegVideoViewHolder(inflater, parent, connection)
@@ -861,7 +861,11 @@ class WidgetAdapter(
         }
     }

-    class VideoViewHolder internal constructor(inflater: LayoutInflater, parent: ViewGroup) :
+    class VideoViewHolder internal constructor(
+        inflater: LayoutInflater,
+        parent: ViewGroup,
+        val connection: Connection
+    ) :
         ViewHolder(inflater, parent, R.layout.widgetlist_videoitem) {
         private val videoView: VideoView = itemView.findViewById(R.id.video)
         private val mediaPlayer = MediaPlayer(parent.context)
@@ -873,7 +877,11 @@ class WidgetAdapter(
[...]

PR #1830, opened 2020-02-27
Preserve authentication information over HTTP redirects.

### LINKED ISSUE #1212: Does not send basic authentication for access to events (reverse proxy)
### Actual behaviour
I am using openHAB 2.4 behind a reverse SSL proxy with basic authenticaion. I can see in the Apache logs that the app does not send the credentials for this URL:

"GET /openhab/rest/sitemaps/events/0280f475-925e-463f-9227-202b094a7bd0?sitemap=default&pageid=default HTTP/1.1" 401 442

This does not happen when using the Chrome browser to access the Basic UI.

### Expected behaviour
The app should always send the credentials (or react properly to the 401).

 
### Steps to reproduce
1. Set up reverse proxy with Apache and Basic Authentication
2. Point the app to that remote URL

No.

#### Client
* Android version: Pie
* Device model: Xiaomi Redmi Note 5
* Stock or customized ROM: Custom, Pixel Experience
* App version : 2.5.0
* Build flavor : Play Store
* Device language: German

#### Server
* Server version: 2.4.0
* Reverse Proxy: Apache
* Authentication method : Basic Auth

#### openHAB Cloud
N/A

#### App log
```
There is no such option!?
```

#### openHAB Server log
```
Nothing in that log.
```

#### Reverse Proxy log
```
Insert your reverse proxy error log here.
For nginx: `/var/log/nginx/error.log`
For Apache 2: `/var/log/apache2/error.log`
```
Nothing in that log.
```
"GET /openhab/rest/sitemaps/events/0280f475-925e-463f-9227-202b094a7bd0?sitemap=default&pageid=default HTTP/1.1" 401 442
```

maniac103:
Fixes #1212

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/HttpClient.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/util/HttpClient.kt
@@ -40,11 +40,26 @@ import java.util.concurrent.TimeUnit
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 
-class HttpClient constructor(private val client: OkHttpClient, baseUrl: String?, username: String?, password: String?) {
+class HttpClient constructor(client: OkHttpClient, baseUrl: String?, username: String?, password: String?) {
+    private val client: OkHttpClient
     private val baseUrl: HttpUrl? = baseUrl?.toHttpUrlOrNull()
     @VisibleForTesting val authHeader: String? = if (!username.isNullOrEmpty() && !password.isNullOrEmpty())
         Credentials.basic(username, password, StandardCharsets.UTF_8) else null
 
+    init {
+        val clientBuilder = client.newBuilder()
+        if (authHeader != null) {
+            // Forcibly put authorization info into request, as redirect/retry interceptor might have removed it
+            clientBuilder.addNetworkInterceptor { chain ->
+                val request = chain.request()
+                    .newBuilder()
+                    .addHeader("Authorization", authHeader)
+                    .build()
+                chain.proceed(request)
+            }
+        }
+        this.client = clientBuilder.build()
+    }
     enum class CachingMode {
         DEFAULT,
         AVOID_CACHE,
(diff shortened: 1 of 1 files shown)


maniac103:
@mueller-ma Can you check with your HTTP proxy setup whether this actually works? I don't have a proxy setup to test with.

mobile/src/main/java/org/openhab/habdroid/util/HttpClient.kt
```diff
@@ -40,20 +40,36 @@ import java.util.concurrent.TimeUnit
...
         DEFAULT,
         AVOID_CACHE,
         FORCE_CACHE_IF_POSSIBLE
     }
 
     fun makeSse(url: HttpUrl, listener: EventSourceListener): EventSource {
-        val request = makeAuthenticatedRequestBuilder()
+        val request = Request.Builder()
             .url(url)
+            .addHeader("User-Agent", "openHAB client for Android")
```
mueller-ma: I'd make constants out of those two strings and "Authorization" above.

maniac103: Is it really that useful? I'm not sure about it. If you think we really should do it, I think I'll change it back to a common method (`makeRequestBuilder`).
For the `Authorization` string, I don't think making it a constant makes sense, as it's only used once.

mueller-ma: "Authorization" is also used in the MainActivity. I'd put it in constants as we need to use these headers for the video widgets as well (#1829) and there we cannot use a common function.

maniac103: I'd keep it as is, following the example of the okhttp internal code ¯\_(ツ)_/¯ Since those header names are well-known names in the HTTP spec, IMHO putting them behind constants makes the code harder to (immediately) understand.

PR #1832, opened 2020-02-27
Restore webview state after rotation

### LINKED ISSUE #1169: HABPanel is reset when device orientation is changed

mueller-ma:
Fixes #1169

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/activity/WebViewFragment.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/activity/WebViewFragment.kt
@@ -162,7 +162,11 @@ class WebViewFragment : Fragment(), ConnectionFactory.UpdateListener {
 
     fun goBack(): Boolean {
         if (webView?.canGoBack() == true) {
+            val oldUrl = webView?.url
             webView?.goBack()
+            if (oldUrl == webView?.url) {
+                goBack()
+            }
             return true
         }
         return false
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/activity/WebViewFragment.kt
```diff
@@ -163,7 +162,11 @@ class WebViewFragment : Fragment(), ConnectionFactory.UpdateListener {
 
     fun goBack(): Boolean {
         if (webView?.canGoBack() == true) {
+            val oldUrl = webView?.url
             webView?.goBack()
+            if (oldUrl == webView?.url) {
+                goBack()
```
maniac103: Isn't this equivalent to returning `webView?.url != oldUrl`? I'd find this more elegant than the recursive call.

mueller-ma: No, it's not the same. If `goBack()` returns `false`, HABPanel is closed.
I added this code to "filter out" redundant entries in the back stack.

maniac103: I see. The recursion still irks me a bit. How about this:
```
val oldUrl = webView?.url
do {
    webView?.goBack()
} while (webView?. url == oldUrl && webView?.canGoBack() == true)
```

And maybe add a comment like `// skip redundant history entries while going back`

mueller-ma: Picked.

PR #1834, opened 2020-02-27
Add 'rate app' to about menu

mueller-ma:
But only for the stable release. The play store is not the recommended way to report issues with the beta version.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/AboutActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/AboutActivity.kt
@@ -39,6 +39,7 @@ import org.openhab.habdroid.model.ServerProperties
 import org.openhab.habdroid.util.HttpClient
 import org.openhab.habdroid.util.ScreenLockMode
 import org.openhab.habdroid.util.Util
+import org.openhab.habdroid.util.openInAppStore
 import org.openhab.habdroid.util.openInBrowser
 import java.text.DateFormat
 import java.text.SimpleDateFormat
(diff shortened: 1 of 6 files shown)


mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
```diff
@@ -309,6 +309,14 @@ fun Context.getHumanReadableErrorMessage(url: String, httpCode: Int, error: Thro
     }
 }
 
+fun Context.openInAppStore(app: String) {
```
maniac103: `openAppInPlayStore()`?

mobile/src/main/java/org/openhab/habdroid/ui/AboutActivity.kt
```diff
@@ -113,6 +114,14 @@ class AboutActivity : AbstractBaseActivity(), FragmentManager.OnBackStackChanged
                 .desc(context.getString(R.string.about_copyright, year))
                 .icon(R.mipmap.icon)
                 .build())
+            if (Util.isFlavorStable) {
```
maniac103: I wonder whether we only should do this in the full-stable flavor, since people probably use the foss flavor because they don't want to use Google's services.

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
mueller-ma: Any app store installed on the device can be opened, not only the play store. E.g. G-Droid (an alternative F-Droid client) supports rating, so it makes sense to offer opening in other app stores as well.

mobile/src/main/java/org/openhab/habdroid/ui/AboutActivity.kt
mueller-ma: As above

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
maniac103: So those other apps also register for the `market` scheme? If so, LGTM then.

mueller-ma: Yes, they do. On my phone I can choose between F-Droid and Play Store.

PR #1836, opened 2020-02-27
Rename Constants to PrefKeys

mueller-ma:
Constants contains only constants that are used as preference keys, except for the two SUBSCREEN_...
IMO the file should be renamed to PrefKeys and the PREFENCE_ prefix can be dropped. I started with that, but want to get some feedback before I continue with the rest.


maniac103:
Makes sense to me.

PR #1848, opened 2020-03-03
Remove newline from value view

### LINKED ISSUE #676: Long value breaks layout
While the client now handles long labels well, long values do break the layout:

27394031_10209075025351550_381319596_o

These are the item definitions for the first two items displayed:

```
String   FD_EMSB_Latest_Title            "EMSB [%s]"                       <emsb>                   {channel="feed:feed:EnglishMTL:latest-title"}

String   FD_EMSB_Latest_Title_LBL        "Just a reminder for parents, students and staff: for inclement weather, we only announce school closures. If there is no specific closure announcement, then all of our schools and centres will be open. "                            <emsb>

```
When we take the long value from the first item and use it as the label for the second item, then this second item looks fine - so the problem seems to be specific to long values.

The item is rendered nicely both on Basic UI and on the iOS client.

mueller-ma:
Fixes #676


mueller-ma:
This should be merged after the next stable release.

PR #1851, opened 2020-03-05
Send battery level to server

### LINKED ISSUE #269: Send device information to openhab
I would like to have a simple way to send information about my device (e.g. gps location, battery status) to openhab. This way it would be easy to have a very good present recognition or to control a charger.

Of course this would need more permissions. See: #244

mueller-ma:
Periodic workers don't have a failed state, so the notification listener
doesn't know if the job has failed or not. Because of that I used a
periodic worker for triggering a one time worker that sends the battery
level.
While charging the frequency is lowered to 10 - 15 minutes, if the
battery isn't low.

Closes #269

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -134,6 +134,30 @@ then
 end
 ```
 
+#### Battery level
+
+The openHAB app will send the current battery level to the server if you enable it in the settings.
+While charging the battery level is sent every 10 to 15 minutes, otherwise every 2 to 6 hours.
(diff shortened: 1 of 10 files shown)


maniac103:
Mostly LGTM. One cleanup suggestion: https://github.com/maniac103/openhab.android/commit/16b323e7075e7c6438015e63d3aca9f6dbecd3fa

mueller-ma:
Currently I'm a bit bussy, but I'll review your commit later this week.

mueller-ma:
Picked your commit, thanks!

mueller-ma:
This should be merged after the next stable release.

Issue #1841, opened 2020-02-28
Unable to scroll sub items in Grafana chart displays


m4rky-m4rk:
openHAB 2.5.2
Galaxy Note 10+ 
Android version 10

I have problems scrolling within Grafana charts when using the app on my phone but not when using a browser to view either Grafana direct or via the via sitemap and basicUI.  

I have seen the problem in several places. For both large chart legends and for the pull down menu for setting the time ranges. I cannot drag the lists to reveal the all the items when using the app.  When viewing charts directly via Grafana a small side bar appears when dragging the lists that does not appear for BasicUI but for both the list does move up a or down when dragged. For the app no sidebar appears and lists cannot be dragged to show the hidden items

![SmartSelect_20200228-100137_openHAB](https://user-images.githubusercontent.com/26285654/75526357-ee77e480-5a11-11ea-87be-d24db8f79ff2.jpg)

![SmartSelect_20200228-100114_openHAB](https://user-images.githubusercontent.com/26285654/75526371-f46dc580-5a11-11ea-9c00-b7b4c7a6a2d1.jpg)

mueller-ma:
Can you post the WebView definition in your sitemap?

m4rky-m4rk:
```
Text label="Temperature charts" icon="line" {
                                        Webview label="Temperature charts" icon="line" url="http://openhab:3000/d/xxxxxx/temperatures?orgId=1&refresh=5m" height=150 // height set for mobile use
                                }
```

mueller-ma:
I think the issue is that the scrolling is consumed by the list view and not passed to the webview. We tried to make the height of webviews auto-adjust, but that didn't work well.
One thing we could do is: If there is only one element on a page, handle it differently, i.e. charts and webviews full height.
For charts it should be quite easy to use the full screen mode, that has been added recently. Auto refresh and voice command menu entry need to be implemented, though.
For webviews we can reuse WebViewFragemt.
@maniac103 WDYT?

mueller-ma:
When a Google Map view is implemented on a website, you can scroll the website with one finger (as usual) and the map view with two fingers. This might be a good fix in our case as well.

PR #1854, opened 2020-03-05
Improve chart activity

mueller-ma:
* Add voice recognition
* Hide widget value in title
* Don't show legend for single items if legend is not set


maniac103:
Legend and title changes make sense to me, but I'm not sure about the voice recognition one. I somewhat doubt we need it in ChartActivity, since that one (IMHO) is a short-lived one that's only shown temporarily. Is there a specific reason for adding that functionality?

mueller-ma:
Yes, I added voice recognition is preparation for https://github.com/openhab/openhab-android/issues/1841#issuecomment-593146432.

maniac103:
For that purpose I wouldn't use ChartActivity as-is anyway though, but rather refactor its logic into a fragment and make the activity a thin wrapper around that fragment.

mueller-ma:
Makes sense. I've removed the voice recognition.

PR #1856, opened 2020-03-06
Run background tasks after boot

mueller-ma:
Just to be sure the server has the latest device information.


mueller-ma:
I don't think this makes much sense for the current two tasks. It also adds a permission, which should be added to the changelog. #1851 and this could be combined to:

> * Send battery level to server. This requires the "Start at boot" permission.

PR #1860, opened 2020-03-09
Simplify and fix API 26+ connection detection logic.

maniac103:
Remove the redundant list of transports sorted by priority. That list
was out of date (Bluetooth was missing), so BT tethering connections
were filtered out before they could reach the when block. Since priority
sorting is done in ConnectionFactory anyway, we don't need the priority
logic anymore anyway.


maniac103:
@mueller-ma Did you test the BT tethering use case on Android 8.0+?

mueller-ma:
I don't have a Android 8+ device, so I cannot test bluetooth tethering. The device I used has Android 6.

PR #1863, opened 2020-03-15
Check if intent is resolvable before starting activity

mueller-ma:


maniac103:
Aren't both approaches pretty much equivalent?

mueller-ma:
Yes, but IMO it's nicer to check for errors instead of catching the exception (if it's easy to do so).

PR #1879, opened 2020-03-23
Fix dynamic icons

### LINKED ISSUE #1874: Dynamic icons do not work
If an item has changed its state the dynamic icon of this item is expected to change its presentation immediately. But it does not in the Android app anymore. If the app's activity is left and got back again  the dynamic icon's presentation is shown as expected. 
_Version 2.11.24-beta_

mueller-ma:
Fixes #1874


maniac103:
Can you keep the intermediate variable? Makes the statements shorter and thus easier to understand.

mueller-ma:
Makes sense.

PR #1888, opened 2020-03-24
Send charging state to server

mueller-ma:

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -186,7 +186,7 @@ then
     } else if (ChargingState.state == "AC") {
         // Device is charging over AC adapter
     } else if (ChargingState.state == "WIRELESS") {
-        // Device is charging wireless
+        // Device is charging wirelessly
     } else if (ChargingState.state == "UNKNOWN_CHARGER") {
         // Device is charging in an unknown way (None of the three above).
         // If you see this state, please report that.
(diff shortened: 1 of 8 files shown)


docs/USAGE.md
```diff
@@ -158,6 +168,35 @@ then
...
+rule "Charging state changed"
+when
+    Item ChargingState changed
+then
+    if (ChargingState.state == "USB") {
+        // Device is charging over USB
+    } else if (ChargingState.state == "AC") {
+        // Device is charging over AC adapter
+    } else if (ChargingState.state == "WIRELESS") {
+        // Device is charging wireless
```
maniac103: Wirelessly

docs/USAGE.md
```diff
@@ -158,6 +168,35 @@ then
...
+    } else if (ChargingState.state == "UNKNOWN_CHARGER") {
+        // Device is charging in an unknown way (None of the three above).
+        // If you see this state, please report that.
+    } else {
+        // Device isn't charging ("UNDEF" is send)
```
maniac103: If that's the case, shouldn't the option/item not rather be named 'ChargerType'?

mueller-ma: It can be useful if you have multiple chargers, but you can control only one with openHAB. My use case is: I have an AC charger connected to a switchable socket. If my smartphone is charging, the charger will be turned off at 90% battery level. However if I charge it at a different charger at home, e.g. transfer files to or from my computer, it makes no sense to turn off the AC charger.

If one is only interested in whether the device is charging, he can check for `ChargingState == UNDEF`.

PR #1893, opened 2020-03-25
Try to avoid OOM on Bitmaps

mueller-ma:
Scale down images on low ram devices to save memory.

### DIFF
--- a/mobile/src/full/java/org/openhab/habdroid/core/FcmMessageListenerService.kt
+++ b/mobile/src/full/java/org/openhab/habdroid/core/FcmMessageListenerService.kt
@@ -138,9 +138,10 @@ class FcmMessageListenerService : FirebaseMessagingService() {
             val connection = ConnectionFactory.cloudConnectionOrNull
             if (connection != null && !applicationContext.isDataSaverActive()) {
                 try {
+                    val size = resources.getDimensionPixelSize(R.dimen.svg_image_default_size)
                     iconBitmap = connection.httpClient
                             .get(icon.toUrl(this, true), timeoutMillis = 1000)
-                            .asBitmap(resources.getDimensionPixelSize(R.dimen.svg_image_default_size), false)
+                            .asBitmap(size, size, false)
                             .response
                 } catch (e: HttpClient.HttpException) {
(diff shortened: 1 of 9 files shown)


maniac103:
Instead of this approach, my suggestion would be being a bit more clever in the bitmap loading logic in `ResponseBody.toBitmap`, in particular

- if no forced size is given, apply the logic from [here](https://developer.android.com/topic/performance/graphics/load-bitmap#load-bitmap) and load the bitmap pre-scaled to roughly match target size
- if that doesn't suffice and we encounter an OOME when loading the bitmap, retry loading once with `inSampleSize` multiplied by e.g. 4

My guess is that most of the time icon loading won't be an issue, even on low RAM devices. I think it's image items (e.g. webcam) that are offenders here. The advantage of the suggested approach over yours would be dealing with the memory constraints in a more dynamic way instead of a 'sledge hammer' approach, which will deteriorate icon image quality even in cases where not strictly needed.

mueller-ma:
Thanks for the suggestions. I'll give it a try.

mueller-ma:
> if no forced size is given, apply the logic from here and load the bitmap pre-scaled to roughly match target size

I've tried this, but now bitmaps aren't shown at all. I feel like the issue is that the input stream isn't reset after the first use, but calling `reset()` doesn't help.

maniac103:
> I feel like the issue is that the input stream isn't reset after the first use

That's correct. Since that stream doesn't support reset(), I think the best idea is to re-fetch it (call byteStream() again). Since the data is buffered, doing the call again should be cheap.

mueller-ma:
> call byteStream() again

This is what's currently in the code. I wanted to use Java's `Object#clone()`, but that doesn't exist in Kotlin.
WDYT about https://github.com/openhab/openhab-android/pull/1893/commits/03e10fc7967c5c80e570cfc8bd59f8dffa5efcd4?

maniac103:
> WDYT about 03e10fc?

This probably works, but may OOM since the full content is loaded before any down sampling that might occur. At least it's loaded in compressed form, though.

I'll have a look what's going on here.

mueller-ma:
Closing this for now.

PR #1894, opened 2020-03-25
Catch OOM in ResponseBody.toBitmap()

mueller-ma:


mueller-ma:
See https://github.com/openhab/openhab-android/pull/1893#issuecomment-604925724

PR #1899, opened 2020-03-26
Show snackbar on SSE subscription failure

mueller-ma:
Basic UI does this as well and it's good to debug some auto refresh issues

### DIFF
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -170,7 +170,7 @@
     <string name="error_openhab_offline">Your openHAB server is offline while the cloud instance is running</string>
     <string name="error_no_browser_found">No browser found on your device</string>
     <string name="error_no_speech_to_text_app_found">No voice recognition app found</string>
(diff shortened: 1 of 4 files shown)


mobile/src/main/res/values/strings.xml
```diff
@@ -170,6 +170,7 @@
     <string name="error_openhab_offline">Your openHAB server is offline while the cloud instance is running</string>
     <string name="error_no_browser_found">No browser found on your device</string>
     <string name="error_no_speech_to_text_app_found">No voice recognition app found</string>
+    <string name="error_sse_failed">Widgets may not refresh automatically (SSE failed)</string>
```
maniac103: How about 'Event subscription failed, widget refresh may be impaired'?

maniac103:
Besides the string, does Basic UI really also fall back to long polling? I thought they only do SSE reconnection, which this snackbar specifically is *not* about?

mueller-ma:
> Besides the string, does Basic UI really also fall back to long polling?

I don't think so. With "Basic UI does this as well" I meant the snackbar that is shown.

PR #1900, opened 2020-03-27
Add support for configuring multiple OH targets

### LINKED ISSUE #10: support multiple openhab server

maniac103:
Only one can be active at a time, though. Selection of targets happens via toggle in the navigation drawer.

Closes #10


maniac103:
This is mostly finished, but the configuration UI needs more work (e.g. confirmation prompts when leaving without saving or deleting servers; also some sort of error indication on incomplete server configurations).
Before progressing, I wanted to collect feedback on the UI aspects of this first.

mobile/src/main/java/org/openhab/habdroid/core/UpdateBroadcastReceiver.kt
```diff
@@ -123,6 +154,7 @@ class UpdateBroadcastReceiver : BroadcastReceiver() {
         private const val SECURE_CREDENTIALS = 190
         private const val DARK_MODE = 200
         private const val WIDGET_ICON = 250
+        private const val MULTI_SERVER_SUPPORT = 274
```
mueller-ma: Can you also bump the versionCode in build.gradle?

mueller-ma:
I few UI suggestions:

- [ ] In demo mode hide the server selector in the drawer.
- [ ] "Show all sitemaps in drawer setting":
If enabled I'm fine with the current behavior. If disabled, I suggest the following menu structure (with servers A and B):
````
Server A (opens default sitemap of server a)
Server B (opens default sitemap of server b)
<Divider>
<Server selector>
HABPanel
Notifications
NFC
<Divider>
Settings
About
````
The default sitemaps of all servers are always directly selectable, but HABPanel, notifications and NFC only from the selected server.

- [ ] If HABPanel, notifications or NFC are open and the user changes the server, the current page should be updated, e.g. show HABPanel from new server.

Things that need to be updated for multi server support:

- [ ] Default sitemap: Do the same than for username, password and url (save on per id basis).
- [ ] Sitemap shortcuts: Save the server id with the shortcut. Users can adjust the label by themselves (if the launcher supports that).
- [ ] Widgets: Save the server id in the widget data
- [ ] Background tasks: Just send them to all servers?
- [ ] NFC tags: That's more tricky as the server id will be different on different devices. Save the local url on the tag (the remote one might not be unique)? If no url is present, use connection 0. If the url doesn't match any configured url, show a toast.

And a small feature request:

- [ ] Add a setting to the local connection called wifi ssid. When the app is opened, the active server is selected based on the wifi ssid, if configured.

**Comment by maniac103** [comment]
> In demo mode hide the server selector in the drawer.

Makes sense.

> "Show all sitemaps in drawer setting":
> If enabled I'm fine with the current behavior. If disabled, I suggest the following menu structure (with servers A and B):

I really don't want to move UI controls based on some (from a user POV) unrelated setting. Besides the (IMHO unnecessary) complication it feels just inconsistent and 'magic' to me.

> If HABPanel, notifications or NFC are open and the user changes the server, the current page should be updated, e.g. show HABPanel from new server.

This seems questionable to me, since I don't think it's clear the user wants to check notifications on server B after he just checked server A. What does Gmail (or other Google app) do in that case?

> Update the title of the HABPanel fragment to HABPanel <server name>? Same for notifications and NFC.

IMHO unneeded and clutter since the current server name is readily available in the drawer.

> Default sitemap: Do the same than for username, password and url (save on per id basis).

Indeed.

> Sitemap shortcuts: Save the server id with the shortcut. Users can adjust the label by themselves (if the launcher supports that).

Right, but what to do with existing ones, and what to do with shortcuts belonging to a server when that server is removed?

> Widgets: Save the server id in the widget data

Same as for shortcuts.

> Background tasks: Just send them to all servers?

Uh, here it starts getting ugly. Automatically sending it to all servers may be a privacy concern, but I don't see how it should work otherwise.

> NFC tags: That's more tricky as the server id will be different on different devices. Save the local url on the tag (the remote one might not be unique)? If no url is present, use connection 0. If the url doesn't match any configured url, show a toast.

Connection 0 should be 'active server', but yeah, makes sense.

> Server discovery on first start: The log says "Service resolved: ....", but the loading indicator keeps spinning. 30 seconds after the first server has been found my second one was found. So maybe let it search for a minute or so and save all found servers?

I would make it stop on the first found server as it's always been the case. If the loading indicator doesn't go away, that's a bug.

> Add a setting to the local connection called wifi ssid. When the app is opened, the active server is selected based on the wifi ssid, if configured.

Maybe in a later PR ... I'd try to keep it simple for now.

**Comment by mueller-ma** [comment]
> I really don't want to move UI controls based on some (from a user POV) unrelated setting. Besides the (IMHO unnecessary) complication it feels just inconsistent and 'magic' to me.

IMO the drawer should always show only one sitemap and IMO if that's the case, there is enough space for the default sitemaps of the other servers.
However there are users don't like this and want to have all their sitemaps shown in the drawer. This is why I added a preference for that (#1554).
I could also live with the current behavior, if you remove the IMO unneeded divider between the sitemap and HABPanel, if only one sitemap is shown.

> Right, but what to do with existing ones, and what to do with shortcuts belonging to a server when that server is removed?

I haven't check yet if it's easily possible to update existing shortcuts and widgets. If so: Add the server id, if not `no id == first server`. If the server is removed, also remove widgets and shortcuts if possible. If not, just keep them and show a toast if pressed.

> Uh, here it starts getting ugly. Automatically sending it to all servers may be a privacy concern, but I don't see how it should work otherwise.

They could be moved to the server preference or simply add a switch to the server prefs with "Send device information to this server" which is enabled by default.

> I would make it stop on the first found server as it's always been the case. If the loading indicator doesn't go away, that's a bug.

Probably better. Two servers at the same wifi aren't that common.

> Maybe in a later PR ... I'd try to keep it simple for now.

Sure.

**Comment by maniac103** [comment]
> I could also live with the current behavior, if you remove the IMO unneeded divider between the sitemap and HABPanel, if only one sitemap is shown.

That we can compromise on :-)

**Comment by maniac103** [comment]
More points:

- [ ] Need to do FCM registration for all servers
- [ ] What server to use for showing registration status?
- [ ] How do we decide which server to go to on notification click?

**Comment by mueller-ma** [comment]
A quite simple way would be to have a primary server and make some features for this server only. Then these features could be made multi server compatible in small patches.
The UX would be quite poor however as it may be difficult to know what's multi server compatible and what not.

> What server to use for showing registration status?

Show it in the server settings?

> How do we decide which server to go to on notification click?

One (summary) notification for each server? Or we show notifications from all servers in the notification fragment. The shown priority could be replaced with the server name.

**Comment by maniac103** [comment]
> One (summary) notification for each server? 

The thing is, I don't think we can detect what server (or rather login) an FCM message came from.

> Or we show notifications from all servers in the notification fragment

Can't meaningfully do that due to the paging.

The more I think about it, and the more features we discover that need adaption, the more I wonder whether all that effort is worth it :-/ I really would be interested how large the part of our user base is who maintains more than one server.

**Comment by mueller-ma** [comment]
> I really would be interested how large the part of our user base is who maintains more than one server.

I use two openHAB server at different locations.

> The more I think about it, and the more features we discover that need adaption, the more I wonder whether all that effort is worth it :-/

Maybe we should really add a pref to set a server as primary and only update a few features, that can be done quite simple (NFC, widgets, shortcuts). If I could view Sitemaps of other servers I would be quite satisfied.
The pref summary could explain what "primary server" means.

**Comment by psmedley** [comment]
I also run openhab in two locations. My primary home and also my holiday house. It would be great to get this finalised and included - my current workaround is to install both the 'openhab' and 'openhab beta' app on my phone.

**Comment by mueller-ma** [comment]
@maniac103 Can I help you with coding for the PR? IMO it would be fine to

1. Finish the UI work on the settings (like you said in the first comment)
2. Add a pref for the primary server with a short explanation (saying that the secondary servers only support viewing the sitemaps) and a link to the docs with a list of features that are multi server ready and those that aren't ready.
3. Mark the multi server support as beta, e.g. by adding "(beta)" to the "Add server" pref
4. Fix default sitemap selection
5. Fix discovery
6. Merge this PR

After that the missing features can be ported to be multi server compatible in separate PRs. For people having multiple servers, but are connected most of the time with their primary one, just viewing sitemaps is good enough.
The "Add server" pref could also be hidden in stable releases.

**Comment by tcgerhard** [comment]
Another dual-home Openhaber here.   A small twist -- the two sites are connected by a site-to-site VPN.   My current workarounds:
1) a few "important" items shadowed from site2 to site1 via MQTT
2) Android app uses the credentials for site1; for site2, use habpanel in Chrome browser on the phone (which is a lousy UX)
I'd be thrilled to have a way to switch credentials easily without manually updating them when switching context.

**Comment by mueller-ma** [comment]
> Add a pref for the primary server with a short explanation (saying that the secondary servers only support viewing the sitemaps) and a link to the docs with a list of features that are multi server ready and those that aren't ready.

Currently `ConnectionFactory` can only handle one server at a time, so it has to be updated to handle at least two servers, primary and active. Example: The app can be open and show the sitemaps of server 2, while some background tasks sends an item update to server 1.

**Comment by maniac103** [comment]
> Currently ConnectionFactory can only handle one server at a time, so it has to be updated to handle at least two servers, primary and active. Example: The app can be open and show the sitemaps of server 2, while some background tasks sends an item update to server 1.

I'll look into that.

**Comment by maniac103** [comment]
>> Currently ConnectionFactory can only handle one server at a time, so it has to be updated to handle at least two servers, primary and active. Example: The app can be open and show the sitemaps of server 2, while some background tasks sends an item update to server 1.

> I'll look into that.

Done. Questions I came across when implementing this:
- Should `AbstractItemPickerActivity` fetch items and icons from active or primary server?
- When opening a cloud notification via clicking it, how do we want to handle that? Change active server to primary one? Just open primary notifications (I've currently done that)? If the latter, do we want to change the title?
- I think we'd need some sort of 'unsaved changes reminder' in the server editor fragment. Either as a clear reminder pref, or as a dialog invoked when going back with unsaved changes.

**Comment by mueller-ma** [comment]
> Should AbstractItemPickerActivity fetch items and icons from active or primary server?

For now the items should be fetched from the primary server. I'm going to adapt `AbstractItemPickerActivity` for multi server support, but keep the possibility to display the items from the primary server only.

> When opening a cloud notification via clicking it, how do we want to handle that? Change active server to primary one? Just open primary notifications (I've currently done that)? If the latter, do we want to change the title?

I wouldn't change the active server, but change the title.

> I think we'd need some sort of 'unsaved changes reminder' in the server editor fragment. Either as a clear reminder pref, or as a dialog invoked when going back with unsaved changes.

Yes, that's definitely needed. I'd add a dialog when trying the exit the fragment with unsaved changes.

File: mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.kt [hunk 2]
```diff
@@ -446,8 +524,8 @@ class ConnectionFactory internal constructor(
             instance.triggerConnectionUpdateIfNeededAndPending()
             val sub = instance.stateChannel.openSubscription()
             do {
-                val (available, reason, cloudInitialized, _) = sub.receive()
```
mueller-ma, line 452: Need to add `PrefKeys.PRIMARY_SERVER_ID` to `UPDATE_TRIGGERING_KEYS`? (I cannot comment on the correct line since it isn't changed in this commit).

mueller-ma:
> I'm going to adapt AbstractItemPickerActivity for multi server support, but keep the possibility to display the items from the primary server only.

https://github.com/mueller-ma/openhab.android/commit/b1f52609f6d57657298994de9f49125316c2dc52

mueller-ma:
I added multi server support to a few features: https://github.com/mueller-ma/openhab.android/commits/multi-server

maniac103:
A few of those commits look like they change the active server as part of a background operation. I don't think that's a good idea - IMHO the active server should only be changed by user's choice. If we want to allow operations for random servers (do we really? I somewhat doubt it, it feels like overkill) we need to add a method in ConnectionFactory for getting a Connection for a given server ID.

mueller-ma:
> If we want to allow operations for random servers (do we really? I somewhat doubt it, it feels like overkill) we need to add a method in ConnectionFactory for getting a Connection for a given server ID.

I think we need that for e.g. the Tasker plugin. Tasker can send item updates at any time to any server, so IMO making a connection to a non-primary and non-active server is required.
The error notification should be extended to contain the server name.

mueller-ma:
What about the preferences that are hidden or shown based on the server flags? IMO this isn't that important to be a blocker for merging this PR.
But we could save the latest server properties in the `ServerConfiguration` and display the pref it at least one server supports that pref.

maniac103:
> Tasker can send item updates at any time to any server

Since Tasker can only do what we allow it to do, that's your personal definition. I'd just a different (and IMO equally valid) one: `Tasker can send item updates at any time to the primary server`.

I think we should think about what we actually want here: Do we want 

1. multi-server support with all bells and whistles, including support for all kinds of background operations or
2. multi-server support for the 'foreground' operations only, with the background operations talking to one specific server only

My thought was that we are aiming at 2), and that the concept of a 'primary' server is meant for precisely the distinction which server background operations talk to. That may (IMHO) the concepts are easy to explain to an average user. You now seem to head towards 1) though, and I'm really not sure whether the additional complications this will bring [1] are really well invested maintainer time [2] considering the feature targets ... maybe 1 or 2% of our user base?

[1] A lot of stuff needs to be re-thought for this, starting at server ID generation (must not reuse them in that case), over ConnectionFactory to doing UI for server selection for each and every feature.
[2] It's not only the initial coding ... we *will* get bug reports for obscure niche use cases we didn't think of in this PR.

> What about the preferences that are hidden or shown based on the server flags? IMO this isn't that important to be a blocker for merging this PR.
> But we could save the latest server properties in the ServerConfiguration and display the pref it at least one server supports that pref.

Then it's not a `ServerConfiguration` anymore. With approach 2) above, we wouldn't need to change anything, since the cached flags match the active server, which seems the right thing to use to me.

In general, my personal opinion is we should take route 2), use the 'active' server for everything 'foreground' related (widget list display, NFC item picker, HABPanel invocation, about activity) and use the 'primary' server for everything background related (Tasker, item uploads, homescreen widgets). Voice commands probably would need to support both, depending on whether they're invoked by icon in activity [use active server] or homescreen widget [use primary server]).

mueller-ma:
I meant the primary server solution as something temporarily, so that this PR gets merged faster and the more complicated features to port can have their own PRs. However I honestly don't see me porting push notifications or send dev info as they are both quite complex.

> use the 'active' server for everything 'foreground' related (widget list display, NFC item picker, HABPanel invocation, about activity)

NFC item picker requires changes in the `ItemUpdateWorker`, like Tasker and homescreen widgets. Just hide it in the drawer if active != primary server?
With #1975 I copied the push/pull notification status to the settings, because it's a preference in the foss flavor. I was thinking that this info doesn't suit on a about page, where the focus is on the creators and maybe getting help. https://github.com/mueller-ma/openhab.android/commit/49533e9d449d962300da45386232f4d67e05cf4a can be changed to set the active server to the server opened in the prefs.

> use the 'primary' server for everything background related (Tasker, item uploads, homescreen widgets)

I'm fine with that.

What about sitemap shortcuts? They don't require changes in the workmanager stuff and are visually connected to one server (at least they are part of a sitemap on one server). The server name could be added to the title if there's more than one server.

maniac103:
> However I honestly don't see me porting push notifications or send dev info as they are both quite complex.

Same here ;-)

> NFC item picker requires changes in the ItemUpdateWorker, like Tasker and homescreen widgets. Just hide it in the drawer if active != primary server?

Sounds good to me.

> With #1975 I copied the push/pull notification status to the settings, because it's a preference in the foss flavor. I was thinking that this info doesn't suit on a about page, where the focus is on the creators and maybe getting help. mueller-ma@49533e9 can be changed to set the active server to the server opened in the prefs.

IMHO we can/should nuke the REST API version from the UI entirely. It's not immediately important or useful to the user anyway. I think it should totally be sufficient to put it in the log.

> What about sitemap shortcuts? They don't require changes in the workmanager stuff and are visually connected to one server (at least they are part of a sitemap on one server). The server name could be added to the title if there's more than one server.

That would be a case where I'd probably be OK with changing the active server automatically (without explicit user choice). What about NFC sitemap links though? Those don't have a server ID at present...
You see, it gets complicated enough with the feature limited version already ;-)

mueller-ma:
> IMHO we can/should nuke the REST API version from the UI entirely. It's not immediately important or useful to the user anyway. I think it should totally be sufficient to put it in the log.

👍 

> That would be a case where I'd probably be OK with changing the active server automatically (without explicit user choice). What about NFC sitemap links though? Those don't have a server ID at present...

For now I would support NFC for the primary server only. You cannot simply save the server id there, but some unique cross-device identifier. Maybe the local url?
My commit https://github.com/mueller-ma/openhab.android/commit/855f88e66372bbbac34c929e70d3a27e3a236391#diff-3423ef95a8fa5c7fd4200ea28f74092bR619 opens the sitemap on the primary server for NFC tags. (There's a bugfix in the next commit: https://github.com/mueller-ma/openhab.android/commit/c6c0a3f812fd862a39272442afad00ccfe0fbb07)

mueller-ma:
I rebased your branch and pushed a few commits with the suggestions from the last few comments. Should be ready for review (and fixing the code style).

mobile/src/main/java/org/openhab/habdroid/ui/preference/BetaPreference.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    override fun onBindViewHolder(holder: PreferenceViewHolder?) {
+        super.onBindViewHolder(holder)
+        if (holder != null) {
+            betaTag = holder.itemView.findViewById(R.id.beta_tag)
+            betaTag?.isGone = !showBetaTag
+        }
+    }
+
+    fun changeBetaTagVisibility(show: Boolean) {
```
maniac103, line 40: Naming suggestion: `setBetaTagVisibility`

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -576,16 +564,262 @@ class PreferencesActivity : AbstractBaseActivity() {
...
+        }
+
+        override fun onOptionsItemSelected(item: MenuItem): Boolean {
+            return when (item.itemId) {
+                R.id.save -> {
+                    saveAndQuit()
+                    true
+                }
+                R.id.delete -> {
+                    AlertDialog.Builder(preferenceManager.context)
```
maniac103, line 614: We should use `DialogFragment`s here (and below), since otherwise the dialog will disappear on device rotation.

maniac103:
@mueller-ma Can you explain the rationale between the server ID constants in ServerConfiguration and the related logic, please? I don't really like those constants, as IMHO they don't belong to the ServerConfiguration class, and AFAICT we can get away just fine without them: https://github.com/maniac103/openhab.android/commit/66253b78bcf605e2979726499eede15eebd493e9

docs/USAGE.md
```diff
@@ -278,6 +278,27 @@ In case of an error the plugin returns an error code.
...
 
+## Multi server support
+
+When adding multiple servers to the app, there's always a primary and an active one.
+The active server is used for foreground operations, e.g. display the Sitemaps, and can be changed in the side menu.
+The primary server is used for all features that don't support multiple servers and can be changed in the settings.
+
+Features, that support multiple servers:
+* Display Sitemaps and HABPanel
+* Voice commands launched from in-app (send to active server) and from widgets (send to primary server)
```
maniac103, line 289: `send` -> `sent`

docs/USAGE.md
```diff
@@ -278,6 +278,27 @@ In case of an error the plugin returns an error code.
 | 11         | The app couldn't establish a connection                                                    |
 | 1000+      | A connection was established, but an error occured. The error code is 1000 + the HTTP code |
 
+## Multi server support
+
+When adding multiple servers to the app, there's always a primary and an active one.
+The active server is used for foreground operations, e.g. display the Sitemaps, and can be changed in the side menu.
+The primary server is used for all features that don't support multiple servers and can be changed in the settings.
```
maniac103, line 285: I'd rather say 'is used for all background operations'.

mueller-ma:
> Can you explain the rationale between the server ID constants in ServerConfiguration and the related logic, please? 

I added them as "open the currently active server when executing this" and as a more meaningful fallback for functions like `Bundle#getInteger()`. With your commit I see that they can be avoided. Picked it 👍

mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
mueller-ma: Can you have a look at my latest commit? It crashes, but I didn't figure out why.

maniac103: You can't pass callbacks like that, but need to use the parent fragment as interface. I'll fix it.

maniac103: Fixed now.

mueller-ma:
I pushed three commits with small improvements. Other than that I'd replace the "At least server name and one connection is required" toast with a snackbar by using the function provided by #2140.
#2140 should be merged before this one and released in 2.16.0. This PR should be for a longer time in beta.

mueller-ma:
RTM from my side.

mobile/src/main/java/org/openhab/habdroid/core/UpdateBroadcastReceiver.kt
mueller-ma: Done

maniac103:
I'll give it a last double check tomorrow.

mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
```diff
@@ -123,8 +123,8 @@ class NotificationHelper constructor(private val context: Context) {
         var iconBitmap: Bitmap? = null
 
         if (message.icon != null) {
-            val connection = ConnectionFactory.cloudConnectionOrNull
-            if (connection != null && context.determineDataUsagePolicy().canDoLargeTransfers) {
+            val connection = ConnectionFactory.primaryCloudConnection?.connection
+            if (connection != null && !context.determineDataUsagePolicy().canDoLargeTransfers) {
```
maniac103, line 127: @mueller-ma Was this change intentional? It looks wrong on first impression.

maniac103 (commented):
I've pushed some additional minor cleanup. Besides the one question, LGTM!

mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
mueller-ma: The `!` is a rebase mistake by myself. Before it was `!context.dataSaverActive()`.

psmedley:
Thanks for all the work on this, looks good to me with my limited testing

cwihne:
Equipped family members in DE and US with individual OH instances, aside of my own installation with >100 things + >1.300 items. Up until now I used released app for remote OHs (reconfiguring connections) + beta app for my system. What a great achievement to have multiple OHs in a single app supported. Thank you.

PR #1903, opened 2020-03-29
Catch IllegalArgumentException if DataSaverStateChangeReceiver isn't …

mueller-ma:
…registered

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
@@ -87,10 +87,8 @@ class WidgetListFragment : Fragment(), WidgetAdapter.ItemClickListener {
     private val suggestedCommandsFactory by lazy {
         SuggestedCommandsFactory(requireContext(), false)
     }
-    private val dataSaverChangeListener: BroadcastReceiver? = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+    private val dataSaverChangeListener by lazy {
         DataSaverStateChangeReceiver()
-    } else {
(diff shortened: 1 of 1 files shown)


mueller-ma:
> doing the API level check

This is needed, because on unsupported api versions, the receiver is null.

> do it in onStop

Thanks! I didn't see that it registered in onStart, but unregistered in onPause.


I moved the registration logic to the receiver itself.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
```diff
@@ -87,10 +87,8 @@ class WidgetListFragment : Fragment(), WidgetAdapter.ItemClickListener {
     private val suggestedCommandsFactory by lazy {
         SuggestedCommandsFactory(requireContext(), false)
     }
-    private val dataSaverChangeListener: BroadcastReceiver? = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+    private val dataSaverChangeListener by lazy {
```
maniac103: Does the 'by lazy' get us anything here, given it'll be constructed unconditionally in onStart? I think the old variant actually was better, just the register call could've been made dependent on null status of the listener.

mueller-ma:
Superseded by #1905

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
mueller-ma: I moved the unregistration to `onStop()` in #1905. The battery saver is available for all api levels the app supports, so no need to check for it in that PR.

PR #1905, opened 2020-03-30
Support battery saver

### LINKED ISSUE #1898: Support battery saver
Currently the app doesn't look for the current battery saver status.

It could do the same than in #1796.

mueller-ma:
Closes #1898

Also closes #1903 as it contains the fix


maniac103:
This is IMHO by far too invasive. Why would we want to stop image loading just because battery saver is active? I think stopping automatic refresh and not auto-playing videos should totally be sufficient.

mueller-ma:
Basically: Less radio usage == less battery usage.
TBH I was going for this approach as I can reuse a lot of code this way.

maniac103:
> Basically: Less radio usage == less battery usage.

When e.g. on Wifi, the additional battery usage for some additional packets is negligible.

> TBH I was going for this approach as I can reuse a lot of code this way.

That may be true, but the behavior doesn't make sense to me. Just because my battery runs below 15% in the evening and battery saver kicks in (automatically), I don't want gimped image loading in the OH app.

mueller-ma:
Closing this for now.

PR #1906, opened 2020-03-30
Clean up data saver related code

maniac103:
I finished this a little late for #1796.

For details, see individual commits.


mueller-ma:
Mostly LGTM, but I disagree with removing the preference. A lot of apps have settings to reduce data usage when actively using the app (e.g. images and videos aren't downloaded/shown by default), but continue working in the background, e.g. receiving messages. If I enable data saver, I'm afraid that some apps stop their background syncs. So I have data saver disabled, but enable the in-app data saving options.

maniac103:
> So I have data saver disabled, but enable the in-app data saving options

What makes you sure those apps won't stop syncing in that case? IOW, what makes you think those apps (got examples for those?) - unlike ours - treat internal option and system wide data saver differently?

The issue I have with that line of argument is that it - when applied to all apps - makes system data saver 100% useless.

mueller-ma:
The Android developer page says
> It is considered good practice for apps to check whether the user has enabled Data Saver and make an effort to limit foreground and background data usage. 

https://developer.android.com/training/basics/network-ops/data-saver

This sounds to me that apps may restrict background syncs. Preferences from other apps clearly says that only images and videos in foreground usage are restricted:
Wikipedia
Screenshot_20200330-232659
Firefox
Screenshot_20200330-232714
Signal
Screenshot_20200330-232851
Screenshot_20200330-233200


The developer page also says

> This ability is especially useful when roaming, near the end of the billing cycle, or for a small prepaid data pack. 

For me data saver is something I'd enable if all apps must use as less data as possible. The settings in the apps are something that degrade the functionality partly, but save a lot of data at long term. Watching YouTube videos in a lower quality is also an example.

> If we're going to keep the pref, we'll have to improve the strings at the very least.

👍

maniac103:
Thanks for those examples. All of them basically distinguish between 'Load everything', 'Only load over Wi-Fi' and 'Don't load' though. I assume that's what you originally had in mind in #1796? If so, I'll give that 3-state pref in place of the current switch a try, as it makes much more sense to me.

mueller-ma:
> I assume that's what you originally had in mind in #1796?

Mostly yes. I don't understand why these applications give you the possibility not to load everything when on an unmetered connection, but I'd be ok with adding this.
I wasn't able to pass the information whether the connection is metered to the connection object, so I removed the "if metered" from the PR.
IMO the best solution would be a three state preference with "Off" (Load everything), "Moderate" (Load charts in lower resolution and images, but without auto refresh) and "Strict" (The current behavior). And it should only apply for metered connections. "Moderate" could also be used for the battery saver.

mueller-ma:
@maniac103 Can you remove https://github.com/openhab/openhab-android/pull/1906/commits/ac34915d5bfddd2eef977856b650d9489559a90f from this PR for now? The crash fixed by https://github.com/openhab/openhab-android/pull/1906/commits/2a85f762e03cf2c5194442defaf5e8151e7ddbd6 occurs quite often and I'd like to fix it. I could also create a new PR with this commit only, if you prefer.

maniac103:
Yeah, I'll do that later today.

maniac103:
Done, reduced that commit to a string improvement.

mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
```diff
@@ -59,13 +79,44 @@ class OpenHabApplication : MultiDexApplication() {
...
+    fun registerDataSaverStateChangedListener(l: OnDataSaverActiveStateChangedListener) {
+        dataSaverListeners.add(l)
+    }
+
+    fun unregisterDataSaverStateChangedListener(l: OnDataSaverActiveStateChangedListener) {
+        dataSaverListeners.remove(l)
+    }
+
+    inner class DataSaverStateChangeReceiver : BroadcastReceiver() {
+        fun isDataSaverEnabled(context: Context): Boolean {
```
mueller-ma, line 103:
```suggestion
        fun isSystemDataSaverEnabled(context: Context): Boolean {
```
IMO this should be renamed to be clear that it's the system data saver, not the app pref.

maniac103:
> IMO the WebViewHolder should be stopped as well when the data saver is enabled.

Since it doesn't implement the `stop()` method, that'd be a no-op.

mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
maniac103: Done.

mueller-ma:
Since we don't know what's inside a webview, it should be stopped by loading "about:blank" in `handleDataSaverChange()`.

mueller-ma:
What do you think about https://github.com/openhab/openhab-android/commit/a3b34ddeadcd103b66f82bfe7d2094f40a038b26? It makes it easier to debug if the data saver hint is shown for a viewholder that doesn't support it.

maniac103:
> Since we don't know what's inside a webview, it should be stopped by loading "about:blank" in handleDataSaverChange().

Discarding already loaded data seems a little much to me.

mueller-ma:
What if the webpage contains a live feed from a webcam and continues using data?

maniac103:
That'd be a problem. Still, I don't like inconsistent UX, and loading about:blank would be inconsistent ... how does one come back from that?
If we want to do that, we'd need to re-show the placeholder for WebViews after enabling data saver. I'm not sure whether it's really worth adding such extra complexity for that edge case though.

mueller-ma:
> I'm not sure whether it's really worth adding such extra complexity for that edge case though.

Since the placeholder is shown after rebinding the view, e.g. through scrolling in, I'm fine with the current behavior.

PR #1907, opened 2020-03-30
Handle video player errors

mueller-ma:
Show an error message and retry button instead of showing nothing. Also add a loading indicator.


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -985,6 +1008,12 @@ class WidgetAdapter(
             }
             return widget.url
         }
+
+        override fun onClick(v: View?) {
+            boundWidget?.let {
+                loadWidget(boundWidget!!)
```
maniac103: it

mueller-ma: 🤦‍♂ Thanks. I found this issue by myself, but you were one minute faster :)

PR #1917, opened 2020-04-01
Use GitHub Action for emulator tests

mueller-ma:


mueller-ma:
GitHub doesn't offer Linux hardware accelerated VMs.

PR #1919, opened 2020-04-01
Run periodic work on application start

mueller-ma:
Especially for testing it's useful to force the app to do periodic background work. This is a nice way to do so and device and radio are already awake.


mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
```diff
@@ -59,6 +59,7 @@ class OpenHabApplication : MultiDexApplication() {
         ConnectionFactory.initialize(this)
         BackgroundTasksManager.initialize(this)
         RemoteLog.initialize(this)
+        BackgroundTasksManager.triggerPeriodicWork(this)
```
mueller-ma, line 62: This is the wrong place. Periodic tasks may trigger themselves over and over again.

PR #1928, opened 2020-04-04
Update USAGE.md

Trysupe:
The "k" has to be removed from the Code. Otherwise the OpenHAB Extension throws the error

"Type mismatch: cannot convert from (Object)=>void to Procedure0(org.eclipse.xtext.xbase.validation.IssueCodes.incompatible_types)"

With the k removed everything works properly

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -113,7 +113,7 @@ then
         } else {
             logInfo("alarm", "New Alarm")
             timerAlarm = createTimer(epoch,
-                [ k |
+                [ |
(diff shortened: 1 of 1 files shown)


mueller-ma:
Thanks for your contribution. Actually the example rule is out of date. The item should be `DateTime`, not `Number`.
Do you want to have a look at this and update the rule? Here's a rule you can use: https://community.openhab.org/t/alarmclock-doesn-t-fire/87202/3

mueller-ma:
Thanks for your PR.

PR #1941, opened 2020-04-09
Fix push status in About

### LINKED ISSUE #1939: Wrong push notification status is shown if no remote server
### Actual behaviour
Wrong push notification status is shown if no remote server is configured.

### Expected behaviour
It should say "No remote server configured"

 
#### Client
* Android version: All
* Device model: All
* App version : 2.12.9-beta
* Build flavor : Full
* Device language: All

#### openHAB Cloud
* Self hosted cloud or myopenhab.org: None

mueller-ma:
* Check status again after connection change
* Show correct message if no remote server is configured

Fixes #1939

I've tested:
* FOSS flavor
* Play Service issue
* Invalid remote credentials
* Push notifications are working
* Registration is progress


mobile/src/foss/java/org/openhab/habdroid/core/CloudMessagingHelper.kt
```diff
@@ -15,25 +15,23 @@ package org.openhab.habdroid.core
...
     val isSupported: Boolean
         get() = false
 
     fun onConnectionUpdated(context: Context, connection: CloudConnection?) {}
 
     fun onNotificationSelected(context: Context, intent: Intent) {}
 
-    fun getPushNotificationStatus(context: Context): String {
-        return context.getString(R.string.info_openhab_notification_status_unavailable)
+    fun getPushNotificationStatus(context: Context): Pair<String, Int> {
```
maniac103, line 31: Maybe use a `data class` for that return value ... advantage over `Pair` would be meaningful naming of the fields.

mueller-ma: Done

maniac103:
Forgot to `git add` the new class file? At least I can't see it anywhere...

BTW, why not define it in the CloudMessagingHelper class itself? It's a one liner after all.

mueller-ma:
I defined it at the end of `AboutActivity.kt`. In `CloudMessagingHelper` I would need to add it to both flavors.

PR #1943, opened 2020-04-11
Define gravity of loading indicator in xml

mueller-ma:


mueller-ma:
Maybe the person who added this comment did only try `gravity` and not `layout_gravity`.

Issue #1938, opened 2020-04-08
Title of page incorrect on tablets (double-views), correct on phones (single views)


tobiasisenberg:
### Actual behaviour
Describe us what happens
When looking at a sitemap on a phone (i.e., the chosen level of the sitemap fills the entire screen), the shown title of the level matches that of the selected level. See screenshots looking at level 0, 1, and 2 of a test sitemap:
[phone top level](https://drive.google.com/file/d/1XLQ7FmAXw4mVT3YRev0ObNh4l_mmjfQk/view?usp=sharing)
[phone first level](https://drive.google.com/file/d/1XPCKrZbAqf2ogxcVjvdcSdD7BjHwekSu/view?usp=sharing)
[phone second level](https://drive.google.com/file/d/16N0Q6HFKhtBRdLWPbal67-64cGZVr5e5/view?usp=sharing)

When using the same sitemap on a tablet that shows a split view with a detailed level and the next-higher level on the left, the title is incorrect. It is the title/label of the next-higher level on the left that is only the context, yet as the focus is the level on the right it should be the title/label of that level that is shown. Examples:
[tablet top level](https://drive.google.com/file/d/1CvLEx3epuyyWimaxK7jdk_2Q1F6h1Et1/view?usp=sharing)
[tablet first level](https://drive.google.com/file/d/1BMljWiJ7Ew0dxq6eRmyfcIvOVUrYB6un/view?usp=sharing)
[tablet second level](https://drive.google.com/file/d/1Ko7hI9oKm-fXi_YLf68_XUduZVVwYl7i/view?usp=sharing)

Here, in "tablet top level" all is good. But then I selected the "First level item 1" and this is shown in "tablet first level" but the title is still the one of the top level, despite the fact that we now focus on the first level. Next, we click on "Second level item 1" shown in "tablet second level" and we see the second level in focus on the right side, yet the title is that of the first level.

### Expected behaviour
The behavior for tablets should be the same as for phones, the correct title of the level that one focuses on should be shown.

 
### Steps to reproduce
1. Use example sitemap below.
2. Observe using phone and tablet.

```
sitemap testmap label="Top level" {
    Text label="First level item 1" {
        Text label="Second level item 1" {
            Text label="Third level item 1"
            Text label="Third level item 2"
            Text label="Third level item 3"
        }
        Text label="Second level item 2" {
            Text label="Third level item 1"
            Text label="Third level item 2"
            Text label="Third level item 3"
        }
[...]

PR #1944, opened 2020-04-11
Update list selector

mueller-ma:
See #1938


maniac103:
Continuing from https://github.com/openhab/openhab-android/issues/1938#issuecomment-612355017 ... why not semi transparent colorSecondary? You invalidated the one concern about that by yourself ;-)

I can't find much guidance about list selection in Material Design docs, but [this page](https://material.io/design/interaction/selection.html#) uses secondary color + alpha, and the default theme uses secondary color as well...

mueller-ma:
> Continuing from #1938 (comment) ... why not semi transparent colorSecondary?

I changed it to pink for the basic ui theme. In the orange theme it makes no difference and in the grey theme the secondary color is different in day and night mode. And having white text on a white background is quite hard to read.

maniac103:
> And having white text on a white background is quite hard to read

Sure, text still needs to be legible ;-)

Did you miss the 12% opacity comment or deliberately choose not to use that value?

mueller-ma:
> Did you miss the 12% opacity comment or deliberately choose not to use that value?

I missed your edit.
How do I calculate "12%"? `FF` is 100% => 255 * 0.12 = 30,6 => 12% ~ `1E`? Using `1E` as alpha produces colors that are too light IMO.

maniac103:
Yeah, that calculation is correct. I'll play round with this.

maniac103:
I think I agree that 12% is way too low. Let's interpret the MD spec creatively and use the 36% opacity (`#5c...`) originally intended for image activation highlight ... that value seems to provide sufficient visibility to me, while still not being too stark.

BTW, your current selector drawable does not work for me on API 28 emulator. I needed this diff:
```diff
diff --git a/mobile/src/main/res/drawable/list_selector.xml b/mobile/src/main/res/drawable/list_selector.xml
index 7226eb81..1f71e190 100644
--- a/mobile/src/main/res/drawable/list_selector.xml
+++ b/mobile/src/main/res/drawable/list_selector.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <selector xmlns:android="http://schemas.android.com/apk/res/android">
-    <item android:state_activated="true" android:drawable="?attr/selectedWidgetBackgroundColor" />
+    <item android:state_activated="true">
+        <color android:color="?attr/selectedWidgetBackgroundColor" />
+    </item>
     <item>
         <ripple android:color="?attr/colorControlHighlight" />
     </item>
```

The problem probably is that `android:drawable` is only resolved as drawable, not as attribute.

mueller-ma:
I picked your diff and updated the colors. RTM now.

Issue #111, opened 2015-05-27
Remove spell correction from openHAB setting input fields


dominicdesu:
Currently my android wants to do spell correction on the openHAB URL (in settings for both for remote and normal URL) as well as the user name. If I am not careful, it even auto-corrects it. This is not necessary neither for a URL nor for the username and should be removed.

belovictor:
This was completely unexpected and I had to switch off autocorrection on my phone. I'm currently looking into this :-)

PR #1945, opened 2020-04-11
Add amoled/black theme

mueller-ma:

### DIFF
--- a/mobile/src/main/res/values/themes.xml
+++ b/mobile/src/main/res/values/themes.xml
@@ -37,27 +37,19 @@
         <item name="selectedWidgetBackgroundColor">#5c455A64</item>
     </style>
 
-    <style name="openHAB.Amoled" parent="openHAB.DayNight">
+    <style name="openHAB.Amoled.orange" parent="openHAB.DayNight.orange">
         <item name="chartTheme">black</item>
         <item name="android:windowBackground">@color/black</item>
     </style>
 
-    <style name="openHAB.Amoled.orange" parent="openHAB.Amoled">
-        <item name="colorPrimary">@color/openhab_orange</item>
-        <item name="colorSecondary">@color/openhab_orange</item>
-        <item name="colorPrimaryDark">@color/openhab_orange_dark</item>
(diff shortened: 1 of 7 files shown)


mobile/src/main/res/values/themes.xml
```diff
@@ -34,6 +34,29 @@
         <item name="colorPrimaryDark">@color/blue_grey_900</item>
     </style>
 
+    <style name="openHAB.Amoled" parent="openHAB.DayNight">
+        <item name="chartTheme">black</item>
+        <item name="android:windowBackground">@color/black</item>
```
mueller-ma, line 39: I'm not sure if changing the charts and background from dark grey to black is enough. I'll ask in the forum for feedback.

maniac103:
What's the reasoning for wanting to add this?
I still see zero point whatsoever in an 'Amoled' theme. People think it provides battery savings, but those savings aren't even worth thinking about, at least if you aren't staring at the OH app all day, which seems highly unlikely. IMHO, the light/dark theme toggle is totally sufficient.

mueller-ma:
A user requested the theme [here](https://community.openhab.org/t/android-app-beta/41382/155) and two users via Play Store review. For wall mounted tablets there is some battery saving.

maniac103:
Is assuming wall mounted tablets operate on battery instead of AC a reasonable assumption?

mueller-ma:
Yes, it is. But saving energy isn't that bad, right? And even if it doesn't save any energy, it looks better if the display is of bad quality.
I won't use it, but IMO an additional entry in the list doesn't hurt. Let's see what feedback we get from the user who requested the theme.

mobile/src/main/res/values/themes.xml
maniac103: I think you'll want to at least change the card background for frames. I think it's `#111` by default.

mueller-ma: I'm not sure what you mean. The app doesn't use cards for frames, only basic ui does this.

maniac103: Never mind, I thought we are using cards for the frames.

Issue #1912, opened 2020-03-31
Title of color picker contains state from label


mueller-ma:
### Actual behaviour
Title of color picker contains state from label

### Expected behaviour
Title of color picker doesn't contain state from label

`label` should be changed to `labelWithoutState`: https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt#L1129

Issue #1248, opened 2019-03-07
Value in page title is missing


lolodomo:
### Actual behaviour
The title of any page is a fixed text defined in the sitemap as label.

### Expected behaviour
The title of any page should include, when defined in the sitemap, an additional (dynamic) value.
This is already correctly handled by Basic UI.

![image](https://user-images.githubusercontent.com/10186704/53936721-27a1a100-40ab-11e9-9c4a-735e455395bf.png)

What I am talking about is the text "légère pluie" in the page title.

Example of sitemap to reproduce that:
```
Text item=OWMCondition label="Météo" icon="rain" {
}
```

Note that the value is already updated through SSE by the OH server.

maniac103:
Filtering out the value of page titles has been around 'forever': https://github.com/openhab/openhab-android/commit/a58245c098c647f4d74429f11c59dd2f708c7131#diff-3f4712530de18c074e8f711b4e38483eR118 ... one could change that, reformatting the title like BasicUI does (remove [ and ]).

Updating the title seems to be done by sending SSE events with widgetId == pageId.

lolodomo:
Yes, I am the one who implemented this SSE event for title on server side. So I can confirm.

PR #1946, opened 2020-04-12
Only show state in label where explicitly wanted

mueller-ma:
I noticed the state may appear in the data saver hint for charts with a single item. Since this has been an issue in the past (#1912), I decided to remove the state completely from `label` and only show it when we really want to show it.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
@@ -36,7 +36,7 @@ import kotlin.math.max
 data class Widget(
     val id: String,
     val parentId: String?,
-    val label: String,
+    private val rawLabel: String,
     val icon: IconResource?,
     val state: ParsedState?,
     val type: Type,
@@ -59,8 +59,8 @@ data class Widget(
     val height: Int,
     val visibility: Boolean
 ) : Parcelable {
-    val labelWithoutState get() = label.split("[", "]")[0]
-    val stateFromLabel: String? get() = label.split("[", "]").getOrNull(1)
+    val label get() = rawLabel.split("[", "]")[0]
+    val stateFromLabel: String? get() = rawLabel.split("[", "]").getOrNull(1)
 
     val mappingsOrItemOptions get() =
         if (mappings.isEmpty() && item?.options != null) item.options else mappings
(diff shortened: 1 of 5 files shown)


mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -131,6 +129,7 @@ data class Widget(
             val icon = iconName.toOH2WidgetIconResource(item, source.type, source.mappings.isNotEmpty())
             return Widget(source.id, source.parentId,
                 eventPayload.optString("label", source.label),
+                eventPayload.optStringOrFallback("label", source.stateFromLabel),
```
mueller-ma, line 132: Need to split the string here.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -257,7 +257,8 @@ fun JSONObject.collectWidgets(parent: Widget?): List<Widget> {
     val widget = Widget(
         getString("widgetId"),
         parent?.id,
-        optString("label", ""),
+        optString("label", "").split("[", "]")[0],
+        optString("label", "").split("[", "]").getOrNull(1),
```
maniac103, line 261: Splitting twice seems unneeded (here and above), I think using an intermediate variable would be better.

maniac103:
This kills off the state from the fragment title (right?), might want to check old issue reports about that (I think I recall a report asking for state in title).

mueller-ma:
I found that issue: #1248
The value is still shown in the title, because it comes [directly from JSON](https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.kt#L405) and the [brackets are just stripped](https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt#L371).

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
mueller-ma: I think it would be better to save the raw label as is in a private var and make it accessible via two functions (like it was before).

mueller-ma: > I think it would be better to save the raw label as is in a private var and make it accessible via two functions (like it was before).


Done

PR #1950, opened 2020-04-14
Improve text readability in dialogs

### LINKED ISSUE #1947: Difficult to see some text when using dark theme
Maybe more of a bug than a feature, but it's very minor. I use a dark them on Android and the dark blue is hard to see on dark grey, especially on a small screen and old eyes :smile:...

Color wheel
Settings

mueller-ma:
Fixes #1947

Also changes background color and use round corners.

### DIFF
--- a/mobile/src/main/res/values/styles.xml
+++ b/mobile/src/main/res/values/styles.xml
@@ -11,4 +11,23 @@
     <style name="WhiteText">
         <item name="android:textColor">#ffffff</item>
     </style>
+
+    <style name="openHAB.Dialog.Base" parent="@style/Theme.MaterialComponents.DayNight.Dialog.Alert">
+        <item name="dialogCornerRadius">16dp</item>
+    </style>
+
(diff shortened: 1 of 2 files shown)


mobile/src/main/res/values/styles.xml
```diff
@@ -11,4 +11,23 @@
     <style name="WhiteText">
         <item name="android:textColor">#ffffff</item>
     </style>
+
+    <style name="openHAB.Dialog.Base" parent="@style/Theme.MaterialComponents.DayNight.Dialog.Alert">
+        <item name="dialogCornerRadius">16dp</item>
```
maniac103, line 16: Where does that 16dp figure come from? IMHO it's a bit much, 8dp would be the value I'd choose.

Other than that, LGTM.

mueller-ma: It comes from https://material.io/components/dialogs#theming 
I can have a look how to looks with 8dp.

mueller-ma: 8dp looks also ok, I reduced the corner size.

PR #1955, opened 2020-04-16
Don't use our own function for hiding prefs

mueller-ma:


mueller-ma:
I discovered this function when looking at the docs for a different reason. No idea when it was added.

PR #1957, opened 2020-04-16
Remove unneeded checks for api 21

mueller-ma:
Minimum api is 21, so no need to check for it.


mueller-ma:
Trigger Stickler

PR #1970, opened 2020-04-22
Alarm clock: Send UNDEF if no alarm is set

### LINKED ISSUE #1964: Alarm clock is set to 0 instead of UNDEF in some cases
[Here's a fallback](https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt#L391) to `0`, which should be changed to fall back to `UNDEF`.

mueller-ma:
Fixes #1964

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
@@ -388,7 +388,7 @@ class BackgroundTasksManager : BroadcastReceiver() {
                 var time: String? = if (sender in IGNORED_PACKAGES_FOR_ALARM) {
                     "UNDEF"
                 } else {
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -388,7 +388,7 @@ class BackgroundTasksManager : BroadcastReceiver() {
                 var time: String? = if (sender in IGNORED_PACKAGES_FOR_ALARM) {
                     "UNDEF"
                 } else {
-                    (info?.triggerTime ?: 0).toString()
+                    (info?.triggerTime ?: "UNDEF").toString()
```
maniac103, line 391: `info?.triggerTime?.toString() ?: "UNDEF"` would be a bit more readable, I think.

PR #1971, opened 2020-04-22
Use same animation when opening prefs

### LINKED ISSUE #1965: Different activity opening animations for settings
### Actual behaviour
When the prefs activity is launched from the drawer, it slides in from the bottom. If launched from `MissingConfigurationFragment`, it slides in from the right. In latter case the status bar is orange for a short period of time and then changes to the correctly themed color.

### Expected behaviour
It should have the same animation.

#### Client
* Android version: 9

mueller-ma:
Fixes #1965

@maniac103 Do you remember why you chose `TaskStackBuilder` over a simple `startActivity()`?

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -43,13 +43,8 @@
             android:label="@string/app_preferences_name">
             <intent-filter>
                 <action android:name="android.intent.action.APPLICATION_PREFERENCES" />
-
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
-
-            <meta-data
-                android:name="android.support.PARENT_ACTIVITY"
-                android:value="org.openhab.habdroid.ui.MainActivity" />
         </activity>
         <activity
             android:name=".ui.AboutActivity"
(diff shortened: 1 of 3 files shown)


maniac103:
> Do you remember why you chose TaskStackBuilder over a simple startActivity()?

I didn't introduce that. It was introduced in #562, specifically https://github.com/openhab/openhab-android/pull/562/commits/1a594e15d7a902199d2db1fb1aa5132bc03f154d
I don't quite get the logic behind that, though, since the prefs activity is only invoked from the main activity, not from outside the app ... it would probably have been sufficient to not use the `FLAG_ACTIVITY_CLEAR_TASK` flag and to not finish.
So I agree one can simply use `startActivity()` here; might want to replace the parent activity metadata attribute in the manifest when removing this though.

mueller-ma:
> the prefs activity is only invoked from the main activity, not from outside the app

It may also be invoked by the settings app of the system.

maniac103:
Good point, but the settings app doesn't use TaskStackBuilder either ... the main activity (IMHO correctly) isn't put in the back stack when invoked from there, I just tested.

mueller-ma:
`NavUtils.navigateUpFromSameTask()` uses the metadata from the manifest, so I replaced it.

PR #1975, opened 2020-04-24
Notifications for foss flavor

mueller-ma:
Now the app may already run a worker on a regular basis, so checking for
notifications then is quite cheap.


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -325,6 +326,9 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         // Handle menu items
         return when (item.itemId) {
             R.id.mainmenu_voice_recognition -> {
+                launch {
```
mueller-ma, line 329: For debugging, should be removed before merging.

mueller-ma:
apk for testing:
[mobile-foss-beta-debug.zip](https://github.com/openhab/openhab-android/files/4529130/mobile-foss-beta-debug.zip)


- Rename it to .apk
- Install my apk

assets/store_descriptions/en-US/strings.xml
```diff
@@ -17,7 +17,7 @@
     <important_note>Important note</important_note>
     <oh_server>You need an openHAB server for this app.</oh_server>
     <short_description>Vendor and technology agnostic open source home automation</short_description>
-    <fdroid>The builds on F-Droid have FCM and crash reporting removed and will not be able to receive push notifications from openHAB Cloud.</fdroid>
+    <fdroid>The builds on F-Droid have FCM and crash reporting removed and therefore won't be able to receive push notifications from openHAB Cloud in real time, but delayed.</fdroid>
```
maniac103, line 20: The old text was more correct, since polled notifications aren't push notifications by definition. I'd add some sentence like `Instead, notifications are polled on a regular basis in these builds.` instead.

mobile/src/foss/java/org/openhab/habdroid/core/MessageListenerService.kt
```diff
@@ -0,0 +1,121 @@
...
+                Log.d(TAG, "Notification response could not be parsed", e)
+                null
+            } catch (e: HttpClient.HttpException) {
+                Log.e(TAG, "Notifications request failure", e)
+                null
+            } ?: return
+
+            val prefs = context.getPrefs()
+
+            val lastShownMessage = prefs.getString(PrefKeys.FOSS_LAST_SHOWN_MESSAGE, "-1")!!.toBigInteger(16)
```
maniac103, line 64: I don't think messing around with `BigInteger` is actually needed. I think it would be simpler and more logical to find the index of the last seen notification ID in the received list (using indexOf) and only use notifications up to that index using `subList()`.
I'd also suggest keeping the pref key name constant internal to this class, as it's not used anywhere else.
Basically this:
```
val lastSeenMessageId = prefs.getString(LAST_SEEN_MESSAGE_KEY, null)
prefs.edit {
    val newestSeenId = messages.firstOrNull()?.id ?: lastSeenMessageId
    putString(LAST_SEEN_MESSAGE_KEY, newestSeenId)
}
if (lastSeenMessageId == null) {
    // Never checked for notifications before
    return
}
val lastSeenIndex = messages.indexOf(lastSeenMessageId)
val newMessages = if(lastSeenIndex >= 0) messages.subList(0, lastSeenIndex) else messages
if (newMessages.isEmpty()) {
    // out
}
// normal handling
```

mobile/src/foss/java/org/openhab/habdroid/core/MessageListenerService.kt
```diff
@@ -0,0 +1,121 @@
...
+                nm.notify(notificationId, n)
+
+                if (HAS_GROUPING_SUPPORT) {
+                    val count = getGcmNotificationCount(nm.activeNotifications)
+                    nm.notify(SUMMARY_NOTIFICATION_ID, makeSummaryNotification(context, count, timestamp, null))
+                }
+            }
+        }
+
+        fun isRequired(): Boolean {
```
maniac103, line 115: I'd rather name this `usesPolling()` or `requiresPolling()`.

mobile/src/foss/res/values/strings.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="MissingTranslation">
-    <string name="info_openhab_notification_status_unavailable">Not available in the FOSS version</string>
+    <string name="push_notification_status_impaired">Notifications are impaired in the FOSS version. Click for more information.</string>
```
maniac103, line 3: `...have limited functionality in the FOSS version.`?

mobile/src/main/java/org/openhab/habdroid/util/NotificationUtils.kt
```diff
@@ -0,0 +1,192 @@
...
+import android.service.notification.StatusBarNotification
+import androidx.core.app.NotificationCompat
+import androidx.core.content.ContextCompat
+import org.openhab.habdroid.R
+import org.openhab.habdroid.background.NotificationUpdateObserver
+import org.openhab.habdroid.core.connection.ConnectionFactory
+import org.openhab.habdroid.model.IconResource
+import org.openhab.habdroid.ui.MainActivity
+
+object NotificationUtils {
```
maniac103, line 34: I really don't like having utility classes in Kotlin. I'll think about a better approach, probably by using a base class or composition.

maniac103: @mueller-ma This is my take on organizing the shared code, please have a look and let me know about your thoughts: https://github.com/maniac103/openhab.android/commit/96fa360b9e36fe848691bd72ed78475fdd53311b

maniac103:
I've added fixes for my inline comments to my branch as well, for your convenience: https://github.com/maniac103/openhab.android/commits/foss-notifications
Open question is the pref for this feature, other than that, I think this should be fine now.

mueller-ma:
I'll review your branch soon.

mueller-ma:
@maniac103 I picked your commits and added a preference. The only issue is, that the pref summary isn't updated right after the pref is changed.

mueller-ma:
What about a new category in the prefs called "Notifications" with the following prefs:

- Status
- NotificationPollingPreference (only foss)
- Ringtone (only Android 7 or prior)
- Vibration (only Android 7 or prior)
- Ringtone & Vibration (only Android 8 or higher)

The status pref shows the same status than under About and contains a help icon leading to https://www.openhab.org/docs/configuration/actions.html#cloud-notification-actions.

mueller-ma:
> The only issue is, that the pref summary isn't updated right after the pref is changed.

That's because I used an `OnPreferenceClickListener` and not `OnPreferenceChangeListener` 🤦‍♂️

PR #1983, opened 2020-04-29
Replace item prefix with device identifier

### LINKED ISSUE #1692: NFC: Ability to send value based on reading device
If I want to have a NFC tag, that every person reads before leaving the house, I need several tags. One for each person.

Instead I'd like to have one tag that sends a value based on the reading device to an item. This item can be used as "proxy" item that triggers a rule and sets the individual items.
The state could be the host name, a mac address or the item prefix from the background tasks.

Add a flag to the nfc tag that instructs the reading device to prefix the item, like it's done for the background tasks.

### LINKED ISSUE #1951: add (device) prefix to voicecommand
This is a feature request. I have installed the app on several smartphones, but in openhab I cannot differentiate from which device the voicecommand was sent.

**It would be great to have an option to add a prefix to the voicecomand.** 
Then in openhab rule, it could be checked whether the voice command begins with the prefix. Based on that information it would be possible to execute different actions depending on the user.

_This would make sense especially in a multi-family house with several floors, so that every user could control the television on his floor without explicitly mentioning the location of the television (on the ground floor)._

mueller-ma:
Rename the existing prefix of "Send device info" to "Device identifier"
and make it available for NFC tags and voice command.

Closes #1692
Closes #1951

For NFC tags a new query parameter is added (`d=true`) if the device
identifier should be send instead of the state on the tag. Older clients
will send "UNSUPPORTED" instead.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/VoiceService.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/VoiceService.kt
@@ -23,6 +23,8 @@ import org.openhab.habdroid.core.connection.ConnectionFactory
 import org.openhab.habdroid.core.connection.exception.ConnectionException
 import org.openhab.habdroid.util.HttpClient
 import org.openhab.habdroid.util.ToastType
+import org.openhab.habdroid.util.getPrefixForVoice
+import org.openhab.habdroid.util.getPrefs
 import org.openhab.habdroid.util.showToast
 import java.util.Locale
 
@@ -31,11 +33,17 @@ import java.util.Locale
  */
 class VoiceService : IntentService("VoiceService") {
     override fun onHandleIntent(intent: Intent?) {
-        val voiceCommand = intent?.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS)?.elementAtOrNull(0)
+        var voiceCommand = intent?.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS)?.elementAtOrNull(0)
             ?: return
 
         Log.i(TAG, "Recognized text: $voiceCommand")
         showToast(getString(R.string.info_voice_recognized_text, voiceCommand))
+
+        val prefix = getPrefs().getPrefixForVoice()
+        if (prefix != null) {
+            voiceCommand = "$prefix|$voiceCommand"
+            Log.d(TAG, "Prefix voice command: $voiceCommand")
+        }
 
         runBlocking {
             ConnectionFactory.waitForInitialization()
--- a/mobile/src/main/java/org/openhab/habdroid/core/VoiceService.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/VoiceService.kt
@@ -39,8 +39,7 @@ class VoiceService : IntentService("VoiceService") {
         Log.i(TAG, "Recognized text: $voiceCommand")
         showToast(getString(R.string.info_voice_recognized_text, voiceCommand))
 
-        val prefix = getPrefs().getPrefixForVoice()
-        if (prefix != null) {
+        getPrefs().getPrefixForVoice()?.let { prefix ->
             voiceCommand = "$prefix|$voiceCommand"
             Log.d(TAG, "Prefix voice command: $voiceCommand")
         }
--- a/mobile/src/main/java/org/openhab/habdroid/ui/preference/DeviceIdentifierPreference.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/preference/DeviceIdentifierPreference.kt
@@ -1,0 +1,148 @@
+/*
+ * Copyright (c) 2010-2020 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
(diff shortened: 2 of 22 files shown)


maniac103:
Generally this seems fine to me, just the `CommandType` handling seems a bit messy to me since `SuggestedCommandsFactory` doesn't care about a `CommandType`. My suggestion for fixing this: https://github.com/maniac103/openhab.android/commit/7a708011dcfe6693eea2870729c94635794ef2c2

Additionally, should the pref get an icon? Maybe [this one](http://materialdesignicons.com/icon/identifier)?

mueller-ma:
> My suggestion for fixing this: maniac103@7a70801

I'll have a look at it.

> Additionally, should the pref get an icon?

👍 
I also think it should get a summary. The device id if set, otherwise "Click for more information about this setting"?

maniac103:
> otherwise "Click for more information about this setting"?

I'd rather make it something like `Data sent to the server and voice commands are not prefixed with a device identifier` ... that way the user is pointed to how the pref is used by the summary already.

mueller-ma:
> My suggestion for fixing this: maniac103@7a70801

Picked.

mobile/src/main/java/org/openhab/habdroid/util/Util.kt
```diff
@@ -54,6 +54,6 @@ object Util {
             (Build.BRAND.startsWith("generic") && Build.DEVICE.startsWith("generic")) ||
             "google_sdk" == Build.PRODUCT
         Log.d(TAG, "Device is emulator: $isEmulator")
-        return isEmulator
+        return true
```
mueller-ma, line 57: For debugging only.

mueller-ma:
I don't really like the icon you suggested. What about these ones: https://materialdesignicons.com/icon/tag-text-outline or https://materialdesignicons.com/icon/tag-outline

maniac103:
tag-text-outline works for me as well.

mobile/src/main/java/org/openhab/habdroid/core/VoiceService.kt
```diff
@@ -31,12 +33,18 @@ import java.util.Locale
...
     override fun onHandleIntent(intent: Intent?) {
-        val voiceCommand = intent?.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS)?.elementAtOrNull(0)
+        var voiceCommand = intent?.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS)?.elementAtOrNull(0)
             ?: return
 
         Log.i(TAG, "Recognized text: $voiceCommand")
         showToast(getString(R.string.info_voice_recognized_text, voiceCommand))
 
+        val prefix = getPrefs().getPrefixForVoice()
+        if (prefix != null) {
```
maniac103, line 43: `getPrefs().getPrefixForVoice()?.let { prefix ->`?

mobile/src/main/java/org/openhab/habdroid/ui/preference/DeviceIdentifierPreference.kt
```diff
@@ -0,0 +1,166 @@
...
+
+    fun updateSummary() {
+        summary = if ((value as String?).isNullOrEmpty()) {
+            context.getString(R.string.device_identifier_summary_not_set)
+        } else {
+            value as String
+        }
+    }
+
+    fun setValue(value: String = (this.value as String?).orEmpty()) {
```
maniac103, line 62: Seeing this cast thrice, I wonder whether it would be worth adding a separate property (`valueString`) for the casted value.

mobile/src/main/java/org/openhab/habdroid/ui/preference/DeviceIdentifierPreference.kt
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        override fun onTextChanged(s: CharSequence, start: Int, before: Int, count: Int) {
+            // no-op
+        }
+
+        override fun afterTextChanged(s: Editable) {
+            val value = s.toString()
+            if (value.contains(" ") || value.contains("\n")) {
+                editorWrapper.error = context?.getString(R.string.error_sending_alarm_clock_item_empty)
```
maniac103, line 137: That used string resource looks a little weird here ... what does alarm clock have to do with this?

mueller-ma: The string res resolves to `Please enter a valid Item name`. I can rename it, though.

mobile/src/main/java/org/openhab/habdroid/ui/preference/DeviceIdentifierPreference.kt
mueller-ma: I prefer not to store the value twice. WDYT about https://github.com/openhab/openhab-android/pull/1983/commits/67661eb654a6c163b8f5400ba7176b1830423921?

maniac103: That one looks good to me.

PR #1990, opened 2020-05-04
Improve device info strings

mueller-ma:
"Alarm time" is better English than "Alarm clock time"

### DIFF
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -196,7 +196,7 @@
     <string name="send_device_info_category_event_based_title">Event based</string>
     <string name="send_device_info_category_event_based_explanation">These Items are updated on corresponding events, e.g. an incoming phone call</string>
     <string name="send_device_info_category_schedule_based_title">Schedule based</string>
-    <string name="send_device_info_category_schedule_based_explanation">These Items are updated on a regular schedule\nNote: Shorter intervals impair battery life more than higher ones and your device may decide to delay updates.</string>
+    <string name="send_device_info_category_schedule_based_explanation">The shorter the chosen interval is, the larger is the potential impact on the battery life of your device. Additionally, your device might not follow the chosen interval exactly, but delay the updates to reduce power consumption across apps.</string>
     <string name="send_device_info_category_general">General</string>
     <string name="send_device_info_item_prefix">Item name prefix</string>
     <string name="send_device_info_item_prefix_summary">The Item names of the following features are prefixed with \'%s\'</string>
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -196,15 +196,15 @@
     <string name="send_device_info_category_event_based_title">Event based</string>
     <string name="send_device_info_category_event_based_explanation">These Items are updated on corresponding events, e.g. an incoming phone call</string>
     <string name="send_device_info_category_schedule_based_title">Schedule based</string>
-    <string name="send_device_info_category_schedule_based_explanation">These Items are updated on a regular schedule\nNote: Lower values impair battery life more than higher ones and your device may decide to delay updates.</string>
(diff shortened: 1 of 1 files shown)


mobile/src/main/res/values/strings.xml
```diff
@@ -196,15 +196,15 @@
     <string name="send_device_info_category_event_based_title">Event based</string>
     <string name="send_device_info_category_event_based_explanation">These Items are updated on corresponding events, e.g. an incoming phone call</string>
     <string name="send_device_info_category_schedule_based_title">Schedule based</string>
-    <string name="send_device_info_category_schedule_based_explanation">These Items are updated on a regular schedule\nNote: Lower values impair battery life more than higher ones and your device may decide to delay updates.</string>
+    <string name="send_device_info_category_schedule_based_explanation">These Items are updated on a regular schedule\nNote: Lower values for the schedule impair battery life more than higher ones and your device may decide to delay updates.</string>
```
maniac103, line 199: How about `Shorter intervals` instead of `Lower values for the schedule`? The latter sounds pretty German ;-)

mobile/src/main/res/values/strings.xml
```diff
@@ -196,15 +196,15 @@
...
     <string name="send_device_info_category_general">General</string>
     <string name="send_device_info_item_prefix">Item name prefix</string>
     <string name="send_device_info_item_prefix_summary">The Item names of the following features are prefixed with \'%s\'</string>
     <string name="send_device_info_item_prefix_summary_not_set">No prefix is set for the Item names of the following features</string>
     <string name="settings_item_update_pref_howto_summary">Tap here to get hints on how to set up this feature server-side</string>
-    <string name="settings_alarm_clock">Alarm clock</string>
-    <string name="settings_alarm_clock_summary_on">Sends alarm clock time to item \'%1$s\'</string>
-    <string name="settings_alarm_clock_summary_off">Don\'t send alarm clock time</string>
+    <string name="settings_alarm_clock">Alarm time</string>
+    <string name="settings_alarm_clock_summary_on">Sends alarm time to item \'%1$s\'</string>
```
maniac103, line 206: While we're at it: capitalize `Item`

mobile/src/main/res/values/strings.xml
mueller-ma: Thanks :)

mobile/src/main/res/values/strings.xml

mobile/src/main/res/values/strings.xml
```diff
@@ -196,31 +196,31 @@
     <string name="send_device_info_category_event_based_title">Event based</string>
     <string name="send_device_info_category_event_based_explanation">These Items are updated on corresponding events, e.g. an incoming phone call</string>
     <string name="send_device_info_category_schedule_based_title">Schedule based</string>
-    <string name="send_device_info_category_schedule_based_explanation">These Items are updated on a regular schedule\nNote: Lower values impair battery life more than higher ones and your device may decide to delay updates.</string>
+    <string name="send_device_info_category_schedule_based_explanation">These Items are updated on a regular schedule\nNote: Shorter intervals impair battery life more than higher ones and your device may decide to delay updates.</string>
```
maniac103, line 199: Another thing: `higher` -> `longer`

Maybe also reword this completely?
`The shorter the chosen interval is, the larger is the potential impact on the battery life of your device. Additionally, your device might not follow the chosen interval exactly, but delay the updates depending on its operating conditions.`

mueller-ma: Maybe `...but delay the updates.` or `...but delay the updates to save battery.`?

maniac103: As this is an explanation, I wouldn't omit (potential) delay reasons. Your latter suggestion is fine with me, I'm just not sure whether it's valid English wording. Maybe `...but delay the updates to reduce power consumption across apps.`? That way we'd kinda explain the additional delay is due to wakeup batching.

PR #1991, opened 2020-05-04
Send DND mode

### LINKED ISSUE #1663: Send -Do Not Disturb mode state- to openHAB
No. But its similarly to "Send alarm clock to openHAB" openhab#1137 and "Send call state to server" openhab#1512 

If the "Do Not Disturb mode state" changed on my android phone the state should stored in a server item value.

No idea.

### Additional context
It could be a string item like CallState and can contain the modes: "" for off, SILENT, ALARM_ONLY, VIP_ONLY or so.
Another Item can contain the configured end time of the mode like alarmClock. If the end time is not set, it could be a very high timestamp.

mueller-ma:
Closes #1663


docs/USAGE.md
```diff
@@ -193,11 +191,10 @@ then
     } else {
         // Device isn't charging ("UNDEF" is send)
     }
-
 end
 ```
 
-#### Wi-Fi Name
+#### Wi-Fi Name (SSID)
```
mueller-ma, line 197: This will invalidate the current anchor for this heading.

mueller-ma: Fixed

PR #1995, opened 2020-05-05
Ask for permission when enabling 'Send wifi name'

### LINKED ISSUE #1987: geo permissions not asked when sending wifi ssid to server
### Actual behaviour
If the option "Send WLAN SSID to server" is activated and no geo permissions have been granted in Android for the app, you get "unknown ssid" in the corresponding openhab item.

### Expected behaviour
As with activating the transmission of the caller status, the authorizations should be checked automatically and the user can simply grant access to geo data.

 
### Steps to reproduce
1. disable sending ssid to server
2. remove geo permissions in android for openhab app
3. enable sending ssid to server
4. resulting in: unknown ssid on android 9

#### Client
* Android version: 9
* Device model: LG G6
* App version: 2.12.15  
* Device language: german

mueller-ma:
Fixes #1987

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -212,6 +212,10 @@ when
 then
     if (WifiName.state == "UNDEF") {
         // Device isn't connected to Wi-Fi
+    } else if (WifiName.state == "LOCATION_OFF") {
+        // On Android 8 and higher "Location" must be turned on, otherwise apps don't have access to the Wi-Fi SSID
+    } else if (WifiName.state == "NO_PERMISSION") {
+        // The location permission has been revoked by the user
     } else {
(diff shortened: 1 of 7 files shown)


mueller-ma:
For Android 8+ "Location" must be turned on to be able to read the Wi-Fi SSID. @maniac103 How do you think we should handle this?
I propose to mention the fact in the docs and the wifi pref summary and send "LOCATION_OFF" to the item.

maniac103:
Sounds good to me. Maybe just put the notice to the pref summary if location is off?

mueller-ma:
I'd display it all the time, so that the user knows that the Item can't be updated if he turns off the location later.

Issue #2014, opened 2020-05-20
Setting to make "schedule based" items "event based"


mueller-ma:
### Is your feature request related to a problem? Please describe.
Several users asked if it might be possible to have the DND mode as event based item, not schedule based. This makes totally sense to me; for wifi name as well. On Android 8 or higher listening to these event isn't possible anymore, when the app is in the background.

### Describe the solution you'd like
Add a setting that enables a foreground service running a broadcast receiver. Currently I've set the schedule to 15 minutes on my device and the battery usage of openHAB is about 1%. So the battery usage might be negligible.

### Describe alternatives you've considered
Having the broadcast receiver at least running when the app is in the foreground.

### Additional context

- https://community.openhab.org/t/android-app-beta/41382/236
- https://github.com/openhab/openhab-android/issues/1663#issuecomment-629592558

mueller-ma:
#2015 didn't close this issue.

PR #2015, opened 2020-05-20
Listen for schedule based item events when in foreground

mueller-ma:
Listen for events of schedule bases items when the app is in the foreground. This applies to all Android versions.

See #2014

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
@@ -82,8 +82,12 @@ class BackgroundTasksManager : BroadcastReceiver() {
                 scheduleWorker(context, PrefKeys.SEND_CHARGING_STATE)
             }
             WifiManager.NETWORK_STATE_CHANGED_ACTION -> {
-                Log.d(TAG, "Wifi state has changed")
+                Log.d(TAG, "Wifi state changed")
                 scheduleWorker(context, PrefKeys.SEND_WIFI_SSID)
+            }
+            NotificationManager.ACTION_INTERRUPTION_FILTER_CHANGED -> {
+                Log.d(TAG, "DND mode changed")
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -236,6 +240,23 @@ class BackgroundTasksManager : BroadcastReceiver() {
...
+            return IntentFilter().apply {
+                // These broadcasts are already defined in the manifest, so we only need them on Android 8+
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                    addAction(Intent.ACTION_POWER_CONNECTED)
+                    addAction(Intent.ACTION_POWER_DISCONNECTED)
+                    addAction(Intent.ACTION_BATTERY_LOW)
+                    addAction(Intent.ACTION_BATTERY_OKAY)
+                    addAction(WifiManager.NETWORK_STATE_CHANGED_ACTION)
+                }
+                // This broadcast is only send to registered receivers, so we need that in any case
```
maniac103, line 253: sent

mueller-ma:
The use case here are, as you already said, wall mounted tablets. It's also a preparation for #2014.

mueller-ma:
> I don't see e.g. battery low events occurring on a wall mounted device

It's probably easier to communicate to the user that _all_ items are updated when in foreground.

> I'd assume when a call comes in while the app is in foreground the item is updated twice with this change. Can you check whether that's the case?

`getIntentFilterForForeground()` adds actions to the intent filter based on the api level. For devices below Android O, it only adds the [dnd mode action](https://developer.android.com/reference/android/app/NotificationManager#ACTION_INTERRUPTION_FILTER_CHANGED). This is the only one that can't be received by a manifest receiver on all api levels (except on L, because it doesn't have the dnd api).
On Android O and higher it also adds actions that are defined in the manifest, but are not implicit broadcast exceptions: https://developer.android.com/guide/components/broadcast-exceptions

PR #2033, opened 2020-05-30
Fix crash on Android L when changing prefix

mueller-ma:
The DND mode pref isn't shown there


mueller-ma:
This commit was included in #2015, but it's actually a different issue, so I moved it to a new PR.

Issue #2022, opened 2020-05-24
Add support for external fonts and CSS files in SVG images


ewaldc:
### Is your feature request related to a problem? Please describe.
Problem: Rapidly change the style for a whole collection of SVG images/icons by having a shared CSS file that is included in each SVG (e.g. change color, dark/light mode switch).

The included SVG rendering library [AndroidSVG ](https://github.com/BigBadaboom/androidsvg) supports external fonts and CSS files (e.g. SVG @import "style.css") but the user application should override/implement the [SVGExternalFileResolver class](https://github.com/BigBadaboom/androidsvg/blob/master/androidsvg/src/main/java/com/caverock/androidsvg/SVGExternalFileResolver.java).
Documentation is [here ](http://bigbadaboom.github.io/androidsvg/api_summary_11.html) and in [official SVG pages](https://www.w3.org/TR/SVG/styling.html)

### Describe the solution you'd like
Extend _SVGExternalFileResolver_  with _resolveFont_ and _resolveCSSStyleSheet_  e.g. in [ExtensionFuncs.kt](https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt).
__Happy to give it a shot (although I am new to Kotlin) and if it works submit a PR__

### Describe alternatives you've considered
Include all styling in each SVG (versus a shared CSS file of all SVG).

### Additional context

Chart with imported CSS and one without (for testing)
[chart_extCSS.zip](https://github.com/openhab/openhab-android/files/4673205/chart_extCSS.zip)
[chart.zip](https://github.com/openhab/openhab-android/files/4673203/chart.zip)

PS. These SVG charts also don't render properly on the AndoridSVG library, but that is an issue with that library and I'm filing a PR with them (or find a workaround)

maniac103:
> Happy to give it a shot (although I am new to Kotlin) and if it works submit a PR

:+1: Go for it!

Some implementation notes:
- Since we can simply extend the resolver class, we should do so instead of using extension functions.
- The source CSS and font files are simply put into the images directory on the server?
- The class constructor likely will need a reference to an `HttpClient` instance to fetch the images from
- Fetching bitmaps and CSS files is easy, but `Typeface` instances can't be created from an InputStream. Probably need to save them to the app's cache directory and construct them from there (and skip download if file already exists)
[...]

Issue #728, opened 2018-03-02
Some svg look pixelated


mueller-ma:
### Actual behaviour
![screenshot_20180223-174050](https://user-images.githubusercontent.com/22525368/36903802-0902adc4-1e2f-11e8-9b2e-3b72c20f5107.png)

### Expected behaviour
Person icon shouldn't be so pixelated, because it's SVG.

 
### Steps to reproduce
1. Use this icon:
````xml
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" baseProfile="full" width="24" height="24" viewBox="0 0 24.00 24.00" enable-background="new 0 0 24.00 24.00" xml:space="preserve">
	<path fill="#E7B48F" fill-opacity="1" stroke-width="0.2" stroke-linejoin="round" d="M 12,2C 13.1046,2 14,2.89543 14,4C 14,5.10457 13.1046,6 12,6C 10.8954,6 10,5.10457 10,4C 10,2.89543 10.8954,2 12,2 Z M 10.5,7.00001L 13.5,7.00001C 14.6046,7.00001 15.5,7.89544 15.5,9.00001L 15.5,14.5L 14,14.5L 14,22L 10,22L 10,14.5L 8.5,14.5L 8.5,9.00001C 8.5,7.89544 9.39543,7.00001 10.5,7.00001 Z "/>
</svg>
````
Source: https://github.com/Templarian/MaterialDesign/blob/master/icons/svg/human-male.svg

2. Set icon format to svg.

### Can you reproduce the issue in demo mode?
No.

#### Client
* Android version: Android 7.1.2
* Device model: Oneplus X
* Stock or customized ROM: LineageOS 14.1
* App version : 2.0.15-beta
* Installation source : Github
* Device language: DE

#### Server
Client issue...

maniac103:
Should probably scale the 24px size given in the SVG with the device's display density. This is how I solved it in the past: https://github.com/slapperwan/gh4a/blob/master/app/src/main/java/com/gh4a/utils/HttpImageGetter.java#L559

PR #2072, opened 2020-06-24
Fix up SVG rendering.

### LINKED ISSUE #913: SVG with height and width attributes aren't shown correctly
### Actual behaviour
See https://community.openhab.org/t/version-2-1-toast-messages-for-the-voice-input/44436/4
 
### Steps to reproduce
1. Use the icon provided in the thread.

No

The issue is that the icon has a larger height than the viewport dimens that are used for [creating the drawable](https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/util/MyWebImage.java#L136). Having a fixed heigth for icons seems wrong to me as every ui should scale the icons according to the available space. I guess the best fix would be to remove the height and width from the svg before it's rendered by the svg lib.

### LINKED ISSUE #2021: None-square SVG files don't render correctly
### Actual behaviour
Non-square-SVG files (e.g. viewBox="0 0 1200 1000") don't render properly (scaling is off, part of image is off screen, see picture). Square images render fine.

### Expected behaviour
Rendering should as in attached image (patched version of the code, see attached).
A user workaround is of course to make the SVG square.

### Steps to reproduce
Render attached SVG. 

Yes, by replacing chart.png with chart.svg for weather_chart

#### Client
* Android version: multiple (7.x, 8.x, 9.x, 10.x)
* Device model: at least 8 different phones and tablets
* App version : most recent stable and beta (2.12.20-beta)
* Build flavor : 
* Device language: english

#### Server
* Server version: 2.5.5
* Reverse Proxy: 
* Authentication method : 

I think I discovered where the issue is: in [ExtensionFuncs.kt](https://github.com/openhab/openhab-android/blob/master/mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt)

```
        if (docWidth < 0 || docHeight < 0) {
            val aspectRatio = svg.documentAspectRatio
            if (aspectRatio > 0) {
                val heightForAspect = targetSize.toFloat() / aspectRatio
                val widthForAspect = targetSize.toFloat() * aspectRatio
                if (widthForAspect < heightForAspect) {
                    docWidth = Math.round(widthForAspect)
                    docHeight = targetSize
                } else {
                    docWidth = targetSize
                    docHeight = Math.round(heightForAspect)
                }
            } else {
                docWidth = targetSize
                docHeight = targetSize
            }
            // we didn't take density into account anymore when calculating docWidth
            // and docHeight, so don't scale with it and just let the renderer
            // figure out the scaling
            density = null
        }
        if (docWidth != targetSize || docHeight != targetSize) {
            val scaleWidth = targetSize.toFloat() / docWidth
            val scaleHeight = targetSize.toFloat() / docHeight
            density = (scaleWidth + scaleHeight) / 2
        }

        val bitmap = Bitmap.createBitmap(targetSize, targetSize, Bitmap.Config.ARGB_8888)
        val canvas = Canvas(bitmap)
        if (density != null) {
            canvas.scale(density, density)
        }
```
It's fixed by changing:
`        if (docWidth != targetSize || docHeight != targetSize) {`
to
`        if (docWidth != targetSize && docHeight != targetSize) {`
Basically there is no need for _density_ since properly build SVG images will autoscale fine and can preserve aspect ratio (e.g. with _preserveAspectRatio="xMidYMid meet"_) but I admit I don't fully understand the whole purpose of _canvas.scale(density, density)_

Screenshot_20200524-100919_openHAB_Beta
Screenshot_20200524-102338_openHAB_Beta
[chart.zip](https://github.com/openhab/openhab-android/files/4673176/chart.zip)

maniac103:
In particular, fix rendering of SVGs with explicit sizes and of
non-square SVGs.

Closes #2021
Closes #913

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -311,6 +311,10 @@ Please make sure `Default Human Language Interpreter` is set to `Rule-based Inte
 Generating charts can be taxing to the server.
 If you experience slow chart loading times and your server isn't powerful, open `Settings` and disable `High resolution charts` to improve loading times.
 
+### Icons look pixelated
+
+For good looking icons, the best approach is using SVG icons. Bitmap icons have a fixed size that doesn't scale with screen pixel density, so depending on the device, they may be scaled up by a large factor. When using SVG icons, ideally use icons that don't have a fixed size (in other words, they shouldn't have a 'width' and 'height' attribute on the root tag), as otherwise scaling might become necessary again: the app renders SVGs at their native size scaled by screen density, but scales them to a common size; when using icons without fixed size, the app can render them at precisely the needed size.
+
 ## Trademark Disclaimer
 
 Google Play and the Google Play logo are trademarks of Google Inc.
--- a/mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
@@ -32,6 +32,7 @@ import org.openhab.habdroid.core.connection.ConnectionFactory
 import org.openhab.habdroid.model.CloudNotification
 import org.openhab.habdroid.ui.MainActivity
 import org.openhab.habdroid.util.HttpClient
+import org.openhab.habdroid.util.ImageConversionPolicy
 import org.openhab.habdroid.util.getNotificationTone
 import org.openhab.habdroid.util.getNotificationVibrationPattern
 import org.openhab.habdroid.util.getPrefs
@@ -125,9 +126,10 @@ class NotificationHelper constructor(private val context: Context) {
             val connection = ConnectionFactory.cloudConnectionOrNull
             if (connection != null && !context.isDataSaverActive()) {
                 try {
+                    val targetSize = context.resources.getDimensionPixelSize(R.dimen.notificationlist_icon_size)
                     iconBitmap = connection.httpClient
                         .get(message.icon.toUrl(context, true), timeoutMillis = 1000)
-                        .asBitmap(context.resources.getDimensionPixelSize(R.dimen.notificationlist_icon_size), false)
+                        .asBitmap(targetSize, ImageConversionPolicy.PreferTargetSize)
                         .response
                 } catch (e: HttpClient.HttpException) {
                     Log.d(TAG, "Error getting icon", e)
--- a/mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
@@ -127,7 +127,7 @@ class NotificationHelper constructor(private val context: Context) {
                 try {
                     iconBitmap = connection.httpClient
                         .get(message.icon.toUrl(context, true), timeoutMillis = 1000)
-                        .asBitmap(context.resources.getDimensionPixelSize(R.dimen.svg_image_default_size), false)
+                        .asBitmap(context.resources.getDimensionPixelSize(R.dimen.notificationlist_icon_size), false)
                         .response
                 } catch (e: HttpClient.HttpException) {
                     Log.d(TAG, "Error getting icon", e)
--- a/mobile/src/main/java/org/openhab/habdroid/ui/ChartActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/ChartActivity.kt
@@ -158,11 +158,6 @@ class ChartActivity : AbstractBaseActivity(), SwipeRefreshLayout.OnRefreshListen
             return
         }
 
-        if (chart.height == 0 || chart.width == 0) {
-            Log.d(TAG, "Height or width is 0")
-            return
-        }
-
         val chartUrl = widget.toChartUrl(
             getPrefs(),
             chart.width,
@@ -174,7 +169,7 @@ class ChartActivity : AbstractBaseActivity(), SwipeRefreshLayout.OnRefreshListen
         ) ?: return
 
         Log.d(TAG, "Load chart with url $chartUrl")
-        chart.setImageUrl(connection, chartUrl, chart.width, refreshDelayInMs = widget.refresh, forceLoad = force)
+        chart.setImageUrl(connection, chartUrl, refreshDelayInMs = widget.refresh, forceLoad = force)
     }
 
     private fun updateHasLegendButtonState(item: MenuItem) {
--- a/mobile/src/main/java/org/openhab/habdroid/ui/CloudNotificationAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/CloudNotificationAdapter.kt
@@ -112,7 +112,6 @@ class CloudNotificationAdapter(context: Context, private val loadMoreListener: (
                 iconView.setImageUrl(
                     conn,
                     notification.icon.toUrl(itemView.context, !itemView.context.isDataSaverActive()),
-                    itemView.resources.getDimensionPixelSize(R.dimen.notificationlist_icon_size),
                     timeoutMillis = 2000
                 )
             } else {
--- a/mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/ItemPickerAdapter.kt
@@ -121,11 +121,9 @@ class ItemPickerAdapter(context: Context, private val itemClickListener: ItemCli
             val connection = ConnectionFactory.usableConnectionOrNull
             val icon = item.category.toOH2IconResource()
             if (icon != null && connection != null) {
-                val size = iconView.resources.getDimensionPixelSize(R.dimen.notificationlist_icon_size)
                 iconView.setImageUrl(
                     connection,
                     icon.toUrl(itemView.context, !itemView.context.isDataSaverActive()),
-                    size,
                     timeoutMillis = 2000
                 )
             } else {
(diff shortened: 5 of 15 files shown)


maniac103:
I've verified this one against the SVGs in
- #728 
- https://community.openhab.org/t/svg-icons-in-the-android-app/46274/29
- https://community.openhab.org/t/version-2-1-toast-messages-for-the-voice-input/44436/6 (#913)
- #2021 (the zip attached there seems off, I used the one attached to #2022)

mueller-ma:
I set up a sitemap page with four text items, one for each icon, and four image items, also one for each icon. Before the image item with the icon from #728 looked good, now it's pixelated.

maniac103:
Well, that icon specifies it wants to have a size of 24x24, so my change obeys to that. Drop that explicit size and everything will be fine.
IMHO we should follow the icon's specifications instead of guessing what the icon author might have wanted and/or the user wants to be displayed.

mueller-ma:
I didn't specify an explicit size in my sitemap, though: `Image url="http://openhab:8080/icon/svg_728?state=UNDEF&format=svg"`

> IMHO we should follow the icon's specifications instead of guessing what the icon author might have wanted and/or the user wants to be displayed.

Isn't this patch about not honoring the icon's specifications and instead display icons "100% height/width"? An explicit size caused the issue in https://community.openhab.org/t/svg-icons-in-the-android-app/46274/27, right?

maniac103:
> I didn't specify an explicit size in my sitemap, though

OK, so the behavior you'd like to see for that icon when used as image is it being sized at 24x24dp instead of full width? If so, makes sense to me, I'll look into that.

maniac103:
Additional question: may this be an artifact of our image caching? We cache by URL, so at the time the (large) image view requests the icon, there already might be a (small) icon view variant in the cache.

mueller-ma:
Basic UI honors the size specified in the svg when used as image, but not when used as icon.
IMO it makes more sense to display the image with full width, when no size is specified.

> Additional question: may this be an artifact of our image caching? We cache by URL, so at the time the (large) image view requests the icon, there already might be a (small) icon view variant in the cache.

I'll check that.

maniac103:
So how about this? I noticed we scaled up plain bitmaps to full width as well, so I also fixed that in the process.

mueller-ma:
I like it, but users may complain and ask for a minimum size of images. We'll see.

> Before the image item with the icon from #728 looked good, now it's pixelated.

This is fixed now. The icon is displayed with 24x24 px and sharp.

> Additional question: may this be an artifact of our image caching? We cache by URL, so at the time the (large) image view requests the icon, there already might be a (small) icon view variant in the cache.

However the same icon as icon is now pixelated. Now I have a sitemap page with this icon as icon only, not as image. Even when I refresh the page, the icon is still pixelated. This isn't a caching issue since I can see in the logs that the icon is downloaded: `<-- 200 OK http://openhab:8080/icon/svg_728?format=SVG&anyFormat=true (35ms)`

> I didn't specify an explicit size in my sitemap, though

BTW: The image widget doesn't take a height parameter at all: https://www.openhab.org/docs/configuration/sitemaps.html#element-type-image

maniac103:
> However the same icon as icon is now pixelated. Now I have a sitemap page with this icon as icon only, not as image. 

Yes, but that's the icon's fault. With numbers for a Pixel 3 emulator:
- display density is 440 dpi, thus `displayMetrics.density` is 2.75
- icon specifies its size as 24x24, thus it gets rendered at 24 * 2.75 = 66 pixels for each side
- that 66x66 pixels bitmap then gets stretched to a 132x104 pixels icon ImageView (*)

I don't see how we could avoid the pixelation without messing with the icon itself. If anybody complains, we can tell them 'fix yo icon' ... after all, icons without explicit document size will be rendered sharp.

(*) Do you know whether there's a specific reason for the icon ImageView to have top/bottom, but not left/right padding? That's where the difference in dimension comes from. Personally, I'd make the padding symmetric for all sides, which has the nice side effect of images with unconstrained sizes being rendered in the correct size. At the moment, I'm looking at the width only (here: 132px) and render at that size (here: 132x132), while the final image will be scaled by height (here: 104x104). I'm not looking at height due to the 'Image' item case, where width = match_parent and height = wrap_content, so without image height = 0.

mueller-ma:
Honestly I think this PR makes SVG icon handling worse. Before icons without explicit size and with explicit size in the root `svg` tag only were rendered well. Many icons have an explicit size and I don't want the user to bother about that.

> I don't see how we could avoid the pixelation without messing with the icon itself.

Then I'd keep it as is.

> In particular, fix rendering [...] of non-square SVGs.

Seems good to me, but non-square SVGsare probably more seldom than SVGswith explicit size.

> Personally, I'd make the padding symmetric for all sides, which has the nice side effect of images with unconstrained sizes being rendered in the correct size.

Go for it 👍

maniac103:
> Honestly I think this PR makes SVG icon handling worse. Before icons without explicit size and with explicit size in the root svg tag only were rendered well.

I respectfully disagree, since this PR fixes the cases in
- https://community.openhab.org/t/svg-icons-in-the-android-app/46274/29
- #2021

Which were plain broken before. Additionally, this PR gets us to a state where SVG rendering is 100% explainable and predictable. IMHO those points are much more important than a pixelated rendering under some circumstances.

Will make an additional commit though to make paddings symmetric. I'll also look at how BasicUI handles icons. We can add back your previous hack to force rendering them in icon view size, but as you can see in the forum link above, that hack also has its side effects ... and correctly sized, but blurry, icons are IMHO preferable over wrongly sized sharp ones.

maniac103:
> I'll also look at how BasicUI handles icons.

BasicUI shows that icon at a smaller size instead of scaling it up. We can do the same if you prefer that. Personally I'm torn about which of the two is the better option.

mueller-ma:
I didn't look at Basic UI before. The icons there aren't full size and sharp, so I think it's ok to do so in the app as well.

> BasicUI shows that icon at a smaller size instead of scaling it up. We can do the same if you prefer that. Personally I'm torn about which of the two is the better option.

I think it's easier to explain when the icons aren't scalled up. Otherwise we honor the explicit size during rendering, but not afterwards.

> We can add back your previous hack to force rendering them in icon view size, but as you can see in the forum link above, that hack also has its side effects

We could add it back, but only for icons that have a 1:1 ratio and no explicit sizes except the one in the root tag.

maniac103:
> I think it's easier to explain when the icons aren't scalled up. Otherwise we honor the explicit size during rendering, but not afterwards.

That's true, but without scaling, it might become impossible to have consistent icon sizes between UIs. We use 40 'px' (dp actually), but who knows what exact value iOS and BasicUI use. I wonder whether making this optional would make sense, or is that too much?

mueller-ma:
I wouldn't add a setting for that. If you add a section to the "Help and Technical Details" section in the docs, you can implement upscaling for icons.

maniac103:
Added the docs, please let me know if that wording is OK.

mueller-ma:
The docs LGTM.

What about my suggestion "We could add it back, but only for icons that have a 1:1 ratio and no explicit sizes except the one in the root tag."?

maniac103:
> What about my suggestion "We could add it back, but only for icons that have a 1:1 ratio and no explicit sizes except the one in the root tag."?

Is there a way to determine whether that condition is true? I'm not sure TBH.

mueller-ma:
I looked at the FAQ and there are three questions interesting for us, starting at https://bigbadaboom.github.io/androidsvg/faq.html#My_SVG_won't_scale_to_fit_the_canvas_or_my_ImageView
It basically says that SVGs are scaled up if the have a viewbox, but no explicit size. It's recommended to override the height/width attribute and set a viewbox if not present.

Something like:
````diff
diff --git a/mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt b/mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
index a8087c14..e297e640 100644
--- a/mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
@@ -156,6 +156,16 @@ fun InputStream.svgToBitmap(targetSize: Int): Bitmap {
         var density: Float? = displayMetrics.density
         val targetSizeFloat = targetSize.toFloat()

+        val height = svg.documentHeight
+        val width = svg.documentWidth
+        if (svg.documentViewBox == null && height > 0 && width > 0) {
+            svg.setDocumentViewBox(0F, 0F, width, height)
+        }
+        if (svg.documentViewBox != null) {
+            svg.setDocumentWidth("100%")
+            svg.setDocumentHeight("100%")
+        }
+
         svg.renderDPI = DisplayMetrics.DENSITY_DEFAULT.toFloat()
         var docWidth = svg.documentWidth * displayMetrics.density
         var docHeight = svg.documentHeight * displayMetrics.density

````

With this commit all four tested icons render sharp. When used as image items they're rendered too large (higher than the screen height), though.

maniac103:
The former part (generate view box from width/height) seems fine to me (and is the fix for the printer icon). I don't want to do the latter part (discard width/height if view box is present) though, for the same reasons as with your original hack - it *is* your original hack after all. ;-) I'll add a parameter that passes conversion size preference (prefer source / prefer target) and let it toggle exactly that piece of code. For icons we can set it to 'prefer target', for images to 'prefer source'.

mueller-ma:
Can you rebase this PR?

maniac103:
Sure, I'll do it once I'm near a PC again.

PR #2075, opened 2020-06-24
Apply app theme to intro activity

### LINKED ISSUE #2074: Intro activity theme incorrect
### Actual behaviour
See screenshot:
Screenshot_1592984819

### Expected behaviour
Describe us what should happen
- Orange status bar
- White navigation bar
 
#### Client
* Android version: 9
* Device model: API 28 emulator
* App version 2.13.6 beta
* Build flavor full
* Device language: en

### Comments
Probably a regression of #2058

maniac103:
Closes #2074


mueller-ma:
When I introduced the intro, it wasn't possible to have any other theme except the default one set. However with backups and the system wide dark mode, it can now have any theme, e.g. basic UI in dark.
Did you check other themes except orange as well?

mueller-ma:
Since all icons are orange, I'd stick to the orange theme:
````diff
diff --git a/mobile/src/main/java/org/openhab/habdroid/ui/IntroActivity.kt b/mobile/src/main/java/org/openhab/habdroid/ui/IntroActivity.kt
index 6c1923d6..5438643c 100644
--- a/mobile/src/main/java/org/openhab/habdroid/ui/IntroActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/IntroActivity.kt
@@ -25,12 +25,11 @@ import com.github.paolorotolo.appintro.AppIntro
 import com.github.paolorotolo.appintro.AppIntroFragment
 import org.openhab.habdroid.R
 import org.openhab.habdroid.util.PrefKeys
-import org.openhab.habdroid.util.Util
 import org.openhab.habdroid.util.getPrefs

 class IntroActivity : AppIntro() {
-        setTheme(Util.getActivityThemeId(this))
+        setTheme(R.style.openHAB_DayNight_orange)
         super.onCreate(savedInstanceState)

         if (getPrefs().getBoolean(PrefKeys.RECENTLY_RESTORED, false)) {

````

IMO it's ok to use the regular dark theme even if the black theme is configured.

**Comment by maniac103** [comment]
> Since all icons are orange, I'd stick to the orange theme

Or make them white and apply primary color to them using tinting?

**Comment by mueller-ma** [comment]
> Since all icons are orange

These icons are only shown during the "real" intro, not after a restore. The only icon during the restore is the openHAB icon, which must be kept grey and orange.

Thus with commit it should be fine: https://github.com/mueller-ma/openhab.android/commit/c69242c1ee8b2af8b28454e14854c89ce7cb506d

**Comment by maniac103** [comment]
I opted for the similarly simple, but future proof, option of coloring those drawables in primary color.

File: mobile/src/main/java/org/openhab/habdroid/ui/IntroActivity.kt [hunk 1]
```diff
@@ -93,17 +95,23 @@ class IntroActivity : AppIntro() {
...
             imageDrawable,
-            greyColor, // Background color
-            blackColor, // Title color
-            ContextCompat.getColor(this, R.color.black))) // Description color
+            colorBackground, // Background color
+            colorText, // Title color
+            colorText)) // Description color
+    }
+
+    private fun resolveThemeColor(@AttrRes colorAttr: Int): Int {
```
mueller-ma, line 111: I'd make a extension function of `context` out of this one and use it in other places as well, but I can do that in a separate PR.

maniac103: I thought about this (and actually started), but wasn't sure what to make out of this check: https://github.com/openhab/openhab-android/blob/d4b6d6dc44ac54631f9ffc5c88ac91877b65ac9a/mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt#L85 ... do you remember what the point of that check is?

mueller-ma: The background might be a drawable. This check makes sure we get a color. Copied from https://stackoverflow.com/a/14468034/13139982

maniac103: Hmm, OK. I wonder what the best way to deal with that difference is, then:
- Make the method signature e.g. `resolveThemedColor(attr: Int): Int?` and return null if it's not a color or
- Have two different methods (`resolveThemedColor` and `resolveThemedColorWithFallback`), where the latter allows passing a fallback color

The former might be a bit more cumbersome to use due to needed null checks ... or pass a fallback color instead of returning a nullable value and pass 0 as fallback if we're sure the color is resolvable?

maniac103: Done.

mueller-ma:
ktlint is failing

PR #2082, opened 2020-07-04
Add setting to make "schedule based" items "event based"

### LINKED ISSUE #2014: Setting to make "schedule based" items "event based"

mueller-ma:
Closes #2014

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/ForegroundBroadcastReceiver.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/ForegroundBroadcastReceiver.kt
@@ -1,0 +1,144 @@
+/*
+ * Copyright (c) 2010-2020 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.openhab.habdroid.background
+
+import android.app.Service
+import android.content.Context
+import android.content.Intent
+import android.os.Build
+import android.os.IBinder
+import android.util.Log
+import androidx.annotation.VisibleForTesting
+import androidx.core.app.NotificationCompat
+import androidx.core.content.ContextCompat
+import org.openhab.habdroid.R
+import org.openhab.habdroid.util.PrefKeys
+import org.openhab.habdroid.util.getPrefs
+import org.openhab.habdroid.util.isItemUpdatePrefEnabled
+
+class ForegroundBroadcastReceiver : Service() {
+    private var isServiceStarted = false
+    private var backgroundTasksManager = BackgroundTasksManager()
+    private var isRegistered = false
+
+    override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
+        Log.d(TAG, "onStartCommand())")
+        when (intent?.action) {
+            ACTION_START -> startService()
+            ACTION_STOP -> stopService()
+        }
+        return START_STICKY
+    }
+
+    override fun onCreate() {
+        super.onCreate()
+        Log.d(TAG, "onCreate")
+
+        NotificationUpdateObserver.createNotificationChannels(this)
+
+        val titlesOfItems = BackgroundTasksManager.KNOWN_PERIODIC_KEYS
+            .filter { key -> getPrefs().isItemUpdatePrefEnabled(key) }
+            .map { key -> getString(getTitleResForDeviceInfo(key)) }
+        val summary = when {
+            Build.VERSION.SDK_INT < Build.VERSION_CODES.O &&
+                getPrefs().isItemUpdatePrefEnabled(PrefKeys.SEND_DND_MODE) -> {
+                getString(R.string.send_device_info_foreground_service_running_summary_dnd)
+            }
+            Build.VERSION.SDK_INT < Build.VERSION_CODES.O -> null
+            titlesOfItems.size == 1 -> {
+                getString(R.string.send_device_info_foreground_service_running_summary_one, titlesOfItems[0])
+            }
+            titlesOfItems.size == 2 -> {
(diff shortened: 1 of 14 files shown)


mobile/src/main/java/org/openhab/habdroid/background/ForegroundBroadcastReceiver.kt
```diff
@@ -0,0 +1,144 @@
...
+import android.util.Log
+import androidx.annotation.VisibleForTesting
+import androidx.core.app.NotificationCompat
+import androidx.core.content.ContextCompat
+import org.openhab.habdroid.R
+import org.openhab.habdroid.util.PrefKeys
+import org.openhab.habdroid.util.getPrefs
+import org.openhab.habdroid.util.isItemUpdatePrefEnabled
+
+class ForegroundBroadcastReceiver : Service() {
```
maniac103, line 30: This class seems a bit misnamed. How about `BroadcastEventListenerService`?

mobile/src/main/java/org/openhab/habdroid/background/ForegroundBroadcastReceiver.kt
```diff
@@ -0,0 +1,144 @@
...
+    private var isServiceStarted = false
+    private var backgroundTasksManager = BackgroundTasksManager()
+    private var isRegistered = false
+
+    override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
+        Log.d(TAG, "onStartCommand())")
+        when (intent?.action) {
+            ACTION_START -> startService()
+            ACTION_STOP -> stopService()
```
maniac103, line 39: This logic seems to be unneeded. I'd instead suggest to
- Call `context.stopService()` in `startOrStopService`
- Drop the actions completely
- Check whether there are matching intent filter actions unconditionally in onStartCommand() and call `stopSelf(startId)` if there aren't any
- Do receiver unregistering (if needed) in `onDestroy`

mobile/src/main/java/org/openhab/habdroid/background/ForegroundBroadcastReceiver.kt
```diff
@@ -0,0 +1,144 @@
...
+                getString(R.string.send_device_info_foreground_service_running_summary_one, titlesOfItems[0])
+            }
+            titlesOfItems.size == 2 -> {
+                getString(R.string.send_device_info_foreground_service_running_summary_two,
+                    titlesOfItems[0], titlesOfItems[1])
+            }
+            else -> {
+                getString(R.string.send_device_info_foreground_service_running_summary_more,
+                    titlesOfItems[0], titlesOfItems[1])
+            }
```
maniac103, line 69: Wouldn't a summary like `Listening for 3 event types` be sufficient?

mobile/src/main/java/org/openhab/habdroid/background/ForegroundBroadcastReceiver.kt
```diff
@@ -0,0 +1,144 @@
...
+import androidx.annotation.VisibleForTesting
+import androidx.core.app.NotificationCompat
+import androidx.core.content.ContextCompat
+import org.openhab.habdroid.R
+import org.openhab.habdroid.util.PrefKeys
+import org.openhab.habdroid.util.getPrefs
+import org.openhab.habdroid.util.isItemUpdatePrefEnabled
+
+class ForegroundBroadcastReceiver : Service() {
+    private var isServiceStarted = false
```
maniac103, line 31: This variable is unused.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -283,7 +286,11 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         updateTitle()
         showMissingPermissionsWarningIfNeeded()
 
-        registerReceiver(backgroundTasksManager, BackgroundTasksManager.getIntentFilterForForeground())
+        val intentFilter = BackgroundTasksManager.getIntentFilterForForeground(this)
+        if (intentFilter.countActions() != 0 &&
+            !prefs.getBoolean(PrefKeys.SEND_DEVICE_INFO_FOREGROUND_SERVICE, false)) {
```
maniac103, line 291: Add an extension func for that getter?

mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/background/ForegroundBroadcastReceiver.kt
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/background/ForegroundBroadcastReceiver.kt
mueller-ma: Thanks, that looks way better.

mobile/src/main/java/org/openhab/habdroid/background/ForegroundBroadcastReceiver.kt
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/background/ForegroundBroadcastReceiver.kt
mueller-ma: I'd like to inform the user why the event listener is running. Mentioning just the count is sufficient, too, but I like my version more.

maniac103:
LGTM now from looking at the code ... I haven't actually tested it though.

PR #2083, opened 2020-07-04
Use WorkManager to send voice commands

### LINKED ISSUE #1958: voicecommand widget not working with energysave
if you close the app by "close all" on app overwiew and then try to send voicecommand with widget, google reacts to voice, but the command is not sent to openhab and  the popup generated by the app which displays the recognized text is not showing up.

if app is closed (same scenario as before) and energy saving options for openhab are disabled, then then voice command is always sent to openhab and the popup shows up with the spoken command.

possible solutions:
1. give user a hint to disable energy saving options for openhab in system settings (open settingspage automatically at install)
2. optimize background logic

tested on android 9

mueller-ma:
Fixes #1958

This is mostly untested as I don't have a voice recognition app currently.

@markusuidl Do you want to test the following apk?

- Rename it to .apk
- Install my apk

[mobile-foss-beta-debug.zip](https://github.com/openhab/openhab-android/files/4874123/mobile-foss-beta-debug.zip)


markusuidl:
Ok will give this a try.

Should I enable energy optimization from Android for the app, because actually I have excluded the application from power save?

mueller-ma:
> Should I enable energy optimization from Android for the app, because actually I have excluded the application from power save?

Yes, this shouldn't be needed any longer with this PR.

markusuidl:
@mueller-ma had no time to uninstall and reinstall your apk, finally managed now and will test...

markusuidl:
@mueller-ma installed your apk, works like a charm!

android system energy optimization is enabled for openhab (by default after reinstall)
placed widget to homescreen and it seems to work better now
voicecommands in app work too

voicecommands of my testing scenarios have been recognized and sent to openhab 👍

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -162,6 +164,27 @@ class BackgroundTasksManager : BroadcastReceiver() {
...
             }
+            ACTION_VOICE_RESULT -> {
+                var voiceCommand = intent.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS)?.elementAtOrNull(0)
+                    ?: return
+
+                Log.i(TAG, "Recognized text: $voiceCommand")
+                context.getPrefs().getPrefixForVoice()?.let { prefix ->
+                    voiceCommand = "$prefix|$voiceCommand"
+                    Log.d(TAG, "Prefix voice command: $voiceCommand")
+                }
```
maniac103, line 175: Since voice commands are handled specially anyway, why not remove adding the prefix here...

mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
```diff
@@ -221,6 +225,33 @@ class ItemUpdateWorker(context: Context, params: WorkerParameters) : Worker(cont
         }
     }
 
+    private fun handleVoiceCommand(connection: Connection, value: ValueWithInfo): Result {
+        val headers = mapOf("Accept-Language" to Locale.getDefault().language)
```
maniac103, line 229: ... and instead do it here, which avoids the `split()` line below.

mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
```diff
@@ -221,6 +225,33 @@ class ItemUpdateWorker(context: Context, params: WorkerParameters) : Worker(cont
...
+                    .post("rest/voice/interpreters", value.value, headers = headers)
+                    .asStatus()
+            }
+        } catch (e: HttpClient.HttpException) {
+            if (e.statusCode == 404) {
+                Log.d(TAG, "Voice interpreter endpoint returned 404, falling back to item")
+                runBlocking {
+                    connection.httpClient
+                        .post("rest/items/VoiceCommand", value.value)
+                        .asStatus()
```
maniac103, line 242: If that statement throws, the app will crash (unlike before, where there was a nested try-catch).

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
mueller-ma: Sure, that's a better solution.

mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
mueller-ma: Added try/catch here

PR #2084, opened 2020-07-06
Always avoid cache when loading images that have refresh set.

### LINKED ISSUE #2066: Image items are not refreshing
### Actual behaviour

In 2.13.5-beta `Image` items with refresh interval set are not refreshing. The issue is not present in 2.13.0

### Expected behaviour
Image items should refresh properly.

 
### Steps to reproduce
1. Add an Image item with refresh interval (eg camera feed) to the sitemap
2. Open the sitemap in the app

The issue has probably originated in #2039 but I'm not sure.

maniac103:
We already avoided the cache for the actual refresh, but not for the
initial load triggered via onAttachedToWindow(). Since we need the
information whether refresh is enabled in that method, but the actual
sequence of calls is setImageUrl() -> onAttachedToWindow() ->
startRefreshing(), move the refresh interval setter to setImageUrl(), so
we have the interval available in onAttachedToWindow().

Closes #2066

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -628,7 +628,6 @@ class WidgetAdapter(
                 imageView.maxHeight = Integer.MAX_VALUE
             }
 
-            @Suppress("LiftReturnOrAssignment")
             if (value != null && value.matches("data:image/.*;base64,.*".toRegex())) {
                 val dataString = value.substring(value.indexOf(",") + 1)
                 val data = Base64.decode(dataString, Base64.DEFAULT)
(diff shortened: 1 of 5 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
```diff
@@ -146,7 +143,9 @@ class WidgetImageView constructor(context: Context, attrs: AttributeSet?) : AppC
         scope = CoroutineScope(Dispatchers.Main + Job())
         lastRequest?.let { request ->
             if (!request.hasCompleted()) {
-                request.execute(false)
+                // Make sure to have an up-to-date image if refresh is enabled (when avoiding cache in that case,
+                // we'll always load a stale image from cache until first refresh otherwise)
```
maniac103, line 147: I need to reword that comment.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -635,19 +634,16 @@ class WidgetAdapter(
                 val data = Base64.decode(dataString, Base64.DEFAULT)
                 val bitmap = BitmapFactory.decodeByteArray(data, 0, data.size)
                 imageView.setImageBitmap(bitmap)
-                refreshRate = 0
```
mueller-ma, line 638: You can remove the `@Suppress("LiftReturnOrAssignment")` now.

PR #2088, opened 2020-07-07
Fix TOGGLE on color items

mueller-ma:
Without this commit the following commands where send, when a toggle
action was used on a turned off color bulb:
1. ON (Light was turned on)
2. ON (Nothing happened)
3. OFF (Light was turned off)

Now it sends:
1. 100
2. 0


maniac103:
Why send 100/0 instead of ON/OFF? IIRC there were cases that specifically want ON/OFF so that on turning off a previously set brightness isn't altered/lost.

mueller-ma:
This is actually a bug in the Tradfri binding. No idea why this PR "fixed" the issue when I opened it.

PR #2101, opened 2020-07-21
Add support for configuring multiple OH targets

### LINKED ISSUE #10: support multiple openhab server

mueller-ma:
Fixes #10
Closes #1900

TODO:

- [ ] Add a pref for the primary server with a short explanation (saying that the secondary servers only support viewing the sitemaps) and a link to the docs with a list of features that are multi server ready and those that aren't ready.
- [ ] Fix discovery


maniac103:
Why a separate PR instead of updating the existing one? With a new one I can't amend to the PR even if I found time for doing so.

mueller-ma:
> Why a separate PR instead of updating the existing one? 

I didn't want to push my commit unasked to your branch, especially since it'd require a force push after rebasing your commit. I can do so, if you want, though.

mueller-ma:
I pushed my commit to #1900.

PR #2105, opened 2020-07-22
Queue snackbars instead of hiding the previous one

mueller-ma:
We might show multiple snackbars on launch, e.g. "Connected to local
server" and a warning that the app doesn't have the permissions for all
active background tasks.
Instead of hiding the previous snackbar when showing a new one, queue
the new one.


maniac103:
Code simplification suggestion: https://github.com/maniac103/openhab.android/commit/6090c10e5acdcefa180d781d02a3c7686925457d

mueller-ma:
I picked your commit, thanks!

PR #2114, opened 2020-07-25
Send dev info: Show last update in pref summary

mueller-ma:


maniac103:
This is a nice idea. I'd move the observing to the pref itself though. I've pushed that change [here](https://github.com/maniac103/openhab.android/commits/item-update-log) together with a minor string change for your picking consideration ;-)

PR #2115, opened 2020-07-25
Item picker: Show snackbar if demo mode is on

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
@@ -246,19 +246,19 @@ abstract class AbstractItemPickerActivity : AbstractBaseActivity(), SwipeRefresh
             }
 
             if (connection is DemoConnection) {
-                Log.d(TAG, "Demo mode is on => Show snackbar")
-                val snackbar = Snackbar.make(
+                Snackbar.make(
                     findViewById(R.id.activity_content),
                     R.string.info_demo_mode_short,
                     Snackbar.LENGTH_INDEFINITE
-                )
-                snackbar.setAction(R.string.turn_off) {
-                    applicationContext.getPrefs().edit {
-                        putBoolean(PrefKeys.DEMO_MODE, false)
+                ).apply {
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
```diff
@@ -240,6 +245,22 @@ abstract class AbstractItemPickerActivity : AbstractBaseActivity(), SwipeRefresh
                 return@launch
             }
 
+            if (connection is DemoConnection) {
+                Log.d(TAG, "Demo mode is on => Show snackbar")
```
maniac103, line 249: This log line doesn't look too useful.

mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
```diff
@@ -240,6 +245,22 @@ abstract class AbstractItemPickerActivity : AbstractBaseActivity(), SwipeRefresh
...
+                val snackbar = Snackbar.make(
+                    findViewById(R.id.activity_content),
+                    R.string.info_demo_mode_short,
+                    Snackbar.LENGTH_INDEFINITE
+                )
+                snackbar.setAction(R.string.turn_off) {
+                    applicationContext.getPrefs().edit {
```
maniac103, line 256: No need for using `applicationContext` here (can just omit is, since `this@AbstractItemPickerActivity` is a context as well)

mobile/src/main/java/org/openhab/habdroid/ui/AbstractItemPickerActivity.kt
```diff
@@ -240,6 +245,22 @@ abstract class AbstractItemPickerActivity : AbstractBaseActivity(), SwipeRefresh
...
+                        putBoolean(PrefKeys.DEMO_MODE, false)
+                        loadItems()
```
maniac103, line 258: Should pull that line out of the `edit` block.

mueller-ma:
I applied your suggestions and used `apply { }` instead of saving the Snackbar object in a variable.

PR #2123, opened 2020-07-30
Fix blurred outline of switches with mapping

### LINKED ISSUE #1615: Outline of section switch blurred
Like the new rendering of the section switch. Really appreciate the work on improving the design of the app ;-)

Unfortunately a tiny flaw has been introduced from 2.9.9-beta to 2.9.10-beta:

The outline of the first and the last section is thicker than inner sections (and thicker than a single button switch). I think, it's even visible in your screenshots posted above. At least these detail picts show the issue. The "inner" sections (as well as the single button) have a **sharp thin outline** while the first and the last section have a **thicker/blurred** outline. Seems there is something different.

2 9 10-day
2 9 10-month
2 9 10-year

The effect is **clearly visible** here on two different tablets with different DPI. But on a Samsung phone (Galaxy S9) it's _not_ visible though. 
It's harder to see on the attached jpg-screenshots (same for your and mine jpg's ;-)), but immediately to notice on the _real_ devices display.

On the **2.9.9-beta it was rendered perfectly**.

_Originally posted by @curlyel in https://github.com/openhab/openhab-android/pull/1606#issuecomment-543746237_

mueller-ma:
Fixes #1615

### DIFF
--- a/mobile/src/main/res/values-mdpi/style.xml
+++ b/mobile/src/main/res/values-mdpi/style.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <style name="openHAB.OutlineButton" parent="@style/Widget.MaterialComponents.Button.OutlinedButton">
+        <item name="strokeWidth">2dp</item>
+    </style>
+</resources>
--- a/mobile/src/main/res/values-mdpi/style.xml
+++ b/mobile/src/main/res/values-mdpi/style.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<resources>
-    <style name="openHAB.OutlineButton" parent="@style/Widget.MaterialComponents.Button.OutlinedButton">
-        <item name="strokeWidth">2dp</item>
-    </style>
-</resources>
--- a/mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.kt
@@ -203,7 +203,7 @@ class LogActivity : AbstractBaseActivity(), SwipeRefreshLayout.OnRefreshListener
             "Device: ${Build.DEVICE}\n" +
             "Product: ${Build.PRODUCT}\n" +
             "OS: ${Build.VERSION.RELEASE}\n" +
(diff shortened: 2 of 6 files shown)


maniac103:
Will this not make the outline thicker on x(x)hdpi devices?

mueller-ma:
Yes, it does. Do you think it's too thick now on higher dpi devices?

maniac103:
I haven't tried yet, so I'll need to see ... but as IMHO the current look is perfect, I think I likely won't like it.
Ideal would be a hdpi-only override.

mueller-ma:
Interesting: `getResources().getDisplayMetrics().density` returns `1.0` on my device where I can reproduce this issue, so this is not only limited to densities, that have a non-integer factor.

maniac103:
That's your 1024x600 7" tablet, right? That one indeed is mdpi, in which case I don't understand why this is happening at all :-/

mueller-ma:
Quite correct: It's a 1024x552 7" tablet.

I added the display size and density to the log. I'd say we merge this and ask @curlyel and in the forum to check the outline switches now and then we can override the stroke width for other densities as well, if needed.

mobile/src/main/res/values-mdpi/style.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <style name="openHAB.OutlineButton" parent="@style/Widget.MaterialComponents.Button.OutlinedButton">
+        <item name="strokeWidth">2dp</item>
```
maniac103, line 4: mdpi is inherited by higher densities if there is no explicit specification for those. I'd probably add this definition in the base styles, make it use a dimension value and set that to 2dp in the base value and 1dp in hdpi.

mueller-ma: TIL about the inheritance here.

curlyel:
> we merge this and ask @curlyel and in the forum to check the outline

Sure, happy to test! 
How to get a build which contains the merged commit? Just updating the beta app?

maniac103:
> How to get a build which contains the merged commit? Just updating the beta app?

Yes, the commit is part of the 2.13.19 beta release.

curlyel:
I can confirm, the outline it's no longer blurred here. The outline appears slightly thicker than before but it is sharp for all selections, see:

PR #2136, opened 2020-08-11
Improve strings related to event listener and dev id

mueller-ma:


mobile/src/main/res/values/strings.xml
```diff
@@ -286,7 +286,7 @@
 
     <!-- Device identifier -->
     <string name="device_identifier_title">Device identifier</string>
-    <string name="device_identifier_prefix_voice"><![CDATA[Prefix voice commands with \'<Device identifier>|\']]></string>
+    <string name="device_identifier_prefix_voice">Prefix voice commands with the device identifier followed by a vertical bar</string>
```
mueller-ma, line 289: `<` and `>` caused Crowdin to see `Device identifier` as untranslatable string.

maniac103: Damn, I missed this. I really don't like the new string, as it's significantly less understandable than the old one. Did you try writing < and > in HTML syntax (`&lt;` and `&gt;`) instead of using CDATA?

mueller-ma: No, I didn't try that, but it should work. Can you open a new PR for this?

maniac103: Yeah, as soon as I'm near a PC again (likely Monday).

mueller-ma: In this case I opened a PR: #2142

mueller-ma: It's still recognized as xml tag, but I added a comment that is shown in Crowdin.

PR #2140, opened 2020-08-14
Android 11 Support

### LINKED ISSUE #2137: Prepare for Android 11
The app needs to be prepared for Android 11. I'm on it.
https://developer.android.com/preview

mueller-ma:
Fixes #2137

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
@@ -103,9 +103,12 @@ class OpenHabApplication : MultiDexApplication() {
 
     @RequiresApi(Build.VERSION_CODES.R)
     private fun registerDataAccessAudit() {
+        class DataAccessException(message: String) : Exception(message)
+
         val appOpsCallback = object : AppOpsManager.OnOpNotedCallback() {
             private fun logPrivateDataAccess(opCode: String, trace: String) {
                 Log.e("DataAudit", "Operation: $opCode\nStacktrace: $trace")
+                RemoteLog.nonFatal(DataAccessException("Operation: $opCode\nStacktrace: $trace"))
             }
 
             override fun onNoted(syncNotedAppOp: SyncNotedAppOp) {
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
@@ -13,6 +13,9 @@
 
 package org.openhab.habdroid.core
 
+import android.app.AppOpsManager
+import android.app.AsyncNotedAppOp
+import android.app.SyncNotedAppOp
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
@@ -92,6 +95,46 @@ class OpenHabApplication : MultiDexApplication() {
             systemDataSaverStatus = listener.getSystemDataSaverStatus(this)
             batterySaverActive = listener.isBatterySaverActive(this)
         }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+            registerDataAccessAudit()
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.R)
+    private fun registerDataAccessAudit() {
+        val appOpsCallback = object : AppOpsManager.OnOpNotedCallback() {
+            private fun logPrivateDataAccess(opCode: String, attributionTag: String?, trace: String) {
+                Log.i("DA_$attributionTag", "Operation: $opCode\nStacktrace: $trace")
+            }
+
+            override fun onNoted(syncNotedAppOp: SyncNotedAppOp) {
+                logPrivateDataAccess(
+                    syncNotedAppOp.op,
+                    syncNotedAppOp.attributionTag,
+                    Throwable().stackTrace.toString()
+                )
+            }
+
+            override fun onSelfNoted(syncNotedAppOp: SyncNotedAppOp) {
+                logPrivateDataAccess(
+                    syncNotedAppOp.op,
+                    syncNotedAppOp.attributionTag,
+                    Throwable().stackTrace.toString()
+                )
+            }
+
+            override fun onAsyncNoted(asyncNotedAppOp: AsyncNotedAppOp) {
+                logPrivateDataAccess(
+                    asyncNotedAppOp.op,
+                    asyncNotedAppOp.attributionTag,
+                    asyncNotedAppOp.message
+                )
+            }
+        }
+
+        val appOpsManager = getSystemService(AppOpsManager::class.java) as AppOpsManager
+        appOpsManager.setOnOpNotedCallback(mainExecutor, appOpsCallback)
     }
 
     override fun onTerminate() {
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
@@ -30,6 +30,7 @@ import androidx.appcompat.app.AppCompatDelegate
 import androidx.multidex.MultiDexApplication
 import androidx.security.crypto.EncryptedSharedPreferences
 import androidx.security.crypto.MasterKeys
+import java.security.InvalidKeyException
 import org.openhab.habdroid.background.BackgroundTasksManager
 import org.openhab.habdroid.core.connection.ConnectionFactory
 import org.openhab.habdroid.util.DataUsagePolicy
@@ -37,7 +38,6 @@ import org.openhab.habdroid.util.RemoteLog
 import org.openhab.habdroid.util.determineDataUsagePolicy
 import org.openhab.habdroid.util.getDayNightMode
 import org.openhab.habdroid.util.getPrefs
-import java.security.InvalidKeyException
 
 class OpenHabApplication : MultiDexApplication() {
     interface OnDataUsagePolicyChangedListener {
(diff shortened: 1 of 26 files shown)


mueller-ma:
The `MainActivity` has only sometimes a white navigation bar when using a light theme and its behavior isn't always `BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE`. Other activities work well.

mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
```diff
@@ -83,6 +86,46 @@ class OpenHabApplication : MultiDexApplication() {
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+            registerDataAccessAudit()
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.R)
+    private fun registerDataAccessAudit() {
+        val appOpsCallback = object : AppOpsManager.OnOpNotedCallback() {
+            private fun logPrivateDataAccess(opCode: String, attributionTag: String?, trace: String) {
+                Log.i("DA_$attributionTag", "Operation: $opCode\nStacktrace: $trace")
```
maniac103, line 99: Looking at https://developer.android.com/preview/privacy/data-access-auditing I don't think we need to use `attributionTag` at all, as we never call `createAttributionContext`.

Also, is implementing that audit callback really useful for us? I don't think we conciously access private data ourselves (or do we?), so is the goal finding unwated accesses by the libraries we're using?

mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
```diff
@@ -120,16 +100,64 @@ abstract class AbstractBaseActivity : AppCompatActivity(), CoroutineScope {
...
-        var uiOptions = window.decorView.systemUiVisibility
-        val flags = (View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
-            or View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY
-            or View.SYSTEM_UI_FLAG_FULLSCREEN)
-        uiOptions = if (isEnabled && !forceNonFullscreen) {
-            uiOptions or flags
+        @Suppress("DEPRECATION")
+        fun checkFullscreenPreR() {
+            var uiOptions = window.decorView.systemUiVisibility
+            val flags = (View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
```
maniac103, line 106: Stickler comments actually seem valid here (or at least can be followed without making the code look ugly)

mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
```diff
@@ -120,16 +100,64 @@ abstract class AbstractBaseActivity : AppCompatActivity(), CoroutineScope {
...
+                black
+            }
+        }
+        window.navigationBarColor = windowColor
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+            window.insetsController
+                ?.setSystemBarsAppearance(flagsR, WindowInsetsController.APPEARANCE_LIGHT_NAVIGATION_BARS)
+        } else {
+            @Suppress("DEPRECATION")
+            window.decorView.systemUiVisibility = uiOptions
```
maniac103, line 159: I find that method a bit hard to follow. I'd suggest doing handling of `windowColor` first, and then having two separate code blocks for flag handling pre-R and post-R. That way we don't need separate `flagsR` and `flagsPreR` variables (which don't make sense since only one of them will be used on a given device anyway).

mobile/src/main/res/values/strings.xml
```diff
@@ -236,7 +236,10 @@
     <string name="settings_dnd_mode_summary_on">Sends DND mode to Item \'%1$s\'</string>
     <string name="settings_dnd_mode_summary_off">Don\'t send DND mode</string>
     <string name="settings_dnd_mode_howto_url" translatable="false">https://www.openhab.org/docs/apps/android.html#dnd-mode</string>
+    <string name="settings_background_tasks_permission_hint">You\'ll be asked for a required permissions after turning on this setting. Please do not deny the permissions.</string>
+    <string name="settings_background_tasks_permission_hint_android_r">You\'ll be asked for a required permissions after turning on this setting. Please allow it for all the time.</string>
```
maniac103, line 240: Wording improvement suggestion: `This feature requires additional permissions to work. After enabling it, you'll be asked for permission approval.` ... and then for pre-R remove the second sentence, and for post-R `Please make sure to approve the permission also for background usage, since the Item updates also happen when the app is in background.`

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -1042,7 +1054,7 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
     }
 
     private fun showMissingPermissionsWarningIfNeeded() {
-        val missingPermissions = BackgroundTasksManager.KNOWN_KEYS
+        var missingPermissions = BackgroundTasksManager.KNOWN_KEYS
```
maniac103, line 1057: Instead of making this `var`, I'd rather keep it as `val`, but as `MutableList` (basically remove line 1067), and to the conversion to `Array` later (in line 1101).

maniac103:
Reading the new docs for `Toast`, it looks like Toasty is dead. We should try to get rid of `showToast` as much as possible, potentially killing unneeded toasts (is e.g. the FCM registration failure one really needed?)

mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
mueller-ma: I didn't see a single data access, but IMO it's nice to have and see unwanted activities by libraries. I changed it to log level `e`.

mobile/src/main/res/values/strings.xml
mueller-ma: I also changed the string names to be consistent with other strings that have different versions on different api levels.

mueller-ma:
> We should try to get rid of showToast as much as possible, potentially killing unneeded toasts

I changed a few toasts to snackbars in this PR already. A few others could be changed, too. But some toasts are shown when the app is in background, so it's not possible to change them to snackbars.

> is e.g. the FCM registration failure one really needed?

IMO we should notify the user when notifications cannot be received. We could do this by a snackbar that's shown when the app is started.

maniac103:
> We could do this by a snackbar that's shown when the app is started.

Yeah, that was what I was heading at (and probably also is what Google wants to achieve by this change - get rid of context less popups coming out of nowhere).

mueller-ma:
The crash was:
````
    Process: org.openhab.habdroid.beta, PID: 9309
    java.lang.RuntimeException: Unable to start activity ComponentInfo{org.openhab.habdroid.beta/org.openhab.habdroid.ui.MainActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'android.view.WindowInsetsController com.android.internal.policy.DecorView.getWindowInsetsController()' on a null object reference
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3449)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3601)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7656)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
     Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'android.view.WindowInsetsController com.android.internal.policy.DecorView.getWindowInsetsController()' on a null object reference
        at com.android.internal.policy.PhoneWindow.getInsetsController(PhoneWindow.java:3880)
        at org.openhab.habdroid.ui.AbstractBaseActivity.setStatusBarColor(AbstractBaseActivity.kt:147)
        at org.openhab.habdroid.ui.AbstractBaseActivity.onCreate(AbstractBaseActivity.kt:73)
        at org.openhab.habdroid.ui.MainActivity.onCreate(MainActivity.kt:161)
        at android.app.Activity.performCreate(Activity.java:8000)
        at android.app.Activity.performCreate(Activity.java:7984)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1309)
````

File: mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt [hunk 6]
```diff
@@ -971,6 +984,26 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         @BaseTransientBottomBar.Duration duration: Int = Snackbar.LENGTH_LONG,
         onClickListener: (() -> Unit)? = null
     ) {
+        showSnackbar(getString(messageResId), actionResId, tag, duration, onClickListener)
+    }
+
+    private fun showSnackbar(
```
maniac103, line 990: I wonder whether moving this to the base class would make sense, so it can be used from the other activities as well?

mueller-ma: Yes, that makes sense. I moved it to the abstract base activity.

maniac103:
I got two cleanup commits on top of this: https://github.com/maniac103/openhab.android/commits/android11

Other than that, LGTM after fixing the conflicts.

mueller-ma:
I picked both commits.
The navigation bar still has white buttons on a white background in the main activity. I haven't figured out why. It's working in all the other activities and the code for that is shared between them.

mueller-ma:
> The navigation bar still has white buttons on a white background in the main activity. I haven't figured out why. It's working in all the other activities and the code for that is shared between them.

The deprecated functions still work on Android 11. IMO we could use them instead, suppress the warnings and open an issue to keep track of that.
I'd like to merge this so I can use the new snackbar function in other PRs as well.

maniac103:
> IMO we could use them instead, suppress the warnings and open an issue to keep track of that.

Yeah, that sounds like a good plan.

PR #2142, opened 2020-08-14
Improve device identifier string

mueller-ma:
See https://github.com/openhab/openhab-android/pull/2136#pullrequestreview-464926327


maniac103:
Did you check whether it works, IOW, looks correct when running the app? If so, LGTM!

PR #2153, opened 2020-08-19
Don't clear icon cache on manual refresh

### LINKED ISSUE #1230: Icon cache apparently cleared when sliding down

mueller-ma:
Don't clear icon cache on manual refresh and increase HTTP cache size.

Fixes #1230

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/CacheManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/util/CacheManager.kt
@@ -16,11 +16,9 @@ package org.openhab.habdroid.util
 import android.content.Context
 import android.graphics.Bitmap
 import android.util.LruCache
-
 import okhttp3.Cache
 import okhttp3.HttpUrl
 import org.openhab.habdroid.model.IconFormat
-
 import java.io.File
 import java.io.FileInputStream
 import java.io.FileOutputStream
@@ -28,8 +26,9 @@ import java.io.IOException
 import java.io.InputStream
 
 class CacheManager private constructor(appContext: Context) {
-    val httpCache: Cache = Cache(File(appContext.cacheDir, "http"), (2 * 1024 * 1024).toLong())
-    private val bitmapCache: LruCache<HttpUrl, Bitmap>
+    val httpCache: Cache = Cache(File(appContext.cacheDir, "http"), (10 * 1024 * 1024).toLong())
+    private val iconBitmapCache: LruCache<HttpUrl, Bitmap>
+    private val temporaryBitmapCache: LruCache<HttpUrl, Bitmap>
     private val widgetIconDirectory = appContext.getDir("widgeticons", Context.MODE_PRIVATE)
 
     init {
@@ -38,7 +37,12 @@ class CacheManager private constructor(appContext: Context) {
         // int in its constructor.
         val maxMemory = (Runtime.getRuntime().maxMemory() / 1024).toInt()
         // Use up to 1/8 of the available VM memory for the bitmap cache
-        bitmapCache = object : LruCache<HttpUrl, Bitmap>(maxMemory / 8) {
+        iconBitmapCache = object : LruCache<HttpUrl, Bitmap>(maxMemory / 8) {
+            override fun sizeOf(key: HttpUrl, value: Bitmap): Int {
+                return value.byteCount / 1024
+            }
+        }
+        temporaryBitmapCache = object : LruCache<HttpUrl, Bitmap>(maxMemory / 8) {
             override fun sizeOf(key: HttpUrl, value: Bitmap): Int {
                 return value.byteCount / 1024
             }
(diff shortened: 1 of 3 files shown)


maniac103:
Sorry, but AFAICT this is wrong on a variety of levels:
- We don't need an on-disk cache, since HTTP responses are cached on disk by OkHttp already
- This also means the cache already *is* persistent
- Removing state from the cache key means there's a chance a wrong icon is displayed (IIRC we don't start a HTTP request in case a cached bitmap was found)
- File I/O should never happen on the UI thread (the purpose of the memory cache is precisely avoiding that)
- #1260 is not a mitigation for (and, in fact, only inspired by, but otherwise not related to) #1230 ... #1260 is about avoid triggering swipe to refresh accidentally, which I find to be universally useful (I find accidentally triggering swipe to refresh by just swiping up in a list highly annoying)

Can't you fix #1230 as well by keeping icons in memory cache when clearing cache for pull-to-refresh?

mueller-ma:
> * We don't need an on-disk cache, since HTTP responses are cached on disk by OkHttp already
> * This also means the cache already is persistent

Good points, however IMO we can improve the caching behavior by
1. cache the icons forever
2. ignore the server url, because local and remote connections (should) go to the same server and serve the same icon, but okhttp doesn't know that.

> * Removing state from the cache key means there's a chance a wrong icon is displayed (IIRC we don't start a HTTP request in case a cached bitmap was found)

The state is included in the file name: `File(cacheDir, "icon-$iconName-$state")` Did you mix this up with [my other PR](https://github.com/openhab/openhab-android/pull/2154/files#diff-f32004aa542272e7fc83727d077b13cdR199-R200)?

> * File I/O should never happen on the UI thread (the purpose of the memory cache is precisely avoiding that)

👍 

> * Can't you fix #1230 as well by keeping icons in memory cache when clearing cache for pull-to-refresh?

AFAIK I can only clear the entire `LruCache`.

An okhttp interceptor is probably a better way to increase the caching time of the icons.

maniac103:
> AFAIK I can only clear the entire LruCache.

There is a [remove method](https://developer.android.com/reference/android/util/LruCache#remove(K))
The problem is rather getting a list of keys currently present in the cache. One could - during `evictAll()` check whether the removed entry is an icon, and if it is, store it to some temporary list and put it right back.

> The state is included in the file name: File(cacheDir, "icon-$iconName-$state")

Right, I misread the code there. Please ignore that remark.

> cache the icons forever

Shouldn't OkHttp already cache them forever? I don't think they'll expire.

maniac103:
Another (more obvious?) approach would be using 2 LruCaches: one for icons, one for all other bitmaps.

mobile/src/main/java/org/openhab/habdroid/util/CacheManager.kt
```diff
@@ -38,19 +37,37 @@ class CacheManager private constructor(appContext: Context) {
         // int in its constructor.
         val maxMemory = (Runtime.getRuntime().maxMemory() / 1024).toInt()
         // Use up to 1/8 of the available VM memory for the bitmap cache
-        bitmapCache = object : LruCache<HttpUrl, Bitmap>(maxMemory / 8) {
+        iconBitmapCache = object : LruCache<HttpUrl, Bitmap>(maxMemory / 8) {
```
maniac103, line 40: Nit: Maybe make this a named class (static class inside CacheManager) instead of duplicating the anonymous class?

mobile/src/main/java/org/openhab/habdroid/util/CacheManager.kt
```diff
@@ -38,19 +37,37 @@ class CacheManager private constructor(appContext: Context) {
...
     fun cacheBitmap(url: HttpUrl, bitmap: Bitmap) {
-        bitmapCache.put(url, bitmap)
+        if (isIconUrl(url)) {
+            iconBitmapCache.put(url, bitmap)
+        } else {
+            temporaryBitmapCache.put(url, bitmap)
+        }
+    }
+
+    private fun isIconUrl(url: HttpUrl): Boolean {
```
maniac103, line 68: Maybe make this method `private fun targetCache(url: HttpUrl): LruCache<Bitmap>` and return the appropriate one here? Doing so would simplify the methods above.

mueller-ma: Good idea!

mobile/src/main/java/org/openhab/habdroid/util/CacheManager.kt
```diff
@@ -16,41 +16,45 @@ package org.openhab.habdroid.util
...
 import java.io.FileOutputStream
 import java.io.IOException
 import java.io.InputStream
 
 class CacheManager private constructor(appContext: Context) {
-    val httpCache: Cache = Cache(File(appContext.cacheDir, "http"), (2 * 1024 * 1024).toLong())
-    private val bitmapCache: LruCache<HttpUrl, Bitmap>
+    val httpCache: Cache = Cache(File(appContext.cacheDir, "http"), (10 * 1024 * 1024).toLong())
+    private val iconBitmapCache: LruCache<HttpUrl, Bitmap>
+    private val temporaryBitmapCache: LruCache<HttpUrl, Bitmap>
```
maniac103, line 31: Nit: use type `BitmapCache` here and for the method signature below?

maniac103 (commented):
LGTM besides the small comment.

maniac103 (approved):

PR #2154, opened 2020-08-19
Fix flickering when loading uncached icons

### LINKED ISSUE #684: Prefetch icons for other states

mueller-ma:
By saving and comparing the url the image view doesn't show the wrong icon when it's recycled.

Fixes #684


mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
```diff
@@ -46,6 +47,8 @@ class WidgetImageView constructor(context: Context, attrs: AttributeSet?) : AppC
     private val adjustViewBoundsForDownscalingOnly: Boolean
     private var internalLoad: Boolean = false
     private var lastRequest: HttpImageRequest? = null
+    private var lastUrl: HttpUrl? = null
```
maniac103, line 50: This should equal `lastRequest?.url`

mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
```diff
@@ -196,7 +199,7 @@ class WidgetImageView constructor(context: Context, attrs: AttributeSet?) : AppC
         if (cached != null) {
             applyLoadedBitmap(cached)
         } else {
-            applyProgressDrawable()
+            applyProgressDrawable(url)
```
maniac103, line 202: Why not just skipping that call if the last URL matches the new one?

mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
```diff
@@ -46,6 +46,7 @@ class WidgetImageView constructor(context: Context, attrs: AttributeSet?) : AppC
     private val adjustViewBoundsForDownscalingOnly: Boolean
     private var internalLoad: Boolean = false
     private var lastRequest: HttpImageRequest? = null
+    private var lastBitmap: Bitmap? = null
```
maniac103, line 49: Can nuke this variable now.

mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
```diff
@@ -195,7 +196,9 @@ class WidgetImageView constructor(context: Context, attrs: AttributeSet?) : AppC
 
         if (cached != null) {
             applyLoadedBitmap(cached)
-        } else {
+        } else if (url.newBuilder().removeAllQueryParameters("state").build() !=
+            lastRequest?.url?.newBuilder()?.removeAllQueryParameters("state")?.build()
```
maniac103, line 200: Maybe add a method `fun statelessUrlEquals(url: HttpUrl): Boolean` to `HttpRequest` and call it like `} else if (lastRequest?.statelessUrlEquals(url) != true) {`
?

PR #2157, opened 2020-08-21
Fix divider not showing when there're invible widgets

mueller-ma:
https://community.openhab.org/t/not-always-lines-under-items-on-sitemap-openhab-app/103961

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -1340,11 +1340,6 @@ class WidgetAdapter(
                         return true
                     }
                 }
-                if (adapter.getItemViewType(position) == TYPE_INVISIBLE &&
-                    adapter.getItemViewType(position + 1) == TYPE_INVISIBLE
-                ) {
-                    return true
-                }
             }
 
             return false
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -1331,15 +1331,19 @@ class WidgetAdapter(
 
             // hide dividers before and after frame widgets
             val adapter = parent.adapter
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -1276,16 +1276,20 @@ class WidgetAdapter(
...
                     return true
                 }
                 if (position < adapter.itemCount - 1) {
-                    if (adapter.getItemViewType(position + 1) in noDividerTypes) {
+                    if (adapter.getItemViewType(position + 1) in intArrayOf(TYPE_FRAME, TYPE_INVISIBLE)) {
                         return true
                     }
                 }
+                if (adapter.getItemViewType(position) == TYPE_INVISIBLE &&
+                    adapter.getItemViewType(position + 1) == TYPE_INVISIBLE
```
maniac103, line 1289: This needs to be in the if check above, to not run into an AIOOBE for the last item in the list, where `position + 1` exceeds the valid list indices.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -1276,16 +1276,20 @@ class WidgetAdapter(
 
             // hide dividers before and after frame widgets
             val adapter = parent.adapter
-            val noDividerTypes = intArrayOf(TYPE_FRAME, TYPE_INVISIBLE)
             if (adapter != null) {
-                if (adapter.getItemViewType(position) in noDividerTypes) {
+                if (adapter.getItemViewType(position) == TYPE_FRAME) {
```
maniac103, line 1280: I do not yet fully get the logic change. Before dividers were hidden before and after both frame and invisible widgets. Now the dividers are hidden
- before and after frames
- before invisible widgets
- after invisible widgets, but only if the next one is an invisible widget.

I don't get the last condition, can you explain that a bit? Shouldn't it be sufficient to remove dividers before invisible widgets, and the divider of the next widget will divide it from the previous one?

maniac103 (approved):

PR #2164, opened 2020-08-23
Add long press behavior for color widgets

### LINKED ISSUE #2161: Add long press behavior for color widgets
In Basic UI long pressing the up and down buttons on color items causes the UI to send `INCREASE` or `DECREASE` about 4 times per second until the button is released.

mueller-ma:
Fixes #2161

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -70,6 +70,11 @@ import com.google.android.material.button.MaterialButtonToggleGroup
 import com.google.android.material.slider.LabelFormatter
 import com.google.android.material.slider.Slider
 import com.google.android.material.switchmaterial.SwitchMaterial
+import java.io.IOException
+import java.util.HashMap
+import java.util.Locale
+import kotlin.math.abs
+import kotlin.math.roundToInt
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
@@ -93,11 +98,6 @@ import org.openhab.habdroid.util.beautify
 import org.openhab.habdroid.util.getPrefs
 import org.openhab.habdroid.util.isDataSaverActive
 import org.openhab.habdroid.util.orDefaultIfEmpty
-import java.io.IOException
-import java.util.HashMap
-import java.util.Locale
-import kotlin.math.abs
-import kotlin.math.roundToInt
 
 /**
  * This class provides openHAB widgets adapter for list view.
@@ -1134,7 +1134,7 @@ class WidgetAdapter(
                 val button = itemView.findViewById<PeriodicSignalImageButton>(id)
                 button.clickCommand = clickCommand
                 button.longClickHoldCommand = longClickHoldCommand
-                button.callback = fun (_: View, value: String?) {
+                button.callback = { _, value: String? ->
                     value?.let { connection.httpClient.sendItemCommand(boundItem, value) }
                 }
             }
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/widget/PeriodicSignalImageButton.kt
```diff
@@ -0,0 +1,84 @@
...
+    override fun onLongClick(v: View?): Boolean {
+        isLongPressed = true
+        scheduleNextSignal()
+        return true
+    }
+
+    @CallSuper
+    override fun onTouch(v: View, event: MotionEvent): Boolean {
+        if (event.actionMasked == MotionEvent.ACTION_UP) {
+            if (!isLongPressed) {
```
maniac103, line 56: Shouldn't this be `if (isLongPressed)`?

Also, couldn't you use the existence of `periodicCallbackExecutor` as indication of active long press?

mueller-ma: > Shouldn't this be if (isLongPressed)?

No, it shouldn't. For short clicks it should send ON/OFF, and when releasing long clicks, it shouldn't send anything.

> Also, couldn't you use the existence of periodicCallbackExecutor as indication of active long press?
 
👍

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -1134,19 +1148,18 @@ class WidgetAdapter(
             boundItem = widget.item
         }
 
-        override fun handleRowClick() {
+        // Up and down buttons callback
+        override fun invoke(v: View, value: String?) {
+            value?.let { connection.httpClient.sendItemCommand(boundItem, value) }
+        }
```
maniac103, line 1154: I'd prefer to have this one as a member variable instead, since the method name is somewhat unspecific (and yes, I realize there's a comment, but still :-) ).

mobile/src/main/java/org/openhab/habdroid/ui/widget/PeriodicSignalImageButton.kt
```diff
@@ -0,0 +1,82 @@
...
+        setOnTouchListener(this)
+    }
+
+    @CallSuper
+    override fun onLongClick(v: View?): Boolean {
+        scheduleNextSignal()
+        return true
+    }
+
+    @CallSuper
```
maniac103, line 51: While this annotation seems correct, shouldn't you follow it by calling the super method below?

mueller-ma: From the java docs: "Denotes that any overriding methods should invoke this method as well."
I added the annotation here, so the methods overriding it need to call super, but not this one.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -1119,13 +1125,22 @@ class WidgetAdapter(
...
+            val buttonCommandInfoList =
+                arrayOf(
+                    Triple(R.id.up_button, "ON", "INCREASE"),
+                    Triple(R.id.down_button, "OFF", "DECREASE")
+                )
+            for ((id, clickCommand, longClickHoldCommand) in buttonCommandInfoList) {
+                val button = itemView.findViewById<PeriodicSignalImageButton>(id)
+                button.clickCommand = clickCommand
+                button.longClickHoldCommand = longClickHoldCommand
+                button.callback = fun (_: View, value: String?) {
```
maniac103, line 1137: Can't you write this as
```
button.callback = { _: View, value: String? ->
    value?.let { connection.httpClient.sendItemCommand(boundItem, value)
}
```
?

mobile/src/main/java/org/openhab/habdroid/ui/widget/PeriodicSignalImageButton.kt
maniac103: Oh, I see now ... I didn't realize that this view is a touch listener for itself. Could probably also override `onTouchEvent` and call the super method instead of using the listener?

mueller-ma: Done

maniac103 (approved):

PR #2171, opened 2020-08-24
Add long press action to copy item name

mueller-ma:
Sometimes it's useful to quickly get the name of an Item, e.g. when configuring quick tiles or Tasker. This new long press action also allows some code cleanup, because "Create widget" now never appears as the only action.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
@@ -15,8 +15,11 @@ package org.openhab.habdroid.ui
 
 import android.app.PendingIntent
 import android.appwidget.AppWidgetManager
+import android.content.ClipData
+import android.content.ClipboardManager
 import android.content.ComponentName
 import android.content.Context
+import android.content.Context.CLIPBOARD_SERVICE
 import android.content.Intent
 import android.graphics.Bitmap
 import android.graphics.Canvas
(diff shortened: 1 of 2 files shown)


mueller-ma:
For long press actions that use `onContextItemSelected()` the widget is always taken from the left pane, even if a widget on the right side was selected.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
```diff
@@ -235,88 +251,75 @@ class WidgetListFragment : Fragment(), WidgetAdapter.ItemClickListener,
...
+
+        if (hasCommandOptions && nfcSupported) {
+            val nfcMenu = menu.addSubMenu(
+                Menu.NONE, CONTEXT_MENU_ID_WRITE_ITEM_TAG,
+                Menu.NONE,
+                R.string.nfc_action_write_command_tag
+            )
+            nfcMenu.setHeaderTitle(R.string.item_picker_dialog_title)
+            populateStatesMenu(nfcMenu, context, suggestedCommands, suggestedCommands.shouldShowCustom) {
+                    state, mappedState, itemId ->
```
mueller-ma, line 302: This indentation is done by Android Studio.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt
```diff
@@ -203,20 +207,32 @@ class WidgetListFragment : Fragment(), WidgetAdapter.ItemClickListener,
     }
 
     override fun onContextItemSelected(item: MenuItem): Boolean {
-        val context = context
+        val activity = activity
```
JohannesPtaszyk, line 210: Can this be invoked while the fragment is detached from the activity?
If not I would prefere to requireActvitiy() here.

mueller-ma: I'm not sure if it might happen due to some bad timing, so I'd like not to throw when `activity == null`.

mueller-ma:
> For long press actions that use onContextItemSelected() the widget is always taken from the left pane, even if a widget on the right side was selected.

`onCreateContextMenu()` is handled by the `WidgetListFragment` from the correct side, while `onContextItemSelected()` is always handled by the `WidgetListFragment` of the left side.

JohannesPtaszyk (approved):

mueller-ma:
> For long press actions that use onContextItemSelected() the widget is always taken from the left pane, even if a widget on the right side was selected.

I added a `setOnMenuItemClickListener()` that already knows the correct item name.

PR #2186, opened 2020-09-01
Add toggle change upscale images

### LINKED ISSUE #2183: Pictures shown small in sitemap
Hello all,

in the latest release of the Android app, my images are shown very small and not as before in full width. This happens both with the latest stable and beta release.
On the Brwoser it is rendered correctly.

Openhab App stable: 2.14.0
Openhab App beta: 2.14.5-beta
Android 9, LG G6
Openhab 2.5.7-1

Here is a picture from the current issue - the picture width is sufficient for the screen to be displayed in full width, as it did before the app update. The picture is within the frame "Wetter".

Screenshot_20200830-094709

mueller-ma:
Fixes #2183

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
@@ -20,6 +20,7 @@ import android.os.SystemClock
 import android.util.AttributeSet
 import android.util.Log
 import androidx.appcompat.widget.AppCompatImageView
+import kotlin.random.Random
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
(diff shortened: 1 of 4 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
```diff
@@ -229,7 +231,9 @@ class WidgetImageView constructor(context: Context, attrs: AttributeSet?) : AppC
 
     private fun applyLoadedBitmap(bitmap: Bitmap) {
         removeProgressDrawable()
-        if (adjustViewBoundsForDownscalingOnly) {
+        if (context.getPrefs().getBoolean(PrefKeys.IMAGE_UPSCALING, true)) {
```
maniac103, line 234: I don't like this pref evaluation here, since `WidgetImageView` is designed to be generic and doesn't know what kind of content it has.

PR #2187, opened 2020-09-01
Add a toggle for image item scaling behavior.

maniac103:


mobile/src/main/java/org/openhab/habdroid/util/PrefExtensions.kt
```diff
@@ -90,6 +91,14 @@ fun SharedPreferences.shouldRequestHighResChart(): Boolean {
     return getBoolean(PrefKeys.CHART_HQ, true)
 }
 
+fun SharedPreferences.getImageWidgetScalingType(): WidgetImageView.ImageScalingType {
+    return if (getBoolean(PrefKeys.IMAGE_WIDGET_SCALE_TO_FIT, true)) {
```
mueller-ma, line 95: This `if` statement should be inversed, right?

maniac103: Good catch, fixed.

mobile/src/main/res/values/strings.xml
```diff
@@ -47,6 +47,9 @@
     <string name="settings_openhab_icon_format_value_svg" translatable="false">SVG</string>
     <string name="settings_openhab_icon_format_png">Bitmap</string>
     <string name="settings_openhab_icon_format_svg">Vector</string>
+    <string name="settings_image_widget_scaling">Scale images to fit screen</string>
+    <string name="settings_image_widget_scaling_summary_off">Only scale images down \'Image\' widgets when they\'re larger than the screen, but don\'t scale them up to match the screen size</string>
```
mueller-ma, line 51:
```suggestion
    <string name="settings_image_widget_scaling_summary_off">Don\'t scale \'Image\' widgets up to match the screen size</string>
```
I don't think any user would expect to see the image only partly, because it's too large.

mueller-ma:
This PR needs a rebase.

mobile/src/main/res/values/strings.xml
maniac103: Reworded it slightly differently, but keeping the gist of the suggestion.

mueller-ma:
One thing I've noticed: Scrolling fast in the item picker list shows the (progress) openHAB icon and than changes to a larger (fallback) openHAB icon. IMO both icons should have the same size.

NoWhere2Go19:
Downloaded the beta app to test it. However I Can't find the switch for the scaling in the settings. I'm happy to test whenever needed
t47b6-qcfrh

maniac103:
Sure you got 2.4.10 and not just 2.4.9? It's there for me:
Screenshot_20200910-110303_openHAB_Beta

mueller-ma:
Usually the Play Store needs a few hours until the new update is distributed.

NoWhere2Go19:
Ahh yeah sorry, it's not 2.14.10. Thought there is a direct pipeline, my fault for not checking the versioning.

NoWhere2Go19:
Another comment: I saw that you released on Github the "2.15.0-release" a few days ago. The apk & app are working without any problems. However as I guess that this is directed to the stable release in the PlayStore, I just wanted to let you know that only the 2.14 stable is released there (at least Playstore localization Germany). The 2.15 wasn't available there over the last couple of days.

maniac103:
@NoWhere2Go19 It's at [20% rollout coverage](https://github.com/openhab/openhab-android/issues/549#issuecomment-690711659). For me, 2.15.0 is available already.

NoWhere2Go19:
> @NoWhere2Go19 It's at [20% rollout coverage](https://github.com/openhab/openhab-android/issues/549#issuecomment-690711659). For me, 2.15.0 is available already.

Ahh ok, thanks for the quick reply. Then it seems like that it hangs somewhere at Google, the beta release was there within a few hours. Let's see, thanks (y)

mueller-ma:
I usually rollout the stable release in stages: 20%, 50% and then 100%. This way I can stop the distribution if any severe error occurs. If I don't see any new crash until today evening, I change it to 100%.

https://support.google.com/googleplay/android-developer/answer/6346149

PR #2192, opened 2020-09-03
Tiles: Display status in subtitle

mueller-ma:
Instead of using Toasts the subtitle displays the status of the item update. Success messages are cleared after 6 seconds.
This applies to Android Q only as the subtitle isn't available on prior versions.


mobile/src/main/java/org/openhab/habdroid/background/tiles/AbstractTileService.kt
```diff
@@ -25,23 +25,53 @@ import android.service.quicksettings.TileService
...
+    var subtitleUpdateJob: Job? = null
 
     override fun onStartListening() {
         Log.d(TAG, "onStartListening()")
         qsTile?.let { updateTile(it) }
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            val workManager = WorkManager.getInstance(applicationContext)
+            val lifeCycleOwner = object : LifecycleOwner {
+                private val lifecycleRegistry = LifecycleRegistry(this).apply {
+                    handleLifecycleEvent(Lifecycle.Event.ON_START)
```
maniac103, line 62: I don't think this is sufficient, since you never emit a destroy event. See here for a reference to a full implementation: https://stackoverflow.com/a/50279360

mobile/src/main/java/org/openhab/habdroid/background/tiles/AbstractTileService.kt
```diff
@@ -77,14 +107,43 @@ abstract class AbstractTileService : TileService() {
...
+            WorkInfo.State.FAILED, WorkInfo.State.CANCELLED -> R.string.item_update_short_status_failed
+            WorkInfo.State.SUCCEEDED -> R.string.item_update_short_status_succeeded
+            null -> R.string.empty_string
+        }
+        qsTile?.apply {
+            subtitle = getString(statusRes)
+            updateTile()
+        }
+        subtitleUpdateJob?.cancel()
+        if (statusRes != R.string.empty_string) {
```
mueller-ma, line 139:
```suggestion
        if (statusRes == R.string.item_update_short_status_succeeded || statusRes == R.string.item_update_short_status_failed) {
```
All other events are triggered by the work info listener. In an error case there's still the error notification.

mobile/src/main/java/org/openhab/habdroid/background/tiles/AbstractTileService.kt
mueller-ma: You're right, but the issue is, that `onStartListening()` and `onStopListening()` aren't called as they should IMO.
In the log I can see, that Android calls `onStartListening()`, `onClick()` and `onStopListening()` with little time between whenever I click on the tile.

From the docs: "The lifecycle of a TileService is different from some other services in that it may be unbound during parts of its lifecycle."
To me this sounds like the OS will kill the tile service anyway. However I could stop `lifecycleRegistry` when the "Success" message is cleared and start it again in `onClick()`.

maniac103: > To me this sounds like the OS will kill the tile service anyway.

Yes, and the point I was trying to make was to emit a destroy event when that happens (and onDestroy is called).

> However I could stop lifecycleRegistry when the "Success" message is cleared and start it again in onClick().

:+1:

PR #2194, opened 2020-09-04
Use Item tags for tile icon suggestion

mueller-ma:
I used an array for the tags, because it can be put in a bundle. Android Studio suggested to override `equals()` and `hashCode()`, so I added the auto generated functions.


maniac103:
> Android Studio suggested to override equals() and hashCode(), so I added the auto generated functions.

I think AS is wrong there, as - AFAIK - `data class`es already implement those.

mueller-ma:
> Android Studio suggested to override equals() and hashCode()

With and without these functions the unit tests I added fail. So both, the auto generated one by AS and the auto generated one by Kotlin, are invalid.

mueller-ma:
I did some more testing: For Items without `member`s the custom `equals()` works well, but not the one generated by `data class`. For Items with `member`s both `equals()` functions fail.

maniac103:
https://github.com/maniac103/openhab.android/commit/0b8e57171b414897451e2615e1d831f75a37aa28 is the fix for the issue - its commit message has the explanation for the failure ;-)

maniac103 (approved):

PR #2201, opened 2020-09-12
Fix fragement manager IllegalStateException

### LINKED ISSUE #1650: ContentController, IllegalStateException: Can not perform this action after onSaveInstanceState
### Steps to reproduce
1. Configure local url with https to a server with an untrusted cert
2. Configure valid remote connection
3. Leave settings
4. Click on "Cancel" in the MTM dialog

#### Client
* Android version: Android 5
* App version : latest
* Build flavor : Full
* Device language: DE

mueller-ma:
Instead of taking the saved state from the activity, use it from the
fragment manager itself.

From the Java docs of `isStateSaved()`:

> Any operations that would change saved state should not be performed if this method returns true. For example, [...] any FragmentTransaction using {@link FragmentTransaction#commit()} instead of {@link FragmentTransaction#commitAllowingStateLoss()} will change the state and will result in an error.

Fixes #1650

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
@@ -451,15 +451,13 @@ abstract class ContentController protected constructor(private val activity: Mai
         activity.showSnackbar(R.string.error_sse_failed, tag = MainActivity.TAG_SNACKBAR_SSE_ERROR)
     }
 
-    internal abstract fun executeStateUpdate(reason: FragmentUpdateReason, allowStateLoss: Boolean)
+    internal abstract fun executeStateUpdate(reason: FragmentUpdateReason)
 
     private fun updateFragmentState(reason: FragmentUpdateReason) {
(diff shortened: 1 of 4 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
```diff
@@ -451,15 +451,15 @@ abstract class ContentController protected constructor(private val activity: Mai
...
 
-    internal abstract fun executeStateUpdate(reason: FragmentUpdateReason, allowStateLoss: Boolean)
+    internal abstract fun executeStateUpdate(reason: FragmentUpdateReason)
 
     private fun updateFragmentState(reason: FragmentUpdateReason) {
         if (fm.isDestroyed) {
             return
         }
         // Allow state loss if activity is still started, as we'll get
         // another onSaveInstanceState() callback on activity stop
```
maniac103, line 461: Can/should also nuke that comment now.

maniac103 (approved):

PR #2207, opened 2020-09-17
Remove channel group "Other"

mueller-ma:
This channel group isn't required, because Android puts notifications under
"Other" if no group is set.


mueller-ma:
This hides the channel group only for new installations. If I remove the channel group with `nm.deleteNotificationChannelGroup("other")`, the channels are also removed from the UI. Setting their group to `null` doesn't make them shown up again.
Since this is only a very minor improvement, I'd say it can be merged as is.

PR #2213, opened 2020-09-22
Add workflow for Update Gradle Wrapper Action

cristiangreco:
Hey there 👋, first of all thanks for your work on OpenHab!

I've got a suggested change: would you be willing to use [this GitHub Action](https://github.com/gradle-update/update-gradle-wrapper-action) to automatically keep Gradle Wrapper updated to latest release?

**What does "Update Gradle Wrapper Action" do?** It can be configured to run at scheduled intervals (e.g. daily or weekly) and will check whether the Wrapper script in the repo is up-to-date to the latest Gradle release: in case a new Gradle version is available, it will create a PR to update the Wrapper. And that's it!

**Why is that a good thing?** Well, first of all it alleviates the chore of manually updating the Wrapper, as you got a task that keeps track of new Gradle releases for you! More importantly, it boosts security around the Wrapper update and usage processes: this actions verifies that the `gradle-wrapper.jar` file has not been tampered with (uses checksum comparison), and it sets the `distributionSha256Sum` property so that the new Gradle binary itself will be verified locally upon download.

**Where can I find more about?** The [README](https://github.com/gradle-update/update-gradle-wrapper-action/blob/master/README.md) contains quite detailed information!

In this PR I propose adding a new workflow which runs the action every day at midnight (but feel free to adjust the frequency as you prefer). I've verified it works correctly in my fork of the repo, and you can [see here](https://github.com/gradle-update/openhab-android/pull/1) how the generated PR looks like.

The action is under active development, you can have a look at the [list of inputs](https://github.com/gradle-update/update-gradle-wrapper-action#action-inputs) currently supported. There's new features coming up soon and if you'd like to request any particular change just let me know!

I'd love to see the action used by OpenHab as such a popular project would definitely help me spread the word, but I also genuinely hope you can find this useful. Would love your feedback! ❤️

### DIFF
--- a/.github/workflows/update-gradle-wrapper.yml
+++ b/.github/workflows/update-gradle-wrapper.yml
@@ -1,0 +1,17 @@
+name: Update Gradle Wrapper
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  update-gradle-wrapper:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Update Gradle Wrapper
+        uses: gradle-update/update-gradle-wrapper-action@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}


mueller-ma:
I'm currently a bit busy, so it may take some time until I fully review this.

cristiangreco:
> I'm currently a bit busy, so it may take some time until I fully review this.

Sure, no worries. Ping me if you need any detail or clarification!

mueller-ma:
Two things I noticed in your Readme:

> It's like Dependabot for Gradle Wrapper.

Did you ask dependabot to integrate the gradle wrapper update in their app?

> So, for example, if you have any on: pull_request or on: push workflow that runs CI checks on Pull Requests, they won't be triggered when using GITHUB_TOKEN.
> A recommended workaround is to create a Personal Access Token (PAT) with repo scope and add it as a secret into your repository.

I don't want to add a personal access token with access to all of my repos to the secrets of openhab-android. The other GitHub actions must run after a gradle wrapper update to check if the build still works fine. Is there another way?

cristiangreco:
@mueller-ma thanks for your reply!

> Two things I noticed in your Readme:
> 
> > It's like Dependabot for Gradle Wrapper.
> 
> Did you ask dependabot to integrate the gradle wrapper update in their app?

Honestly not, but could do in the future (though their support for Gradle doesn't seem to be out of beta yet).

> > So, for example, if you have any on: pull_request or on: push workflow that runs CI checks on Pull Requests, they won't be triggered when using GITHUB_TOKEN.
> > A recommended workaround is to create a Personal Access Token (PAT) with repo scope and add it as a secret into your repository.
> 
> I don't want to add a personal access token with access to all of my repos to the secrets of openhab-android. The other GitHub actions must run after a gradle wrapper update to check if the build still works fine. Is there another way?

Unfortunately the only practical workaround, if you don't want to use a PAT, is to manually close and re-open the PR generated by the action. A "human intervention" will trigger the `pull_request` event and let other workflow run. I'm looking into the possibility of using Deploy keys (which have repo access only) but haven't made any progress yet. Would that be more suitable for you?

mueller-ma:
Today I manually updated gradle wrapper to fix #2257. This update also required a new version of Android studio. IMO it's probably better to manually update the wrapper when required.

PR #2231, opened 2020-10-01
Add hint why connection is marked as insecure

### LINKED ISSUE #2000: Add info why connection is marked as insecure
When using http or no authentication, the connection summary in the settings says "Insecurely connected to example.com". Some users think it's insecure, because they didn't configure a ssl client cert.

Add a info pref to the local and remote connection subscreen explaining why the connection is marked as insecure.

mueller-ma:
Fixes #2000

I'm not sure about the prefs title, it could be dropped.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
@@ -930,15 +930,22 @@ class PreferencesActivity : AbstractBaseActivity() {
         }
 
         private fun updateConnectionSummary(key: String, path: ServerPath?) {
+            fun insecureMessage(beautyUrl: String, reason: Int) =
+                getString(R.string.settings_insecure_connection_summary, beautyUrl, getString(reason))
+
             val pref = getPreference(key)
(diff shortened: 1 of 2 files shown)


maniac103 (commented):
I don't really like this approach. If we add a second pref with the security hint, we could as well put the same hint to the location connection pref summary. I'm not sure it's really needed either, since the reason becomes obvious once the user enters the connection prefs.

mueller-ma:
> I'm not sure it's really needed either, since the reason becomes obvious once the user enters the connection prefs.

Some people think their connection is marked as insecure, because they don't provide a ssl client certificate. This is the reason I want to add some additional hint.

> If we add a second pref with the security hint, we could as well put the same hint to the location connection pref summary.

That might be a better solution. I give it a try.

JohannesPtaszyk:
> I don't really like this approach. If we add a second pref with the security hint, we could as well put the same hint to the location connection pref summary. I'm not sure it's really needed either, since the reason becomes obvious once the user enters the connection prefs.

I actually think it is a nice addition to add a hint here. I know some people that got confused by the insecure connection.

mueller-ma:
> If we add a second pref with the security hint, we could as well put the same hint to the location connection pref summary.

Done

PR #2232, opened 2020-10-01
Only show confirm leave dialog if the config was changed

mueller-ma:
When creating a new server the inital config is a ServerConfiguration with empty or null values. The equal check with the server config from prefs always fails, because `null` is saved.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
@@ -650,12 +650,14 @@ class PreferencesActivity : AbstractBaseActivity() {
         ConfirmationDialogFragment.Callback,
         ServerEditorConfirmLeaveDialogFragment.Callback {
         private lateinit var config: ServerConfiguration
+        private lateinit var initialConfig: ServerConfiguration
         private var markAsPrimary = false
 
         override val titleResId: Int get() = R.string.settings_edit_server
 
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -650,12 +650,14 @@ class PreferencesActivity : AbstractBaseActivity() {
...
         ServerEditorConfirmLeaveDialogFragment.Callback {
         private lateinit var config: ServerConfiguration
+        private lateinit var initialConfig: ServerConfiguration
         private var markAsPrimary = false
 
         override val titleResId: Int get() = R.string.settings_edit_server
 
         override fun onCreate(savedInstanceState: Bundle?) {
             config = requireArguments().getParcelable("config")!!
+            initialConfig = config
```
maniac103, line 660: Dies this actually work? I'd expect only the reference to be copied and thus later modifications to `config` also being applied to `initialConfig`.

mueller-ma: It copies the reference here, but when `config` is changed, a new object is created and only assigned to `config`, not `initialConfig`.

maniac103: Ah, right, due to `ServerConfiguration` being immutable. Nevermind then :-)

maniac103 (approved):

PR #2237, opened 2020-10-02
Add confirm leave dialog to tiles prefs

### LINKED ISSUE #2234: Add confirm leave dialog to quick tile preferences
Currently when the user changes something in the quick tile settings and leaves the tile preference page without saving, the change is lost.

Add a dialog that asks if the user really wants to leave without saving.

mueller-ma:
Closes #2234

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
@@ -1275,21 +1275,7 @@ class PreferencesActivity : AbstractBaseActivity() {
         }
 
         override fun onBackPressed(): Boolean {
-            val currentData: TileData? = if (enabledPref.isChecked) {
-                TileData(
-                    itemAndStatePref.item.orEmpty(),
-                    itemAndStatePref.state.orEmpty(),
-                    itemAndStatePref.label.orEmpty(),
-                    namePref.text.orEmpty(),
-                    itemAndStatePref.mappedState.orEmpty(),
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt
```diff
@@ -1341,6 +1338,41 @@ class PreferencesActivity : AbstractBaseActivity() {
             }
         }
 
+        override fun onLeaveAndSave() {
+            Log.d(TAG, "Save tile $tileId")
+            val context = preferenceManager.context
+            val data: TileData? = if (enabledPref.isChecked) {
+                val itemName = itemAndStatePref.item
```
maniac103, line 1345: Is there a reason for constructing the `TileData` instance slightly different than above here? I think it would be preferable to have a common method for constructing it.

mueller-ma: Yes, there's a reason. Here I only want to create the `TileData` only if no value is null or empty. Above I don't care if values are missing, I just need a `TileData` to compare it to the saved one.

maniac103: Yes, I understand that part, but can't you create the `TileData` instance and check validity of its members afterwards?

mueller-ma: Yes, that's probably better.

PR #2255, opened 2020-10-10
closes #2209 Trim username and password fields 

### LINKED ISSUE #2209: Trim username and password fields
By accident a user may enter a white space in the username or password field which leads to a 401 error.

Trim the username and password fields when saving.

Show a warning when a white space is entered.

Gosunet:
Closes #2209

Add warning in all CustomInputTypePreference if there is a whitespace at the end of input.

image

I've had the warning il all CustomInputTypePreference as you suggested. If you want it only for password i will make the change !

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
@@ -70,7 +70,7 @@ open class CustomInputTypePreference constructor(context: Context, attrs: Attrib
     }
 
     class PrefFragment : EditTextPreferenceDialogFragmentCompat(), TextWatcher {
-        private var wrapper: TextInputLayout? = null
+        private lateinit var wrapper: TextInputLayout
         private lateinit var editor: EditText
 
         override fun onBindDialogView(view: View?) {
@@ -85,7 +85,7 @@ open class CustomInputTypePreference constructor(context: Context, attrs: Attrib
                 editor.inputType = type
             }
             arguments?.getCharSequence(KEY_TITLE)?.let { title ->
-                wrapper?.hint = title
+                wrapper.hint = title
             }
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
                 val hints = arguments?.getStringArray(KEY_AUTOFILL_HINTS)
@@ -96,6 +96,11 @@ open class CustomInputTypePreference constructor(context: Context, attrs: Attrib
                     editor.setAutofillHints(*hints)
                 }
             }
+        }
+
+        override fun onStart() {
+            super.onStart()
+            afterTextChanged(editor.text)
         }
 
         override fun beforeTextChanged(charSequence: CharSequence, start: Int, before: Int, count: Int) {
@@ -112,8 +117,8 @@ open class CustomInputTypePreference constructor(context: Context, attrs: Attrib
             val isLastCharWhitespace = value.lastOrNull()?.isWhitespace() ?: false
             val isFirstCharWhitespace = value.firstOrNull()?.isWhitespace() ?: false
 
-            val res = editor.resources
-            wrapper?.error = when {
+            val res = wrapper.resources
+            wrapper.error = when {
                 isFirstCharWhitespace -> res.getString(R.string.error_first_char_is_whitespace)
                 isLastCharWhitespace -> res.getString(R.string.error_last_char_is_whitespace)
                 else -> null
--- a/mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
@@ -15,6 +15,8 @@ package org.openhab.habdroid.ui.preference
 
 import android.content.Context
 import android.os.Build
+import android.text.Editable
+import android.text.TextWatcher
 import android.util.AttributeSet
 import android.view.View
 import android.widget.EditText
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
```diff
@@ -67,26 +69,52 @@ open class CustomInputTypePreference constructor(context: Context, attrs: Attrib
         return PrefFragment.newInstance(key, title, inputType, autofillHints)
     }
 
-    class PrefFragment : EditTextPreferenceDialogFragmentCompat() {
+    class PrefFragment : EditTextPreferenceDialogFragmentCompat(), TextWatcher {
+        private lateinit var wrapper: TextInputLayout
```
mueller-ma, line 73: This cannot be `lateinit`, because it isn't initialized below if `view == null`. Change it to
```suggestion
        private var wrapper: TextInputLayout? = null
```

mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
```diff
@@ -67,26 +69,52 @@ open class CustomInputTypePreference constructor(context: Context, attrs: Attrib
...
-            }
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                val hints = arguments?.getStringArray(KEY_AUTOFILL_HINTS)
-                if (hints == null) {
-                    editor?.importantForAutofill = View.IMPORTANT_FOR_AUTOFILL_NO
-                } else {
-                    editor?.importantForAutofill = View.IMPORTANT_FOR_AUTOFILL_YES
-                    editor?.setAutofillHints(*hints)
+            super.onBindDialogView(view)
+            if (view != null) {
```
mueller-ma, line 78: Better check for `view == null` and return if so.

mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
```diff
@@ -67,26 +69,52 @@ open class CustomInputTypePreference constructor(context: Context, attrs: Attrib
...
+        }
+
+        override fun onTextChanged(charSequence: CharSequence, start: Int, before: Int, count: Int) {
+            // no-op
+        }
+
+        override fun afterTextChanged(editable: Editable) {
+            val value = editable.toString()
+
+            val whitespaceAtLastCharacter = value.lastOrNull()?.isWhitespace() ?: false
```
mueller-ma, line 111:
```suggestion
            val isLastCharWhitespace = value.lastOrNull()?.isWhitespace() ?: false
```

mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
```diff
@@ -67,26 +69,52 @@ open class CustomInputTypePreference constructor(context: Context, attrs: Attrib
...
+
+            val res = editor.resources
+            wrapper.error = when {
+                whitespaceAtLastCharacter -> res.getString(R.string.error_trailing_whitespace)
```
mueller-ma, line 115: I'd add a check for the first character as well.

mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
```diff
@@ -101,8 +129,10 @@ open class CustomInputTypePreference constructor(context: Context, attrs: Attrib
                 autofillHints: Array<String>?
             ): PrefFragment {
                 val f = PrefFragment()
-                f.arguments = bundleOf(ARG_KEY to key, KEY_TITLE to title,
-                    KEY_INPUT_TYPE to inputType, KEY_AUTOFILL_HINTS to autofillHints)
+                f.arguments = bundleOf(
+                    ARG_KEY to key, KEY_TITLE to title,
```
mueller-ma, line 133: Can you add new lines after each `Pair`?

mobile/src/main/res/values/strings.xml
```diff
@@ -140,6 +140,7 @@
     <string name="error_invalid_url">Please enter a valid URL in a \'http(s)://host(:port)/\' form</string>
     <string name="error_invalid_https_url">Please enter a valid URL in a \'https://host(:port)/\' form</string>
     <string name="error_port_seems_invalid">The port you have entered seems invalid for the selected protocol</string>
+    <string name="error_trailing_whitespace">Be careful ! Whitespace at the end of line. Delete it if unwanted.</string>
```
mueller-ma, line 143:
```suggestion
    <string name="error_trailing_whitespace">The last character is a whitespace, which is probably unwanted</string>
```

mueller-ma (changes requested):
I thought a bit more about this: `CustomInputTypePreference` should have a new property called `whitespaceBehavior`. Possible values are `ignore` (no error message), `trim` (call `.trim()` on the value when saving) and `warn` (like this PR does).
`trim` is useful for the server urls and server names, `warn` for passwords and usernames.

I can do that in a new PR or you can give it a try.

mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
maniac103: Or keep initialization of `editor` and `wrapper` as they are (can keep them lateinit then) and just return if they're null.

mueller-ma:
@Gosunet Can you address my comments above? The thing with the new parameter can be done in a new PR by myself.

Gosunet:
Yes sorry for the delay ! I made the change you requested. I let you do the 'whitespaceBehavior' in another PR.

mobile/src/main/java/org/openhab/habdroid/ui/preference/CustomInputTypePreference.kt
Gosunet: Ok, i added : 

`val isFirstCharWhitespace = value.firstOrNull()?.isWhitespace() ?: false`

to the check :

wrapper?.error = when {
                isFirstCharWhitespace -> res.getString(R.string.error_first_char_is_whitespace)
                isLastCharWhitespace -> res.getString(R.string.error_last_char_is_whitespace)
                else -> null
            }

mueller-ma:
Mostly looks good to me. I changed the `wrapper` back to `lateinit`, as maniac mentioned above, and re-validate the string when opening the pref. Can you pick my commit: https://github.com/mueller-ma/openhab.android/commit/8b5e55ef65829511f355a15f9bb93251a9a69dd7?

Gosunet:
Yep done :)

mueller-ma:
Looks good so far, but I'd like to add the `whitespaceBehavior` here before merging. I added the label `hacktoberfest-accepted`, so this PR counts for your Hacktoberfest progress.

PR #2276, opened 2020-10-27
Make drawer entries configurable

mueller-ma:


maniac103:
Can you explain the rationale for this a bit? On a first glance it seems rather pointless to me, given the drawer is not exactly crammed with a lot of items. AFAICT removing items will just increase the amount of whitespace in the drawer.

mueller-ma:
There're two reasons:

- I don't want every family member to be able to use every UI. They usually don't touch the settings, so this is enough to keep them out.
- It allows us to add more UIs that are less frequent used and thus be hidden by default, e.g. HABot, Frontail (Server log viewer bundled with openhabian) or maybe even external links, like status.openhab.org.

maniac103:
Okay, makes  sense for the UI selection part. I'd keep out NFC and (especially) notifications though.

mueller-ma:
NFC is quite technical, so I'd like an option to hide it. Notifications should be fine to be always shown.

Issue #146, opened 2015-12-29
Force remote URL


staehler:
When logged in in a WLAN with similar adress range than the own intranet, habdroid tries to use the local URL, which will not work.
Please make it possible to force the use of the external URL.
Thanks and regards,
Michael

digitaldan:
@staehler I have not tried this, but what happens if you leave internal blank and put something in the external field?

mueller-ma:
[Here](https://github.com/mueller-ma/openhab.android/blob/master/mobile/src/main/java/org/openhab/habdroid/core/OpenHABTracker.java#L110) is checked if the local url is reachable by a simple tcp connection. It should check if this is a openhab server.

mueller-ma:
@maniac103 Do you think an option to limit the local connection to one wifi is usefull here?

maniac103:
@mueller-ma What do you mean by 'limit to one WiFi'? If you mean limiting the local connection to a specific (B)SSID, I'm not sure that makes sense. I, for one, had my previous router set up to use different SSIDs for 2.4 GHz and 5 GHz, and I doubt that's an uncommon setup.

mueller-ma:
Right, I haven't thought at those setups.

mueller-ma:
In case the tcp connection couldn't be opened to the local url, the remote url is used. But if the tcp connection is successful and the http connection fails, the error (mostly 400 or 404) is shown.

PR #2281, opened 2020-10-30
Auto select server based on Wi-Fi SSID

### LINKED ISSUE #2229: Select server instance based on wifi ssid
Add a setting to the local connection called wifi ssid. When the app is opened, the active server is selected based on the wifi ssid, if configured.

_Originally posted by @mueller-ma in https://github.com/openhab/openhab-android/pull/1900#issuecomment-604945035_

mueller-ma:
An optional Wi-Fi SSID can be configured for each server. If the app is
started and connected to one of these Wi-Fis, it switches to the
corresponding server.

Later this preference could be used force connecting to a remote url,
see #146.

Closes #2229

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/ServerConfiguration.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ServerConfiguration.kt
@@ -62,7 +62,8 @@ data class ServerConfiguration(
     val localPath: ServerPath?,
     val remotePath: ServerPath?,
     val sslClientCert: String?,
-    val defaultSitemap: DefaultSitemap?
+    val defaultSitemap: DefaultSitemap?,
+    val wifiSsid: String?
 ) : Parcelable {
     fun saveToPrefs(prefs: SharedPreferences, secretPrefs: SharedPreferences) {
         val serverIdSet = prefs.getConfiguredServerIds()
@@ -72,6 +73,7 @@ data class ServerConfiguration(
             putString(PrefKeys.buildServerKey(id, PrefKeys.LOCAL_URL_PREFIX), localPath?.url)
             putString(PrefKeys.buildServerKey(id, PrefKeys.REMOTE_URL_PREFIX), remotePath?.url)
             putString(PrefKeys.buildServerKey(id, PrefKeys.SSL_CLIENT_CERT_PREFIX), sslClientCert)
+            putString(PrefKeys.buildServerKey(id, PrefKeys.WIFI_SSID_PREFIX), wifiSsid)
             if (!serverIdSet.contains(id)) {
                 serverIdSet.add(id)
                 putConfiguredServerIds(serverIdSet)
@@ -100,6 +102,7 @@ data class ServerConfiguration(
             remove(PrefKeys.buildServerKey(id, PrefKeys.SSL_CLIENT_CERT_PREFIX))
             remove(PrefKeys.buildServerKey(id, PrefKeys.DEFAULT_SITEMAP_NAME_PREFIX))
             remove(PrefKeys.buildServerKey(id, PrefKeys.DEFAULT_SITEMAP_LABEL_PREFIX))
+            remove(PrefKeys.buildServerKey(id, PrefKeys.WIFI_SSID_PREFIX))
             putConfiguredServerIds(serverIdSet)
             if (prefs.getActiveServerId() == id) {
                 putActiveServerId(if (serverIdSet.isNotEmpty()) serverIdSet.first() else 0)
@@ -139,7 +142,16 @@ data class ServerConfiguration(
                 return null
             }
             val clientCert = prefs.getStringOrNull(PrefKeys.buildServerKey(id, PrefKeys.SSL_CLIENT_CERT_PREFIX))
-            return ServerConfiguration(id, serverName, localPath, remotePath, clientCert, getDefaultSitemap(prefs, id))
+            val wifiSsid = prefs.getStringOrNull(PrefKeys.buildServerKey(id, PrefKeys.WIFI_SSID_PREFIX))
+            return ServerConfiguration(
+                id,
+                serverName,
+                localPath,
+                remotePath,
+                clientCert,
+                getDefaultSitemap(prefs, id),
+                wifiSsid
+            )
         }
 
         fun saveDefaultSitemap(prefs: SharedPreferences, id: Int, defaultSitemap: DefaultSitemap?) {
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -111,6 +111,7 @@ import org.openhab.habdroid.util.areSitemapsShownInDrawer
 import org.openhab.habdroid.util.determineDataUsagePolicy
 import org.openhab.habdroid.util.getActiveServerId
 import org.openhab.habdroid.util.getConfiguredServerIds
+import org.openhab.habdroid.util.getCurrentWifiSsid
 import org.openhab.habdroid.util.getDefaultSitemap
 import org.openhab.habdroid.util.getGroupItems
 import org.openhab.habdroid.util.getHumanReadableErrorMessage
@@ -275,7 +276,9 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
             queryServerProperties()
         }
 
-        val switchToServer = determineServerIdToSwitchTo()
+        val currentWifiSsid = getCurrentWifiSsid()
+        val switchToServer = determineServerIdToSwitchToBasedOnWifi(currentWifiSsid, wifiSsidDuringLastOnStart)
+        wifiSsidDuringLastOnStart = currentWifiSsid
         if (pendingAction == null && switchToServer != -1) {
             switchServerBasedOnWifi(switchToServer)
         }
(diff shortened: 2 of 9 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -524,6 +527,85 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
...
     }
 
+    private fun selectServerBasedOnWifi() {
+        fun getPermissionForSsidRead() = when {
+            Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q -> Manifest.permission.ACCESS_FINE_LOCATION
+            Build.VERSION.SDK_INT >= Build.VERSION_CODES.O -> Manifest.permission.ACCESS_COARSE_LOCATION
+            else -> null
+        }
+
+        val serverHaveWifiSet = prefs
```
JohannesPtaszyk, line 537: I would suggest to prefix this with "does"

mobile/src/main/java/org/openhab/habdroid/model/ServerConfiguration.kt
```diff
@@ -62,7 +62,8 @@ data class ServerConfiguration(
     val localPath: ServerPath?,
     val remotePath: ServerPath?,
     val sslClientCert: String?,
-    val defaultSitemap: DefaultSitemap?
+    val defaultSitemap: DefaultSitemap?,
+    val wifiSsid: String?
```
JohannesPtaszyk, line 66: We seem to initialize this with lots of "nulls". I would suggest to insert some default null values.

JohannesPtaszyk (approved):
Overall this is fine for me. I just added some hints and suggestions where I would make changes :)

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -524,6 +527,85 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         }
     }
 
+    private fun selectServerBasedOnWifi() {
+        fun getPermissionForSsidRead() = when {
```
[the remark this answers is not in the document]
mueller-ma: It only returns the required permissions for reading the wifi ssid when the app is in foreground. There's already a public function for foreground and background access: `BackgroundTasksManager.getRequiredPermissionsForTask()`.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
mueller-ma: What about `hasAnyServerWifiSet`?

mobile/src/main/java/org/openhab/habdroid/model/ServerConfiguration.kt
mueller-ma: I only found two cases where `wifiSsid` is set to `null`: During upgrade from a pre-multi server version and when adding a new server. In latter case every argument except the name is null, so I wouldn't initialize `wifiSsid` with `null` by default.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
maniac103: `anyServerHasSetWifi`? (Putting `Set` in the middle to not imply this variable actually is a set)

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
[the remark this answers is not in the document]
JohannesPtaszyk: Fine for me :)

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
JohannesPtaszyk: Sounds reasonable

mobile/src/main/java/org/openhab/habdroid/model/ServerConfiguration.kt
JohannesPtaszyk: 👍 was not sure if it makes sense. Thanks for clarification!

JohannesPtaszyk (approved):

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -242,6 +244,7 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         }
 
         EventListenerService.startOrStopService(this)
+        selectServerBasedOnWifi()
```
mueller-ma, line 247: I'm not sure if `onCreate()` is the best location to call `selectServerBasedOnWifi()`. Maybe do it in `onResume()`, but only if the last time the app was open was some time ago?

JohannesPtaszyk: Why not onCreate tho? 🤔

mueller-ma: The user might close the app by pressing the home button, go to a location with a different known wifi, open the app again and it might be still in memory, so `onCreate()` might not be called.

JohannesPtaszyk: Not sure about the permission state on newer Android phones, but wouldn't it make sense to do it in a BroadcastReceiver?

JohannesPtaszyk: Never mind. Thats not possible anymore. Sorry.

maniac103: > Maybe do it in onResume(), but only if the last time the app was open was some time ago?

I'd do it in onStart, and compare current Wifi SSID to the one from the last check ... so base it on actual SSID change, not time.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -524,6 +527,85 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
...
+
+        val anyServerHasSetWifi = prefs
+            .getConfiguredServerIds()
+            .map { id -> ServerConfiguration.load(prefs, getSecretPrefs(), id) }
+            .any { config -> config?.wifiSsid?.isNotEmpty() == true }
+
+        val locationManager = getSystemService(Context.LOCATION_SERVICE) as LocationManager
+        val wifiManager = applicationContext.getSystemService(Context.WIFI_SERVICE) as WifiManager
+        val wifiInfo = wifiManager.connectionInfo
+        val wifiSsid = if (wifiInfo.networkId == -1) null else wifiInfo.ssid.removeSurrounding("\"")
```
maniac103, line 545: Can't you shorten this to `val wifiSsid = wifiManager.connectionInfo.ssid?.removeSurrounding("\"")`?

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -524,6 +527,85 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
...
+            wifiSsid.isNullOrEmpty() -> {
+                Log.d(TAG, "Cannot auto select server: SSID empty, probably not connected to wifi")
+                return
+            }
+        }
+
+        val serverForCurrentWifi = prefs
+            .getConfiguredServerIds()
+            .map { id -> ServerConfiguration.load(prefs, getSecretPrefs(), id) }
+            .firstOrNull { config -> config?.wifiSsid == wifiSsid } ?: return
```
maniac103, line 585: Nit: elvis operator on new line?

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
```diff
@@ -455,13 +456,18 @@ fun ServiceInfo.addToPrefs(context: Context) {
     val port = port.toString()
     Log.d(Util.TAG, "Service resolved: $address port: $port")
 
+    val wifiManager = context.applicationContext.getSystemService(Context.WIFI_SERVICE) as WifiManager
+    val wifiInfo = wifiManager.connectionInfo
+    val wifiSsid = if (wifiInfo.networkId == -1) null else wifiInfo.ssid.removeSurrounding("\"")
```
maniac103, line 461: This seems kind of magical. I think I'd rather leave the SSID empty for automatically created servers.

mobile/src/main/res/values/strings.xml
```diff
@@ -116,6 +116,10 @@
     <string name="settings_server_primary_url" translatable="false">https://www.openhab.org/docs/apps/android.html#multi-server-support</string>
     <string name="settings_server_default_name">openHAB %d</string>
     <string name="click_here_for_more_information">Touch here for more information</string>
+    <string name="settings_multi_server_wifi_ssid_summary">The app will select this server if it\'s connected to this Wi-Fi name</string>
```
maniac103, line 119: Consider on/off strings:
`The app will select this server when connected to Wi-Fi '%s'`
`The app will not automatically select this server based on connected Wi-Fi network`

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
mueller-ma: Should the last check be saved in memory or shared prefs?

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
mueller-ma: If the user manually configures a second server and at some point the app switches to the first server magically, a snackbar is shown to inform the user. Then the user knows about this feature or he has seen it before when adding the second server.

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
mueller-ma: `getSSID()` always returns a non-null string:

> The SSID may be {@link WifiManager#UNKNOWN_SSID}, if there is no network currently connected or if the caller has insufficient permissions to access the SSID.

````java
public static final String UNKNOWN_SSID = "<unknown ssid>";
````

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
maniac103: Still, I don't think second-guessing the user's intent is a good idea.

maniac103: Maybe a dialog or snackbar asking the user whether he wants to bind the found server to the current WiFi network would be a decent compromise?

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
maniac103: IMHO memory, since you *do* want this to happen when opening the app when it wasn't running before, regardless of the state before app shutdown.

mobile/src/main/res/values/strings.xml
mueller-ma: Done

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
JohannesPtaszyk: Guess, since it is important a Dialog would be sufficient 👍

maniac103:
2 commits for your consideration: https://github.com/maniac103/openhab.android/commits/2229-multi-server-wifi

mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
mueller-ma: I'd rather not add the wifi ssid at all instead of showing a dialog. This is probably too special for most users.

Issue #2320, opened 2020-11-30
OH 3 Item control widget doesn't work


MikeJMajor:
Item control widget isn't working with OH 3:

![Screenshot_20201130-214751](https://user-images.githubusercontent.com/24585489/100669972-18ba5200-3356-11eb-80f0-bab7ab568fae.jpg)

mueller-ma:
I cannot reproduce this. Can you post the log?

MikeJMajor:
Hopefully this is the right section

`0,org.openhab.habdroid.ui.ItemUpdateWidgetItemPickerActivity,topStateChangedWhenResumed]
12-01 08:06:33.890 26180 29607 I okhttp.OkHttpClient: --> GET https://<openhab-local-address-openHAB>:8443/rest/items
12-01 08:06:33.890 26180 29607 I okhttp.OkHttpClient: User-Agent: openHAB client for Android
12-01 08:06:33.890 26180 29607 I okhttp.OkHttpClient: Cache-Control: no-cache
12-01 08:06:33.890 26180 29607 I okhttp.OkHttpClient: --> END GET
12-01 08:06:34.440 26180 29607 I okhttp.OkHttpClient: <-- 200 OK https://<openhab-local-address-openHAB>:8443/rest/items (549ms)
12-01 08:06:34.440 26180 29607 I okhttp.OkHttpClient: Content-Type: application/json
12-01 08:06:34.440 26180 29607 I okhttp.OkHttpClient: Date: Tue, 01 Dec 2020 08:06:33 GMT
12-01 08:06:34.440 26180 29607 I okhttp.OkHttpClient: Transfer-Encoding: chunked
12-01 08:06:34.440 26180 29607 I okhttp.OkHttpClient: Server: Jetty(9.4.20.v20190813)
12-01 08:06:34.440 26180 29607 I okhttp.OkHttpClient: <-- END HTTP
12-01 08:06:35.185 26180 26180 D AbstractItemPickerActivity: Item response could not be parsed
12-01 08:06:35.185 26180 26180 D AbstractItemPickerActivity: org.json.JSONException: No value for label
12-01 08:06:35.185 26180 26180 D AbstractItemPickerActivity: 	at org.json.JSONObject.get(JSONObject.java:399)
12-01 08:06:35.185 26180 26180 D AbstractItemPickerActivity: 	at org.json.JSONObject.getString(JSONObject.java:560)
[...]

PR #2329, opened 2020-12-09
Make label optional

### LINKED ISSUE #2305: State option without label should not break the rendering
### Actual behaviour
If you define with metadata (OH3) a state option for an item without a label, this breaks the page loading in Android app. The page is displayed until this item.

### Expected behaviour
State option without label should be supported. In this case, the state value should be considered as label.
 
### Steps to reproduce
Define such an item:
```
String Mode "Mode" (Salon) [ "Control", "Power" ] { stateDescription="" [options="TV,PS3,OFF=Eteint"] }
```
And in sitemap, each of these lines are breaking the page rendering:
```
		Switch item=Mode mappings=[TV="TV", PS3="Console", OFF="Eteint"]
		Default item=Mode
```
If the item definition is changed to the following, there is no problem:
```
String Mode "Mode" (Salon) [ "Control", "Power" ] { stateDescription="" [options="TV=TV,PS3=PS3,OFF=Eteint"] }
```

mueller-ma:
Fixes #2305 and #2320

@MikeJMajor @lolodomo Can you please test the following apk?

- Rename it to .apk
- Install my apk

[mobile-foss-beta-debug.zip](https://github.com/openhab/openhab-android/files/5669285/mobile-foss-beta-debug.zip)

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/LabeledValue.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/LabeledValue.kt
@@ -24,5 +24,6 @@ data class LabeledValue internal constructor(val value: String, val label: Strin
 
 @Throws(JSONException::class)
 fun JSONObject.toLabeledValue(keyName: String, valueName: String): LabeledValue {
-    return LabeledValue(getString(keyName), getString(valueName))
+    val value = getString(keyName)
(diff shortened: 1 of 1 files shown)


MikeJMajor:
@mueller-ma It works for me.

maniac103 (approved):

PR #2330, opened 2020-12-10
Improve notification status strings

mueller-ma:
* Put "notification" in every string, so it isn't out of context when
  shown as Snackbar in MainActivity
* Remove FCM from the success string, because it's quite technical

### DIFF
--- a/mobile/src/full/res/values/strings.xml
+++ b/mobile/src/full/res/values/strings.xml
@@ -3,6 +3,6 @@
     <string name="info_openhab_gcm_in_progress">Device registration is in progress</string>
     <string name="info_openhab_gcm_failed">Push notification registration failed</string>
     <string name="info_openhab_gcm_failed_play_services">Push notification registration failed due to a Google Play Services issue: %s</string>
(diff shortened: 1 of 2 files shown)


mobile/src/full/res/values/strings.xml
```diff
@@ -3,6 +3,6 @@
     <string name="info_openhab_gcm_in_progress">Device registration is in progress</string>
     <string name="info_openhab_gcm_failed">Push notification registration failed</string>
     <string name="info_openhab_gcm_failed_play_services">Push notification registration failed due to a Google Play Services issue: %s</string>
-    <string name="info_openhab_gcm_connected">Device successfully registered with Firebase Cloud Messaging</string>
+    <string name="info_openhab_gcm_connected">Notifications are enabled</string>
```
maniac103, line 6: This seems a bit overly generic. How about `Successfully registered for notification reception` or `Notification registration completed successfully`?

mueller-ma: I updated these strings to `Notification notification...`.

maniac103 (approved):

Issue #2359, opened 2021-01-10
[UX] Settings->Connection->Local authentication failure resolution is not clear


13rac1:
### Actual behaviour

1. Go to Settings->Connection->Local
2. Enter
    * Local server URL
    * Username
    * Password
3. Authentication Failed -> No Help Message
![Screenshot_20210109-194608](https://user-images.githubusercontent.com/195061/104113949-a2ddca80-52b3-11eb-99eb-e2cb4f3e56be.png)

### Expected behaviour

1. Go to Settings->Connection->Local
2. Enter
    * Local server URL
    * Username or API Token
    * Password
3. Authentication Failed -> Display help or something else containing the following information "Is Basic Authentication Enabled or have you provided an API Token as the Username?"

A summary of the Authentication section from the docs would be extremely helpful: https://github.com/openhab/openhab-docs/blob/b8c8ca28/configuration/restdocs.md#authentication

> ## Authentication
>  Starting with version 3, openHAB supports password protection for sensible contents such as parts of the semantic model. To access this kind of information, the REST API provides the common mechanism Basic authentication and OAuth authorization. Both mechanisms can be used out of the box by the most programming languages and frameworks, but with regard to the curl examples from above, there are two alternative ways of authenticating yourself:
> 
> ### With username/password
> To the complete command, add -u {USER_NAME}, then enter your password as prompted. You can also do this in a single, non-interactive command, in this case add -u {USER_NAME}:{PASSWORD} instead.
> Note: This approach uses basic authentication which needs to be enabled in the org.openhab.restauth settings (accessible under Main UI > Settings > API Security).
> 
> ### With an API token
> This method is often recommended in order to keep your passwords safe and avoid to store them without encryption in any public places. To authenticate with an API token, append -u '{API_TOKEN}:' to the commandline.
> 
> You can manage all access tokens in your profile settings in the Main UI.
 
### Steps to reproduce
See above

### Can you reproduce the issue in demo mode?
N/A

#### Client
* Android version: 10
* Device model: Google Pixel
* App version: 2.15.0
* Build flavor Play Store
* Device language: English

#### Server
* Server version: 3.0.0
* Reverse Proxy: N/A
* Authentication method: Username/Password

### Logs
N/A

### Related

This is similar to #2339 but different enough as a UX-specific bug report I thought it should be separate. Please close if you disagree. I wanted the issue specifically documented so people could resolve the authentication failure more easily.

13rac1:
[...]

PR #2360, opened 2021-01-10
Add hint where to add API token

mueller-ma:
See #2359


maniac103:
I'd rather keep the pref title as is and instead add an info pref stating that if an API token is used, it should be entered into the user name field and the password needs to be empty in that case.

Issue #2358, opened 2021-01-09
Error when /api/v1/settings/notifications is not found


Flole998:
This is an issue that is related to openhab/openhab-core#2079. While the other issue covers unexpected behavior on the server this one is for unexpected behavior in the App.

When using a local openHAB Server the App requests /api/v1/settings/notifications, which is for myopenHAB and doesn't exist on a local installation. This causes the App to display a message at the bottom (. We should either remove this call (at least when contacting a local server) or silence the warning (at least when contacting a local server). I am not sure if it isn't trying to use my local server or if it is also trying to request this on a local server (which it shouldn't in my opinion). I would completely silence this warning for this URL as some might have the local server configured with the local address and the remote server configured with an external IP Address.

maniac103:
Did you enter the URL of your local server as 'remote server' in the app? The app will only do this request on the remote server.

Flole998:
I entered a locally reachable URL for local server and a remotely reachable URL for the remote URL. It should be able to contact the local URL in my case, but seems like it doesn't.

Also I can't remove the remote URL anymore, if I empty the field it doesn't seem to save it.

maniac103:
Please post the app log after a connection attempt.

Flole998:
Seems like it is actually using the local connection?

```
--------- beginning of main
01-09 19:35:37.860 10434 10434 I AppCompatViewInflater: app:theme is now deprecated. Please move to using android:theme instead.
01-09 19:35:37.890 10434 10434 I MediaPlayer: constructor
01-09 19:35:37.975 10434 10434 D Ringtone: Successfully created local player
01-09 19:35:37.991 10434 10434 D PreferencesActivity: Removing notification prefs for < 25
01-09 19:35:38.155 10434 10434 D OnePlusJankManager:  Chor uploadMDM JANK_TYPE_ONCE mViewTitle = org.openhab.habdroid.beta/org.openhab.habdroid.ui.MainActivity--- jank level = 1
01-09 19:35:38.207 10434 10434 D OnePlusJankManager:  Chor uploadMDM JANK_TYPE_ONCE mViewTitle = org.openhab.habdroid.beta/org.openhab.habdroid.ui.PreferencesActivity--- jank level = 1
01-09 19:35:38.615 10434 10434 W ListPreference: Setting a summary with a String formatting marker is no longer supported. You should use a SummaryProvider instead.
01-09 19:35:38.654 10434 10434 D MainActivity: onStop()
[...]

PR #2361, opened 2021-01-10
Ignore canceled cloud checks

mueller-ma:
See https://github.com/openhab/openhab-android/issues/2358#issuecomment-757358601

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.kt
@@ -26,6 +26,7 @@ import java.security.Principal
 import java.security.PrivateKey
 import java.security.cert.X509Certificate
 import java.util.HashSet
+import java.util.concurrent.CancellationException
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.kt
```diff
@@ -372,6 +373,8 @@ class ConnectionFactory internal constructor(
                     active?.remote?.toCloudConnection()
                 }
                 updateActiveCloud(CloudConnectionResult(result, null))
+            } catch (e: CancellationException) {
```
maniac103, line 376: From the logs I posted, it seems to be JobCancellationException that is thrown, not CancellationException.

mueller-ma: The `internal` `JobCancellationException` is a child class of `CancellationException`, so I had to use that.

maniac103 (approved):

PR #2363, opened 2021-01-11
Retry sitemap connection issues

mueller-ma:


maniac103:
Why would we want to retry a (previously successful) server property query on page load failure?

mueller-ma:
When a sitemap loading error occurs, a `CommunicationFailureFragment` is shown. This fragment has a retry button that calls `retryServerPropertyQuery()` on click.
I agree that this isn't the best behavior, but I just copied the existing behavior.

PR #2381, opened 2021-01-24
Hide widget divider for invisible widgets

### LINKED ISSUE #2378: App draws unnecessary lines at the top of frames
### Actual behaviour
The App draws unnecessary and unwanted lines between items when they are not visible because of their config. 
Example: 

1. 
  ```
  Frame
    {
        Text item=Netzwerk_DLANStatus label="DLAN muss wiederhergestellt werden" visibility=[Netzwerk_DLANStatus==OFF]
        Text item=System_Updates_OpenHAB label="OpenHAB Update verfügbar" visibility=[System_Updates_OpenHAB!="OFF"]
        Text item=Android_Wecker_Formatiert visibility=[Android_Wecker_Gestellt==ON]
    }
```
The first Item is not visible since `Netzwerk_DLANStatus=OFF`.
Screenshot_20210124-151149
You can see a line and some spacing between the top of the page and "OpenHAB Update verfügbar". Moreover you can see the exact same problem at the beginning of the frame "Multimedia". Draw attention to the line at the beginning of the "Multimedia" frame and the missing line in the "Steckdosen" frame.

2. When I now comment out the first line the line disappears:
Screenshot_20210124-151210
These unwanted lines only appear at the top of a frame if the first item is not visible.

### Expected behaviour
The lines should not be visible

 
### Steps to reproduce
1. See config as above

#### Client
* Android version: 11 (just upgraded yesterday, was on 10 as well)
* Device model: OnePlus 7TPro HD1913
* App version : 2.15.15-beta
* Build flavor : Full
* Device language: German

#### Server
* Server version: 2.5-10 (happens on 2.5-11 as well)
* Reverse Proxy: No
* Important Parts of the sitemap: 
[sitemap.txt](https://github.com/openhab/openhab-android/files/5862429/sitemap.txt)

### PR DESCRIPTION
**Pull request by mueller-ma**
Fixes #2378

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -1357,14 +1357,19 @@ class WidgetAdapter(
                 return false
             }
 
-            // hide dividers before and after frame widgets
+            // Hide dividers before and after frame widgets
             val adapter = parent.adapter
-            if (adapter != null) {
+            val viewTypeCurrentPos = adapter?.getItemViewType(position)
+            if (viewTypeCurrentPos != null) {
                 if (adapter.getItemViewType(position) == TYPE_FRAME) {
                     return true
                 }
                 if (position < adapter.itemCount - 1) {
                     if (adapter.getItemViewType(position + 1) in intArrayOf(TYPE_FRAME, TYPE_INVISIBLE)) {
+                        return true
+                    } else if (viewTypeCurrentPos == TYPE_INVISIBLE &&
+                        adapter.getItemViewType(position - 1) == TYPE_FRAME
+                    ) {
                         return true
                     }
                 }
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -1357,21 +1357,26 @@ class WidgetAdapter(
                 return false
             }
 
-            // Hide dividers before and after frame widgets
             val adapter = parent.adapter
             val viewTypeCurrentPos = adapter?.getItemViewType(position)
             if (viewTypeCurrentPos != null) {
-                if (adapter.getItemViewType(position) == TYPE_FRAME) {
+                // Hide divider after frame widgets
+                if (viewTypeCurrentPos == TYPE_FRAME) {
                     return true
                 }
-                if (position < adapter.itemCount - 1) {
-                    if (adapter.getItemViewType(position + 1) in intArrayOf(TYPE_FRAME, TYPE_INVISIBLE)) {
-                        return true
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt [hunk 1]
```diff
@@ -1357,15 +1357,20 @@ class WidgetAdapter(
                 return false
             }
 
-            // hide dividers before and after frame widgets
+            // Hide dividers before and after frame widgets
             val adapter = parent.adapter
-            if (adapter != null) {
+            val viewTypeCurrentPos = adapter?.getItemViewType(position)
+            if (viewTypeCurrentPos != null) {
```
maniac103, line 1363: I think I preferred the previous version, since that new variant is just an obscured null check of `adapter` as well.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -1357,15 +1357,20 @@ class WidgetAdapter(
...
                 if (adapter.getItemViewType(position) == TYPE_FRAME) {
                     return true
                 }
                 if (position < adapter.itemCount - 1) {
                     if (adapter.getItemViewType(position + 1) in intArrayOf(TYPE_FRAME, TYPE_INVISIBLE)) {
                         return true
+                    } else if (viewTypeCurrentPos == TYPE_INVISIBLE &&
+                        adapter.getItemViewType(position - 1) == TYPE_FRAME
```
maniac103, line 1371: This needs to be put into a separate if block with condition `position > 0`, so the check matches the used indices.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: It saves one `adapter?.getItemViewType(postion)` call. Not sure how expensive this call is.

maniac103: How? AFAICT the variable is used exactly once below.

mueller-ma: I forgot to use it in https://github.com/openhab/openhab-android/pull/2381/files#diff-b5b8a2ffbcde1bf9caec46b35e310bb0cb1f47171a75e0482527c06680d330f3R1364

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: Done

Trysupe:
In what state is this PR currently?

mueller-ma:
I'll release a new stable version soon and merge this afterwards as it may add new cases where the dividers are handled incorrectly.

PR #2386, opened 2021-01-25
Don't go back to sitemaps when app opened by shortcut

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
@@ -259,7 +259,7 @@ abstract class ContentController protected constructor(private val activity: Mai
         }
     }
 
-    fun showWebViewUi(ui: WebViewUi) {
+    fun showWebViewUi(ui: WebViewUi, isStackRoot: Boolean) {
         val prefs = activity.getPrefs()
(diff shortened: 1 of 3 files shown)


maniac103:
I think it would be more elegant to pass a boolean parameter `isStackRoot` to `showWebViewUi` in `ContentController` and use that information in `goBack` / `canGoBack`.

mueller-ma:
I made `isStackRoot` a property of `WebViewFragment`.

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
```diff
@@ -410,7 +409,11 @@ abstract class ContentController protected constructor(private val activity: Mai
      * @return true if back key can be consumed, false otherwise
      */
     fun canGoBack(): Boolean {
-        return temporaryPage != null || !pageStack.empty()
+        return if ((temporaryPage as? WebViewFragment)?.isStackRoot == true) {
+            return (temporaryPage as? WebViewFragment)?.goBack() == true
```
maniac103, line 413: `goBack` -> `canGoBack`

Also probably should put the conversion result into an intermediate variable ... doing so is a bit more readable (IMHO) and saves one null check.

maniac103 (approved):

PR #2420, opened 2021-01-31
Hide server selector if there's no server configured

mueller-ma:
Fixes a regression of #2379


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -897,7 +897,8 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
             inServerSelectionMode = false
         }
 
-        drawerModeSelectorContainer.isVisible = connection !is DemoConnection
+        drawerModeSelectorContainer.isVisible =
+            connection !is DemoConnection && prefs.getConfiguredServerIds().isNotEmpty()
```
maniac103, line 901: Wenn probably also don't need it if there's only one server configured.

mueller-ma: I should be looked at the `if` right above this. I'll simplify the code.

maniac103 (approved):
That makes more sense, LGTM! :+1:

Issue #2289, opened 2020-11-07
Add shortcut to specific Habpanel panel


politas:
I have a default panel I always want to open, which has all the most frequently used controls. Having to find out along the list of panels adds a significant delay to the time to do anything, when I just want to hit one button to trigger a rules chain. 

There's a lovely "Pin to home" option, but unfortunately, it ignores the current panel, and just makes a shortcut to the panel selector

"Pin to home" should create a shortcut to the panel I'm looking at, not the level up from that.

mueller-ma:
Instead of passing multiple arguments via a `Bundle` to `WebViewFragment`, it's probably better to make `WebViewFragment` abstract and use subclasses for every UI. This way things like this should be easier to implement.

politas:
There doesn't seem to be any UI element to allow this to actually work? The pin icon goes away when I choose a panel.

mueller-ma:
You're right. I tested my commit with openHAB 2.5 where app bar is always shown. Starting with openHAB 3 HABPanel is shown in fullscreen and therefore the button has to be added to its subpages as well.
This is tracked in #2506.

PR #2426, opened 2021-02-01
Use child classes of WebViewFragment for each UI

mueller-ma:
IMO this makes it easier to
1. pass arguments to the fragment, like `callback`
2. implement #2289 and #2337

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
@@ -179,6 +179,7 @@ abstract class ContentController protected constructor(private val activity: Mai
             pageStack.add(Pair(page, f ?: makePageFragment(page)))
         }
         temporaryPage = fm.getFragment(state, STATE_KEY_TEMPORARY_PAGE)
+        (temporaryPage as? AbstractWebViewFragment)?.setCallback(this)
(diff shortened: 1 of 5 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
```diff
@@ -261,27 +256,12 @@ abstract class ContentController protected constructor(private val activity: Mai
...
-            activeServerId,
-            isStackRoot,
-            ui.shortcutAction,
-            title,
-            ui.shortcutIconRes
+        val webViewFragment = ui.fragment.newInstance()
+        webViewFragment.init(
+            activity,
+            this,
+            isStackRoot
```
maniac103, line 263: This doesn't work. Fragments must be reinstatiable only from fragment state (e.g. after screen rotation), in which case that init() method isn't called. The arguments must be inferred from life cycle methods (e.g. onAttach) or the argument bundle.

maniac103:
The parent callback needs to passed differently as well (as mentioned, best idea probably is to route via MainActivity; I believe the connection holder fragment does something like this already).

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
mueller-ma: Ok, I removed `isStackRoot` and `activity` for now and I'll do the same with `callback` later.

maniac103 (approved):

PR #2435, opened 2021-02-05
Reduce data access warnings

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/Util.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Util.kt
@@ -57,12 +57,4 @@ object Util {
         Log.d(TAG, "Device is emulator: $isEmulator")
         return isEmulator
     }
-
-    fun createAttributionContext(context: Context, tag: String): Context {
-        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
-            context.createAttributionContext(tag)
-        } else {
-            context
-        }
-    }
 }
--- a/mobile/src/main/java/org/openhab/habdroid/util/Util.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/util/Util.kt
@@ -57,4 +57,12 @@ object Util {
         Log.d(TAG, "Device is emulator: $isEmulator")
         return isEmulator
     }
+
+    fun createAttributionContext(context: Context, tag: String): Context {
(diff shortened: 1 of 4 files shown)


mobile/src/main/java/org/openhab/habdroid/util/Util.kt
```diff
@@ -57,4 +57,12 @@ object Util {
         Log.d(TAG, "Device is emulator: $isEmulator")
         return isEmulator
     }
+
+    fun createAttributionContext(context: Context, tag: String): Context {
```
maniac103, line 61: Maybe make this an extension method? `fun Context.withAttribution(tag: String): Context`

mueller-ma: I want to avoid having it an extension function of `Context`, so there's no confusion on what's the version for all Android version. I wanted to make it an extension of `ContextCompat`, but wasn't able to create a `static` extension function.

maniac103: > so there's no confusion on what's the version for all Android version.

I don't understand that part, since the method body would be the same as in the utility function. AFAICT your argument could be brought up for *any* extension method?

mueller-ma: Usually we add extension functions to add a useful feature to a class, e.g. for converting `ResponseBody` to a `Bitmap`, we added `ResponseBody.toBitmap()`.
In this case there's already a function of `Context`, that creates an attribution context. However this works only for Android R and above. In such case there's usually a static function in `ContextCompat` that provides this feature on all (relevant) api levels, e.g. `ContextCompat.getColor()`.
This is why I wanted to add an extension function for `ContextCompat`, but because it doesn't work, I added this function to `Util`.

maniac103: `ContextCompat` only exists because there is no such thing as extension functions in Java. Essentially extension functions are just static methods with some syntactic sugar on top.
The reason why I'm so insisting here is that I think the `Util` class should ideally die completely (`getActivityThemeId` would also make sense as extension function) ... it just feels like an anti pattern in Kotlin.

mueller-ma: Ok, makes sense to me now. I made it an extension function.

maniac103 (approved):

PR #2441, opened 2021-02-07
Only lock swipe on drawer

mueller-ma:
When the app is opened by a shortcut to e.g. HABPanel, and it doesn't go
fullscreen, the drawer button is disabled as well.
With this commit only disable swipe-to-open, but keep the drawer button
active.


maniac103:
Why is the drawer locked at all in the first place?

mueller-ma:
It's locked, so it doesn't interfere with the swipe-to-open gesture of the HABPanel drawer.

mueller-ma:
I just noticed that the swipe gesture is now broken. I'll have a closer  look at this.

PR #2447, opened 2021-02-09
Hide frames with no visible children

### LINKED ISSUE #2445: App draws little spaces for empty frames
### Actual behaviour
The app draws little spaces for frames even though all elements of the frame are not visible because of `visibility=[...]`. You can see this at the top (before the frame "Licht"):
Screenshot_20210209-151959

================================

I added a label to the Frame to test this further:
Screenshot_20210209-151950

================================

This does not only happen on top of the sitemap. I injected an empty frame after the frame "Licht" before "Steckdosen" to test this:
Screenshot_20210209-152201

### Expected behaviour
Don't draw these spaces. I guess the app should check if at least of the elements of the frame are visible currently and if not skip the entire frame.

 
### Steps to reproduce
```
sitemap smartphone label="Felix"
{
    Frame
    {
        Text item=Netzwerk_DLANStatus label="DLAN muss wiederhergestellt werden" visibility=[Netzwerk_DLANStatus==OFF]
        Text item=System_Updates_OpenHAB label="OpenHAB Update verfügbar" visibility=[System_Updates_OpenHAB!="OFF"]
        Text item=Wecker_Formatiert visibility=[Wecker_Gestellt==ON]
        Text item=System_SSH_Versuche_Fehlgeschlagen labelcolor=["red"] valuecolor=["red"] visibility=[System_SSH_Versuche_Fehlgeschlagen!=0]
        Text item=System_SSH_FehlgeschlageneIPs visibility=[System_SSH_Versuche_Fehlgeschlagen!=0]
        Switch item=System_SSH_Versuche_Fehlgeschlagen_Bestaetigen mappings=[ON="OK"] visibility=[System_SSH_Versuche_Fehlgeschlagen!=0]
    }
....
```
All of the visibility checks fail so no element is visible.

#### Client
* Android version: 11
* Device model: OnePlus 7TPro HD1913
* App version : 2.16.3-beta
* Build flavor : Full
* Device language: German

mueller-ma:
Fixes #2445

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -109,7 +109,7 @@ class WidgetAdapter(
     val itemList: List<Widget> get() = items
     private val widgetsById = mutableMapOf<String, Widget>()
     val hasVisibleWidgets: Boolean
-        get() = items.any { widget -> isWidgetIncludingAllParentsVisible(widget) }
+        get() = items.any { widget -> shouldShowWidget(widget) }
 
     private val inflater = LayoutInflater.from(context)
     private val chartTheme: CharSequence
@@ -257,7 +257,7 @@ class WidgetAdapter(
         }
     }
 
-    private tailrec fun isWidgetIncludingAllParentsVisible(widget: Widget): Boolean {
+    private tailrec fun shouldShowWidget(widget: Widget): Boolean {
         if (!widget.visibility) {
             return false
         }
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -261,6 +261,12 @@ class WidgetAdapter(
         if (!widget.visibility) {
             return false
         }
+        if (widget.type == Widget.Type.Frame) {
+            // Hide frames with no visible children
+            return items
+                .filter { it.parentId == widget.id }
+                .any { it.visibility }
```
maniac103, line 268: This should only return false if there's no visible child, and otherwise defer to the lines below. Otherwise a non-enpty frame will be shown even if one of its parents is invisible (even though I know this is a theoretical case).

mueller-ma: Yes, I saw this issue as well. Since we don't have nested frames, this shouldn't be an issue, but I can change it.

mueller-ma: Maybe the function should be renamend now? `shouldShowWidget`?

maniac103 (approved):

PR #2459, opened 2021-02-15
Determine actual position of first visible widget

### LINKED ISSUE #2455: App draws buffer on top of the first visible frame
The App draws little spaces (buffers) between the frames to differentiate them. So far so good
### Actual behaviour
However the app draws them between all frames and on top of the first one. I think it would look nicer if there would be no buffer on top of the first visible frame. Consequently the top bar (where the sitemap name is written) would line up with the first frame.

I've marked red what should disappear:
Screenshot_20210213-174657

I made a quick sketch and I think it looks way better:
Screenshot_20210213-175014

#### Client
* Android version: 11
* Device model: OnePlus 7T Pro
* App version : 2.16.5-beta
* Build flavor : Full
* Device language: GER

maniac103:
Don't hardcode it to 0, since they might be invisible widgets at the top
of the list.

Fixes #2455

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -114,6 +114,7 @@ class WidgetAdapter(
     private val inflater = LayoutInflater.from(context)
     private val chartTheme: CharSequence
     private var selectedPosition = -1
+    private var firstVisibleWidgetPosition = -1
     private val colorMapper = ColorMapper(context)
 
     interface ItemClickListener {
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -114,6 +114,7 @@ class WidgetAdapter(
     private val inflater = LayoutInflater.from(context)
     private val chartTheme: CharSequence
     private var selectedPosition = -1
+    private var firstVisibleWidgetPosition = -1
```
mueller-ma, line 117: Maybe use a constant here:

```suggestion
    private var selectedPosition = RecyclerView.NO_POSITION
    private var firstVisibleWidgetPosition = RecyclerView.NO_POSITION
```

mueller-ma: @maniac103 What do you think about my suggestion?

maniac103: It's fine ... I just forgot to update the branch :-/ I'll update it tomorrow.

PR #2462, opened 2021-02-15
Load widgets after activity has been restored

### LINKED ISSUE #2220: No widgets are shown when activity is destroyed
### Actual behaviour
When the dev option "Don't keep activities" is enabled, the app shows "This page doesn’t contain any widgets that are set to be visible" after re-entering the app.

### Expected behaviour
When no widgets are cached, the app should get the widget list from the server.

 
### Steps to reproduce
1. Enable "Don't keep activities"
2. Open app
3. Go to home screen
4. Go back to app

Yes

mueller-ma:
Fixes #2220

`PageConnectionHolderFragment`'s `connections` was null after it has
been recreated.


maniac103:
So what's the difference when using this dev option compared to device rotation (which recreates the activity) or swiping the app out of recents (which destroys the activity)? The former case should lead to the connection fragment being retained, the latter case should lead to a new instance creation of both fragment and activity.

mueller-ma:
When using this dev option `onRestoreInstanceState()` of `ContentController` gets a bundle with the fragments, but all the connections are lost.

PR #2470, opened 2021-02-22
Avoid adding two servers during discovery

mueller-ma:
The discovery is started in the intro and MainActivity, but only if no server is configured.
However the second discovery can be started before the first one finds a
server. Then both runs may add a server to the prefs.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
@@ -113,7 +113,7 @@ fun String?.toNormalizedUrl(): String? {
             .toString()
         if (url.endsWith("/")) url else "$url/"
     } catch (e: IllegalArgumentException) {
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/util/ExtensionFuncs.kt
```diff
@@ -499,9 +499,14 @@ fun Uri.Builder.appendQueryParameter(key: String, value: Boolean): Uri.Builder {
 }
 
 fun ServiceInfo.addToPrefs(context: Context) {
+    if (context.getPrefs().getConfiguredServerIds().isNotEmpty()) {
```
maniac103, line 502: I'd prefer doing this at the caller site, since the method name doesnt imply only one service info can ever be added to prefs.

mueller-ma: Done. I added the check to `MainActivity` only, because it's executed after `IntroActivity`.

maniac103 (approved):

PR #2511, opened 2021-04-05
Allow horizontal swipes to trigger swipe-to-refresh

### LINKED ISSUE #1288: Cannot trigger pull to refresh sometimes
### Actual behaviour
When the widget list is scrolled to the top, it is sometimes not possible to swipe down to trigger a refresh. The app behaves like it has no swipe to refresh implemented.

### Expected behaviour
When the widget list is scrolled to the top, it should always be possible to swipe down to trigger a refresh.

 
### Steps to reproduce
1. Swipe to refresh (works)
2. Swipe down again (doesn't work)
3. Swipe down again and again (may work or not)

Yes

#### Client
* Android version: 7.1.2
* Device model: Oneplus X
* Stock or customized ROM: Lineage 14.2
* App version : 2.6.13

mueller-ma:
I noticed that I cannot swipe-to-refresh when there's too much
movement in horizontal direction, e.g. (X is my finger trail):

````
-------------------
|                 |
|                 |
|    XXX          |
|       XX        |
|         X       |
|         X       |
|         X       |
|         X       |
|                 |
-------------------
````

I don't see a reason why horizontal movement is disallowed here, so I
remove the restriction.

Fixes #1288

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/widget/RecyclerViewSwipeRefreshLayout.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/widget/RecyclerViewSwipeRefreshLayout.kt
@@ -22,16 +22,13 @@ import androidx.core.view.NestedScrollingChildHelper
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import androidx.swiperefreshlayout.widget.SwipeRefreshLayout
-import kotlin.math.abs
 
 class RecyclerViewSwipeRefreshLayout(context: Context, attrs: AttributeSet) : SwipeRefreshLayout(context, attrs) {
     private val touchSlop: Int = ViewConfiguration.get(context).scaledTouchSlop
-    private var downX: Float = 0F
     private var downY: Float = 0F
     private var childScrollableOnDown: Boolean = false
     private val parentOffsetInWindow = IntArray(2)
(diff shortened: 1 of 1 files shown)


maniac103:
The reason is disambiguation between horizontal swipes and vertical ones. Child views might want to consume the former, so we shouldn't consume it here.

mueller-ma:
Ah, I see. But IMO there's still place for improvements. I have a webview that can be swiped horizontally, but the swipe may be caught by `RecyclerViewSwipeRefreshLayout`.
With the webview being the first and only item, `X` is where the webview is scrolled horizontally, `Y` is where `RecyclerViewSwipeRefreshLayout` catches the swipe and doesn't swipe-to-refresh. You can see that once a swipe is caught, it's never released to the child again, which makes it IMO a bit hard to scroll webviews horizontally.

````
-------------------
|                 |
|                 |
|    XXX          |
|       YYYYYYY   |
|                 |
|                 |
|                 |
|                 |
-------------------
````

mueller-ma:
The reason why `RecyclerViewSwipeRefreshLayout` was added is because the icon cache had been cleared on every swipe-to-refresh. This isn't the case anymore, so maybe remove `RecyclerViewSwipeRefreshLayout` at all?
I'll check some other apps if they have a similar functionality.

mueller-ma:
Closing for now

PR #2598, opened 2021-05-19
Redirect Gadgetbridge Device Actions to openHAB server

mueller-ma:

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
@@ -101,6 +101,10 @@ class BackgroundTasksManager : BroadcastReceiver() {
                 Log.d(TAG, "DND mode changed")
                 scheduleWorker(context, PrefKeys.SEND_DND_MODE)
             }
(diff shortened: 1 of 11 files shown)


mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -252,27 +256,36 @@ class BackgroundTasksManager : BroadcastReceiver() {
...
+            PrefKeys.SEND_DND_MODE,
+            PrefKeys.SEND_GADGETBRIDGE
         )
         internal val KNOWN_PERIODIC_KEYS = listOf(
             PrefKeys.SEND_BATTERY_LEVEL,
             PrefKeys.SEND_CHARGING_STATE,
             PrefKeys.SEND_WIFI_SSID,
-            PrefKeys.SEND_DND_MODE
+            PrefKeys.SEND_DND_MODE,
+            PrefKeys.SEND_GADGETBRIDGE // Required for EventListenerService
```
mueller-ma, line 280: TODO: This causes the periodic workers to run, but `triggerPeriodicWork()` will never trigger an update for Gadgetbridge, because there's no intent as input.

PR #2624, opened 2021-06-06
Refresh FullscreenImageActivity

### LINKED ISSUE #2571: Honor refresh for fullscreen images
I consider this a follow-up on PR #2411 

I really like the feature to display images in fullscreen in the app.
I use it to show images of the ip cams and zoom in to interesting regions when needed.
However, zooming disables the refresh which is bad for ip cams.

1. Use an widget like ```Image url="http://10.0.0.xxx/webcapture.jpg?command=snap&channel=0" refresh=1000```
2. Enjoy the every-second refreshed image stream
3. Click on the widget to get to fullscreen view
4. The app should honor the refresh interval even in fullscreen and should allow to zoom to regions of interest and still refresh the image

Using the video widget does not (yet, #320 ) allow for fullscreen, so this is no successful route.

### Additional context
Should be clear enough already ;-)

Thank you very much for your great efforts!!! :)

mueller-ma:
Closes #2571

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/FullscreenImageActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/FullscreenImageActivity.kt
@@ -14,6 +14,7 @@
 package org.openhab.habdroid.ui
 
 import android.graphics.BitmapFactory
+import android.graphics.Matrix
 import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/FullscreenImageActivity.kt
```diff
@@ -127,7 +144,21 @@ class FullscreenImageActivity : AbstractBaseActivity() {
         bitmap ?: return finish()
 
         Handler(Looper.getMainLooper()).post {
```
maniac103, line 146: Now that I see it, why does this use a Handler instead of `launch(Dispatchers.Main)`?

mueller-ma: `loadImage()` does some IO, so I originally thought it's better to use the `Dispatchers.IO` here.

PR #2650, opened 2021-07-09
Fix NPE if Base64 bitmap is null

mueller-ma:


mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
```diff
@@ -105,7 +105,13 @@ class WidgetImageView constructor(context: Context, attrs: AttributeSet?) : AppC
     fun setBase64EncodedImage(base64: String) {
         prepareForNonHttpImage()
         val data = Base64.decode(base64, Base64.DEFAULT)
-        val bitmap = BitmapFactory.decodeByteArray(data, 0, data.size)
+        val bitmap: Bitmap? = BitmapFactory.decodeByteArray(data, 0, data.size)
+
+        if (bitmap == null) {
+            removeProgressDrawable()
```
maniac103, line 111: This is already done by `prepareForNonHttpImage()`.

maniac103 (approved):

PR #2657, opened 2021-07-16
Don't needlessly open about:blank for WebView widgets

### LINKED ISSUE #2633: If opening text item which contains also webview(s), then webviews are opened in new browser window
### Actual behaviour
I have a sitemap with expandable Text items.
These text items contains sub-items and one or more webviews.
When I open that text item, new browser window is opened

### Expected behaviour
When opening (expanding) text item, sub-items should be displayed and in-line webview along with them, not in new browser window.

 
### Steps to reproduce
Part of sitemap -
````
	Frame label="label of frame" {
		Text item=Pagrabs_Summary icon="supermarket" {
			Default item=Pagrabs_temperature label="Temperature"
			Default item=Pagrabs_humidity label="Humidity"
			Default item=Pagrabs_lastUpdate
			Webview url="http://192.168.xx.xxx:3000/dashboard-solo/db/saimnieciba?orgId=1&panelId=4&theme=light" height=8 icon=none
			Webview url="http://192.168.xx.xxx:3000/dashboard-solo/db/saimnieciba?orgId=1&panelId=5&theme=light" height=8 icon=none
		}		
		
		Text item=Workshop_temperature {
			Default item=Workshop_lastUpdate
			Webview url="http://192.168.xx.xxx:3000/dashboard-solo/db/saimnieciba?orgId=1&panelId=3&theme=light" height=8 icon=none
		}
	}
````

When clicking on text item "Workshop_temperature" or "Pagrabs_summary", new browser window will open.

#### Client
* Android version: 10 and 9
* Device model: Huawei P30 Pro and Samsung Galaxy Tab A
* App version: 2.17.4
* Build flavor: Full
* Device language: lv, en-us
On both devices Google Chrome is set as default browser.

#### Server
* Server version: 
* Reverse Proxy: No
* Authentication method: None

### PR DESCRIPTION
**Pull request by maniac103**
Doing so causes the the comparisons which check whether to use the WebView or an external browser in ConnectionWebViewClient to fail, as they'll compare the actual URL to about:blank.

Fixes #2633

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -1143,9 +1143,9 @@ class WidgetAdapter(
             }
             with(webView) {
                 adjustForWidgetHeight(widget, 0)
-                loadUrl("about:blank")
 
                 if (url == null) {
+                    loadUrl("about:blank")
                     return
                 }
                 setUpForConnection(connection, url) { progress ->
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -44,6 +44,7 @@ import androidx.appcompat.app.AlertDialog
 import androidx.core.view.children
 import androidx.core.view.get
 import androidx.core.view.isGone
+import androidx.core.view.isInvisible
 import androidx.core.view.isVisible
 import androidx.core.widget.ContentLoadingProgressBar
 import androidx.recyclerview.widget.RecyclerView
@@ -1155,7 +1156,9 @@ class WidgetAdapter(
                         progressBar.show()
                     }
                     progressBar.progress = progress
+                    webView.isInvisible = progress == 0
                 }
+                webView.isInvisible = true
                 loadUrl(url.toString())
             }
         }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt [hunk 1]
```diff
@@ -1143,9 +1143,9 @@ class WidgetAdapter(
             }
             with(webView) {
                 adjustForWidgetHeight(widget, 0)
-                loadUrl("about:blank")
```
mueller-ma, line 1146: AFAIK this is done, because recycled WebViews would show the previous page for a moment otherwise. Maybe change `shouldOverrideUrlLoading()`, so it never opens the browser if the previous or current page is `about:blank`?

maniac103: Maybe set the WebView to invisible until the loading progress is, say, 50%?

mueller-ma: Maybe set the webview to invisible for 0% only. It has a progress bar which should be shown when loading.

maniac103: Done.

mueller-ma: Several WebViews still show the previous page until ~70% progress, so hiding at 0% progress doesn't help. I'd still prefer my suggestion in my first comment here.

mueller-ma: @maniac103 This comment was resolved, but IMO it's currently still open.

PR #2694, opened 2021-08-30
Render section switch as selection when too wide

### LINKED ISSUE #1326: Render switch with mappings as selection

mueller-ma:
Fixes #1326


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -823,6 +825,12 @@ class WidgetAdapter(
             } else {
                 group.check(checkedId)
             }
+
+            // Layout has to be drawn before measuring the width
+            handler.post {
```
maniac103, line 830: There's zero guarantee for the layout to be finished in the next main loop iteration. This needs to be solved by using a layout listener.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -823,6 +825,12 @@ class WidgetAdapter(
...
+                group.isVisible = group.measuredWidth < (parent.width * 0.8)
+                spinner.isVisible = !group.isVisible
```
maniac103, line 832: No further spinner setup needed here? Hardly can believe that...

maniac103 (changes requested):

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: The spinner setup is done in the super class: `SelectionViewHolder`

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mueller-ma: I'll check that.

Flole998:
This commit (or at least I believe it's this one) isn't working at all for me. I'm running version 2.19.1 and I am seeing various results when scrolling up and down so that the items become invisible and visible again. It's a huge mess, I've attached a few screenshots
Screenshot_20210929-023055__01
Screenshot_20210929-023023__01
Screenshot_20210929-023035__01

PR #2718, opened 2021-10-01
Reset section switch visibility to default on bind

### LINKED ISSUE #2710: Section switch rendered as *both* switch and selection
### Actual behaviour
A section switch with exactly one item is (since #2694) sometimes rendered as both switch and selection in one line:

Screenshot_20210925-195338_openHAB_Beta~2

### Expected behaviour
Only one of the items should be shown; since in this particular case there's only one switch it never should be rendered as selection.

 
### Steps to reproduce
Sitemap entry:
```
        Frame label="Squeezebox" {
                Switch item=GroupSqueezeboxPlayers label="Wiedergabe [(%d)]" mappings=[PAUSE="Pause"]
         }
```

No.

#### Client
* Android version: 11
* Device model: OnePlus 6
* App version: 2.19.1 beta
* Build flavor : Full
* Device language: DE

#### Server
* Server version: 3.1
* Reverse Proxy: None
* Authentication method : None

#### openHAB Cloud
* Self hosted cloud or myopenhab.org: myoh
* openHAB Cloud version : 
* Authentication method : 

### Comment

I guess it's related to the layout not changing after the first layout pass. Does the initial visibility of the parts of the row need to be adjusted?

mueller-ma:
Fixes #2710

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -793,7 +793,6 @@ class WidgetAdapter(
         View.OnClickListener {
         private val group: MaterialButtonToggleGroup = itemView.findViewById(R.id.switch_group)
         private val spareViews = mutableListOf<View>()
-        private val handler: Handler = Handler(Looper.getMainLooper())
         private var boundItem: Item? = null
 
         override fun bind(widget: Widget) {
@@ -840,6 +839,9 @@ class WidgetAdapter(
             } else {
                 group.check(checkedId)
             }
+
+            group.isVisible = true
+            spinner.isVisible = false
         }
 
         override fun onLayoutChange(


maniac103:
I know I suggested that approach myself, but I think we'll have to go a step back. What we want to know is whether the button group width exceeds 80% of the `itemView`s width. After writing the suggestion in #2710, I realized that I would expect a relayout to happen anyway whenever we change the contents of the buttons (that is, either button count or button text). However, it looks like the layout change listener isn't called as expected - probably because the parent's view bounds simply don't change after the first layout pass.

My suggestion would be this diff (lightly tested only):
```diff
diff --git a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
index e7195e57..e8d38d32 100644
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -30,6 +30,7 @@ import android.view.MotionEvent
 import android.view.View
 import android.view.ViewConfiguration
 import android.view.ViewGroup
+import android.view.ViewTreeObserver
 import android.webkit.WebView
 import android.widget.ArrayAdapter
 import android.widget.Button
@@ -331,7 +332,7 @@ class WidgetAdapter(
         inflater: LayoutInflater,
         val parent: ViewGroup,
         @LayoutRes layoutResId: Int
-    ) : RecyclerView.ViewHolder(inflater.inflate(layoutResId, parent, false)), View.OnLayoutChangeListener {
+    ) : RecyclerView.ViewHolder(inflater.inflate(layoutResId, parent, false)) {
         open val dialogManager: DialogManager? = null
         var started = false
             private set
@@ -340,7 +341,6 @@ class WidgetAdapter(
         fun start() {
             if (!started) {
                 onStart()
-                parent.addOnLayoutChangeListener(this)
                 started = true
             }
         }
@@ -349,25 +349,12 @@ class WidgetAdapter(
                 return false
             }
             onStop()
-            parent.removeOnLayoutChangeListener(this)
             started = false
             return true
         }
         open fun onStart() {}
         open fun onStop() {}
         open fun handleRowClick() {}
-
-        override fun onLayoutChange(
-            v: View?,
-            left: Int,
-            top: Int,
-            right: Int,
-            bottom: Int,
-            oldLeft: Int,
-            oldTop: Int,
-            oldRight: Int,
-            oldBottom: Int
-        ) {}
     }
 
     abstract class LabeledItemBaseViewHolder internal constructor(
@@ -790,10 +777,9 @@ class WidgetAdapter(
         private val connection: Connection,
         colorMapper: ColorMapper
     ) : SelectionViewHolder(inflater, parent, connection, colorMapper, R.layout.widgetlist_sectionswitchitem),
-        View.OnClickListener {
+        ViewTreeObserver.OnGlobalLayoutListener, View.OnClickListener {
         private val group: MaterialButtonToggleGroup = itemView.findViewById(R.id.switch_group)
         private val spareViews = mutableListOf<View>()
-        private val handler: Handler = Handler(Looper.getMainLooper())
         private var boundItem: Item? = null
 
         override fun bind(widget: Widget) {
@@ -842,18 +828,18 @@ class WidgetAdapter(
             }
         }
 
-        override fun onLayoutChange(
-            v: View?,
-            left: Int,
-            top: Int,
-            right: Int,
-            bottom: Int,
-            oldLeft: Int,
-            oldTop: Int,
-            oldRight: Int,
-            oldBottom: Int
-        ) {
-            group.isVisible = group.measuredWidth < (parent.width * 0.8)
+        override fun onStart() {
+            super.onStart()
+            itemView.viewTreeObserver.addOnGlobalLayoutListener(this)
+        }
+
+        override fun onStop() {
+            super.onStop()
+            itemView.viewTreeObserver.removeOnGlobalLayoutListener(this)
+        }
+
+        override fun onGlobalLayout() {
+            group.isVisible = group.measuredWidth < (itemView.width * 0.8)
             spinner.isVisible = !group.isVisible
         }
 
```

mueller-ma:
I applied your patch locally. The visibility issue is fixed with it, but scrolling isn't as smooth as before, at least on my testing sitemap with ~20 section switches. I can see that `onGlobalLayout()` is called very often.

maniac103:
It is called pretty often, but the code inside the method should not consume much CPU time, assuming the visibility of `group`/`spinner` in reused views doesn't change. If the visibility *does* change, I don't think there's anything we can do because it's the relayout passes which are consuming the time in that case.
Unfortunately there's no 'subtree layout changed' listener, the only alternative to my patch is subclassing the `Spinner` and `RadioGroup` classes and listening for layout there. I doubt that's worth the effort though, I'd rather remove this auto-change magic altogether in that case.

mueller-ma:
> but scrolling isn't as smooth as before, at least on my testing sitemap with ~20 section switches

Nevermind, it's also not smooth with current main.

PR #2738, opened 2021-10-27
Android 11 device controls

### LINKED ISSUE #2210: Support external device controls on Android 11
There's a new API for controlling external devices on Android 11 and above: https://developer.android.com/guide/topics/ui/device-control
It might be a bit tricky, because there's a mandatory function to generate all available controls and we probably don't want to publish all Items there.

wiomoc:
Implements basic functionality for controlling items over the [device control API](https://developer.android.com/guide/topics/ui/device-control#kotlin) released in Android 11

closes #2210


mueller-ma:
I tested this with the demo server and have a few questions/suggestions:

- [ ] Item `RGBLight` does always show `0%` when the light is `ON`.
- [ ] Do you want to add support for more item types in this PR?
- [ ] When long pressing a control it always only opens MainActivity, but I'm not sure if there's another practical solution. Maybe ensure it opens the primary server at least? Look for `EXTRA_SERVER_ID` in `MainActivity`.
- [ ] Can you pick https://github.com/openhab/openhab-android/commit/cdd8aab1f7d0d3c5dfe5a0cb8be2dab889d876e2?
- [ ] The lint issue `Missing PendingIntent mutability flag` is showing up in line 237.

wiomoc:
I have taken into account all your suggestions in the last commit. I have also added support for the number type. Now I don't want to add more types 😉

mobile/src/main/java/org/openhab/habdroid/model/Item.kt
```diff
@@ -175,6 +169,15 @@ fun JSONObject.toItem(): Item {
         }
     }
 
+    fun JSONObject.getDoubleOrNull(name: String) = if (has(name)) {
```
mueller-ma, line 172: This function should be placed next to `JSONObject.optStringOrNull()` and be renamend to `JSONObject.optDoubleOrNull`.

mobile/src/main/java/org/openhab/habdroid/model/Item.kt
```diff
@@ -192,7 +195,10 @@ fun JSONObject.toItem(): Item {
         members,
         options,
         state.toParsedState(numberPattern),
-        tags
+        tags,
+        minimum,
+        maximum,
+        step
```
mueller-ma, line 201: I'd parse the json here, no need for these variables.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -177,6 +179,12 @@ class ItemsControlsProviderService : ControlsProviderService() {
         }
     }
 
+    private fun createRangeTemplate(item: Item, format: String) = RangeTemplate(
+        item.name, item.minimum?.toFloat() ?: 0F, item.maximum?.toFloat() ?: 100F,
```
mueller-ma, line 183: `Number` items may have a larger value than 100, but no `maximum` defined, e.g. `AlarmSystem_PinCode` on the demo server:
````
AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-3
    Process: org.openhab.habdroid, PID: 6132
    java.lang.IllegalArgumentException: currentValue=1234.000000 > maxValue=100.000000
        at android.service.controls.templates.RangeTemplate.validate(RangeTemplate.java:181)
        at android.service.controls.templates.RangeTemplate.<init>(RangeTemplate.java:83)
        at org.openhab.habdroid.background.ItemsControlsProviderService.createRangeTemplate(ItemsControlsProviderService.kt:182)
        at org.openhab.habdroid.background.ItemsControlsProviderService.maybeCreateStatefulControl(ItemsControlsProviderService.kt:203)
        at org.openhab.habdroid.background.ItemsControlsProviderService.access$maybeCreateStatefulControl(ItemsControlsProviderService.kt:55)
        at org.openhab.habdroid.background.ItemsControlsProviderService$createPublisherForAllAvailable$job$1.invokeSuspend(ItemsControlsProviderService.kt:100)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
````

**Review by mueller-ma — CHANGES_REQUESTED** [review]
I see the following line multiple times in the log:

> W/ControlsProviderService: onNext(): control is not stateless. Use the Control.StatelessBuilder() to build the control.

When `createPublisherForAllAvailable()` is called, it should use `Control.StatelessBuilder()`: https://developer.android.com/guide/topics/ui/device-control

File: mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt [hunk 4]
```diff
@@ -0,0 +1,337 @@
...
+
+            return Control.StatefulBuilder(item.name, mainActivityPendingIntent)
+                .setTitle(item.label)
+                .setDeviceType(getDeviceType(item))
+                .setControlTemplate(controlTemplate)
+                .setStatus(Control.STATUS_OK)
+                .build()
+        } else {
+            return when (item.type) {
+                Item.Type.Switch, Item.Type.Dimmer, Item.Type.Color, Item.Type.Rollershutter, Item.Type.Number ->
```
mueller-ma, line 267: This list of valid item types has to be maintained here and in the assignment of `controlTemplate` above. I'd assign `controlTemplate` unconditionally and keep the `return null` for unsupported item types there.
Then this list of types can be removed here.

mueller-ma:
Thanks!

PR #2752, opened 2021-11-11
Don't redact myopenhab.org in LogActivity

### LINKED ISSUE #2746: Don't redact myopenhab.org in LogActivity
It's useful to see if a user is using myopenhab.org as remote server. Since this isn't a private address, it doesn't have to be redacted.

References: 
* Function that redacts the urls: https://github.com/openhab/openhab-android/blob/2.19.5-beta/mobile/src/main/java/org/openhab/habdroid/ui/LogActivity.kt#L222-L225
* Regex matching valid myopenhab.org urls: https://github.com/openhab/openhab-android/blob/2.19.5-beta/mobile/src/main/java/org/openhab/habdroid/ui/PreferencesActivity.kt#L962

wiomoc:
Closes #2746


mueller-ma (approved):
LGTM!

PR #2753, opened 2021-11-11
Parse semantic tags

### LINKED ISSUE #2742: Update Item tag detection
The demo server uses some tags that cannot be parsed by the app, e.g. https://demo.openhab.org/rest/items/Light_FF_Daughter_Ceiling

grafik

These tags should be handled in `String?.toItemTag()`, so they can be used in features like #2210.

mueller-ma:
Use these tags to select a device type for device control.

Closes #2742

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Item.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Item.kt
@@ -353,6 +353,14 @@ fun String?.toItemType(): Item.Type {
 }
 
 fun String?.toItemTag(): Item.Tag {
+    this ?: return Item.Tag.Unknown
+
+    try {
+        return Item.Tag.valueOf(this)
+    } catch (e: IllegalArgumentException) {
+        // No 1:1 mapping possible, fall through
+    }
+
     return when (this) {
         "Lighting" -> Item.Tag.Light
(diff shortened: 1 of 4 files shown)


mobile/src/main/java/org/openhab/habdroid/model/Item.kt
```diff
@@ -222,15 +354,149 @@ fun String?.toItemType(): Item.Type {
...
-        "Thermostat" -> Item.Tag.Thermostat
+        "CurrentTemperature" -> Item.Tag.Temperature
+        "CurrentHumidity" -> Item.Tag.Humidity
+        "Thermostat" -> Item.Tag.Temperature
         "homekit:HeatingCoolingMode", "homekit:TargetHeatingCoolingMode" -> Item.Tag.HeatingCoolingMode
         "homekit:TargetTemperature", "TargetTemperature" -> Item.Tag.TargetTemperature
-        "WindowCovering" -> Item.Tag.Blind
+        "WindowCovering" -> Item.Tag.Blinds
+
+        "Equipment" -> Item.Tag.Equipment
```
maniac103, line 367: I'd suggest handling the 1:1 mapping in a default case, using `Item.Tag.valueOf()`. That way one can get rid of the boilerplate code.

Issue #2583, opened 2021-05-13
Android white screen on startup with openHAB 3 UI


cybermcm:
### Actual behaviour
openHAB configured to startup with 3 UI, on startup just a white screen with a loading progress bar appears, see screenshot
![Screenshot_20210513-133218_openHAB_Beta](https://user-images.githubusercontent.com/13907742/118120781-b5e2e180-b3f0-11eb-87c6-1310f1962afa.png)
current workaround: doing a task switch to another app and back and UI works

tried to clean app cache, deleted and reinstalled app, cleaned Android webview cache -> no success
startup with HABPanel -> working
startup with HABPanel -> switching to 3UI -> not working (unless I switch apps)

### Expected behaviour
app should startup without issue
 
### Steps to reproduce
happens every time

### Can you reproduce the issue in demo mode?
happens with my own server and demo mode

#### Client
* Android version: 11, LineageOS 18.1, build date 2021-05-09
* Device model: Samsung SM-G900F klte (very old, I know ;-))
* App version: 2.16.20-beta
* Build flavor: Full
* Device language: English

#### App log
```
-----------------------
Device information
Model: SM-G900F
Manufacturer: samsung
Brand: samsung
Device: klte
Product: klte
OS: 11
Display: 1080x1920, 3.0 density
Data usage policy: DataUsagePolicy(canDoLargeTransfers=true, loadIconsWithState=true, autoPlayVideos=true, canDoRefreshes=true), data saver: 1, battery saver: false
-----------------------

--------- beginning of events
05-13 13:36:21.830 16928 16928 I menu_item_selected: [0,Clear log]
05-13 13:36:22.828 16928 16928 I wm_on_top_resumed_lost_called: [58275911,org.openhab.habdroid.ui.LogActivity,topStateChangedWhenResumed]
05-13 13:36:22.838 16928 16928 I wm_on_paused_called: [58275911,org.openhab.habdroid.ui.LogActivity,performPause]
05-13 13:36:23.206 16928 16928 I wm_on_stop_called: [58275911,org.openhab.habdroid.ui.LogActivity,STOP_ACTIVITY_ITEM]
05-13 13:36:23.790 16928 16928 I wm_on_destroy_called: [58275911,org.openhab.habdroid.ui.LogActivity,performDestroy]
--------- beginning of main
05-13 13:36:25.147 17814 17814 I chatty  : uid=10056(org.openhab.habdroid.beta) expire 44 lines
[...]

PR #2783, opened 2021-12-30
WIP: Use GeckoView instead of WebView

mueller-ma:
See https://github.com/openhab/openhab-android/issues/2583#issuecomment-998923964

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -127,6 +127,9 @@ repositories {
     maven {
         url "https://jitpack.io"
     }
+    maven {
+        url "https://maven.mozilla.org/maven2/"
+    }
     google()
 }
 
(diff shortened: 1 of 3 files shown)


mobile/build.gradle
```diff
@@ -168,6 +171,7 @@ dependencies {
     implementation "com.mikepenz:aboutlibraries-core:$about_libraries_version"
     implementation "com.mikepenz:aboutlibraries:$about_libraries_version"
     implementation "com.github.GrenderG:Toasty:1.5.2"
+    implementation "org.mozilla.geckoview:geckoview-beta:96.0.20211219185855"
```
JohannesPtaszyk, line 174: Suggestion: Switch to stable channel to prevent developing against features that are not available yet

mueller-ma:
My build error was caused by too low heap size...

I haven't found a way to replace the `Webview#addJavascriptInterface()` calls yet.

JohannesPtaszyk:
> My build error was caused by too low heap size...
>I haven't found a way to replace the Webview#addJavascriptInterface() calls yet.

Seems like one would have to write a browser extension for injecting JS.
https://bugzilla.mozilla.org/show_bug.cgi?id=1580454

PR #2794, opened 2022-01-10
Improve hiding section switches

mueller-ma:
Use the actual available space for the switches, not 'just' 80% of the device width.


maniac103:
And what about the label? This calculation might hide it completely AFAICT?

mueller-ma:
`labelView.measuredWidth` is always larger than 0, even for items without any label. It's width is set to `0dp`. Setting it to `wrap_content` doesn't seem to make it smaller. Therefore I currently cannot get the real width of the label or value.

mueller-ma:
Sitemaps need a larger rework, see #3097

PR #2833, opened 2022-02-20
Improve 'Open settings' action on error screens

mueller-ma:
Should be rebased after #2832.


mueller-ma:
The server preference screen cannot be left by pressing save.

PR #2847, opened 2022-03-05
Connect to Wi-Fi if permissions are missing

mueller-ma:
Otherwise when the permissions for location access isn't granted, the
background tasks silently fail.


mueller-ma:
#2881 is a better fix for this issue.

PR #2892, opened 2022-04-29
Hide null label in frames

### LINKED ISSUE #2889: Frame without item shows "null" in label
### Actual behaviour
I use `Frame label="Text" { ... }` in my sitemaps. In the app it appears as `Text null`.

### Expected behaviour
There should be no `null` in the label when there is no item.
 
Yes. It shows ´Weather null´ and ´Demo null´ as frame label.

### Environment data
I am using version 2.20.6-beta. I skip the rest as it seems just a bug (a missing condition) related to changes due to issues #2868 and #2880

mueller-ma:
Fixes #2889

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -480,7 +480,7 @@ class WidgetAdapter(
 
         override fun bind(widget: Widget) {
             @SuppressLint("SetTextI18n")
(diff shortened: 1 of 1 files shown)


maniac103:
While at it, I wonder whether we should also visually separate label and state, e.g. by making the state a little smaller or something?

mueller-ma:
Basic UI displays Frames with label as `Windows [3]`, but everything in the same text size. We could do that as well.

mueller-ma:
I copied the behavior from Basic UI.

PR #2894, opened 2022-04-29
Hide microphone icon when voice recognizer is unavailable

### LINKED ISSUE #2887: Option to hide the microphone icon
I do not use voice control, since for my language no voice recognition software exists.  I want to hide the microphone icon from the top right corner of openhab-android.

Screenshot_20220424-105204_openHAB

The screenshot shows at the top a group name, followed by a group value (a value of a real Item), and finally a microphone icon.  The microphone icon prevents that the Group value it fully shown.

* add an option to hide the microphone, or detect automatically that the OpenHAB instance does not utilize voice recognition, and hide the microphone automatically.

mueller-ma:
Fixes https://github.com/openhab/openhab-android/issues/2887

I added the link to the Google app to the docs, because the "No voice recognition app found" snackbar will probably not show up anymore.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -368,7 +368,8 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
 
     override fun onPrepareOptionsMenu(menu: Menu): Boolean {
         CrashReportingHelper.d(TAG, "onPrepareOptionsMenu()")
-        menu.findItem(R.id.mainmenu_voice_recognition).isVisible = connection != null
+        menu.findItem(R.id.mainmenu_voice_recognition).isVisible =
+            connection != null && SpeechRecognizer.isRecognitionAvailable(this)
         menu.findItem(R.id.mainmenu_crash).isVisible = BuildConfig.DEBUG
         return true
     }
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -368,7 +368,8 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
 
     override fun onPrepareOptionsMenu(menu: Menu): Boolean {
         CrashReportingHelper.d(TAG, "onPrepareOptionsMenu()")
-        menu.findItem(R.id.mainmenu_voice_recognition).isVisible = connection != null
+        val isSpeechRecognizerAvailable = SpeechRecognizer.isRecognitionAvailable(this)
+        menu.findItem(R.id.mainmenu_voice_recognition).isVisible = connection != null && isSpeechRecognizerAvailable
         menu.findItem(R.id.mainmenu_crash).isVisible = BuildConfig.DEBUG
```
maniac103, line 373: I think omitting the intermediate variable makes more sense, as that way the check can be skipped if `connection` is `null`.

mueller-ma: Done

maniac103 (approved):

PR #2901, opened 2022-05-06
DOCS: Fix broken link

Trysupe:
Link and its label were reversed

Signed-off-by: Felix Schneider <fs@felix-schneider.org>


mueller-ma:
Thanks

PR #2925, opened 2022-05-26
Make subtitle in device controls configurable

### LINKED ISSUE #2904: Android Device Controls: subtitle
Android Device Controls, subtitles missing. Opportunity to display useful Information like "item.name" missed.

Use` .setSubtitle() `to add additional Information like location or name of parent item. 

I added: 
`.setSubtitle(item.name)`
under line 319 and under line 326 in the file `ItemsControlsProviderService.kt` (path: \openhab-android-main\mobile\src\main\java\org\openhab\habdroid\background)

The results can be seen in the screenshots below.

Showing the item location, instead of item name, would also be a good alternative.

### Additional context
**Current state (used paint to remove subtitles, as I use the app with my the implemented code now):**

**With implemented Code:**

mueller-ma:
Fixes #2904

Use the semantic model for getting the location: https://github.com/openhab/openhab-core/issues/2977#issuecomment-1134772074


mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -201,6 +206,30 @@ class ItemsControlsProviderService : ControlsProviderService() {
...
     }
 
+    private fun getItemLocation(item: Item, allItems: List<Item>): String? {
+        val groups = mutableListOf<Item>()
+        item.groupNames.forEach { groupName ->
+            val group = allItems.first { item -> item.name == groupName }
+            // Check if item is in a location group
+            val isLocation = group.tags.any { tag -> tag in LOCATION_TAGS }
+            if (isLocation) {
+                return group.label
```
mueller-ma, line 216: I'm not sure if we should show the label of the group tagged with a location tag or if we should translate the tag to a string.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -323,10 +352,12 @@ class ItemsControlsProviderService : ControlsProviderService() {
             else -> return null
         }
 
+        val location = getItemLocation(item, allItems).orEmpty()
+
         return if (stateful) {
             Control.StatefulBuilder(item.name, mainActivityPendingIntent)
                 .setTitle(item.label)
-                .setSubtitle(item.name)
+                .setSubtitle(location)
```
mueller-ma, line 360: Also set location in `setZone()`, so the long item list when adding items is divided into multiple pages?

maniac103:
Suggestion: not only use location, but also equipment tags. At least for my personal tagging this works perfectly, because the equipment also contains the location name (e.g. `Squeezebox Bad`). Implementation here: https://github.com/maniac103/openhab.android/commits/item-control-location

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
maniac103: I'd use group label for `Item.Tag.Location` and a localized name in case of the specialized tags.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
maniac103: I'm torn here: on one hand grouping the long list makes sense, on the other hand, when implementing the suggestion to prefer equipment over location tags, it probably won't help much.

maniac103: Added another commit to my repo that combines both: use location for zone and subtitle for equipment.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
mueller-ma: Sounds like a good idea. I guess we have to add a section to the docs on how the subtitle is generated.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -140,17 +140,22 @@ class ItemsControlsProviderService : ControlsProviderService() {
...
+                emptyList()
+            }
+
+            itemNames
+                .map { itemName ->
+                    allItems.first { item -> item.name == itemName }
                 }
+                .forEach { item ->
+                    maybeCreateControl(item, allItems, primaryServerName, true)?.let { control ->
+                        items[item.name] = item
```
maniac103, line 156: Seeing this now: could probably move this out and just create the map from `allItems`.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -201,6 +206,34 @@ class ItemsControlsProviderService : ControlsProviderService() {
...
+            }
+        }
+
+        // If none of the groups is location or equipment, recursively check parent groups
+        groups.forEach { group ->
+            val label = getItemTagLabel(group, allItems, type)
+            if (label != null) {
+                return label
+            }
+        }
```
maniac103, line 232: Can simplify this using `firstNotNull`.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
mueller-ma: At least on my device (LineageOS 18) the item are grouped by zone only in the "Add elements" screen. After adding them, the zones are gone and therefore I don't see the location.

I pushed a commit that displays the location and equipment in the subtitle.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -323,22 +358,23 @@ class ItemsControlsProviderService : ControlsProviderService() {
             else -> return null
         }
 
+        val location = getItemTagLabel(item, allItems, Item.Tag.Location).orEmpty()
+        val equipment = getItemTagLabel(item, allItems, Item.Tag.Equipment).orEmpty()
+
+        val statefulControl = Control.StatefulBuilder(item.name, mainActivityPendingIntent)
+            .setTitle(item.label)
+            .setSubtitle("$location $equipment")
```
maniac103, line 366: FWIW, I don't like this, because
- putting location there seems redundant because it's already the zone label
- it may be just me, but at least I already refer to the location in the equipment name.

For me, subtitles will have the form `Wohnzimmer Squeezebox Wohnzimmer` now.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
maniac103: > I pushed a commit that displays the location and equipment in the subtitle.

As commented, that doesn't match (at least) my style of tagging. As I doubt I'm the only one tagging that way, I wonder whether we could/should make this optional? Can one have settings activities for device controls?

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
mueller-ma: IMO both options (include location in equipment name or not) seems valid to me. We should add a setting here to control the content of the subtitle:

* Equipment 
* Location and equipment
* Item name?
* Empty?

I'm not sure if the two options marked with `?` are useful.

maniac103: > I'm not sure if the two options marked with ? are useful.

Empty IMHO isn't, item name seems semi useful to me. What's missing in the list is 'Location only' ... wouldn't that also make sense?

mueller-ma: I pushed a commit with a settings.

What about making the zone depending on the subtitle? That's for users that don't use a semantic model at all: https://github.com/openhab/openhab-android/pull/2925/commits/ba6ad274f47cebb090efea12f6cf730e5919fd21#diff-45e816015be826032960e39d5cd817fff58308b24645ad9db3af2374d831840fR376-R379

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
mueller-ma: What exactly do you mean with that?

maniac103: Pull filling `items` out of the `forEach` and fill it separately, using a simple `map`.

maniac103: Can even remove `items` completely and make `allItems` a `Map<String, Item>`.

maniac103: Basically this: https://github.com/maniac103/openhab.android/commit/9da126d1502242e6f94e17179a26108c9c81f76f

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
maniac103: I didn't try yet, but sounds like a good idea.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
mueller-ma: Done: https://github.com/openhab/openhab-android/pull/2925/commits/3e16b0a8c59b4c65dc1c8f76b4e34e43fed13474

docs/USAGE.md
```diff
@@ -313,6 +314,11 @@ In case of an error the plugin returns an error code.
 | 11         | The app couldn't establish a connection                                                                                                                                                                    |
 | 1000+      | A connection was established, but an error occurred. The error code is 1000 + the HTTP code, e.g. 1401 means [Unauthenticated](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors). |
 
+### Quick Access Device Controls
+
+[On Android 11 and above](https://www.android.com/intl/en_US/android-11/#a11-device-controls-article) you can long press the power button to control openHAB Items.
```
maniac103, line 319: Referring to the power button may be a little too specific ... In Android 12 they moved device controls to the quick settings tiles :-/

maniac103 (approved):

Nadahar:
As I also noted on Crowdin, I don't think this string describes the setting very well:
https://github.com/openhab/openhab-android/blob/ab128c13d8227bd24c33a71ec58b70d753dc7391/mobile/src/main/res/values/strings.xml#L141

This option is found under the "main settings" in the app, and there's not much context there that give it more meaning. `Device control subtitle` leaves a lot open for interpretation, and my first understanding was that it was about "Device controlled subtitles/caption" when playing video media. The German translation is `Beschreibung für Android-Gerätesteuerung`, which means something very different as I understand it, something like "Description for Android device-control". It's better, but still not very clear IMO.

I've tried to study the source code, and it seems to me like this is about a "tile feature" introduced in recent versions of Android. It's still unclear to me if this only applies to watches/wearables, or if it can also be useful on mobile devices. Regardless, I think it should include "tile" in some form or shape to make the connection clear.

Since I'm still not familiar with the "tile concept", I'm not sure, but I'm thinking that `Item tile subtitle` or maybe `Item tile description` would be a better fit..?

mueller-ma:
I created a new issue for this: #2970

PR #2963, opened 2022-06-29
Clean up / simplify device controls service

maniac103:
Make the service implementation use coroutines, and convert SSE handling to coroutines as well.
Additionally, refactor device control creation to happen from a factory object which encapsulates all information required in the creation process.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
@@ -35,20 +35,15 @@ import kotlin.math.max
 import kotlin.math.min
 import kotlin.math.roundToInt
 import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.jdk9.flowPublish
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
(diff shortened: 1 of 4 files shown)


mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -62,395 +59,315 @@ import org.openhab.habdroid.util.orDefaultIfEmpty
...
-            )
-            Item.Type.Rollershutter -> createRangeTemplate(item, "%.0f%%")
-            Item.Type.Number -> createRangeTemplate(
-                item,
-                item.state?.asNumber?.unit?.let { "%.0f $it" } ?: "%.0f"
+            val pi = PendingIntent.getActivity(
+                context,
+                requestCode,
+                intent,
+                PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
```
mueller-ma, line 232:
```suggestion
                PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent_Immutable
```

mueller-ma (changes requested):
LGTM except my comment below.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
maniac103: Done

PR #2964, opened 2022-06-29
Refactor data saver handling

maniac103:
- Make data transfer policy depend on whether the used connection is
  metered or not, and only apply data saver logic on metered connections
- Make sure battery saver does not override data saver constraints
- Remove unneeded NetworkTypeApiHelper variants; all needed APIs are
  available in API 21 and newer

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
@@ -183,7 +183,7 @@ class OpenHabApplication : MultiDexApplication() {
         }
 
         override fun onReceive(context: Context, intent: Intent?) {
-            val wasDataSaverActive = systemDataSaverStatus
+            val prevSystemDataSaverStatus = systemDataSaverStatus
             val wasBatterySaverActive = batterySaverActive
             when (intent?.action) {
                 ConnectivityManager.ACTION_RESTRICT_BACKGROUND_CHANGED -> {
@@ -196,7 +196,7 @@ class OpenHabApplication : MultiDexApplication() {
                 }
                 else -> return
             }
-            if (wasDataSaverActive != systemDataSaverStatus || wasBatterySaverActive != batterySaverActive) {
+            if (prevSystemDataSaverStatus != systemDataSaverStatus || wasBatterySaverActive != batterySaverActive) {
                 dataUsagePolicyListeners.forEach { l -> l.onDataUsagePolicyChanged() }
             }
         }
--- a/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
@@ -36,14 +36,12 @@ import org.openhab.habdroid.BuildConfig
 import org.openhab.habdroid.background.BackgroundTasksManager
 import org.openhab.habdroid.core.connection.ConnectionFactory
 import org.openhab.habdroid.util.CrashReportingHelper
-import org.openhab.habdroid.util.DataUsagePolicy
-import org.openhab.habdroid.util.determineDataUsagePolicy
 import org.openhab.habdroid.util.getDayNightMode
 import org.openhab.habdroid.util.getPrefs
 
 class OpenHabApplication : MultiDexApplication() {
     interface OnDataUsagePolicyChangedListener {
-        fun onDataUsagePolicyChanged(newPolicy: DataUsagePolicy)
+        fun onDataUsagePolicyChanged()
     }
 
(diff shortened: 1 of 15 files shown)


mobile/src/main/java/org/openhab/habdroid/core/OpenHabApplication.kt
```diff
@@ -185,7 +183,8 @@ class OpenHabApplication : MultiDexApplication() {
         }
 
         override fun onReceive(context: Context, intent: Intent?) {
-            val oldPolicy = determineDataUsagePolicy()
+            val wasDataSaverActive = systemDataSaverStatus
```
mueller-ma, line 186:
```suggestion
            val prevSystemDataSaverStatus = systemDataSaverStatus
```

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt
```diff
@@ -155,45 +148,7 @@ interface ConnectionManagerHelper {
...
-                        ConnectivityManager.TYPE_MOBILE -> ConnectionType.Mobile(network)
-                        else -> ConnectionType.Unknown(network)
-                    } }
-        }
-    }
-
-    @TargetApi(26)
-    private class NetworkTypeHelperApi26 constructor(context: Context) {
+    private class NetworkTypeHelper constructor(context: Context) {
         private val connectivityManager = context.getSystemService(ConnectivityManager::class.java)!!
```
mueller-ma, line 152: AS: `Call requires API level 23 (current min is 21): android.content.Context#getSystemService`

mobile/src/main/java/org/openhab/habdroid/ui/ChartWidgetActivity.kt
```diff
@@ -148,12 +158,7 @@ class ChartWidgetActivity : AbstractBaseActivity(), SwipeRefreshLayout.OnRefresh
     }
 
     private fun loadChartImage(force: Boolean) {
-        val connection = ConnectionFactory.activeUsableConnection?.connection
-        if (connection == null) {
-            finish()
-            return
-        }
-
+        val conn = connection ?: return
```
mueller-ma, line 161:
```suggestion
        val conn = connection ?: return finish()
```

as in `mobile/src/main/java/org/openhab/habdroid/ui/ImageWidgetActivity.kt`

mueller-ma (changes requested):
Overall looks good to me.

Maybe pick the following patch, so the number in the class name is equal to the min api?

````diff
diff --git a/mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt b/mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt
index ce3880913..6213fa1fa 100644
--- a/mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt
@@ -58,18 +58,19 @@ interface ConnectionManagerHelper {
     }

     private class HelperApi21 constructor(context: Context) :
-        ConnectionManagerHelper, ChangeCallbackHelperApi25(context) {
+        ConnectionManagerHelper, ChangeCallbackHelperApi21(context) {
         private val typeHelper = NetworkTypeHelper(context)
         override val currentConnections: List<ConnectionType> get() = typeHelper.currentConnections
     }

+    @TargetApi(26)
     private class HelperApi26 constructor(context: Context) :
         ConnectionManagerHelper, ChangeCallbackHelperApi26(context) {
         private val typeHelper = NetworkTypeHelper(context)
         override val currentConnections: List<ConnectionType> get() = typeHelper.currentConnections
     }

-    private open class ChangeCallbackHelperApi25 constructor(context: Context) : BroadcastReceiver() {
+    private open class ChangeCallbackHelperApi21 constructor(context: Context) : BroadcastReceiver() {
         var changeCallback: ConnectionChangedCallback? = null
         private val context = context.applicationContext

````

mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionManagerHelper.kt
maniac103: Weird that it didn't show that error for me, because it surely is one. Fixed now.

mobile/src/main/java/org/openhab/habdroid/ui/ChartWidgetActivity.kt
maniac103: This seems a bit redundant, since `onResume()`, which is the only place updating `connection`, already calls `finish()` if the connection is `null`. But it doesn't hurt I guess, so I added it here as well.

PR #2969, opened 2022-07-05
Remove untranslatable string from strings.xml

Nadahar:
There is one string in `strings.xml` where there's nothing to translate. I see that you have added some explanatory context, but I argue that it doesn't really help a translator much in this case. What exactly should a translator do? Replace the colon? It's very hard for a translator to know what the consequences of that would be, and I don't know if there's a language where such a change would even be viable.

I see it instead as an "obstacle" that confuse translators. Not a big deal by itself, but many small streams...

I have never seen Kotlin code before, and have absolutely no idea about its syntax. I don know Java well though, and this change was simple enough that I gave it a shot. I managed to get it running on an emulator in Android Studio, although I didn't figure out how to trigger using that exact string. At least it's compiling/running.

I made this as a consequence of [this post](https://community.openhab.org/t/general-translation-directives/109812/9), but I see now that even this tiny change exceeds the [tiny patch exception](https://www.openhab.org/docs/developer/contributing.html#small-patch-exception), so it probably will never be merged even if you happen to agree with the change itself. I'll let it stay for a little while anyway, since I've already done it, so that one of you can use it as a "guide" to make the same change if you so wish.


mobile/src/main/res/values/strings.xml
```diff
@@ -551,8 +551,6 @@
     <string name="item_update_widget_content_description">Widget for changing Item state</string>
     <string name="item_update_widget_content_description_edit">Edit widget</string>
     <string name="item_update_widget">Widget</string>
-    <!-- %1$s: Item label, %2$s: Action, e.g. "Lights: On" -->
-    <string name="item_update_widget_text">%1$s: %2$s</string>
```
maniac103, line 555: Since this string is used from 3 different places, it's arguably useful, if only for consistency. For hiding it from translation, it should be sufficient to add the `translatable="false"` attribute, which makes your patch a tiny one again ;-)

wborn: Perhaps it would also make sense to keep it translatable for a [right to left](https://en.wikipedia.org/wiki/Right-to-left_script) language like Arabic?

Nadahar: > Since this string is used from 3 different places, it's arguably useful, if only for consistency.
You could argue that from a developer point of view, but from a translator point of view it's just a confusing.

> For hiding it from translation, it should be sufficient to add the translatable="false" attribute, which makes your patch a tiny one again ;-)
Knowing that a little earlier would have saved me more than an hour of Android Studio updates to make the code compile... but yeah, if that's possible it would seem to be the best solution. There are some names in there too, like `openHAB` and the like, that probably should have the attribute too.

> Perhaps it would also make sense to keep it translatable for a [right to left](https://en.wikipedia.org/wiki/Right-to-left_script) language like Arabic?
I don't want to be too assertive since it's been a couple of years since I've debugged RTL UIs, but I'm pretty sure that it works just fine RTL too. The strings are "printed" the other way as well, so everything is "reversed" and the labels end up on the right and the values on the left.

mueller-ma: What I learned when working on localization is: Don't make assumptions about other languages. There are so many languages with so many special cases.

To me it also seems like there's nothing to translate, but in French you add additional spaces around colons: `%1$s : %2$s` Also see https://storylearning.com/learn/french/french-tips/french-punctuation

Because of that I'd keep the string as is to be able to provide a correct French translation.

mueller-ma: > I know about the French spacing habits, but I consider it more "decoration" that punctuation to be honest. 

IMO this should be decided by a French speaker/translator.

> Not translating strings that are equal to the source usually works fine, since that means the source string will be used 

IMO it should be "translated" by using the same string as the source as translation. This way other translators see this string as "done".

> So, this probably means that all translators should copy the source string manually for these?

Crowdin suggests to just copy the string, so it's only one click:

grafik

> I understand that there's no interest in doing anything to make this easier to understand/handle for translators, so I'm closing this.

TBH that's not true. This string should be different in at least one language. I agree that an additional hint like `Use source string as translation if punctuation in your language is similar to English` is useful here.

PR #2976, opened 2022-07-17
Improve strings

### LINKED ISSUE #2972: Add more context to string "item_update_widget_text"
> IMO this should be decided by a French speaker/translator.

I agree, it was just a statement to explain why I hadn't "taken this to heart" enough to remember it.

> TBH that's not true. This string should be different in at least one language. I agree that an additional hint like Use source string as translation if punctuation in your language is similar to English is useful here.

I agree that it should stay translatable, I would just wish that some extra context could have been given.

_Originally posted by @Nadahar in https://github.com/openhab/openhab-android/pull/2969#discussion_r916069318_

mueller-ma:
* Add context to all strings with more than one placeholder
* Write some terms with a capital first letter

Fixes #2972

### DIFF
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -558,7 +558,7 @@
     <string name="item_update_widget_content_description">Widget for changing Item state</string>
     <string name="item_update_widget_content_description_edit">Edit widget</string>
     <string name="item_update_widget">Widget</string>
-    <!-- %1$s: Item label, %2$s: Action, e.g. "Lights: On". Use the source string as translation if the punctuation rules are same as in English. -->
+    <!-- %1$s: Item label, %2$s: Action, e.g. "Lights: On". Use the source string as translation if the punctuation rules are the same as in English. -->
     <string name="item_update_widget_text">%1$s: %2$s</string>
     <string name="item_update_widget_item_picker_title">Select an Item</string>
     <string name="widget_show_state">Append current state to name</string>
--- a/mobile/src/main/res/values/strings.xml
+++ b/mobile/src/main/res/values/strings.xml
@@ -558,7 +558,7 @@
     <string name="item_update_widget_content_description">Widget for changing Item state</string>
     <string name="item_update_widget_content_description_edit">Edit widget</string>
     <string name="item_update_widget">Widget</string>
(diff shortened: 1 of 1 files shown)


mueller-ma:
@Nadahar FYI

Nadahar:
@mueller-ma Great work :) I know it's not merged yet, but I still went through and updated the affected strings at Crowdin - so I hope it will be merged :angel:

mobile/src/main/res/values/strings.xml
```diff
@@ -551,13 +558,14 @@
     <string name="item_update_widget_content_description">Widget for changing Item state</string>
     <string name="item_update_widget_content_description_edit">Edit widget</string>
     <string name="item_update_widget">Widget</string>
-    <!-- %1$s: Item label, %2$s: Action, e.g. "Lights: On" -->
+    <!-- %1$s: Item label, %2$s: Action, e.g. "Lights: On". Use the source string as translation if the punctuation rules are similar to English. -->
```
Nadahar, line 561:
```suggestion
    <!-- %1$s: Item label, %2$s: Action, e.g. "Lights: On". Use the source string as translation if the punctuation rules are the same. -->
```

mueller-ma: Maybe "Use the source string as translation if the punctuation rules are the same as in English"?

Nadahar: Sure, I just wanted to get rid of "similar".

mobile/src/main/res/values/strings.xml
```diff
@@ -551,13 +558,14 @@
     <string name="item_update_widget_content_description">Widget for changing Item state</string>
     <string name="item_update_widget_content_description_edit">Edit widget</string>
     <string name="item_update_widget">Widget</string>
-    <!-- %1$s: Item label, %2$s: Action, e.g. "Lights: On" -->
+    <!-- %1$s: Item label, %2$s: Action, e.g. "Lights: On". Use the source string as translation if the punctuation rules are same as in English. -->
```
Nadahar, line 561:
```suggestion
    <!-- %1$s: Item label, %2$s: Action, e.g. "Lights: On". Use the source string as translation if the punctuation rules are the same as in English. -->
```

PR #2982, opened 2022-07-24
Item control: Support option selection

mueller-ma:


mobile/src/main/java/org/openhab/habdroid/ui/SelectionItemActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    companion object {
+        private val TAG = SelectionItemActivity::class.java.simpleName
+
+        const val EXTRA_ITEM = "item"
+    }
+}
+
+class SelectionAdapter(context: Context, val item: Item) : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
```
mueller-ma, line 72: Maybe there's a better solution than re-inventing a radio button list with RecyclerView, but I haven't found one.

PR #3026, opened 2022-09-01
Minify app

mueller-ma:
On a test device this reduced the app size from 60MB to 30MB.

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -56,7 +56,8 @@ android {
     }
     buildTypes {
         release {
-            minifyEnabled false
+            minifyEnabled true
+            shrinkResources true
             proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
         }
         debug {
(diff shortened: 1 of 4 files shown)


mobile/build.gradle
```diff
@@ -56,11 +56,14 @@ android {
...
-            minifyEnabled false
+            minifyEnabled true
+            shrinkResources true
             proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
         }
         debug {
-            minifyEnabled false
+            minifyEnabled true
+            shrinkResources true
+            proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
```
maniac103, line 66: Why minify the debug build? We don't ship it to user's, and from my experience it'll make debugging more complicated (e.g. due to optimized out variables and the like).

mueller-ma: I wanted to minify the debug build, so issues with removing too much resources or classes can be seen when debugging the app, but I can revert it.

PR #3040, opened 2022-09-24
Improve execution of periodic work

mueller-ma:
* Add a function to do several kind of periodic work: `doPeriodicWork()`
* Call `doPeriodicWork()` instead of `BackgroundTasksManager.triggerPeriodicWork()` and `ItemUpdateWidget.updateAllWidgets()`

With this change notifications are polled when starting the app.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/PeriodicItemUpdateWorker.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/PeriodicItemUpdateWorker.kt
@@ -24,21 +24,21 @@ import org.openhab.habdroid.ui.homescreenwidget.ItemUpdateWidget
 class PeriodicItemUpdateWorker(val context: Context, params: WorkerParameters) : Worker(context, params) {
     override fun doWork(): Result {
         Log.d(TAG, "doWork()")
-        doPeriodicWork(context)
+        runBlocking {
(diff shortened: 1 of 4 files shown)


mobile/src/main/java/org/openhab/habdroid/background/PeriodicItemUpdateWorker.kt
```diff
@@ -22,13 +23,23 @@ import org.openhab.habdroid.ui.homescreenwidget.ItemUpdateWidget
...
+
+    companion object {
+        private val TAG = PeriodicItemUpdateWorker::class.java.simpleName
+
+        fun doPeriodicWork(context: Context) {
+            Log.d(TAG, "doPeriodicWork()")
+            BackgroundTasksManager.scheduleUpdatesForAllKeys(context)
+            ItemUpdateWidget.updateAllWidgets(context)
+            if (CloudMessagingHelper.needsPollingForNotifications(context)) {
+                runBlocking {
```
maniac103, line 39: This needs to happen in a background thread, but not all callers of this run in a background thread.

mueller-ma: Good catch 👍

Issue #2799, opened 2022-01-15
Migrating to Material Design 3


mueller-ma:
https://material.io/blog/migrating-material-3

Issue #1058, opened 2018-10-21
"Info ssl client cert" tooltip has bright background in dark themes


mueller-ma:
### Actual behaviour
"Info ssl client cert" tooltip has bright background in dark themes, which makes the tooltip hard to read.

### Expected behaviour
"Info ssl client cert" tooltip should have a dark background.

#### Client
* Android version: 5.1.1

mueller-ma:
Still happens in API 25

maniac103:
I'll check it.

PR #3061, opened 2022-10-20
Migrate to Material 3 

mueller-ma:
Theme has been generated with https://m3.material.io/theme-builder#/custom and the openHAB orange as seed color.

To reduce complexity, remove the accent color preference.

Fixes https://github.com/openhab/openhab-android/issues/2799
Fixes https://github.com/openhab/openhab-android/issues/1058

TODO:

- [x] Up/Down buttons of color widgets aren't shown
- [x] Change app bar text color to white in light mode
- [x] Update color of app bar action buttons to match text color of app bar
- [X] ~Update color used in `ActivityManager.TaskDescription()`~ Remove custom task description

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
@@ -14,18 +14,15 @@
 package org.openhab.habdroid.ui
 
 import android.Manifest
-import android.app.ActivityManager
 import android.app.KeyguardManager
 import android.content.Intent
 import android.content.pm.PackageManager
 import android.content.res.Configuration
-import android.graphics.BitmapFactory
 import android.os.Build
 import android.os.Bundle
 import android.os.SystemClock
 import android.provider.Settings
 import android.util.Log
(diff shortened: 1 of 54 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
```diff
@@ -170,6 +173,21 @@ abstract class AbstractBaseActivity : AppCompatActivity(), CoroutineScope {
...
+        menu.forEach { item ->
+            val icon = item.icon
+            icon?.let {
+                icon.mutate()
+                DrawableCompat.setTint(icon, iconColor)
+                item.icon = icon
+            }
+        }
+        return super.onPrepareOptionsMenu(menu)
+    }
```
maniac103, line 189: Can't this be done in XML? https://stackoverflow.com/a/54961263

mueller-ma: I didn't know the "app" prefix and tried "android" instead, which isn't backwards compatible.

PR #3081, opened 2022-11-06
Support brotli compression

### LINKED ISSUE #3059: Document what `Accept-Encoding:`/`Content-Encoding:` is supported
When I am not at home I connect to the home-internal openHAB system with openHAB-android. I want to compress the transferred data openHAB server ⇔ openHAB-Android, but only if openHAB-Android can handle the compressed data.  Between openHAB-server and openHAB-Android there is an nginx-proxy, which can compress the data.

https://www.openhab.org/docs/apps/android.html and https://github.com/openhab/openhab-android/blob/main/README.md do not contain `Accept-Encoding`, `gz` (and thus gzip), `br` (and thus brotli), `deflate`.

Please write down, whether openHAB-Android supports compressions via Accept-Encoding/Content-Encoding and which compressions:
- br
- gz
- zstd

Consider adding more compressions from the above list to openHAB-Android.

mueller-ma:
Fixes #3059

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -155,6 +155,7 @@ dependencies {
     implementation "com.squareup.okhttp3:okhttp:$ok_http_version"
     implementation "com.squareup.okhttp3:logging-interceptor:$ok_http_version"
     implementation "com.squareup.okhttp3:okhttp-sse:$ok_http_version"
+    implementation("com.squareup.okhttp3:okhttp-brotli:$ok_http_version")
     implementation "com.github.QuadFlask:colorpicker:0.0.15"
     implementation "com.caverock:androidsvg-aar:1.4"
     implementation "com.github.AppIntro:AppIntro:6.2.0"
--- a/mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/connection/ConnectionFactory.kt
@@ -41,6 +41,7 @@ import kotlinx.coroutines.channels.onClosed
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
+import okhttp3.brotli.BrotliInterceptor
 import okhttp3.internal.tls.OkHostnameVerifier
 import okhttp3.logging.HttpLoggingInterceptor
 import org.openhab.habdroid.core.CloudMessagingHelper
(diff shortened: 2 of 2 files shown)


mueller-ma:
@dilyanpalauzov

Can you please test the following apk?

- Rename it to .apk
- Install my apk

[mobile-foss-beta-debug.zip](https://github.com/openhab/openhab-android/files/9946457/mobile-foss-beta-debug.zip)

dilyanpalauzov:
I have configured my Nginx/openHAB server to return brotli compressed data, when the requested content type is text/html, text/css, application/javascript, or application/json.  Calling /basicui/app from a browser shows that the requests contain `Accept-Encoding: gzip, deflate, br` and the answers return `content-encoding: br`.

For some items in my sitemap, which are supposed to have no-icon, once brotli is enabled, the icon shown in the browser is “broken image“, but I cannot figure out which request/response causes this.  However this “broken image” is displayed only in Chromium, not in Firefox or Epiphany/Web/WebKitGtk.

I have installed the above debug.zip file.  It sends Accept-Encoding: br,gzip and receives Content-Encoding: br, according to the logs of the server.  What shall I verify in that image?

dilyanpalauzov:
> I have configured my Nginx/openHAB server to return brotli compressed data, when the requested content type is text/html, text/css, application/javascript, application/json

or  text/event-stream

dilyanpalauzov:
When I toggle switches in openHAB-android, it calls `POST /rest/items/r1 HTTP/2.0` sending Accept-Encoding: br, gzip.  I see the server returns no Content-Encodig, but Content-Type “text/plain”.  The switches do not work.and I assume, that the reasor are the not-compressed text/plain messages.

dilyanpalauzov:
Hm, on that POST /rest/items/r1  the returned answer contains
```
content-length: 0
content-type: text/html
```
what is the purpose of setting content-type, when the content-length is 0?

In any case, in the browser (firefox) this operation - toggle a switch while the server does the same compression, as for openHAB-Android - does work.

dilyanpalauzov:
When br-compression is enabled and I use that debug-zip file, and toggle a switch in a group, the state of the toggle is not updated immediately.  I have to close the group, open it again, then then toggle has the changed state.

When the server offers no br-compression, or when visiting /basicui/app with a browser, these anomalies do not happen.

mueller-ma:
It seems that more work is required than just to integrate `okhttp-brotli`. Therefore I'm closing this PR, but I'd accept a PR that implements a working brotli integration.

dilyanpalauzov:
> When br-compression is enabled and I use that debug-zip file, and toggle a switch in a group, the state of the toggle is not updated immediately. I have to close the group, open it again, then the toggle has the changed state.

In the mean time the above described anomalies happen often, they do not depend on the brotli compression.  In particular with `Switch mappings=[]` for `autoupdate=false` channel, the UI changes the state only after the device reports a change.  Sometimes changes in groups: `Text { items }` are not shown immediately, I have to go back and forth to see the changes.  Today I think that this does not depend on brotli.

PR #3085, opened 2022-11-08
Allow pref titles to use multiple lines

### LINKED ISSUE #3083: Preference labels are truncated
The title of the pref  "Use as start app" is truncated in German.

mueller-ma:
Fixes #3083

### DIFF
--- a/mobile/src/main/res/xml/preferences.xml
+++ b/mobile/src/main/res/xml/preferences.xml
@@ -35,10 +35,12 @@
             android:defaultValue="true"
             android:key="imageWidgetScaleToFit"
             android:title="@string/settings_image_widget_scaling"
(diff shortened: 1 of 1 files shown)


maniac103:
Do you have a screenshot at hand? I'd imagine it looks rather ugly? In that case I think we should rather shorten the strings.

maniac103:
It's OK, but I think title + summary would be better nevertheless: 'Register as launcher' + 'Allow the app to be used as launcher. After enabling, the app still needs to be selected as default launcher to take effect' or something among these lines ... should also work for DE.

mueller-ma:
I opened #3086 to update the title of this preference.

For smaller devices or other languages some of these titles might be truncated as well and in that case using two line seems to be a better choice.

PR #3091, opened 2022-11-11
Increase padding in sitemap

mueller-ma:
Before the icon had only 4dp padding to the start of the display. This is way less than icon e.g. in the Android settings have.


mobile/src/main/res/layout/widgetlist_videoitem.xml
```diff
@@ -3,7 +3,8 @@
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_height="wrap_content"
-    android:layout_width="match_parent">
+    android:layout_width="match_parent"
+    android:paddingHorizontal="@dimen/widgetlist_margin_full_width_items">
```
maniac103, line 7: Padding should be at child, as progress bar should not be padded (same for WebView etc.)

mueller-ma: Setting `widgetlist_margin_full_width_items` here (which is `0dp`) is more like documentation, because it would be `0dp` anyway. Every `widgetlist_.xml` has `paddingHorizontal` set in the most outer layout.

mueller-ma:
Sitemaps need a larger rework, see #3097

PR #3102, opened 2022-11-27
Fallback to min/max/step of Item if not set for Widget

mueller-ma:
Sliders require values to be Floats, so I changed
Item.minimum/maximum/step to Float.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
@@ -18,6 +18,7 @@ import android.net.Uri
 import android.os.Parcelable
 import kotlin.math.abs
 import kotlin.math.max
+import kotlin.math.min
 import kotlinx.parcelize.Parcelize
 import org.json.JSONException
 import org.json.JSONObject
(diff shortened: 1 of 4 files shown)


mueller-ma:
@maniac103 FYI the new setpoint sliders won't cover the range from 0 to 100 with this change anymore, at least for my setup.

maniac103:
What do you mean by 'won't cover'?

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -285,9 +286,9 @@ fun JSONObject.collectWidgets(parent: Widget?): List<Widget> {
     val type = getString("type").toWidgetType()
     val icon = optStringOrNull("icon")
     val (minValue, maxValue, step) = Widget.sanitizeMinMaxStep(
-        optDouble("minValue", 0.0).toFloat(),
-        optDouble("maxValue", 100.0).toFloat(),
-        optDouble("step", 1.0).toFloat()
+        optFloatOrNull("minValue") ?: item?.minimum ?: 0f,
+        optFloatOrNull("maxValue") ?: item?.maximum ?: 100f,
+        optFloatOrNull("step") ?: item?.step ?: 1f
```
maniac103, line 291: Need to update `updateFromEvent` with this logic, or convert min/max/step to a getter function.

mueller-ma:
> What do you mean by 'won't cover'?

I found this issue, because the new setpoint slider of my heating had the range 0°C - 100°C 😓

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
mueller-ma: Done by adding getters.

PR #3106, opened 2022-11-30
Material Design 3 updates

### LINKED ISSUE #3097: Material Design 3 TODO
In the MD3 update, a few places were forgotten/still need updating:
- The title of the chart scaling pref is not bold (or has wrong weight)
- The header of the NFC activity needs updating (title text color, maybe whole header should become a card or something)
- Notification list probably shouldn't have lines now
- Same for sitemap widget list?
- With the left side drawer now being rounded, it looks somewhat weird now that it doesn't span the room below the action bar

More generally, I'm not sure the MD3 update is a good idea in its present state, since for that to look good, quite a few additional updates are needed:
- Sitemap widgets need a bit of an overhaul, maybe using pills? But certainly different font sizes, similar to e.g. lists in the Google Drive app
- Sitemap widgets opening a dialog (e.g. selection) should open a bottom panel instead 
- List/activity background needs to use a system color (compare to Google apps) or keep our previous color theme

Thoughts?

maniac103:
- Fix ChartScalingPreference for MD3 theme update
- Redesign widget, notification, item picker lists to better match MD3
- Use bottom sheets instead of dialogs where possible

TODO:
- [x] check why elevation overlay flickers and (when found) reenable elevation overlay
- [x] rethink item picker title area?
- [x] fix elevation overlay in notification list

Optional, can be deferred to later PR:
- [ ] convert widget list item long-press menu to bottom sheet?
- [ ] convert permission request dialog to bottom sheet?
- [ ] fix map bottom sheet drag handle not working when expanded

Fixes #3097


maniac103:
Some screenshots (after on top, before on bottom):
screenshots

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
```diff
@@ -508,6 +525,34 @@ abstract class ContentController protected constructor(private val activity: Mai
...
+        if (shouldShowActionBar) {
+            actionBar?.show()
+        } else {
+            actionBar?.hide()
+        }
+        updateContentViewForInsets()
+    }
+
+    private fun updateContentViewForInsets() {
+        val i = insets?.getInsets(WindowInsets.Type.systemBars())
```
mueller-ma, line 545: `WindowInsets.Type` has been added in API 30 (> minSdk).

mueller-ma (commented):
* In demo mode I can open the server selector in the drawer.

maniac103:
> In demo mode I can open the server selector in the drawer.

Can you provide some steps to reproduce?

mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
maniac103: Fixed by using `WindowInsetsCompat`.

mueller-ma:
> Can you provide some steps to reproduce?

* Open settings
* Enable demo mode
* Go back to MainActivity
* Open drawer
* Server selector can be opened by pressing on the caret icon

maniac103:
Doesn't happen here (?)

mueller-ma:
> I guess you have configured multiple servers?

Yes

maniac103:
>> I guess you have configured multiple servers?
>
> Yes

Makes sense I couldn't reproduce then ;-) Fixed now.

mueller-ma:
- On a API 21 device toasts (e.g. "Running in demo mode") are shown behind the navigation bar.
- The attribute `selectedWidgetBackgroundColor` is much brighter now, even the actual value didn't change with this PR. In dark mode the label of the selected widget is hardly visible.

maniac103:
> On a API 21 device toasts (e.g. "Running in demo mode") are shown behind the navigation bar.

Fixed.

mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
```diff
@@ -169,7 +169,7 @@ abstract class AbstractBaseActivity : AppCompatActivity(), CoroutineScope {
             throw IllegalArgumentException("Tag is empty")
         }
 
-        val snackbar = Snackbar.make(findViewById(android.R.id.content), message, duration)
+        val snackbar = Snackbar.make(findViewById(R.id.coordinator), message, duration)
```
maniac103, line 172: Still need to make sure all activity layouts have a CoordinatorLayout with that ID.

mueller-ma:
I updated the section switch and sliders: https://github.com/mueller-ma/openhab.android/commit/779b8cf7c474b578d289436aff5405ee6ce754df

IMO the section switches look better that way. Not sure about the slider (and it's broken when scrolling as the label doesn't stick to the slider then).

mueller-ma:
Maybe set the label visibility to `gone` if the label is empty for items that have the control below the label, e.g. sliders?

maniac103:
> IMO the section switches look better that way.

Agreed, I've taken that change, with the variation of putting the switch back to one-row layout (simlar to `Switch` widget) if there's only a single mapping, as spanning one button over the whole screen width seems like a waste of space.

> Not sure about the slider (and it's broken when scrolling as the label doesn't stick to the slider then).

I don't really like it either, for two reasons (besides the brokenness)
- The sliders need more space that way, for no obvious gain
- The value appears too prominent in the widget list due to the color

> Maybe set the label visibility to gone if the label is empty for items that have the control below the label, e.g. sliders?

Done.

mueller-ma:
> Agreed, I've taken that change, with the variation of putting the switch back to one-row layout (simlar to Switch widget) if there's only a single mapping, as spanning one button over the whole screen width seems like a waste of space.

Way better now 👍 

One small thing: The action bar is black if the first list item is shown. When scrolling the action bar becomes a bit more gray. However if the first sitemap page is shorter than the screen height, the action bar is gray, but should be black.

Other than that: Great work! I'd say merge it and get feedback from beta users.

maniac103:
> One small thing: The action bar is black if the first list item is shown. When scrolling the action bar becomes a bit more gray. However if the first sitemap page is shorter than the screen height, the action bar is gray, but should be black.

As far as I can tell, this is an artifact of the first frame being invisible:
- the action bar background becomes 'elevated' (colored) when the RecyclerView is determined to be scrollable upwards
- for determining that, the RecyclerView basically checks whether the first visible item is actually the first item (see LinearLayoutManager.computeScrollOffset)
- the invisible group header is discarded in this check because of not being visible
- thus the checked item is (in demo mode) the 'First Floor' one, which has adapter position 1
- because of that, the method thinks scrolling up (to adapter position 0) is possible

So all in all, this seems like a RecyclerView bug/shortcoming, which we could hack around in our code (leave a 1dp spacer in case of empty frame at start), but is it worth it?

maniac103:
One other thing: I wanted to make the status bar overlay color match the (web) UI color for main UI / HabPanel fragments, because the white status bar doesn't really fit well there. Changing the status bar color is relatively straightforward, but adjusting status bar icons and navigation bar color accordingly doesn't seem to be trivial, so I'd like to leave this out for now.

mueller-ma (approved):
LGTM! Ready to merge from your side?

maniac103:
Yes.

Issue #3012, opened 2022-08-29
Beta version has been removed by Google from Play Store


mueller-ma:
Mail from Google:

![grafik](https://user-images.githubusercontent.com/22525368/187280071-0244487f-8d28-45e6-8864-ce273e7f7c0a.png)

I submitted a appeal as Google publishes beta versions as an own app as well: https://play.google.com/store/apps/details?id=com.chrome.beta

rkoshak:
With the release of 2.22 I assume that now the main app is now ahead of any beta version app we've installed from the play store previously, correct?

It's totally bogus that they removed it. *Lots* of apps have a beta version on the store). I anxiously await their resolution.

mueller-ma:
In the appeal I submitted I mentioned the Google Chrome Beta, but it seems that the Play Store rules don't count for them....

The beta version could also be published in the beta channel of the app. This means that you cannot install beta alongside the stable version. For this the `.beta` suffix of the package id has to be removed for fullBeta, but kept for fossBeta (because the beta version is still on F-Droid and F-Droid doesn't support channels).

> With the release of 2.22 I assume that now the main app is now ahead of any beta version app we've installed from the play store previously, correct?

Correct. You could download the beta version from https://github.com/openhab/openhab-android/releases, which can be installed alongside the stable version.

mueller-ma:
I will submit future stable version in the open test track before deploying it as production.

You can join at https://play.google.com/apps/testing/org.openhab.habdroid

PR #3107, opened 2022-11-30
Add 'BETA' to icon of beta builds

### LINKED ISSUE #403: Icon of beta version should contain a badge with "beta"

mueller-ma:
Fixes #403

I created the app icons with the following options:
* Asset: assets/openhab-icon.svg
* Trim: No
* Resize: 62%
* Background color: #FFFFFF


cwihne:
For years now I appreciate the parallel release/beta app concept for a range of reasons, while being aware that this is not the standard approach and might mean multiplied work especially for maintainers:
1. testing and comparing both branches to see if issue was already present in release or popped up only recently in beta is way easier than switch between main and release branch - especially in the (rare) case of changes in saved cfg's
2. some level of backup strategy in the case of breaking issues in beta
3. having both installed allows registration to different openHAB instances, with both promoting notifications to the same mobile

mueller-ma:
I also like to have both build variants installed in parallel. The beta build is still distributed on F-Droid and GitHub.

I was thinking about adding a new build version `betaPlay` that is equal to `beta`, but doesn't add `.beta` to the package name. But that should be discussed in #3012.

maniac103:
> For years now I appreciate the parallel release/beta app concept for a range of reasons

Don't get me wrong, I don't think it's a bad idea ... it's just that *Google* thinks it's a bad idea ;-)

PR #3115, opened 2022-12-05
Send slider update while sliding

### LINKED ISSUE #654: Slider: Send state while holding
I'm not sure if thats a good idea (because of the traffic), but I think its better, when the slider item is change, while the user is sliding.

mueller-ma:
Fixes #654


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -619,10 +624,26 @@ class WidgetAdapter(
...
+            isTracking = false
+            sendUpdate()
+        }
+
+        override fun onValueChange(slider: Slider, value: Float, fromUser: Boolean) {
+            if (!fromUser) {
+                return
+            }
+            Log.e(TAG, "${System.currentTimeMillis() - lastChange}")
+            if (System.currentTimeMillis() - lastChange < 1000) {
```
maniac103, line 640: I don't really like this because it will send updates at essentially random times ... I added an alternative approach/suggestion to #3116

PR #3129, opened 2022-12-16
WIP: Fix cancelling widget creation

### LINKED ISSUE #3125: Unable to cancel widget creation
### Actual behaviour
Unable to leave PreferenceActivity after widget has been added to the home screen. A popup shows up and asks how to deal with unsaved changes. The button "Discard" just re-opens the popup.

### Steps to reproduce
1. Use the drawer to create a `ItemUpdateWidget`
2. Press back button when the activity is open.

mueller-ma:
Make `widgetLabel` not-null as null had a special meaning (auto generated label) some time ago, but this had been migrated > 1 year ago.

`null` vs. empty string is causing an issue when checking if data has been changed.

Fixes #3125

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/homescreenwidget/ItemUpdateWidget.kt
@@ -338,7 +338,7 @@ open class ItemUpdateWidget : AppWidgetProvider() {
         ): RemoteViews {
             val darkTheme = data.theme == "dark"
             val layout = when {
-                data.widgetLabel?.isEmpty() == true -> {
+                data.widgetLabel.isNullOrEmpty() -> {
                     if (darkTheme) R.layout.widget_item_update_dark_no_text
                     else R.layout.widget_item_update_light_no_text
                 }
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/core/UpdateBroadcastReceiver.kt
```diff
@@ -170,26 +170,6 @@ class UpdateBroadcastReceiver : BroadcastReceiver() {
...
-                            ?: context.getString(R.string.item_update_widget_text, oldData.label, oldData.mappedState),
-                        oldData.mappedState,
-                        oldData.icon,
-                        oldData.theme,
-                        oldData.showState
-                    )
-
-                    ItemUpdateWidget.saveInfoForWidget(context, newData, id)
-                }
-            }
```
maniac103, line 192: What's the point of removing that migration code now?

mueller-ma: `oldData.widgetLabel` cannot be null anymore, so `?: context.getString(R.string.item_update_widget_text, oldData.label, oldData.mappedState)` has to be removed.
Then the migration would be a no-op.

maniac103: Isn't the migration code just buggy then? Reading the existing data must happen taking the old format into account (which by definition can not change), not the new one.

mueller-ma: > Reading the existing data must happen taking the old format into account

Sure I can do that. But I wondering how long `widgetLabel` should be kept nullable only because of the old format.

maniac103: > But I wondering how long widgetLabel should be kept nullable only because of the old format.

It doesn't need to be. It's just the conversion code that needs to be able to consume a null value. Storing the data and the app widget code itself doesn't need to deal with null values.

PR #3135, opened 2022-12-23
Update firebase to latest version

mueller-ma:

### DIFF
--- a/mobile/src/full/java/org/openhab/habdroid/core/FcmRegistrationWorker.kt
+++ b/mobile/src/full/java/org/openhab/habdroid/core/FcmRegistrationWorker.kt
@@ -31,7 +31,6 @@ import androidx.work.WorkManager
 import androidx.work.WorkRequest
 import androidx.work.WorkerParameters
 import com.google.firebase.messaging.FirebaseMessaging
-import com.google.firebase.messaging.RemoteMessage
 import java.io.IOException
 import java.net.URLEncoder
 import java.util.Locale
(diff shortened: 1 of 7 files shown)


mobile/src/full/java/org/openhab/habdroid/core/FcmRegistrationWorker.kt
```diff
@@ -119,8 +118,7 @@ class FcmRegistrationWorker(private val context: Context, params: WorkerParamete
     // HttpException is thrown by our HTTP code, IOException can be thrown by FCM
     @Throws(HttpClient.HttpException::class, IOException::class)
     private suspend fun registerFcm(connection: CloudConnection) {
-        val token = FirebaseInstanceId.getInstance().getToken(connection.messagingSenderId,
-                FirebaseMessaging.INSTANCE_ID_SCOPE)
```
mueller-ma, line 123: I didn't find a way to pass `connection.messagingSenderId` to Firebase.

maniac103: If it works nevertheless, that's not a problem I guess. What *is* a problem, AFAICT, is that `token` will have type `Task<String>`, not `String`, that is, you'll need to add a listener to retrieve the actual token.

mueller-ma: I haven't tested this yet. I added `.result` which returns a `String`.

maniac103: > .result which returns a String.

... or throws an `IllegalStateException` if the task isn't done yet ;-)

mueller-ma: Fetching the token now works and the registration at myopenhab is successful, but notifications aren't received.

digitaldan: I just tested this with my upcoming changes to openHAB cloud, and i can confirm it works and is getting notifications.  Once i make the change to the server, both the older code and this code will work.  The only gotcha is that users will need to launch the app to register again regardless of this change as we are using a new FCM id from our openhab-mobile project and not the original one.

mueller-ma:
I removed the broken 'hideNotification' feature for now. It has to be replaced by the new endpoint in myopenHAB. As that is free software, it can be used in the foss flavor as well: #3686

PR #3146, opened 2023-01-05
Remove state from Item label

mueller-ma:
For #3143 I set the label of an Item to `Color [JS(hsb-to-brightness.js):%s]`. When this Item is fetched via /rest/items/ the brackets aren't resolved to the actual value. When fetched via /rest/sitemaps/ the brackets are placed. Since `[JS(hsb-to-brightness.js):%s]` isn't useful to be shown in the device control, I added a splitting of the label as it's done for Widgets.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
@@ -186,7 +186,8 @@ class ItemsControlsProviderService : ControlsProviderService() {
         }
 
         fun maybeCreateControl(item: Item): Control? {
-            if (item.label.isNullOrEmpty() || item.readOnly) return null
+            val label = item.label
+            if (label.isNullOrEmpty() || item.readOnly) return null
             val controlTemplate = if (item.options != null) {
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -186,7 +186,8 @@ class ItemsControlsProviderService : ControlsProviderService() {
         }
 
         fun maybeCreateControl(item: Item): Control? {
-            if (item.label.isNullOrEmpty() || item.readOnly) return null
+            val label = item.label
+            if (label.isNullOrEmpty() || item.readOnly) return null
```
maniac103, line 190: This change doesn't look like it's needed and/or really helpful?

mueller-ma: It is required. `label` is now a nullable `String?` property, so no smart cast to `String` is possible anymore. Below `Control.StatefulBuilder().setTitle()` wants a non-null String.

maniac103 (approved):

PR #3165, opened 2023-01-28
Fix coroutine handling in sitemap widgets

maniac103:
As a Job can only run (and be cancelled) once, widgets depending on the CoroutineScope only worked when initially attached, but ceased to work once they were recycled. Fix this by having the scope as member variable instead of subclass and by recreating the scope when attaching the widget.


maniac103:
@mueller-ma This bug certainly deserves a 3.0.2.

PR #3170, opened 2023-01-29
Don't hide frames without label

### LINKED ISSUE #3160: Make nameless frames usefull for separation at sitemaps again
> This is the case, as I wrote: with nameless frames yo can separate with that shadow line. Now all my "separations" not shown any longer...

>That's understood and something we'll think about (and maybe worth opening a separate issue for, because it's actually actionable). Not sure yet what the best implementation for it with the 'flat' design is, maybe a thin orange line with a bit less padding around it?

_Originally posted by @maniac103 in https://github.com/openhab/openhab-android/issues/3159#issuecomment-1406305983_
 
Placed in a new issue like suggested.
The suggestion of a thin orange line would work perfect for me

mueller-ma:
Closes #3160

### DIFF
--- a/mobile/src/main/res/layout/widgetlist_frameitem.xml
+++ b/mobile/src/main/res/layout/widgetlist_frameitem.xml
@@ -15,22 +15,22 @@
         android:id="@+id/container"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:paddingTop="8dp"
-        android:paddingBottom="8dp"
-        android:paddingEnd="24dp">
+        android:layout_marginTop="8dp"
+        android:layout_marginBottom="8dp"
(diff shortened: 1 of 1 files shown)


maniac103:
I assume the look of a frame with label is unchanged, right? Got a screenshot of a frame without label?

mueller-ma:
The labeled frames didn't change. Here are two widgets divided by a frame without label.

grafik

maniac103:
LGTM :+1: It's hidden if it's the first visible widget in a page, right?

mueller-ma:
> It's hidden if it's the first visible widget in a page, right?

Correct

mkrathome:
Thanks a lot 🖖🏻

PR #3177, opened 2023-02-01
Add layout mode selection

### LINKED ISSUE #3173: possibility of show text with values in  "one line"
Hello again, 

After 3 Days with the new Design, I agree with you, that it is a pretty nice style (in the meantime I love that flat and clear look) also I understood the need of the 2-line dessicion. 
I read many times here and also at the openhab-community, that you are possibly think about a way to have the opportunity to show text with a value-item in one line.
To prevent of misunderstood:

I don't ask for a change of the complete design  and also not for a general switch for all types of widgets.
But I think it is worth to think about a possibility for text  which has additional item+value. 

Please understand this as an politely whish from my side🖖🏻

All the best 

Manuel

maniac103:
Fixes #3173


mobile/src/main/res/values-w600dp/dimens.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <integer name="section_switch_max_buttons">6</integer>
+    <dimen name="eidgetlist_compact_slider_width">200dp</dimen>
```
mueller-ma, line 4:
```suggestion
    <dimen name="widgetlist_compact_slider_width">200dp</dimen>
```

mobile/src/main/res/values/strings.xml
```diff
@@ -63,6 +63,9 @@
     <string name="settings_dynamic_colors_title">Dynamic colors</string>
     <string name="settings_dynamic_colors_summary_off">Use color style matching openHAB style throughout the app</string>
     <string name="settings_dynamic_colors_summary_on">Use colors based on your wallpaper throughout the app</string>
+    <string name="settings_sitemap_compact_mode_title">Compact mode</string>
+    <string name="settings_sitemap_compact_mode_off">Use a two-line layout for Sitemap widgets, which makes controls larger and thus easier to use, but requires more scrolling for longer lists</string>
+    <string name="settings_sitemap_compact_mode_on">Use a one-line layout for Sitemap widgetsm, which puts as many widgets as possible on the screen, but reduces size of the controls </string>
```
mueller-ma, line 68:
```suggestion
    <string name="settings_sitemap_compact_mode_on">Use a one-line layout for Sitemap widgets, which puts as many widgets as possible on the screen, but reduces size of the controls</string>
```

mueller-ma (commented):
What about a bit less padding inside the section buttons?

````diff
diff --git a/mobile/src/main/res/layout/widgetlist_sectionswitchitem_button.xml b/mobile/src/main/res/layout/widgetlist_sectionswitchitem_button.xml
index baefe4bcb..35292d1b0 100644
--- a/mobile/src/main/res/layout/widgetlist_sectionswitchitem_button.xml
+++ b/mobile/src/main/res/layout/widgetlist_sectionswitchitem_button.xml
@@ -6,9 +6,10 @@
     android:layout_width="0dp"
     android:layout_weight="1"
-    android:minWidth="0dp"
-    android:paddingHorizontal="16dp"
+    android:minWidth="8dp"
+    android:paddingHorizontal="8dp"
     android:paddingTop="4dp"
     android:paddingBottom="4dp"
     android:textAppearance="?attr/textAppearanceLabelMedium"
-    tools:text="Action 1" />
+    tools:text="Action 1"
+    tools:layout_width="wrap_content" />
diff --git a/mobile/src/main/res/layout/widgetlist_sectionswitchitem_button_compact.xml b/mobile/src/main/res/layout/widgetlist_sectionswitchitem_button_compact.xml
index 6953ecd37..4dd85fd5b 100644
--- a/mobile/src/main/res/layout/widgetlist_sectionswitchitem_button_compact.xml
+++ b/mobile/src/main/res/layout/widgetlist_sectionswitchitem_button_compact.xml
@@ -5,8 +5,8 @@
     style="@style/Widget.Material3.Button.OutlinedButton"
     android:layout_width="wrap_content"
-    android:minWidth="0dp"
-    android:paddingHorizontal="16dp"
+    android:minWidth="8dp"
+    android:paddingHorizontal="8dp"
     android:paddingTop="2dp"
     android:paddingBottom="2dp"
     android:textAppearance="?attr/textAppearanceLabelMedium"

````

**Comment by maniac103** [comment]
> What about a bit less padding inside the section buttons?

It's a bit too small for my taste:

grafik

(Also left and right side look different with that diff applied)

For comparison, the one in the PR, without the diff applied:

grafik

IMO, reducing the padding between buttons is good, but I'm not sure about the outer sides. At a minimum, the rounded corners left and right should look the same. I checked the (MDC library) code for possible reasons, and it looks like corner radius is calculated as `50% of min(button width, button height)`. So if button width is smaller than height, corners aren't fully rounded anymore.

(The change in the non-compact version likely won't do much, since the layout weight will add much more horizontal size than the padding)

File: mobile/src/main/res/layout/widgetlist_sectionswitchitem_compact.xml [hunk 3]
```diff
@@ -0,0 +1,18 @@
...
+
+    <include layout="@layout/widgetlist_icontext_compact" />
+
+    <com.google.android.material.button.MaterialButtonToggleGroup
+        android:id="@+id/switch_group"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="16dp"
+        android:layout_marginTop="4dp"
+        app:singleSelection="true" />
```
mueller-ma, line 16: When the widgetLabel is hidden, these buttons are left aligned.

maniac103: Note to self: Place a `Space` in front of the buttons with `layout_weight` > 0

mobile/src/main/res/layout/widgetlist_switchitem_compact.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    style="@style/WidgetListItemContainerCompact">
+
+    <include layout="@layout/widgetlist_iconvaluetext_compact" />
```
maniac103, line 7: https://github.com/openhab/openhab-android/issues/3173#issuecomment-1413402630

maniac103:
Fixed the section switch padding by using a meaningful `minWidth`:

grafik

maniac103:
Open question: compact mode on or off by default?

mueller-ma:
> Open question: compact mode on or off by default?

I vote for off by default.

mkrathome:
> Open question: compact mode on or off by default?

Hi, if I'm allowed to take part here:
I would also vote for off by default😉

mobile/src/main/res/layout/widgetlist_sectionswitchitem_compact.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    style="@style/WidgetListItemContainerCompact">
+
+    <include layout="@layout/widgetlist_icontext_compact" />
```
maniac103, line 8: Need `iconvaluetext` here.

mobile/src/main/res/values/dimens.xml
```diff
@@ -6,6 +6,7 @@
     <dimen name="widget_margin">0dp</dimen>
     <dimen name="row_height">36dp</dimen>
     <dimen name="widgetlist_icon_size">40dp</dimen>
+    <dimen name="widgetlist_compact_slider_width">100dp</dimen>
```
maniac103, line 9: TODO: make this shorter

maniac103: Actually this value wasn't the issue, but a `layout_weight` being assigned to the slider. Fixed now.

Issue #3163, opened 2023-01-28
Blank popup after a new OH version installation (probable theme issue)


lolodomo:
After installing a new OH version, the first time I run the Android app, I have 2 popups to validate, the second is about using an IP IIRC but the first one is always blank. This is certainly a problem of theme. My setup theme is "Sombre". The second popup has a grey/black background with white text, while the first has a white background with probably white text.

I open an issue only now but this is like that already since a long time. But it worked in the past.

lolodomo:
In case you need to know what is exactly in that popup, I will change my theme to a white theme and I can see its content the next time I install a new OH version.
But I am almost sure this popup is about SSL certificate. I setup an HTTPS server using an IP.

mueller-ma:
Should be fixed by #3232.

PR #3178, opened 2023-02-01
Reintroduce theme selection

maniac103:
TODO:

- [ ] Adjust grey theme colors, because they're teal, not grey
- [ ] Actually test dynamic color selection
- [ ] Test dynamic colors with MTM (how?)


maniac103:
>  Adjust grey theme colors, because they're teal, not grey

It looks like grey highlight color is something the Material theme builder is able to generate ... on purpose, I guess. What to do now? Rename 'grey' to something else? Remove the theme and keep only default + BasicUI (+ dynamic)?

mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
```diff
@@ -75,11 +74,7 @@ abstract class AbstractBaseActivity : AppCompatActivity(), CoroutineScope {
 
     @CallSuper
     override fun onCreate(savedInstanceState: Bundle?) {
-        setTheme(getActivityThemeId())
-        if (getPrefs().getBoolean(PrefKeys.DYNAMIC_COLORS, false)) {
-            DynamicColors.applyToActivityIfAvailable(this)
-        }
-
+        applyUserSelectedTheme()
```
maniac103, line 77: Should additionally apply dynamic colors in `onStart` (palette can change at runtime)

mueller-ma:
> Adjust grey theme colors, because they're teal, not grey -> not possible, palette does not generate grey highlight. Remove? Change name?

Is this theme required any longer? Before the action bar was orange, but with MD3 it's always white or dark. Colors are used less, so they might not "hurt" anymore.
I'd remove it.

Maybe rename "Basic UI" to blue and add a few other colors?

maniac103:
> Is this theme required any longer? Before the action bar was orange, but with MD3 it's always white or dark. Colors are used less, so they might not "hurt" anymore.
> I'd remove it.

Good point -> removed. (NB: it's kinda ironic we introduced a separate theme just because the openHAB orange was too bright in dark mode, and now people complain the primary accent in dark mode isn't the openHAB orange anymore...)

> Maybe rename "Basic UI" to blue and add a few other colors?

We discussed this in #1547 already - my opinion hasn't changed since then ;-)
I chose 'Basic UI' as name because that's what we had for the blue/pink theme before #1547.

mobile/src/main/java/org/openhab/habdroid/ui/AbstractBaseActivity.kt
maniac103: Further research suggests this isn't actually needed (e.g. `DynamicColors.applyToActivitiesIfAvailable` also does this only in `onPreCreate`). I've seen an old color scheme being stuck one or two times, though; will monitor this.

mueller-ma (approved):
LGTM!

> Enable dynamic colors by default if supported? (https://github.com/openhab/openhab-android/issues/3185)

Can be discussed in that issue.

> Test dynamic colors with MTM (how?)

That seems to be broken currently: #3163 
Can be tracked in that issue.

PR #3187, opened 2023-02-03
Allow read-only items in device controls

mueller-ma:
These tiles can display item states.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
@@ -188,7 +188,7 @@ class ItemsControlsProviderService : ControlsProviderService() {
         fun maybeCreateControl(item: Item): Control? {
             val label = item.label
             if (label.isNullOrEmpty()) return null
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -273,6 +237,73 @@ class ItemsControlsProviderService : ControlsProviderService() {
...
+                    "${item.name}_toggle",
+                    ControlButton(item.state?.asBoolean ?: false, context.getString(R.string.nfc_action_toggle)),
+                    createRangeTemplate(item, "%.0f%%")
+                )
+                item.type == Item.Type.Rollershutter -> createRangeTemplate(item, "%.0f%%")
+                item.type == Item.Type.Number -> createRangeTemplate(
+                    item,
+                    item.state?.asNumber?.unit?.let { "%.0f $it" } ?: "%.0f"
+                )
+                else -> ControlTemplate.getNoTemplateObject()
```
mueller-ma, line 303: For some item types, like Location, no useful text-only representation is possible. For these items no tile should be created.

PR #3202, opened 2023-02-07
Do not use item min/max values for dimmer sliders

### LINKED ISSUE #3194: Slider max value more than 100%
I have a "Dimmer" item for brightness that uses an mqtt Thing as a channel. The Thing has a min value of 0 and a max value of 254.
```
Type dimmer : brightness "Helligkeit" [
	stateTopic="zigbee2mqtt/LampLiving/brightness",
	commandTopic="zigbee2mqtt/LampLiving/set/brightness",
	min=0, max=254
]
```
In my sitemap I use a slider to display it:
```
Slider item=LampLivingBrightness label="Helligkeit [%d %%]"
```

In BasicUI (on my PC and on my Android phone) this shows up as a slider with the values from 0% to 100% selectable:
image

In the recent App 3.1.0 I can select more than 100%:
image

I have the same behavior with other slider, and all of them use an mqtt-channel with min and max values. Sliders, that don't have a min or max value set, seem to work fine. 

#### Client
* Android version: 13
* Device model: Oneplus 9 Pro
* App version 3.1.0
* Build flavor: "Full"
* Device language: DE

#### Server
* Server version: openHAB 2.5.12
* Reverse Proxy: NginX

maniac103:
Those have an implicit range of 0..100, because they're fed with PercentageType values.

Fixes #3194

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
@@ -66,9 +66,25 @@ data class Widget(
 
     val mappingsOrItemOptions get() = if (mappings.isEmpty() && item?.options != null) item.options else mappings
 
-    val minValue get() = min(rawMinValue ?: item?.minimum ?: 0f, rawMaxValue ?: item?.maximum ?: 100f)
-    val maxValue get() = max(rawMinValue ?: item?.minimum ?: 0f, rawMaxValue ?: item?.maximum ?: 100f)
-    val step get() = abs(rawStep ?: item?.step ?: 1f)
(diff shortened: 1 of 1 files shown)


rkoshak:
What about the case where someone wants a Dimmer but only to go between 5 and 75. Shouldn't the min/max be OK in that case? I can't imagine why someone would want to do that but I guarantee we'll find those users with this change. ;-)

maniac103:
In that case he needs to set the min and max values on the sitemap widget. If those are set, they override everything else. This change basically is about whether or not to use the item (state description) min/max values if the sitemap min/max values are *not* set.

mueller-ma (approved):

PR #3203, opened 2023-02-07
Refine section switch with two mappings

maniac103:
Those should be small enough to fit into one line with label and value.


maniac103:
Not entirely sure yet whether it's a good idea or inconsistent, but a larger number of users seems to use a double mapping as toggle switch. Open for :+1: :-1:

mueller-ma:
I'd say 👍, but what should we do when the next person comes with: I have a switch with three mappings, but all of them are only one letter?
Add another special case?

maniac103:
> but what should we do when the next person comes with: I have a switch with three mappings, but all of them are only one letter?

That's precisely why I'm torn ;-) But this specific case should be OK, I think, as it's pretty similar to a `Switch` without mappings.

alltime84:
Thanks a lot for implementing! It looks much more consistent now:
image

before:
image

PR #3224, opened 2023-02-10
Fix window background on API <= 28

### LINKED ISSUE #3219: UI broken on Android API 22
https://github.com/openhab/openhab-android/commit/d7db481e40f9f622612316bf2cbf93969c94d516 introduced a sever bug on API 22. UI elements won't vanish anymore when they should.

Directly after start: Progress indicator still shown
Screenshot_2023-02-09-16-36-29

When scrolling the Sitemap:
Screenshot_2023-02-09-16-36-33

maniac103:
Fixes #3219

### DIFF
--- a/mobile/src/main/res/values-night/themes.xml
+++ b/mobile/src/main/res/values-night/themes.xml
@@ -21,7 +21,7 @@
         <item name="colorErrorContainer">#93000A</item>
         <item name="colorOnError">#690005</item>
         <item name="colorOnErrorContainer">#FFDAD6</item>
-        <item name="android:colorBackground">#201A18</item>
+        <item name="android:colorBackground">@color/background_default_theme_dark</item>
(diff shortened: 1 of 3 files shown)


maniac103:
@mueller-ma If you got a bit of time on your hands, and maybe have a better internet download rate then me, it would be great to try on which API level the code before this PR starts working, so we know exactly what 'older API level' means.

maniac103:
Change on the Android side probably happened [in 2018](https://android.googlesource.com/platform/frameworks/base/+/56c68bc296933fb5efbdce95499ecdef6fecafb6), so 'older API level' probably means '<= 28'.

mueller-ma:
The current main (without this fix) works on API 29 and is broken on 28. Your assumption was right.

mueller-ma (approved):

PR #3231, opened 2023-02-13
Enable dynamic theme by default

### LINKED ISSUE #3185: Activate "Dynamic colors" by default
Version 3 changes the typical orange color of openHAB in the headers to a slightly weird brownish tone, which does not really seem to match the taste of many users and which also has no association to openHAB itself.

I noticed the setting "Dynamic colors", which in my case immediately improved my impression of the new design, as it chose the colors that I have anyhow selected for my OS. If a user does not like those colors, it is then clearly their own fault. 😎 

I'd therefore like to propose to enable that setting by default instead of having users to find it first. Wdyt?

mueller-ma:
Requires to apply default preference values in `OpenHabApplication`.

Closes #3185


mobile/src/main/res/values-v33/strings.xmk.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
maniac103, line 1: File seems misnamed

PR #3243, opened 2023-02-15
Improve section switch overflow handling

### LINKED ISSUE #3238: SectionSwitchViewHolder: MaterialButton.isCheckable not always set when view is recycled.
`MaterialButton.isCheckable` is set for the overflow button, but not for regular ones. This can cause issues when the view is recycled.

https://github.com/openhab/openhab-android/blob/main/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt#L771-L786

maniac103:
- Only assign needed width to overflow button
- Allow more buttons to be shown on large tablets

Fixes #3238

### DIFF
--- a/mobile/src/main/res/values/dimens.xml
+++ b/mobile/src/main/res/values/dimens.xml
@@ -13,5 +13,5 @@
     <dimen name="outlineButtonStrokeWidth">2dp</dimen>
     <dimen name="widget_background_radius">8dp</dimen>
 
-    <integer name="section_switch_max_buttons">5</integer>
+    <integer name="section_switch_max_buttons">4</integer>
 </resources>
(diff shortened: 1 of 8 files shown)


mobile/src/main/res/values/dimens.xml
```diff
@@ -13,5 +13,5 @@
     <dimen name="outlineButtonStrokeWidth">2dp</dimen>
     <dimen name="widget_background_radius">8dp</dimen>
 
-    <integer name="section_switch_max_buttons">5</integer>
+    <integer name="section_switch_max_buttons">4</integer>
```
mueller-ma, line 16: At least on my device with 450dp width 5 regular buttons can be shown without any issue. Does it make sense to add resource files with `section_switch_max_buttons = [45678]`?

maniac103: Sure, why not? A regular phone is about 360dp wide, I think.

mueller-ma (approved):

PR #3244, opened 2023-02-15
Fine tune widget value slider value update behavior

maniac103:
Always doing live update has issues because we essentially only want those for lights, but not for e.g. roller shutters or setpoints. Try to be smart about it by
- not doing live updates for setpoints at all
- doing them only for some item types (number, dimmer) for slider widgets


mueller-ma:
Looks good for sliders on my server. Let's beta test this.

PR #3250, opened 2023-02-15
Use yAxisDecimalPattern Chart query parameter

maniac103:
Pretty ugly due to the (very limited) format conversion :-/

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
@@ -20,6 +20,7 @@ import java.util.IllegalFormatException
 import java.util.Locale
 import java.util.regex.Pattern
 import kotlin.math.roundToInt
+import kotlinx.parcelize.IgnoredOnParcel
 import kotlinx.parcelize.Parcelize
 
 @Parcelize
@@ -146,6 +147,9 @@ data class ParsedState internal constructor(
         val unit: String? = null,
         val format: String? = null
     ) : Parcelable {
+        @IgnoredOnParcel
+        val actualFormat = format?.replace("%unit%", unit.orEmpty())
+
         override fun toString(): String {
             return toString(Locale.getDefault())
         }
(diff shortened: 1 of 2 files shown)


mueller-ma:
Do we need to parse it? I think you can just pass it the query string.

maniac103:
Try it ... I did ;-)
Seriously, no, the formats aren't compatible. For e.g. 2 decimal digits, the state description one (printf-like) is e.g. `%.2f °C`, while the chart one (DecimalFormat) wants `0.00 °C`.

J-N-K:
Exactly. That's because we use XChart and that requires a format string that is compatible with `DecimalFormat`. I see five options here:

1. Write a format-translator in every consumer (like you did)
2. Write a format-translator and integrate it in core (OH4 would be a good candidate for breaking changes)
3. Try to incorporate changes in xchart (IMO not really an option, given that there have been nearly no contributions in the past year and we wait for a 3-line fix to be merges since two month)
4. Fork and use a customized version 
5. Find another library

maniac103:
Option 2 sounds at least better than 1, if possible, since that requires the translation code only once and at the same time improves the OH API.
Given option 3, option 4 might also be an option since there won't be too much effort tracking upstream changes?

mueller-ma:
I prefer option 2 as it avoids writing the same code several times. I think the breaking change shouldn't cause to much issues as the parameter isn't even documented yet: https://next.openhab.org/docs/ui/sitemaps.html#element-type-chart

mueller-ma:
Actually I did try ;)

According to https://github.com/openhab/openhab-core/pull/2937#issuecomment-1114018238 you need to put the pattern in the sitemap that can be read by the chart.

I tried:
* `Chart item=... yAxisDecimalPattern="#"` which make Basic UI to call `https://.../chart?...&yAxisDecimalPattern=%23`
* `Chart item=... yAxisDecimalPattern="#.##E0"` which make Basic UI to call `https://.../chart?...&yAxisDecimalPattern=%23.%23%23E0`

So it seems that the pattern is just passed (and encoded), but Basic UI doesn't convert it in any way.

mueller-ma:
I opened #3274 that just passes this parameter to the chart URL. I'd merge that one as it implements the same behavior than Basic UI.

In the future there could be an enhancement in the chart library or somewhere else on the server to accept printf-like format as well or maybe fallback to the Item state description, if `yAxisDecimalPattern` is absent.

maniac103:
Makes sense. I missed the sitemap parameter part of this.

Issue #3147, opened 2023-01-09
[FR] OpenHAB UI 3 URL is hard-coded to myopenhab.orgmyopenhab.org


lsafelix75:
Hi,

I plan to modify the following line of code in (org.openhab.habdroid.ui.activity.Oh3UiWebViewFragment) and add in a new setting to append user-configurable subdomain (etc “home.”) so that it will work for private OpenHab Cloud. What you think?
```

override fun modifyUrl(orig: HttpUrl): HttpUrl {
        if (orig.host == "myopenhab.org") {
            return orig.newBuilder()
                .host("home.myopenhab.org")
                .build()
        }
        return orig
    }
```

mueller-ma:
I'd rather add a new API endpoint on the cloud server that returns the URL for main UI or hard-code a new path, e.g. `https://cloud/mainui` -> return main UI.

mueller-ma:
I can tell for the official server:

- `myopenhab.org` is required for notifications
- `home.myopenhab.org` is required for Main UI

Everything else works with both URLs. There's also an error message when you enter  the wrong URL in the settings: https://github.com/openhab/openhab-android/blob/main/mobile/src/main/res/values/strings.xml#L180

mueller-ma:
It seems that the Main UI domain is hardcoded to `home.`: https://github.com/openhab/openhab-cloud/issues/332

PR #3257, opened 2023-02-18
Main UI URL is hard-coded to home.myopenhab.org

lsafelix75:
Refer to this issue and this is a temporary workaround until a full solution is proposed. 
https://github.com/openhab/openhab-android/issues/3147

Signed-off-by: Felix Lo <lsafelix75@gmail.com>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.1.1'
+        classpath 'com.android.tools.build:gradle:8.1.2'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }


mobile/src/main/java/org/openhab/habdroid/ui/activity/Oh3UiWebViewFragment.kt
```diff
@@ -28,9 +28,9 @@ class Oh3UiWebViewFragment : AbstractWebViewFragment() {
     override val shortcutAction = MainActivity.ACTION_OH3_UI_SELECTED
 
     override fun modifyUrl(orig: HttpUrl): HttpUrl {
-        if (orig.host == "myopenhab.org") {
+        if (orig.host.substringBefore('.') != "home") {
```
mueller-ma, line 31:
```suggestion
        if (!orig.host.startsWith("home.")) {
```

mueller-ma:
I'm not sure about this workaround. There could be setups that currently work.

Maybe it's better to add a server-specific URL setting, under advanced? It should be pre-filled for myopenhab.org users.

mueller-ma:
@maniac103 WDYT here? Add this workaround or add a new setting to override the Main UI url? If latter is done, something similar could be done for #3260.

maniac103:
I don't know the logic in openhab-cloud too well, but if the 'home' prefix is indeed hardcoded there as implied by https://github.com/openhab/openhab-cloud/issues/332 I think this workaround should be fine? Maybe it should be extended to not prepend 'home' if the hostname starts with 'remote.', though, so it matches the openhab-cloud logic.

mueller-ma:
👍 

@lsafelix75 Can you implement the "hostname doesn't start with remote." check?

lsafelix75:
> 👍
> 
> @lsafelix75 Can you implement the "hostname doesn't start with remote." check?

The problem is we need to create a link URL when user clicks on "it", it will be proxied to openhab instance. So, may be instead of hardcoded to home. we hardcode it to remote. because both will enter the same logic in openhabcloud app.js below:

 ```
// If host matches names for full /* proxying, go ahead and just proxy it.
    if (host.indexOf('remote.') === 0 || host.indexOf('home.') === 0) {
        //make sure this was not set by another server
        if (req.url.indexOf('/remote') != 0) {
            req.url = '/remote' + req.url;
        }
    }
```

dastrix80:
When is this being fixed and why is it closed?

PR #3282, opened 2023-02-28
Color SVG icons

### LINKED ISSUE #3204: Change color of "colorless" SVG icons dependent on design
If you use own SVG icons without any "color information" (like fill, stroke, ...) the used color is always "black" independent to the the design you use (light or dark). For light design this maybe ok, but for dark design, you cannot see the icon anymore.

Screenshot_20230207-182555  Screenshot_20230207-182635

Of course, it is possible to add colors in the SVG files, but if you want to be flexible (one user at home is using the light design, the other the dark design), you cannot use the same icons.

It would be great, if you can choose the default color (for colorless SVG icons) on the basis of the design or perhaps on the "dynamic colors based on the wallpaper".

Thank you.

_[This is my first feature request based on the app comparison with other smarthome apps for other smarthome-systems]._

mueller-ma:
The keyword `currentColor` can be used to color SVGs. This seems to be the best solution for the "colorless" icon issue.

Closes #3204

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
@@ -33,9 +33,11 @@ import org.openhab.habdroid.model.CloudNotification
 import org.openhab.habdroid.model.IconResource
 import org.openhab.habdroid.ui.MainActivity
 import org.openhab.habdroid.util.HttpClient
+import org.openhab.habdroid.util.IconBackground
 import org.openhab.habdroid.util.ImageConversionPolicy
 import org.openhab.habdroid.util.PendingIntent_Immutable
 import org.openhab.habdroid.util.determineDataUsagePolicy
+import org.openhab.habdroid.util.getIconFallbackColor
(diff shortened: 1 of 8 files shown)


maniac103:
> The keyword currentColor can be used to color SVGs.

What do you mean by that? The patch doesn't contain any reference to 'currentColor'.
Besides that, LGTM.

maniac103:
OK, so that CSS you're setting only applies in the `fill=currentColor` case and not to e.g. the example icon in https://github.com/BigBadaboom/androidsvg/issues/264 which doesn't provide any color?

I also assume we want don't want to mix #3289 with this and use color filters there, right?

mueller-ma:
> OK, so that CSS you're setting only applies in the fill=currentColor case and not to e.g. the example icon in https://github.com/BigBadaboom/androidsvg/issues/264 which doesn't provide any color?

Correct.

> I also assume we want don't want to mix https://github.com/openhab/openhab-android/issues/3289 with this and use color filters there, right?

For that I'd just tint the whole bitmap. Basic UI doesn't support that for the `oh` icon source yet: https://github.com/openhab/openhab-webui/pull/1751

PR #3293, opened 2023-03-04
Refactor window insets handling

maniac103:
Fix cases where padding wasn't correct with or without fullscreen mode on some API levels.
While at it, also replace some related deprecated code and deduplicate toolbar handling.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/AboutActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/AboutActivity.kt
@@ -44,9 +44,6 @@ class AboutActivity : AbstractBaseActivity(), FragmentManager.OnBackStackChanged
         setContentView(R.layout.activity_about)
         supportFragmentManager.addOnBackStackChangedListener(this)
 
-        setSupportActionBar(findViewById(R.id.openhab_toolbar))
-        supportActionBar?.setDisplayHomeAsUpEnabled(true)
-
         if (savedInstanceState == null) {
             val f = AboutMainFragment()
             f.arguments = intent.extras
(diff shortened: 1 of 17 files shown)


maniac103:
@mueller-ma I *hiope* this fixes all issues about status/navigation bar handling that we had, but there's a lot of possible cases (API level, fullscreen yes/no, different activities...). I window how we can get this the maximum amount of testing without breaking anybody's workflow?

mueller-ma:
I did some testing with several emulators and found these issues:

1. API 24, light theme, not fullscreen: Snackbar is shown behind app bar:
grafik

2. API 24, 28, 29 and 33, fullscreen: Microphone icon is shown completely:
grafik

3. API 33, not fullscreen: There's empty space under the app bar:
grafik

maniac103:
Your level of thorough testing is amazing :+1: 

> API 24, light theme, not fullscreen: Snackbar is shown behind app bar:

Should be fixable by applying the bottom padding to the `R.id.coordinator` instead of the `R.id.activity_content` view.

> API 24, 28, 29 and 33, fullscreen: Microphone icon is shown completely:

Looks like the toolbar has a top padding set ... but it definitely shouldn't have that in fullscreen mode.

> API 33, not fullscreen: There's empty space under the app bar

Display cutout or something like that? `adb shell dumpsys window displays` should say what's up there (see the `WindowInsetsStateController` section) - the top space should be equal to height of status bar + height of display cutout.

In any case, I'll check it.

maniac103:
Fixed all 3 (I hope) ... the insets behavior among different API levels + insufficient ways of overriding AppBarLayout's behavior is such a clusterfuck :-/

mueller-ma (approved):
LGTM! Let's merge and wait for the next issue :/

PR #3305, opened 2023-03-10
Add compact layout for frames

### LINKED ISSUE #3298: Use less space for Frames
I received this request by mail: in compact mode the padding of the frames could be less.

mueller-ma:
Closes #3298

Comparison between default and compact view:


maniac103:
Maybe it's just me, but I find the compact version rather ugly. Maybe reduce outer margin to 4dp and use 4dp padding for the TextView?

mueller-ma:
Updated margin and padding (before and after):

Issue #3306, opened 2023-03-10
Release version 3.4.0


mueller-ma:
### Pre-release checklist

- [x] Check for major crashes in latest stable version
- [x] Check for major crashes in beta versions since last stable version
- [x] Merge new translations

### Release notes

• Fix issues with fullscreen mode
• Support icon sets

### Blocking issues/PRs

- [x] #3298 
- [x] Wait until [3.3.5-beta](https://github.com/openhab/openhab-android/releases/tag/3.3.5-beta) or newer is on F-Droid

### Post-release checklist

- [x] Release in Play Store in the open test track
- [x] Release in Play Store to 20%
- [x] Release in Play Store to 50%
- [x] Release in Play Store to 100%
- [x] Check for new crashes

PR #3311, opened 2023-03-14
Allow custom Frontail URL

### LINKED ISSUE #3260: Allow different port than the default 9001 for Frontail
I'd love to see a setting where the user can change the port of frontail. I am running frontail in a docker container with its internal port `9001` not exposed to the external port `9001` but to a different port to have port ranges per service (e.g. everything related to openhab gets ports between `5000` and `5999`)

Currently the port is [hardcoded](https://github.com/openhab/openhab-android/blob/main/mobile/src/main/java/org/openhab/habdroid/ui/activity/FrontailWebViewFragment.kt#L34). I suggest to add a setting in the [server configuration](https://github.com/openhab/openhab-android/blob/main/mobile/src/main/res/xml/preferences_server.xml). Maybe under `Advanced`

mueller-ma:
I implemented it this way so users can not only change port, but also scheme and/or host. It also allows us to re-use existing code, e.g. UrlInputPreference.

Closes #3260


mueller-ma:
I want to fully release [3.4.0](https://github.com/openhab/openhab-android/issues/3306) first.

Issue #3285, opened 2023-02-28
Support Sitemap widget Input


mueller-ma:
~~I just discovered the Sitemap widget "List". What does it do and should it be supported?~~

A new sitemap widget `Input` was introduced: https://github.com/openhab/openhab-core/pull/3398
We need to add support for it.
NB: Also take care of https://github.com/openhab/openhab-core/pull/3418 in case it's merged.

mueller-ma:
And there's also a new widget called Input: https://github.com/openhab/openhab-core/commits/main/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext

We definitely need a place where all maintainers of Sitemap UIs come together, so new features aren't discovered "by accident".

mherwege:
@mueller-ma @maniac103
Sorry if I took you by surprise on this. It has been on my wish list for a while, so I implemented the Input element myself in core and BasicUI.
I absolutely want to see have it in the Android App as well. I never looked at the Android app code and never touched Kotlin, but I am still willing to have a go at it. It will take time though, so don’t expect it soon from my side. And I wouldn’t mind if someone else takes it on. There is enough I can still contribute to.

maniac103:
>  It is probably implemented in BasicUI but I have to confirm it and test it.

It is (I at least found it in the code). I wonder, though, whether the better option is 
- documenting it or
- removing the code for it

All it does is splitting up the (`String`, I suppose) item state at some separator defined in the widget, and then displaying the lines resulting from that beneath each other, so it's (AFAICT) essentially a multi-line text widget. Is this really that useful?

lolodomo:
I tested the List widget in Basic UI and I should admit it does not work well ! No dynamic update of the value and more important it breaks the layout !
I was very annoyed with the idea to remove it (breaking existing setups) but if it was really used, we will have issue opened to fix it.

That being said, I believe we got a 4.0 wish to add such widget in 4.0. To be verified. If there is a real need, I can also try to fix it in Basic UI.

maniac103:
> (breaking existing setups)

Given it was never documented, was it ever officially supported? Neither Android nor iOS apps do support it.

> I believe we got a 4.0 wish to add such widget in 4.0. To be verified.

If you find it, please also link it here :-)

maniac103:
Support for List was removed from core: https://github.com/openhab/openhab-core/pull/3423
Changing scope of the issue to only cover the input widget.

Issue #1787, opened 2020-01-23
CI: Fail on compiler warnings


mueller-ma:
### Is your feature request related to a problem? Please describe.
The project is free of compiler warnings and a this shouldn't change.

### Describe the solution you'd like
Our GitHub action "build" should fail if there are any compiler warnings.

mueller-ma:
Also check if the actions is faster if the daemon is not disabled (https://github.com/openhab/openhab-android/blob/master/.github/workflows/build.yml#L17).

PR #3337, opened 2023-04-12
support for Input widget

mherwege:
closes [#3285](https://github.com/openhab/openhab-android/issues/3285)
depends on [#3431](https://github.com/openhab/openhab-core/pull/3431)

This PR adds support for the new input widget to the android app.

This is my first time writing for an Android app and writing code in Kotlin. So I will need help to get this finalized, but I hope this is a good basis to start.
I have been running the app in an Android emulator from Android studio, not on a real device. The full syntax is supported (see the BasicUI PR [#1787](https://github.com/openhab/openhab-webui/pull/1787) for my test data and expected behaviour).
Note you will need to run a core with the core PR included to be able to test this.

Known issues at this point observed in the emulator, probably due to my limited understanding of Android development:

- Keyboard popups don't automatically go away after input.
- My sample sitemap has a text input field as the first element in the sitemap. After using one of the date/time input fields, I also get the date/time input field as a popup on this first field, rather than a keyboard.

Appreciate all feedback on this.

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="11" />
-  </component>
-</project>
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="11" />
+    <bytecodeTargetLevel target="17" />
   </component>
 </project>
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="CompilerConfiguration">
+    <bytecodeTargetLevel target="11" />
+  </component>
+</project>
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="17" />
+    <bytecodeTargetLevel target="11" />
   </component>
 </project>
--- a/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
@@ -17,10 +17,8 @@ import android.graphics.Color
 import android.location.Location
 import android.os.Parcelable
 import java.time.LocalDateTime
-import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 import java.time.format.DateTimeParseException
-import java.time.format.FormatStyle
 import java.util.IllegalFormatException
 import java.util.Locale
 import java.util.regex.Pattern
@@ -42,7 +40,7 @@ data class ParsedState internal constructor(
     val asHsv: HsvState?,
     val asBrightness: Int?,
     val asLocation: Location?,
-    val asDateTime: DateTimeState?
+    val asDateTime: LocalDateTime?
 ) : Parcelable {
     override fun equals(other: Any?): Boolean {
         return other is ParsedState && asString == other.asString
@@ -97,9 +95,11 @@ data class ParsedState internal constructor(
             val stateSplit = state.split(",")
             if (stateSplit.size == 3) { // We need exactly 3 numbers to operate this
                 try {
-                    return HsvState(stateSplit[0].toFloat(),
+                    return HsvState(
+                        stateSplit[0].toFloat(),
                         stateSplit[1].toFloat() / 100,
-                        stateSplit[2].toFloat() / 100)
+                        stateSplit[2].toFloat() / 100
+                    )
                 } catch (e: NumberFormatException) {
                     // fall through
                 }
@@ -145,30 +145,13 @@ data class ParsedState internal constructor(
 
         private val HSB_PATTERN = Pattern.compile("^([0-9]*\\.?[0-9]+),([0-9]*\\.?[0-9]+),([0-9]*\\.?[0-9]+)$")
 
-        internal fun parseAsDateTime(state: String, format: String?): DateTimeState? {
+        internal fun parseAsDateTime(state: String): LocalDateTime? {
             return try {
-                var st = state.trim().split(".")[0]
-                val formatter = if (format != null) {
-                    DateTimeFormatter.ofPattern(format)
-                } else {
-                    st = st.replace(" ", "T")
-                    if (TIME_PATTERN.matcher(st).find()) {
-                        val refDate = LocalDateTime
-                            .ofEpochSecond(0, 0, ZoneOffset.UTC)
-                            .format(DateTimeFormatter.ISO_LOCAL_DATE)
-                        st = "${refDate}T$st"
-                    }
-                    if (!st.contains("T")) st = "${st}T00:00:00"
-                    DateTimeFormatter.ISO_LOCAL_DATE_TIME
-                }
-                val dt = LocalDateTime.parse(st, formatter)
-                DateTimeState(dt, format)
+                LocalDateTime.parse(state.split(".")[0], DateTimeFormatter.ISO_LOCAL_DATE_TIME)
             } catch (e: DateTimeParseException) {
                 null
             }
         }
-
-        private val TIME_PATTERN = Pattern.compile("^\\d{2}:\\d{2}")
     }
 
     @Parcelize
@@ -205,59 +188,10 @@ data class ParsedState internal constructor(
             return if (format != null && format.contains("%d")) value.roundToInt() else value
         }
     }
-
-    @Parcelize
-    class DateTimeState internal constructor(
-        val value: LocalDateTime,
-        val format: String? = null
-    ) : Parcelable {
-        override fun toString(): String {
-            return toString(Locale.getDefault())
-        }
-
-        /**
-         * Like [toString][.toString], but using a specific locale for formatting.
-         */
-        fun toString(locale: Locale): String {
-            if (!format.isNullOrEmpty()) {
-                try {
-                    return value.format(DateTimeFormatter.ofPattern(format, locale))
-                } catch (e: IllegalFormatException) {
-                    // State format pattern doesn't match the actual data type
-                    // -> ignore and fall back to our own formatting
-                }
-            }
-            return value.format(DateTimeFormatter.ISO_LOCAL_DATE_TIME).replace("T", " ")
-        }
-
-        fun toISOLocalDateTime(): String {
-            return value.format(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
-        }
-
-        fun toLocalDate(): String {
-            return value.format(DateTimeFormatter.ofLocalizedDate(FormatStyle.SHORT))
-        }
-
-        fun toLocalTime(): String {
-            return value.format(DateTimeFormatter.ofLocalizedTime(FormatStyle.SHORT))
-        }
-
-        fun toLocalDateTime(): String {
-            return value.format(DateTimeFormatter.ofLocalizedDateTime(FormatStyle.SHORT))
-        }
-
-        fun getActualValue(): LocalDateTime {
-            return value
-        }
-    }
 }
 
 fun ParsedState.NumberState?.withValue(value: Float): ParsedState.NumberState {
     return ParsedState.NumberState(value, this?.unit, this?.format)
-}
-
-fun ParsedState.DateTimeState?.withValue(value: LocalDateTime): ParsedState.DateTimeState {
-    return ParsedState.DateTimeState(value, this?.format)
 }
 
 /**
@@ -276,6 +210,6 @@ fun String?.toParsedState(formatPattern: String? = null): ParsedState? {
         ParsedState.parseAsHsv(this),
         ParsedState.parseAsBrightness(this),
         ParsedState.parseAsLocation(this),
-        ParsedState.parseAsDateTime(this, formatPattern)
+        ParsedState.parseAsDateTime(this)
     )
 }
--- a/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
@@ -16,6 +16,10 @@ package org.openhab.habdroid.model
 import android.graphics.Color
 import android.location.Location
 import android.os.Parcelable
+import java.time.LocalDateTime
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
+import java.time.format.DateTimeParseException
 import java.util.IllegalFormatException
 import java.util.Locale
 import java.util.regex.Pattern
@@ -36,7 +40,8 @@ data class ParsedState internal constructor(
     val asNumber: NumberState?,
     val asHsv: HsvState?,
     val asBrightness: Int?,
-    val asLocation: Location?
+    val asLocation: Location?,
+    val asDateTime: DateTimeState?
 ) : Parcelable {
     override fun equals(other: Any?): Boolean {
         return other is ParsedState && asString == other.asString
@@ -138,6 +143,31 @@ data class ParsedState internal constructor(
         }
 
         private val HSB_PATTERN = Pattern.compile("^([0-9]*\\.?[0-9]+),([0-9]*\\.?[0-9]+),([0-9]*\\.?[0-9]+)$")
+
+        internal fun parseAsDateTime(state: String, format: String?) : DateTimeState? {
+            return try {
+                var st = state.trim().split(".")[0]
+                val formatter = if (format != null) {
+                    DateTimeFormatter.ofPattern(format)
+                } else {
+                    st = st.replace(" ","T")
+                     if (TIME_PATTERN.matcher(st).find()) {
+                         val refDate = LocalDateTime
+                             .ofEpochSecond(0, 0, ZoneOffset.UTC)
+                             .format(DateTimeFormatter.ISO_LOCAL_DATE)
+                         st = "${refDate}T${st}"
+                    }
+                    if (!st.contains("T"))  st = "${st}T00:00:00"
+                    DateTimeFormatter.ISO_LOCAL_DATE_TIME
+                }
+                val dt = LocalDateTime.parse(st, formatter)
+                DateTimeState(dt, format)
+            } catch (e: DateTimeParseException) {
+                null
+            }
+        }
+
+        private val TIME_PATTERN = Pattern.compile("^\\d{2}:\\d{2}")
     }
 
     @Parcelize
@@ -174,6 +204,47 @@ data class ParsedState internal constructor(
             return if (format != null && format.contains("%d")) value.roundToInt() else value
         }
     }
+
+    @Parcelize
+    class DateTimeState internal constructor(
+        val value: LocalDateTime,
+        val format: String? = null
+    ) : Parcelable {
+        override fun toString(): String {
+            return toString(Locale.getDefault())
+        }
+
+        /**
+         * Like [toString][.toString], but using a specific locale for formatting.
+         */
+        fun toString(locale: Locale): String {
+            if (!format.isNullOrEmpty()) {
+                try {
+                    return value.format(DateTimeFormatter.ofPattern(format, locale))
+                } catch (e: IllegalFormatException) {
+                    // State format pattern doesn't match the actual data type
+                    // -> ignore and fall back to our own formatting
+                }
+            }
+            return value.format(DateTimeFormatter.ISO_LOCAL_DATE_TIME).replace("T", " ")
+        }
+
+        fun toISOLocalDateTime(): String {
+            return value.format(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+        }
+
+        fun toISOLocalDate(): String {
+            return value.format(DateTimeFormatter.ISO_LOCAL_DATE)
+        }
+
+        fun toISOLocalTime(): String {
+            return value.format(DateTimeFormatter.ISO_LOCAL_TIME)
+        }
+
+        fun getActualValue(): LocalDateTime {
+            return value
+        }
+    }
 }
 
 fun ParsedState.NumberState?.withValue(value: Float): ParsedState.NumberState {
@@ -183,17 +254,18 @@ fun ParsedState.NumberState?.withValue(value: Float): ParsedState.NumberState {
 /**
  * Parses a state string into the parsed representation.
  *
- * @param numberPattern Format to use when parsing the input as number
+ * @param formatPattern Format to use when parsing the input as number or date
  * @return null if state string is null, parsed representation otherwise
  */
-fun String?.toParsedState(numberPattern: String? = null): ParsedState? {
+fun String?.toParsedState(formatPattern: String? = null): ParsedState? {
     if (this == null) {
         return null
     }
     return ParsedState(this,
         ParsedState.parseAsBoolean(this),
-        ParsedState.parseAsNumber(this, numberPattern),
+        ParsedState.parseAsNumber(this, formatPattern),
         ParsedState.parseAsHsv(this),
         ParsedState.parseAsBrightness(this),
-        ParsedState.parseAsLocation(this))
+        ParsedState.parseAsLocation(this),
+        ParsedState.parseAsDateTime(this, formatPattern))
 }
--- a/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
@@ -251,6 +251,10 @@ fun ParsedState.NumberState?.withValue(value: Float): ParsedState.NumberState {
     return ParsedState.NumberState(value, this?.unit, this?.format)
 }
 
+fun ParsedState.DateTimeState?.withValue(value: LocalDateTime): ParsedState.DateTimeState {
+    return ParsedState.DateTimeState(value, this?.format)
+}
+
 /**
  * Parses a state string into the parsed representation.
  *
--- a/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
@@ -20,6 +20,7 @@ import java.time.LocalDateTime
 import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 import java.time.format.DateTimeParseException
+import java.time.format.FormatStyle
 import java.util.IllegalFormatException
 import java.util.Locale
 import java.util.regex.Pattern
@@ -233,12 +234,16 @@ data class ParsedState internal constructor(
             return value.format(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
         }
 
-        fun toISOLocalDate(): String {
-            return value.format(DateTimeFormatter.ISO_LOCAL_DATE)
+        fun toLocalDate(): String {
+            return value.format(DateTimeFormatter.ofLocalizedDate(FormatStyle.SHORT))
         }
 
-        fun toISOLocalTime(): String {
-            return value.format(DateTimeFormatter.ISO_LOCAL_TIME)
+        fun toLocalTime(): String {
+            return value.format(DateTimeFormatter.ofLocalizedTime(FormatStyle.SHORT))
+        }
+
+        fun toLocalDateTime(): String {
+            return value.format(DateTimeFormatter.ofLocalizedDateTime(FormatStyle.SHORT))
         }
 
         fun getActualValue(): LocalDateTime {
--- a/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
@@ -144,20 +144,20 @@ data class ParsedState internal constructor(
 
         private val HSB_PATTERN = Pattern.compile("^([0-9]*\\.?[0-9]+),([0-9]*\\.?[0-9]+),([0-9]*\\.?[0-9]+)$")
 
-        internal fun parseAsDateTime(state: String, format: String?) : DateTimeState? {
+        internal fun parseAsDateTime(state: String, format: String?): DateTimeState? {
             return try {
                 var st = state.trim().split(".")[0]
                 val formatter = if (format != null) {
                     DateTimeFormatter.ofPattern(format)
                 } else {
-                    st = st.replace(" ","T")
-                     if (TIME_PATTERN.matcher(st).find()) {
-                         val refDate = LocalDateTime
-                             .ofEpochSecond(0, 0, ZoneOffset.UTC)
-                             .format(DateTimeFormatter.ISO_LOCAL_DATE)
-                         st = "${refDate}T${st}"
+                    st = st.replace(" ", "T")
+                    if (TIME_PATTERN.matcher(st).find()) {
+                        val refDate = LocalDateTime
+                            .ofEpochSecond(0, 0, ZoneOffset.UTC)
+                            .format(DateTimeFormatter.ISO_LOCAL_DATE)
+                        st = "${refDate}T$st"
                     }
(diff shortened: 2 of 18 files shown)


.idea/compiler.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="11" />
+    <bytecodeTargetLevel target="17" />
```
mueller-ma, line 4: I'm not sure if this causes any unwanted issues. Did you change this on purpose?

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -60,7 +60,8 @@ data class Widget(
     val yAxisDecimalPattern: String?,
     val switchSupport: Boolean,
     val height: Int,
-    val visibility: Boolean
+    val visibility: Boolean,
+    val inputHint: String
```
mueller-ma, line 64: I'd parse it as an enum where unexpected values are parsed as `null`.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -217,6 +228,7 @@ fun Node.collectWidgets(parent: Widget?): List<Widget> {
     var step = 1f
     var refresh = 0
     var height = 0
+    var inputHint = "text"
```
mueller-ma, line 231: No need to updated `Node.collectWidgets()`. `Node` refers to xml nodes, so this function is used to parse the OH1 api.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -118,6 +137,20 @@ class WidgetAdapter(
...
+            widget: Widget
+        ): MaterialDatePicker<Long>
+    }
+
+    interface TimePickerPresenter {
+        fun showTimePicker(
+            dateTime: LocalDateTime,
+            widget: Widget
+        ): MaterialTimePicker
+    }
```
mueller-ma, line 152: The pickers should be shown in a bottom sheet and therefore you can re-use `DetailBottomSheetPresenter` here.

.idea/compiler.xml
mherwege: No, I didn’t do this on purpose. This sits in the .idea folder. Should changes there be included?

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
mherwege: Removed

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
mherwege: Done

.idea/compiler.xml
mherwege: I removed the 2 changes in the .idea folder and it still looks like it compiles and runs fine.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mherwege: I don't know how I have to do that. The MaterialDatePicker and MaterialTimePicker are model components. I don't know how to change that to be part of a bottom sheet.

mherwege:
@mueller-ma Thank you for your very quick first feedback. I made a few adjustments already. However, I am struggling with making the pickers a bottom sheet. I have no idea on how to do that.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -557,6 +604,180 @@ class WidgetAdapter(
...
+            } else ""
+
+            val dataState = if (widget.state != null) {
+                if (widget.inputHint == Widget.Hint.Number)
+                    widget.state.asNumber?.formatValue()
+                else if (displayState.isNotEmpty()) displayState
+                else if (widget.item?.isOfTypeOrGroupType(Item.Type.DateTime) == true)
+                    when (widget.inputHint) {
+                        Widget.Hint.Date -> widget.state.asDateTime?.toISOLocalDate()
+                        Widget.Hint.Time -> widget.state.asDateTime?.toISOLocalTime()
```
maniac103, line 666: Wouldn't a locale dependent format be nicer/better here?

maniac103 (commented):
Besides the comments, I have some cleanup commits here: https://github.com/maniac103/openhab.android/commits/input

I plan to review/test this more closely once there's a server snapshot with support for `Input` and `inputHint`.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
maniac103: Yes, unfortunately they're a `DialogFragment` instead of a `ViewGroup` (or something like that). One interface is sufficient nevertheless, I think: https://github.com/maniac103/openhab.android/commit/09c5652d3bd45c03ff5d3d9c30b2f16302de5476

mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
```diff
@@ -138,6 +143,31 @@ data class ParsedState internal constructor(
...
+
+        internal fun parseAsDateTime(state: String, format: String?): DateTimeState? {
+            return try {
+                var st = state.trim().split(".")[0]
+                val formatter = if (format != null) {
+                    DateTimeFormatter.ofPattern(format)
+                } else {
+                    st = st.replace(" ", "T")
+                    if (TIME_PATTERN.matcher(st).find()) {
+                        val refDate = LocalDateTime
```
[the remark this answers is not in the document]
mherwege: > All of this `LocalDateTime` stuff seems to be API 26+ only.

It indeed is. That's why core library desugaring is used to support it on older API versions ([see](https://developer.android.com/studio/write/java8-support)). The alternative is to not use the Java Date and Time classes and use the pre Java 8 Calendar class (for a discussion, [see](https://medium.com/nanogiants/handling-dates-on-android-1fccccde9d54)). I needed date handling and didn't like to go for old classes that have their own issues, or create an extra external dependency.

The logic is indeed quite complex. The default representation in the UI (without a specific format pattern provided in the item or sitemap definition) would be "YYYY-MM-DD" for date, "hh:mm:ss" for time and "YYYY-MM-DD hh:mm:ss" for datetime. This code interprets any of these and converts them into an ISO local date time format (YYYY-MM-DDThh:mm:ss). It also drops nanoseconds and timezones/offset from the received time string (anything after the .). It looks like we get the full date time with nanoseconds and timezone from OH, but are expected to send back ISO local date time. If we only have time, the resulting datetime will be aligned to eppoch 0.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -557,6 +604,180 @@ class WidgetAdapter(
...
+            inputText.doAfterTextChanged {
+                if (!isBinding) {
+                    hasChanged = true
+                }
+            }
+            inputText.setOnFocusChangeListener { _, hasFocus ->
+                if (!hasFocus && hasChanged) {
+                    updateValue()
+                }
+            }
```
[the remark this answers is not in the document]
mherwege: BasicUI gets away without the need for an extra update button. And I would prefer that. Also notice that it is not relevant for date/time, that have their own buttons on the popup. For me, the change on focus didn't give me a problem (apart from the keyboard not being close automatically, which I think it should be possible to resolve).

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -557,6 +604,180 @@ class WidgetAdapter(
...
+
+            inputTextLayout.suffixText = if (widget.inputHint == Widget.Hint.Number) {
+                widget.state?.asNumber?.unit
+            } else null
+
+            // Don't directly edit field for date/time when inputHint set, but open popup when clicked
+            if (
+                (widget.inputHint == Widget.Hint.Date) or
+                (widget.inputHint == Widget.Hint.Time) or
+                (widget.inputHint == Widget.Hint.Datetime)
```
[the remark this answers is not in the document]
mherwege: Thanks.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mherwege: Yes, potentially. I will check that out.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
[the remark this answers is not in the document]
maniac103: > BasicUI gets away without the need for an extra update button

Well, yes, it's certainly *possible*. I'm just not sure whether it's *good UX*. The button would automatically give you a 'hook' for hinding the keyboard. I don't see how it would be possible to auto-hide the keyboard otherwise. Maybe [this API](https://developer.android.com/reference/android/widget/TextView.html#setOnEditorActionListener%28android.widget.TextView.OnEditorActionListener%29), handling keyboard 'Enter' key press, could be a compromise.

> Also notice that it is not relevant for date/time, that have their own buttons on the popup

I am aware, it's why I suggested the button: that way, *all* inputs would have an explicit way of saving.

mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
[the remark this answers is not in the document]
maniac103: > It indeed is. That's why core library desugaring is used to support it on older API versions ([see](https://developer.android.com/studio/write/java8-support)).

Nevermind that part. It looks like a 'Sync project with Gradle files' resolved the errors AS showed me. Not sure why that didn't happen automatically.

> It looks like we get the full date time with nanoseconds and timezone from OH, but are expected to send back ISO local date time

So the date/time sent to us isn't ISO 8601? Wouldn't it be better to use a custom format string for parsing in that case instead of massaging the input into the ISO format?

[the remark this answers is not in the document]
maniac103: Actually, isn't the response always ISO 8601? I just created a test item named 'TestDate' of type 'DateTime', then updated it via CLI (`openhab:update TestDate "2023-04-30"`) and then fetched it via REST API. Result:

```json
{"link":"http://127.0.0.1:8080/rest/items/TestDate","state":"2023-04-30T00:00:00.000+0200","editable":true,"type":"DateTime","name":"TestDate","label":"New Item","category":"","tags":[],"groupNames":[]}
```

So where do(es) the other format(s) come from?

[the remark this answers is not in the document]
mherwege: > So the date/time sent to us isn't ISO 8601? Wouldn't it be better to use a custom format string for parsing in that case instead of massaging the input into the ISO format?

I tried but ran into trouble. The function is called from various places with different formats of the input, both date/time coming from OH and from updates done in the input fields. I couldn't configure one consistent format for parsing, and would have to rely on a sequence of try/catches with multiple formats. I think that only makes it worse.

mherwege:
@maniac103 The changes you proposed in https://github.com/openhab/openhab-android/commit/9ae709ccaa530b5f8345b3ae7b5bc8018045f404 did not solve the issue with the text inputs. It does not show the date or time pickers anymore on the input field, but the field is not editable anymore after updating a date or time ones.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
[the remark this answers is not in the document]
mherwege: The keyboard also has a return button, so I would prefer that. Looking at Google apps, I don't see them having an explicit confirm button, only on a popup or overlay page with multiple input elements. But I am OK to go the other way as well.

mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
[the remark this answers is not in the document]
maniac103: > both date/time coming from OH and from updates done in the input fields.

I don't understand the latter part. As far as I am aware, there's only two possible cases here:
- input field has format 'text' -> no parsing needed, we send the input to the server verbatim
- input field has any of the date/time formats -> no parsing needed, since we can construct the date/time state from numeric input

Which case am I missing here?

maniac103:
> The changes you proposed in https://github.com/openhab/openhab-android/commit/9ae709ccaa530b5f8345b3ae7b5bc8018045f404 did not solve the issue with the text inputs.

Those changes were only meant to cover the case of views being recycled - say, you have two `Input` widgets (one of type `date`, the other of type `text`) with some other widgets in between and scroll up and down. Result before the change: the `date` widget will set the view to be clickable etc., and when scrolling down to the `text` widget, it will still be clickable because it never was set to not be clickable anymore.

But as mentioned: I did *not* actually try this in practice, because there's no snapshot that allows for that ;-) I'll comment further on behavioral things once there's an easy way to test it.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -557,6 +584,185 @@ class WidgetAdapter(
...
+                Widget.InputTypeHint.Number -> widget.state?.asNumber?.unit
+                else -> null
+            }
+
+            // Don't directly edit field for date/time when inputHint set, but open popup when clicked
+            val isEditable = !(
+                    widget.inputHint == Widget.InputTypeHint.Date ||
+                    widget.inputHint == Widget.InputTypeHint.Time ||
+                    widget.inputHint == Widget.InputTypeHint.Datetime
+                )
```
maniac103, line 672: I think I'd prefer `val isEditable = widget.inputHint != Widget.InputTypeHint.Date && widget.inputHint != ...`

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -557,6 +584,185 @@ class WidgetAdapter(
...
+            inputText.isCursorVisible = isEditable
+            inputText.isFocusable = isEditable
+            inputText.isClickable = isEditable
```
maniac103, line 675: Won't this lead to the picker not shown anymore, since the `OnClickListener` isn't called if the view isn't clickable?
If the EditText needs to be clickable to be focusable, we should leave `isClickable` alone here and just set/unset the `OnClickListener`.

mherwege: It does show when Clickable is false. Unless this is something with the emulator, I think that is what it should be. I found discussions about it [here](https://stackoverflow.com/questions/14933330/datepicker-how-to-popup-datepicker-when-click-on-edittext).

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -557,6 +584,185 @@ class WidgetAdapter(
...
+                if (!isBinding) {
+                    hasChanged = true
+                }
+            }
+            inputText.setOnFocusChangeListener { _, hasFocus ->
+                if (!hasFocus && hasChanged) {
+                    updateValue()
+                }
+            }
+            inputText.setOnClickListener {
```
maniac103, line 610: Thinking about it, I think it would make more sense to instead override `handleRowClick()` and there to
- open date or time picker if appropriate
- if no picker is needed, focus the input field (`inputText.requestFocus()`)

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -557,6 +584,185 @@ class WidgetAdapter(
...
     }
 
+    class InputViewHolder internal constructor(initData: ViewHolderInitData) :
+        LabeledItemBaseViewHolder(initData, R.layout.widgetlist_inputitem, R.layout.widgetlist_inputitem_compact) {
+        private val inputTextLayout: TextInputLayout = itemView.findViewById(R.id.widgetinput)
+        private val inputText: TextInputEditText = itemView.findViewById(R.id.widgetinputvalue)
+        private var isBinding = false
+        private var hasChanged = false
+        private var showingDatePicker = false
+        private var showingTimePicker = false
```
maniac103, line 594: I don't think these are needed. The `FragmentManager` will replace the respective dialogs anyway when a new one is shown. Additionally, as long as the dialog is shown, the widget list won't receive click events.

mherwege: In the emulator, which is slow, I clicked a few times before the picker came up. The result was it came up multiple times. If the FragmentManager replaces and there is no risk, fine to remove.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mherwege: Done, I still needed to inputText.setOnClickListener to pass on the handleRowClick. If not, it only worked when clicking outside of the text field.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mherwege: Done.

mherwege:
> depends on [#3431](https://github.com/openhab/openhab-core/pull/3431)

@maniac103 @mueller-ma The core PR has been merged. The next snapshot should include it.

mherwege:
I don't understand where the build error is coming from.

mherwege:
> * The 'Enter' IME action moves to the next focusable field if there is any, while still hiding the keyboard. I think this is because the `android:imeAction` attribute needs to be set on the `EditText`, not the `TextInputLayout`.

Indeed, modified and improved the selection logic in the input field.

> * IMO, the 'Date', 'Time' and 'Datetime' shouldn't be shown as input fields, but as text with a dropdown button (similar to 'Selection' widgets). After all, it's not possible to (manually) edit them. It's probably useful to introduce a separate type of `ViewHolder` for them in that case.

I actually like the paradigm it uses. It is an input, and clicking on or in the input field opens a picker to allow editing it. It also allows me to reuse some of the logic should one wish to use the date/time input in a pure time field for whatever reason. This is of course personal opinion.

> * For 'Number' inputs, maybe set some alpha on the suffix text view (or use a secondary text style)?

What do you mean here? At the moment it is colored with the value and would also obey the sitemap color syntaxt settings. Could that still be the case if doing something special here? I don't know enough about this.

maniac103:
> Indeed, modified and improved the selection logic in the input field.

Was the intention to select the whole text when focussing it? If so, it didn't work for me -> https://github.com/maniac103/openhab.android/commit/dce155bd9442fa9f5a8563a8ed2ae595544e7cfd

> I actually like the paradigm it uses. It is an input, and clicking on or in the input field opens a picker to allow editing it. It also allows me to reuse some of the logic should one wish to use the date/time input in a pure text field without inputHint for whatever reason. This is of course personal opinion.

In my mind, it shouldn't look like an input field, because it does not actually behave like an input field, no matter whether the sitemap definition says 'Input' or not. It very much behaves like a 'Selection' widget (opens editor fragment on click), so (IMHO) should behave similar to that one. Doing so also allows getting rid of the back-and-forth when converting the selected date/time to string.
For your consideration/test, I've implemented it: https://github.com/maniac103/openhab.android/commit/1825cd017778c37e5eb356a4ccd9af69c9446daf

> What do you mean here? At the moment it is colored with the value and would also obey the sitemap color syntaxt settings. Could that still be the case if doing something special here? I don't know enough about this.

Something like this: https://github.com/maniac103/openhab.android/commit/fc6b7dc02e604f3e1d859eff0043c1765c42ea3b

BTW, after saving a value like 'abc' to a string input field, next time I got the text 'ParsedState(...)' filled into that one. Fix: https://github.com/maniac103/openhab.android/commit/9a49f9d1ffccb9f1d8a480a4569a26b1f5dbbfc3

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
maniac103: I can reproduce this when removing the booleans and clicking very fast. However, this would probably happen with all other fragments as well (everything that displays bottom sheets), and we never got any complaints about this from users, likely because one can't double touch as fast as one can double click with a mouse.
My suggestion would be removing this logic from the adapter in any case, and handling it in the `FragmentPresenter` (IOW, in `WidgetListFragment`) if needed at some point (that is, when it turns out to be a problem in practice).

mherwege:
> Was the intention to select the whole text when focussing it? If so, it didn't work for me -> [maniac103@dce155b](https://github.com/maniac103/openhab.android/commit/dce155bd9442fa9f5a8563a8ed2ae595544e7cfd)

My intention was to put the cursor at the end of the input field, instead of at the beginning. I think that works. I often end up only changing part of the existing field (meter readings go up, so need to change last values). Soft keyboards have a backspace, but most often no arrow keys. And I always find touch selection more difficult. Either way, I am undecided if selecting all would be the better option. What would the most used paradigm be?

mherwege:
> In my mind, it shouldn't look like an input field, because it does not actually behave like an input field, no matter whether the sitemap definition says 'Input' or not. It very much behaves like a 'Selection' widget (opens editor fragment on click), so (IMHO) should look similar to that one. Doing so also allows getting rid of the back-and-forth when converting the selected date/time to string.
> For your consideration/test, I've implemented it: [maniac103@1825cd0](https://github.com/maniac103/openhab.android/commit/1825cd017778c37e5eb356a4ccd9af69c9446daf)

I didn’t have the ability to test anything yet. My first observation is the code currently tries to even allow date/time input on a text field without inputHint. It needs to interpret the string for that (or the user would have to enter an ISO 8601 date). That interpretation is not perfect, but works in most cases (if it fails, value would be returned to previous). Splitting in a separate widget and removing the datestring interpretation code removes that behaviour. I don’t have a problem with that, but than we should force a default inputHint for datetime items, probably default to datetime. I will look into creating that behaviour.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mherwege: I will remove this.

maniac103:
> but than we should force a default inputHint for datetime items, probably default to datetime

I think that makes sense ... I somewhat doubt many people will want to edit date/time values by entering a string.

> I am undecided if selecting all would be the better option. What would the most used paradigm be?

Not sure. I probably depends on expectations ... whether the user is more likely to want to replace the value completely or to just want to edit parts of it. I really don't know which one is more likely.

mherwege:
@maniac103 I adopted most all of the changes you proposed, except the focus all. It think it is easier with touch to select all than to unselect all and position the cursor at the end.
I also default inputHint to datetime for any input widget with an item of type datetime, effectively forcing datetime input. That also allowed me to further remove code from the input widget.

Another point I want your opinion about regarding datetime input: Current code will have the following behaviour:
- If inputHint = date: allow to set date and set time to midnight.
- If inputHInt = time: allow to set time and set date to eppoch 0.
- If inputHInt = datetime: set both date and time.
An alternative could be:
- If inputHint = date: set date and keep current value of time, default to midnight of no time set.
- If inputHint = time: set time and keep current date, default to midnight if no date set.
- If inputHint = datetime: set both date and time.

I see advantages in the second approach, whereby it would be possible to have 2 separate widgets, one for date and one for time, with the possibility to choose what to change without having to go through 2 pickers. I am not sure what the best approach is. What is your opinion?

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -214,13 +214,15 @@ fun String?.toWidgetType(): Widget.Type {
...
-fun String?.toInputHint(): Widget.InputTypeHint {
+fun String?.toInputHint(item: Item?): Widget.InputTypeHint {
     if (this != null) {
         try {
             return Widget.InputTypeHint.valueOf(this.toString().lowercase().replaceFirstChar { c -> c.uppercase() })
         } catch (e: IllegalArgumentException) {
             // fall through
         }
+    } else {
+        if (item?.isOfTypeOrGroupType(Item.Type.DateTime) == true) return Widget.InputTypeHint.Datetime
```
maniac103, line 225: I would rather do this when consuming the input hint. IMO conversion methods shouldn't introduce policy.

maniac103:
> I see advantages in the second approach, whereby it would be possible to have 2 separate widgets, one for date and one for time, with the possibility to choose what to change without having to go through 2 pickers. I am not sure what the best approach is. What is your opinion?

I also think the second approach is favorable, for the very same reasons :+1:

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
mherwege: @maniac103 I have changed it, please check my implementation is in line with what you had in mind here.

mherwege:
> > I see advantages in the second approach, whereby it would be possible to have 2 separate widgets, one for date and one for time, with the possibility to choose what to change without having to go through 2 pickers. I am not sure what the best approach is. What is your opinion?
> 
> I also think the second approach is favorable, for the very same reasons 👍

Implemented this approach.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
maniac103: It roughly is what I had in mind ;-) I'd do this: https://github.com/maniac103/openhab.android/commit/ac93cd67d3cd7e915b2879d4818f7a928a224588

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -207,7 +227,13 @@ class WidgetAdapter(
...
     override fun onBindViewHolder(holder: ViewHolder, position: Int) {
         val wasStarted = holder.stop()
-        holder.vhc = ViewHolderContext(connection, bottomSheetPresenter, colorMapper, serverFlags, chartTheme)
+        holder.vhc = ViewHolderContext(
+            connection,
+            fragmentPresenter,
+            colorMapper,
+            serverFlags,
+            chartTheme
+        )
```
maniac103, line 236: Can revert the reformatting now, I guess?

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -561,6 +588,177 @@ class WidgetAdapter(
...
+
+        private var updateJob: Job? = null
+        private var oldValue: String? = null
+
+        init {
+            inputText.doAfterTextChanged { if (!isBinding) hasChanged = true }
+            inputText.setOnFocusChangeListener { view, hasFocus ->
+                if (!hasFocus) {
+                    hideKeyboard(view)
+                    if (hasChanged) updateValue()
```
maniac103, line 609: Didn't we want to require explicit confirmation (via IME ACTION_DONE)? If so, I'd rather revert to the previous value here.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -561,6 +588,177 @@ class WidgetAdapter(
...
+                }
+            }
+            inputText.setOnEditorActionListener { view, action, _ ->
+                if (action == EditorInfo.IME_ACTION_DONE) {
+                    hideKeyboard(view)
+                    if (hasChanged) updateValue()
```
maniac103, line 615: Nit: Could you please add braces for if and else blocks? I know the Kotlin code style doesn't strictly require them, but (at least) I prefer using them for the additional clarity added by new line + indention.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -561,6 +588,177 @@ class WidgetAdapter(
...
+                .datePicker()
+                .setSelection(date.atOffset(ZoneOffset.UTC).toInstant().toEpochMilli())
+                .build()
+            datePicker.addOnPositiveButtonClickListener {
+                val newDate = LocalDateTime
+                    .ofInstant(Instant.ofEpochMilli(datePicker.selection ?: 0), ZoneOffset.UTC)
+                    .withHour(date.hour)
+                    .withMinute(date.minute)
+                if (showTime) showTimePicker(widget, newDate)
+                else sendUpdate(widget, newDate)
```
maniac103, line 738: Please add braces here as well.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mherwege: done

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mherwege: I intended it also to just work on a focus change, but I will change.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mherwege: done

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
mherwege: done

mobile/src/main/java/org/openhab/habdroid/model/ParsedState.kt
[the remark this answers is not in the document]
mherwege: Yes, it can be, and I just pushed a commit with that.

.idea/compiler.xml
mueller-ma, line 1: Can you restore this file?

mueller-ma (changes requested):
UoM isn't visible in all cases (compared to Basic UI):

grafik

grafik


grafik


grafik

Maybe add the UoM in that case to a read-only text view next to the edittext?

mherwege:
> UoM isn't visible in all cases (compared to Basic UI):

Did you test with a OH snapshot, or with the outstanding BasicUI inputHint PR https://github.com/openhab/openhab-webui/pull/1787 included? The snapshot does not consider the inputHint yet.

When using a number inputHint the unit should be in the suffix. If it is not there, I need to investigate. I doubt putting it in a separate texfield would make a difference. It would then also be more difficult to put it nicely aligned inside the input box, which I find visually easier on the eye than having it after the input box.

.idea/compiler.xml
mherwege: Done

mueller-ma:
I'm using a snapshot and a `Number` item. Using `Number:Dimensionless` displays the `%` in both input fields:

grafik

I missed the suffix part, so you can ignore my suggestion with the text field.

mherwege:
> I'm using a snapshot and a `Number` item. Using `Number:Dimensionless` displays the `%` in both input fields:
> 
> grafik
> 
> I missed the suffix part, so you can ignore my suggestion with the text field.

The issue with a ‘%d %%’ format on a pure Number type is that the % in this case is not a unit, but just a visualisation. It gets stripped when using a number inputHint, but shown with default or text inputHint. This is correct, as you may even have a format string like ‘This is a number %d %%’. This would show the full text when inputHint=text with the number, but can only receive a numeric value as valid input. For the number inputHint, it would only show the number. I believe there is some user configuration responsability here to provide a relevant state description for input. the same problem would appear with string inputs. The problem is that the state descriptions were meant for visualisation, not to set a valid format for input. This whole contraint could be documented better.
Thinking more about it, but needs to be tested if feasible, put the wider format string beyond the format of the pure format of the number (or string) in prefix and suffix, might be something to consider. That would require analyzing the format string in the code, something I have refrained from so far. If we do this, could we do this in a follow up PR. My PR for BasicUI does not try this either (and I am even less convinced it is easy to do there, as we don’t have the unparsed format string available).

maniac103:
@mherwege  I have one cleanup commit for your consideration/review: https://github.com/maniac103/openhab.android/commit/0f2445cfa9300c9a06dba82027a1f7ce48cdb4b2

Other than that, I think we're good to go as well.

mherwege:
@maniac103 I left a question in your cleanup commit about removing the different representations for date/time when no format is provided and inputHint is provided. The other suggestions make sense.

PR #3370, opened 2023-05-02
Display a question mark on switches with 'null' state

### LINKED ISSUE #3358: Show NULL value of Switch toggles
As described at https://github.com/openhab/openhab-webui/issues/1555 openHAB-Android shall highlight somehow Switch items, if they have NULL or UNDEFINED value.  That is: neither ON nor OFF.

I mean here sitemaps.

mueller-ma:
Closes #3358

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -32,13 +32,13 @@ import android.view.inputmethod.EditorInfo
 import android.webkit.WebChromeClient
 import android.webkit.WebView
 import android.widget.Button
-import android.widget.CompoundButton
 import android.widget.ImageView
 import android.widget.LinearLayout
 import android.widget.TextView
 import androidx.annotation.LayoutRes
 import androidx.annotation.StringRes
 import androidx.annotation.VisibleForTesting
+import androidx.core.content.ContextCompat
 import androidx.core.view.children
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -562,21 +563,28 @@ class WidgetAdapter(
...
-        private val switch: CompoundButton = itemView.findViewById(R.id.toggle)
+        private val switch: MaterialSwitch = itemView.findViewById(R.id.toggle)
         private var isBinding = false
 
         init {
             switch.setOnCheckedChangeListener { _, checked ->
                 if (!isBinding) {
                     connection.httpClient.sendItemCommand(boundWidget?.item, if (checked) "ON" else "OFF")
                 }
+                switch.thumbIconDrawable = null
```
mueller-ma, line 574: This shouldn't be required as `bind()` should be called when the value of an Item changes, right? But for `SwitchViewHolder` `bind()` isn't called when pressing on the switch.

maniac103: The sequence should be
- switch is used
- -> command sent to server
- -> SSE event with new state sent by server
- SSE event causes updateWidget() -> updateWidgetAtPosition() -> notifyItemChanged()
- -> onBindViewHolder() is called, which in turn calls bind()

PR #3384, opened 2023-05-07
Don't add ColorItemActivity and SelectionItemActivity to back stack

### LINKED ISSUE #3369: Don't add ColorItemActivity and SelectionItemActivity to back stack
### Actual behaviour
After opening ColorItemActivity or SelectionItemActivity through the device control tiles they are added to the back stack. Pressing back doesn't close the app, but opens the activity displayed before.

### Expected behaviour
Don't add ColorItemActivity and SelectionItemActivity to back stack

mueller-ma:
Closes #3369

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -107,11 +107,13 @@
             android:label="@string/widget_type_color"
             android:showWhenLocked="true"
             android:exported="false"
+            android:taskAffinity=".ui.ColorItemActivity"
             tools:targetApi="o_mr1" />
         <activity
             android:name=".ui.SelectionItemActivity"
             android:showWhenLocked="true"
             android:exported="false"
+            android:taskAffinity=".ui.SelectionItemActivity"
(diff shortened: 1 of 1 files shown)


mobile/src/main/AndroidManifest.xml
```diff
@@ -107,11 +107,13 @@
             android:label="@string/widget_type_color"
             android:showWhenLocked="true"
             android:exported="false"
+            android:taskAffinity=".ui.ColorItemActivity"
             tools:targetApi="o_mr1" />
         <activity
             android:name=".ui.SelectionItemActivity"
             android:showWhenLocked="true"
             android:exported="false"
+            android:taskAffinity=".ui.SelectionItemActivity"
```
maniac103, line 116: In Octodroid I used `android:excludeFromRecents="true" android:taskAffinity=""` for a similar case (URL handler activity). I think we definitely should add the former here as well, and I also think 'don't group into any other activity' is a bit more correct than grouping into itself, because I think the latter becomes similar to the `singleTop` or `singleTask` launch modes.

Issue #3404, opened 2023-05-18
Dynamic icons causing way too many requests


TheTrueRandom:
### Actual behaviour
Each time a value is updated (with a new value which is not cached) a request is sent to openhab to load the icon. This causes incredible many request in some cases:
I am using a sitemap to show power information, this includes: grid power, PV power, battery power, power of devices, etc. The values are updated 1-2 times per second, resulting in about 10 updates per second distributed on those items. This looks incredible nice when it is actually updating. However on the app I can often see icons are flickering and also item updates are stuck until manual refresh. This github issue is for the icons, for the items not updating I still need to analyze why this is happening. In the android studio app inspection I can see that incredible many requests are done to resolve the same icon (poweroutlet, which I am using for all those items).

Generally I have the feeling the app is not really optimized when values are updated multiple times a second (as the UI gets stuck quite often), which is a pity as this looks incredible nice when it actually works.

![Screenshot_2023-05-18_23-57-30](https://github.com/openhab/openhab-android/assets/43322965/d8cd6472-09c6-4173-8eac-b4006c8e559f)

### Expected behaviour
Icons should be cached more aggressively, I'm not sure how an optimal solution could look like but maybe considering something like this
- poweroutlet icon is not even dynamic in openhab, why is it requested every time?
- I'm not sure how dynamic icons work, but I'm quite sure issuing one request every time something changes is very inefficient. Maybe all icon states could be retrieved at once in the beginning. And this could even be persisted in the app and openhab only asked/subscribed for changes to the icons
- on the screenshot you can see the items are updated with many digits, even though on the UI no digits are rendered (as specified in the openhab item file), this seems to be ignored for the image cache. But even if it would not be ignored, there are incredible many different values as seen on the screenshot

 
### Steps to reproduce
1. create an item of type number 
2. define a sitemap with the item and poweroutlet icon
3. create a script to update the item multiple times a second, slightly changing the values (like on the screenshot)
4. notice for every change a request is sent to openhab to fetch the icon

(app is connected to local url)

#### Client
* Android version: 13.0
* Device model: Pixel XL
* App version 3.6.1
* Build flavor play store and from github on an emulator.
* Device language: english

#### Server
* Server version: openhab 3.4.4 release build
* Reverse Proxy: 
* Authentication method: None

maniac103:
> I'm not sure how dynamic icons work, but I'm quite sure issuing one request every time something changes is very inefficient.

[...]

PR #3456, opened 2023-08-14
Improve state generation logic when fetching icons

maniac103:
Add support for https://github.com/openhab/openhab-core/pull/3735, format numeric state according to pattern (as mentioned [here](https://github.com/openhab/openhab-android/issues/3404#issuecomment-1554071638)), and make logic a bit easier to understand.


maniac103:
Draft because I didn't test it yet.

maniac103:
Gave it a try on my home sitemap, looks OK as far as I can tell.

Fejitatete:
Thank you for your work

maniac103:
@mueller-ma OK. I guess/hope we'll get feedback from next beta if anything is off.

lolodomo:
@maniac103 @mueller-ma : regarding the state passed to the icon servlet, do you know if we are passing always the same state value in the Android app and Basic UI ? I hope we are consistent.

maniac103:
I know that I don't know ;-) I would hope so, but I don't have a good idea on how to find out or even enforce it.

PR #3471, opened 2023-09-11
Fix applying first map location

### LINKED ISSUE #3412: Map shows location only after a refresh
### Actual behaviour
The app shows the location of a `Location` item only after a refresh:
1. Open the app
2. Access a site with a `Location` item
3. --> Shows a map of the whole earth
4. Refresh (Slide down or go back and then back into the site)
5. Correct location is shown

### Expected behaviour
Show the location on first sight

 
### Steps to reproduce
See above
See [this](https://drive.google.com/file/d/1eRivv0viszKc7WwwmGNl-5n_fJOelvi4/view) screenshare to further demonstrate the issue

#### Client
* Android version: 3.4.0
* Device model: Samsung Galaxy S22 Ultra
* App version : 3.4.0
* Build flavor : Full
* Device language: DE

#### Server
* Server version: 3.4.3

#### App log
<details>
  <summary>Click to expand</summary>

[Log.txt](https://github.com/openhab/openhab-android/files/11593437/Log.txt)

</details>

maniac103:
If the map isn't loaded yet when the first map widget is bound, the bind attempt for that widget was discarded. Fix this by making sure to apply the widget bind attempt after initially loading the map. While at it, also make sure not to load the map until after the data saver check happened to not cause data transfer before that check.

Fixes #3412

### DIFF
--- a/mobile/src/full/java/org/openhab/habdroid/ui/MapViewHelper.kt
+++ b/mobile/src/full/java/org/openhab/habdroid/ui/MapViewHelper.kt
@@ -98,23 +98,18 @@ object MapViewHelper {
 }
 
 fun GoogleMap.applyPositionAndLabel(widget: Widget, zoomLevel: Float, allowDrag: Boolean) {
-    if (widget.item == null) {
-        return
-    }
-    val canDragMarker = allowDrag && !widget.item.readOnly
-    if (widget.item.members.isNotEmpty()) {
-        val positions = ArrayList<LatLng>()
-        for (member in widget.item.members) {
-            val position = member.state?.asLocation?.toLatLng()
-            if (position != null) {
-                setMarker(position, member, member.label, canDragMarker)
-                positions.add(position)
-            }
-        }
-        if (positions.isNotEmpty()) {
+    val item = widget.item ?: return
+    val canDragMarker = allowDrag && !item.readOnly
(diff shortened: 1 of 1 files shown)


mueller-ma:
The maps key is here https://github.com/openhab/openhab-android/blob/main/.github/workflows/build.yml#L38 but only usable in apps signed by the release key. Some years ago I was able to generate an API key for development in my personal Google account (not a paid account). Did you check if that's still possible?

maniac103:
I think that was the case for me too, but it looks like this isn't possible anymore. At least [the SDK guide] (https://developers.google.com/maps/documentation/android-sdk/get-api-key?hl=en) mentions billing to be a prerequisite.

maniac103:
@mueller-ma While I couldn't test it myself, I'm fairly certain this will fix #3412. My suggestion would be putting this into a beta for an actual live test now, and at some point in the future think about improving the debugging setup, e.g. by sharing a signing certificate we then use when debugging the app, and adding that signing cert to the Google account?

mueller-ma:
Let's give it a try in the beta version: https://github.com/openhab/openhab-android/releases/tag/3.7.5-beta

felixschndr:
I can confirm, it works as intended 👍🏻

PR #3474, opened 2023-09-12
Don't use device locale when formatting state for icons

### LINKED ISSUE #3472: Dynamic icons for numeric values do only work with integers in beta version
### Actual and expected behavior
In version 3.7.3-beta (my be earlier) I recognized, that my temperature icons are not more _dynamic_ and only the default icon is displayed. In version 3.7.0 it is OK. In version 3.7.4-beta the problem persists. The Basic UI via browser is also OK.
As the temperatures are numeric values, I had a look to my switch values and they work as they should. Also the build-in icons are looking OK. In my case humidity.

This is from version 3.7.3-beta:
image

And so it looks in version 3.7.0 (green because it is a nice outside temperature):
image

Item definition:
```
Number Aussentemperatur "Temperatur draußen [%.1f °C]" {channel="mqtt:topic:mosquitto:Balkonlicht:Temperatur"}
```
Sitemap entry:
```
Text item=Aussentemperatur label="Außentemperatur [%.1f °C]" icon="temperature_outdoor"
```
Icons:
image

 
### Steps to reproduce
1. Create own icons
2. Use the icons on a sitemap numeric item

I think, it is not testable in demo mode

#### Client
* Android version: 9 and 13
* Device language: german

#### Server
* Server version: 4.0.2 (Just updated to 4.1.0 M1; just the same)

#### openHAB Cloud
* myopenhab.org

maniac103:
The server expects to be able to parse the passed state as number, which is not possible in e.g. German locale: in that case, e.g. '20,0 °C' instead of the expected '20.0 °C' was passed.

Fixes #3472

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
@@ -142,7 +142,7 @@ internal fun String?.toOH2WidgetIconResource(
         stateToUse == null -> "null"
         // Number items need to use state formatted as per their state description
         item.isOfTypeOrGroupType(Item.Type.Number) || item.isOfTypeOrGroupType(Item.Type.NumberWithDimension)-> {
-            stateToUse.asNumber.toString()
+            stateToUse.asNumber?.toString(Locale.US)
         }
         item.isOfTypeOrGroupType(Item.Type.Color) -> when {
             // Color sliders just use the brightness part of the color


mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
```diff
@@ -142,7 +142,7 @@ internal fun String?.toOH2WidgetIconResource(
         stateToUse == null -> "null"
         // Number items need to use state formatted as per their state description
         item.isOfTypeOrGroupType(Item.Type.Number) || item.isOfTypeOrGroupType(Item.Type.NumberWithDimension)-> {
-            stateToUse.asNumber.toString()
+            stateToUse.asNumber?.toString(Locale.US)
```
mueller-ma, line 145: What about `Locale.ROOT`? I haven't checked the docs about "US vs. ROOT".

maniac103: The docs suggest `Locale.US`. See [here](https://developer.android.com/reference/java/util/Locale), section 'be wary of the default locale'.

PR #3476, opened 2023-09-15
Fix status bar being invisible in chart and image activities

maniac103:
Do enableDrawingBehindStatusBar() unconditionally, as otherwise our app bar will be laid out on top of the status bar.


mueller-ma:
I cannot reproduce this issue, but feel free to merge if it fixes it for you.

maniac103:
Maybe I explained the issue poorly. Try opening a ChartActivity and check the status bar contents: are those visible for you? I used to only get a background, but no contents (clock, icons etc.) prior to this patch.

mueller-ma:
> Try opening a ChartActivity and check the status bar contents: are those visible for you? 

Yes, it can see it correctly on [3.7.4-beta](https://github.com/openhab/openhab-android/releases/tag/3.7.4-beta).

maniac103:
Just tried with exactly that version and got this:

Screenshot_20230917-153000_openHAB Beta

... so it definitely reproduces the issue. This sounds like an Android version thing then. What version does your device have? Android 13 here.

maniac103:
Double checked the change doesn't break anything in Android 9 ... I think this should be safe to merge.

SteWers:
The last beta works in Android 9 and 13.

PR #3503, opened 2023-10-19
Fix parsing of SSE topic for groups

mueller-ma:
There can be an optional group in the topic path.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
@@ -97,7 +97,13 @@ class ItemsControlsProviderService : ControlsProviderService() {
                     }
                     val topic = event.getString("topic")
                     val topicPath = topic.split('/')
-                    if (topicPath.size != 4) {
+                    /* Possible formats:
+                    * - openhab/items/<item>/statechanged
+                    * - openhab/items/<group item>/<item>/statechanged
+                    * When an update for a group is sent, there's also one for the individual item.
+                    * Therefore always take the element on index two.
+                    */
+                    if (topicPath.size !in intArrayOf(4, 5)) {
                         throw JSONException("Unexpected topic path $topic")
                     }
                     val item = allItems[topicPath[2]]


mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -96,15 +96,14 @@ class ItemsControlsProviderService : ControlsProviderService() {
...
-                    }
-                    val item = allItems[topicPath[2]]
-                    if (item != null) {
-                        val payload = JSONObject(event.getString("payload"))
-                        val newItem = item.copy(state = payload.getString("value").toParsedState())
-                        factory.maybeCreateControl(newItem)?.let { control -> send(control) }
+                    val items = getItemsFromTopic(topic)
+                    items.forEach {
+                        val item = allItems[it]
+                        if (item != null) {
```
maniac103, line 102: Untested:
`items.map { i -> allItems[i] }.filterNotNull.forEach { item ->`

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -115,6 +114,22 @@ class ItemsControlsProviderService : ControlsProviderService() {
         }
     }
 
+    /**
+     * Get items from topic string.
+     * Possible formats:
+     * openhab/items/<item>/statechanged
+     * openhab/items/<group item>/<item>/statechanged
```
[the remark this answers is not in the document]
mueller-ma: You're right, that's wrong. There's actually also one event for the item itself, so no need to update it again with the wrong group state.

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -97,7 +97,13 @@ class ItemsControlsProviderService : ControlsProviderService() {
...
                     val topic = event.getString("topic")
                     val topicPath = topic.split('/')
-                    if (topicPath.size != 4) {
+                    /* Possible formats:
+                    * - openhab/items/<item>/statechanged
+                    * - openhab/items/<group item>/<item>/statechanged
+                    * When an update for a group is sent, there's also one for the individual item.
+                    * Therefore always take the element on index two.
+                    */
+                    if (topicPath.size !in intArrayOf(4, 5)) {
```
maniac103, line 106: `4..5` instead of `intArrayOf`?

mobile/src/main/java/org/openhab/habdroid/background/ItemsControlsProviderService.kt
```diff
@@ -97,7 +97,13 @@ class ItemsControlsProviderService : ControlsProviderService() {
                     }
                     val topic = event.getString("topic")
                     val topicPath = topic.split('/')
-                    if (topicPath.size != 4) {
+                    /* Possible formats:
+                    * - openhab/items/<item>/statechanged
```
maniac103, line 101: The asterisk alignment bothers my OCD ;-) Maybe use normal comments instead of block comment?

mueller-ma:
I forgot to push the last changes 🙄

Issue #3234, opened 2023-02-13
Delay splash screen


mueller-ma:
### Is your feature request related to a problem? Please describe.
It's not really a problem. When cold-starting the app the progress indicator is shown for a very short time until it's replaced by the loaded Sitemap.

### Describe the solution you'd like
Keep the splash screen on the screen until sitemaps (or other UIs) are loaded: https://developer.android.com/develop/ui/views/launch/splash-screen?hl=en#suspend-drawing

Google recommend to show the splash screen one second at maximum, so the splash screen should be discarded after one second even if the UI isn't loaded yet.

### Additional context
There's a compat class for that: https://developer.android.com/reference/kotlin/androidx/core/splashscreen/SplashScreen#setKeepOnScreenCondition(androidx.core.splashscreen.SplashScreen.KeepOnScreenCondition)

PR #3512, opened 2023-10-24
added some library to overcome splash screen delay and kotlin null pointer exception.

KUSHAL-RAJ-PAREEK:
code change in main activity ans gradle app level.
solving issue no  #3234
thank you.

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -181,6 +181,9 @@ dependencies {
     implementation "com.mikepenz:aboutlibraries-core:$about_libraries_version"
     implementation "com.mikepenz:aboutlibraries:$about_libraries_version"
 
+    // splash screeen deplay improvment
+    implementation "androidx.core:core-splashscreen:1.0.1"
+    
     // Firebase
     implementation platform("com.google.firebase:firebase-bom:25.11.0")
     fullImplementation "com.google.firebase:firebase-messaging-ktx"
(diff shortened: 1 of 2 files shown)


mobile/build.gradle
```diff
@@ -181,6 +181,9 @@ dependencies {
     implementation "com.mikepenz:aboutlibraries-core:$about_libraries_version"
     implementation "com.mikepenz:aboutlibraries:$about_libraries_version"
 
+    // splash screeen deplay improvment
+    implementation "androidx.core:core-splashscreen:1.0.1"
```
mueller-ma, line 185: How does adding the library without using it solve the issue?

mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -135,6 +135,7 @@ import org.openhab.habdroid.util.isEventListenerEnabled
 import org.openhab.habdroid.util.isScreenTimerDisabled
 import org.openhab.habdroid.util.openInAppStore
 import org.openhab.habdroid.util.parcelable
+import kotlinx.android.synthetic.main.activity_add_sales.*
```
mueller-ma, line 138: This import seems unused.

mueller-ma (changes requested):

PR #3520, opened 2023-11-12
Implement icon in button mappings

### LINKED ISSUE #3493: Implement icon in button mappings
https://github.com/openhab/openhab-core/commit/f7d176570cabb081ce38cb93be4f667d9da52a99

mueller-ma:
Closes #3493

First switch has two mappings, second one 4 (with one invalid icon name) and the last two mappings with text only:

grafik

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
@@ -130,10 +130,11 @@ fun MaterialButton.setTextAndIcon(connection: Connection, mapping: LabeledValue)
         text = mapping.label
         return
     }
+    val iconSize = context.resources.getDimensionPixelSize(R.dimen.section_switch_icon)
     CoroutineScope(Dispatchers.IO + Job()).launch {
         val drawable = try {
             connection.httpClient.get(iconUrl, caching = HttpClient.CachingMode.DEFAULT)
-                .asBitmap(100, 0, ImageConversionPolicy.ForceTargetSize).response
+                .asBitmap(iconSize, 0, ImageConversionPolicy.ForceTargetSize).response
                 .toDrawable(resources)
         } catch (e: HttpClient.HttpException) {
             Log.d(WidgetAdapter.TAG, "Error getting icon for button", e)
--- a/mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
@@ -16,6 +16,7 @@ package org.openhab.habdroid.ui
 import android.annotation.SuppressLint
 import android.content.Context
 import android.os.Build
+import android.util.Log
 import android.view.View
 import android.view.inputmethod.InputMethodManager
 import android.webkit.WebSettings.MIXED_CONTENT_COMPATIBILITY_MODE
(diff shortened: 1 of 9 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
```diff
@@ -111,3 +122,26 @@ fun RemoteViews.duplicate(): RemoteViews {
...
+            connection.httpClient.get(iconUrl, caching = HttpClient.CachingMode.DEFAULT)
+                .asBitmap(100, 0, ImageConversionPolicy.ForceTargetSize).response
+                .toDrawable(resources)
+        } catch (e: HttpClient.HttpException) {
+            Log.d(WidgetAdapter.TAG, "Error getting icon for button", e)
+            null
+        }
+        withContext(Dispatchers.Main) {
+            icon = drawable
+            text = if (drawable == null) mapping.label else null
```
maniac103, line 144: Why hide the label if an icon is given, instead of showing both?

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
```diff
@@ -111,3 +122,26 @@ fun RemoteViews.duplicate(): RemoteViews {
...
+    val iconUrl = mapping.icon?.toUrl(context, true)
+    if (iconUrl == null) {
+        icon = null
+        text = mapping.label
+        return
+    }
+    CoroutineScope(Dispatchers.IO + Job()).launch {
+        val drawable = try {
+            connection.httpClient.get(iconUrl, caching = HttpClient.CachingMode.DEFAULT)
+                .asBitmap(100, 0, ImageConversionPolicy.ForceTargetSize).response
```
maniac103, line 136: Probably should use a `dimen` resource for the bitmap size (16dp?)

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
```diff
@@ -111,3 +122,26 @@ fun RemoteViews.duplicate(): RemoteViews {
...
 }
+
+fun MaterialButton.setTextAndIcon(connection: Connection, mapping: LabeledValue) {
+    val iconUrl = mapping.icon?.toUrl(context, true)
+    if (iconUrl == null) {
+        icon = null
+        text = mapping.label
+        return
+    }
+    CoroutineScope(Dispatchers.IO + Job()).launch {
```
maniac103, line 133: Assign an intermediate drawable (loading indicator) here?

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
mueller-ma: I copied the behavior of Basic UI. The label should be used as fallback only. In our case it could be added as content description of the icon.

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
mueller-ma: Sure. The `100` was just added for faster development and I forgot to replace it...

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
mueller-ma: Not sure. We don't do that for the icon of a widget/item as well and icons are/should be small, so fast to load and they are cached.

maniac103 (approved):

PR #3528, opened 2023-11-25
Only show chart/image/video label if specified in sitemap

maniac103:
This uses the new labelSource JSON response field introduced in openhab-core#3804. If the element isn't present (for older server versions), we'll hide the label by default.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
@@ -38,6 +38,7 @@ data class Widget(
     val id: String,
     val parentId: String?,
     private val rawLabel: String,
+    val labelSource: LabelSource,
     val icon: IconResource?,
     val state: ParsedState?,
     val type: Type,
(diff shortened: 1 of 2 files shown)


maniac103:
Only compile time tested so far.

mueller-ma:
Is this ready to be reviewed?

maniac103:
To be reviewed: definitely. To be merged: should probably give it a try, but I didn't find time yet to do the test (server) setup. I hope to get to it by the end of the week.

maniac103:
Actually gave this a try now, and it seems to work. Using the following sitemap:
```
sitemap page_58d8615235 label="New Sitemap" {
    Chart item=TestItem period=1D
    Chart item=TestItem label="Sitemap label" period=1D
}
```
(where `TestItem` is a `Number` item) yields no label on the first chart and a label (+ icon) on the second chart. For older servers the `labelSource` JSON element yields a `null` value, so the label should be hidden for those.

mueller-ma:
My Android Studio either crashes or says there was an error without showing one when I try to compile this branch. I guess this is some issue with my setup....

PR #3529, opened 2023-11-26
Custom sitemaps for device control

### LINKED ISSUE #3140: Allow for more customization for device controls entries

I find the device controls quite inflexible and inconvenient for most use cases.

- A custom name can't be applied; my items are configured in groups such as:
  >	- Air purifier [Group]
  >	   - Power [Switch]
  >	   - Speed [Dimmer]
  >  - Desk light [Group]
  >	   - Power [Switch]
  >	   - Brightness [Dimmer]
  >	   - ColorTemperature [Dimmer]
  
  So I can't simply add the "Power" item cause I'd have no idea of what's being toggled
- Things can't be grouped; I'd like the control for lights to be swipable (to control the brightness) and tappable (to toggle it); at the current stage I can only add two controls, and tapping the brightness one would turn it off by setting it to zero (losing memory of the previous brightness setting)   

I think it'd be really useful to have a new metadata type similar to that for Google Assistant that allows to:

- Override names and other settings
- Group multiple items together to create controls with complex traits

I'm okay with having creating a few proxy items if necessary to match the controls behavior, and/or with adding additional groups to match the semantics of the controls.

The alternative I considered is going through Google Home, and that's what I'm currently doing, though that's not ideal since I'd rather talk to openHAB directly, for performance and privacy reasons.

mueller-ma:
We already have links to Sitemap subpages (NFC tags), so re-use them for the device control.

Closes #3140

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Item.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Item.kt
@@ -13,7 +13,9 @@
 
 package org.openhab.habdroid.model
 
+import android.net.Uri
 import android.os.Parcelable
+import androidx.core.net.toUri
 import java.util.Locale
 import kotlinx.parcelize.Parcelize
 import org.json.JSONException
@@ -43,6 +45,7 @@ data class Item internal constructor(
     val minimum: Float?,
     val maximum: Float?,
     val step: Float?,
+    val linkToMore: Uri?
 ) : Parcelable {
     val label get() = rawLabel?.split("[", "]")?.getOrNull(0)?.trim()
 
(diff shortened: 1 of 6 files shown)


mobile/src/main/java/org/openhab/habdroid/model/Item.kt
```diff
@@ -338,6 +342,12 @@ fun JSONObject.toItem(): Item {
         emptyList()
     }
 
+    val linkToMore = try {
+        optJSONObject("metadata")?.optJSONObject("link_to_more")?.optStringOrNull("value")?.toUri()
```
mueller-ma, line 346: Not sure if this should be made a well-known metadata or keep it as a custom one.

mueller-ma:
Let's discuss about these links before using them here: https://community.openhab.org/t/create-deep-links-to-openhab/151576

PR #3551, opened 2023-12-26
Fix parsing numbers below 1

mueller-ma:


mobile/src/test/java/org/openhab/habdroid/model/ParsedStateTest.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class ParsedStateTest {
+    @Test
+    fun testParseAsBoolean() {
+        listOf("ON", "100", "1", "0.1").forEach {
```
mueller-ma, line 24: 0.1 is detected as brightness and converted to int -> becomes 0.

This commit only contains the new failing test.

mobile/src/test/java/org/openhab/habdroid/model/ParsedStateTest.kt
```diff
@@ -0,0 +1,69 @@
...
+            assertEquals("${it.key} should be parsed as ${it.value}", expected, actual)
+        }
+    }
+
+    @Test
+    fun testParseAsBrightness() {
+        mapOf(
+            "100" to 100,
+            "0" to 0,
+            "0.1" to 0, // TODO
```
mueller-ma, line 52: Not sure about this one. A light bulb with `0.1` brightness isn't off, so maybe parse everything between 0 and 1 as brightness 1?

PR #3554, opened 2024-01-04
Support f7, material, and iconify icons

### LINKED ISSUE #3339: Support material and iconify as new icon sources
The icon at item level or sitemap widget level can now reference new icon sources (material, framework7, iconify).
As I am almost certain that the Android app is already using mainly the material font, it would be nice if the app could also consider "material" as new icon source.
I will not be surprised that it is only few lines of code to add.

jimtng:
Resolves #3339  

Perform translations to request material and f7 icons from iconify:

material:iconname -> iconify:ic:baseline-iconname
material:set:iconname -> iconify:ic:set-iconname
f7:iconname -> iconify:f7:iconname

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
@@ -70,8 +70,8 @@ data class IconResource internal constructor(
         when (iconSource) {
             "material" -> {
                 iconSource = "iconify"
-                iconSet = "ic"
                 iconName = "$iconSet-$iconName"
+                iconSet = "ic"
             }
             "f7" -> {
                 iconSource = "iconify"
--- a/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
@@ -23,6 +23,7 @@ import java.util.Locale
 import kotlinx.parcelize.Parcelize
 import org.json.JSONException
 import org.json.JSONObject
+import org.openhab.habdroid.R
 import org.openhab.habdroid.util.appendQueryParameter
 import org.openhab.habdroid.util.getIconFormat
 import org.openhab.habdroid.util.getPrefs
@@ -35,11 +36,12 @@ data class IconResource internal constructor(
     internal val customState: String
 ) : Parcelable {
     fun toUrl(context: Context, includeState: Boolean): String {
-        return toUrl(includeState, context.getPrefs().getIconFormat())
+        val iconSize = context.resources.getDimensionPixelSize(R.dimen.widgetlist_icon_size)
+        return toUrl(includeState, context.getPrefs().getIconFormat(), iconSize)
     }
 
     @VisibleForTesting
-    fun toUrl(includeState: Boolean, iconFormat: IconFormat): String {
+    fun toUrl(includeState: Boolean, iconFormat: IconFormat, desiredSizePixels: Int): String {
         if (!isOh2) {
             return "images/$icon.png"
         }
@@ -88,8 +90,8 @@ data class IconResource internal constructor(
                 builder.scheme("https")
                        .authority("api.iconify.design")
                        .path(iconSet)
-                       .appendPath(iconName + ".svg")
-                       .appendQueryParameter("height", "64")
+                       .appendPath("$iconName.svg")
+                       .appendQueryParameter("height", desiredSizePixels.toString())
             }
             else -> builder.path("icon/")
         }
--- a/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
@@ -56,11 +56,26 @@ data class IconResource internal constructor(
             2 -> {
                 iconSource = segments[0]
                 iconName = segments[1]
+                if (iconSource == "material") {
+                    iconSet = "baseline"
+                }
             }
             3 -> {
                 iconSource = segments[0]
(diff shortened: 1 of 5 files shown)


mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
```diff
@@ -44,41 +44,54 @@ data class IconResource internal constructor(
...
+                }
+            }
+            "if", "iconify", "material" -> {
+                if (iconSource == "material") {
+                    iconSet = "mdi"
+                }
+                builder.scheme("https")
+                       .authority("api.iconify.design")
+                       .path(iconSet)
+                       .appendPath(iconName + ".svg")
```
maniac103, line 91:
```suggestion
                       .appendPath("$iconName.svg")
```

mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
```diff
@@ -44,41 +44,54 @@ data class IconResource internal constructor(
...
+                       .appendQueryParameter("height", "64")
```
maniac103, line 92: 64 pixels are probably not enough to look sharp with todays display densities. We need to fetch in `dp`. Suggestion for that would be:
- making the signature of this method `fun toUrl(includeState: Boolean, iconFormat: IconFormat, desiredSizePixels: Int)`
- make the first method this:
```
    fun toUrl(context: Context, includeState: Boolean): String {
        val iconSize = context.resources.getDimensionPixelSize(R.dimen.widgetlist_icon_size)
        return toUrl(includeState, context.getPrefs().getIconFormat(), iconSize)
    }
```

mobile/src/main/java/org/openhab/habdroid/util/HttpClient.kt
```diff
@@ -54,10 +54,11 @@ class HttpClient constructor(client: OkHttpClient, baseUrl: String?, username: S
...
             clientBuilder.addNetworkInterceptor { chain ->
-                val request = chain.request()
-                    .newBuilder()
-                    .addHeader("Authorization", authHeader)
-                    .build()
+                var request = chain.request()
+                // Don't add auth header to requests to external icons
+                if (baseUrl != null && request.url.toString().startsWith(baseUrl)) {
+                    request = request.newBuilder().addHeader("Authorization", authHeader).build()
+                }
```
maniac103, line 61: I find working with immutable values preferable. Thus my suggestion would be something like
```
            clientBuilder.addNetworkInterceptor { chain ->
                val request = chain.request()
                // Make sure not to send authentication credentials to external servers
                chain.proceed(if (this.baseUrl?.host == request.url.host) {
                    request.newBuilder().addHeader("Authorization", authHeader).build()
                } else {
                    request
                })
            }
```

mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
jimtng: Thanks, this is my first time using Kotlin

mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
jimtng: Updated

mobile/src/main/java/org/openhab/habdroid/util/HttpClient.kt
jimtng: Nice! Thanks for the suggestion

mueller-ma:
FYI: I requested the f7 icon set at iconify, so we can support them as well quite easy. AFAIK the official clients for f7 icons are all web-based. https://github.com/iconify/icon-sets/issues/104

jimtng:
I'll add f7 support.

The current material -> mdi isn't right. I'll change that too.

mueller-ma:
I'll review this when f7 is live on iconify.

jimtng:
iconify f7 is live now

mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
```diff
@@ -35,50 +36,76 @@ data class IconResource internal constructor(
...
+                }
+            }
+            "if", "iconify" -> {
+                builder.scheme("https")
+                       .authority("api.iconify.design")
+                       .path(iconSet)
+                       .appendPath("$iconName.svg")
+                       .appendQueryParameter("height", desiredSizePixels.toString())
+            }
+            else -> builder.path("icon/")
```
mueller-ma, line 108: Maybe return the `none` icon of `oh` instead?

jimtng: Done

mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
```diff
@@ -35,50 +36,81 @@ data class IconResource internal constructor(
...
         }
 
-        val suffix = when (iconFormat) {
-            IconFormat.Png -> "PNG"
-            IconFormat.Svg -> "SVG"
+        when (iconSource) {
+            "material" -> {
+                iconSource = "iconify"
+                iconSet = "ic"
+                iconName = "$iconSet-$iconName"
```
mueller-ma, line 74: You need to swap these two lines. Otherwise the `iconName` is always `ic-$iconName`. Maybe you can extend the current tests to check all possible icon source / icon set combinations?

jimtng: oops! fixed and tests added

mueller-ma (approved):
LGTM!

PR #3558, opened 2024-01-09
Don't set app name in notifications

### LINKED ISSUE #3130: App name set as notification title
"openHAB" or "openHAB Beta" is set as notification title: https://github.com/openhab/openhab-android/blob/main/mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt#L219

This causes two issues:
* It's redundant as the app name is already part of the notification layout provided by Android.
* It hides the real message when the notification is minimized. This can be done by muting and minimizing the notification channel.

mueller-ma:
Closes #3130

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
@@ -240,18 +240,15 @@ class NotificationHelper constructor(private val context: Context) {
     private fun makeNotificationBuilder(
         channelId: String,
         timestamp: Long
-    ): NotificationCompat.Builder {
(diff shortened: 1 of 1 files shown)


maniac103:
Doesn't this obscure longer notification texts? I have a rule to send me a notification with a (potentially) multi line list of devices that need battery replacement ... I assume those additional lines won't be shown with this change?

mueller-ma:
Only two lines would were shown. I changed this PR and now the full message is shown. Screenshot with two messages, one with "Test" and one with several "Test":

grafik

and minimized:
grafik

maniac103:
And 'Wecker' in those cases was the severity, I guess? If so, LGTM!

PR #3565, opened 2024-01-13
Only use item state when fetching icons

### LINKED ISSUE #3564: Bug/incompatability with dynamic icon behavior for <battery> and custom icons
I am currently using openhab 4.1.0 and noticed that the following icons do not show the 'dynamic' behavior anymore. The percentage value is **not** shown in the icon. The icon just show the default. This applies for:
* \<battery>
* \<batterylevel>
* \<lqi> --> this is a custom icon I use for linkquality (lqi.svg, lqi-0.svg, lqi-10.svg, etc).

The behavior occurs with item definitions like this:
```
Number ContactSensor1_Battery     "[%d%%]"      <battery> { channel="mqtt:topic:aqaraContactSensor1:battery"}
Number ContactSensor1_LinkQuality "Link [%d%%]" <lqi> { channel="mqtt:topic:aqaraContactSensor1:linkquality"}
```

However, when you add a **space** between the value and the percentage it works fine.
```
Number ContactSensor1_Battery     "[%d %%]"      <battery> { channel="mqtt:topic:aqaraContactSensor1:battery"}
Number ContactSensor1_LinkQuality "Link [%d %%]" <lqi> { channel="mqtt:topic:aqaraContactSensor1:linkquality"}
```

Note that some icons still work as expected without space, for example <light>
```
Dimmer GF_Living_Main_Brightness "Main Level [%d%%]" <light> { channel="<snip>" }
```

I am sure \<battery> and \<lqi> worked fine on previous versions on openhab (but I am not sure when this bug/incompatability was introduced, just noticed now).

## Expected Behavior
Icons should also update when no space is defined between value and percentage symbol: `"[%d%%]"`

## Your Environment
* OpenHab v4.1.0 in a docker container.
* Android openHAB app v3.8.0.

maniac103:
sending icon requests, if available. Apparently this is wrong and we always need to use the item state for fetching icons, so roll back the respective code.

Fixes #3564

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt
@@ -157,7 +157,6 @@ fun String?.toOH2IconResource(): IconResource? {
 }
 
 internal fun String?.toOH2WidgetIconResource(
-    state: ParsedState?,
     item: Item?,
     type: Widget.Type,
     hasMappings: Boolean,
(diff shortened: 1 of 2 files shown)


maniac103:
@lolodomo Does this logic look correct to use (only use item state, never use widget state as formatted by state description), especially in the light of cases like mentioned in #3445 (item state = `0.45`, widget state = `45 %`)?

mueller-ma (approved):
Let's try it in beta.

lolodomo:
> @lolodomo Does this logic look correct

This is the current login adopted by all UIs  Note that the icon servlet needs a deterministic value as current state. The widget pattern was imagined to control the format of the displayed value, but anything else, neither to control the persisted value nor the icon.
Changing and switching to the presented value for the icon, even if I can understand the logic, would probably introduced more difficulties than it solves, this issue seems to be a proof of this.

Issue #3417, opened 2023-06-13
On Try Again spell if local or remote connection is retried


dilyanpalauzov:
The screenshot says: The connection with the host was not successful. [Try again]

In openhab-Android I have a local and a remote connection.  When I am in the home network, I cannot connect to the remote connection.  This is not an openHab  problem, but a matter of routing fact.

When in this state, and I click on [Try again] openhab-Android tries to reconnect, but does not spell, whether it tries to reconnect fist to the local and then to the remote connection.  My colclusion is, that it tries to connect only to the remote connection.

* the programme shall try to reconnect to both local and remote connections
* it shall display to the user which connection it is currently trying to utilize.

![Screenshot_20230612_113519_org openhab habdroid beta](https://github.com/openhab/openhab-android/assets/4992947/de2f2968-953f-49af-a31e-16f186e73803)

dilyanpalauzov:
> > it shall display to the user which connection it is currently trying to utilize.
> 
> At the time of pressing 'retry', it doesn't even know that yet (see above).

Once the programme knows, it shall show it, using a snackbar or however it was called.

maniac103:
> Once the programme knows, it shall show it, using a snackbar or however it was called.

If you configured both local and remote connections, it already does.

dilyanpalauzov:
Yes, it does show it on start.  But from the above screenshot state it did not show it.

I observed this several times and I will not be soon in that network to retry it again.

dilyanpalauzov:
With openhab-Android 3.7.0 I changed the remote URI to unresolvable name and created random local URI (http://192.168.168.168).  I connect over public Wi-Fi.  The error message is “Cannot connect to the domain”.  (The error message in the screenshot above is: The connection with the host is unsuccessful.  So not exactly the same)

I click on Try again.  The system does not show a hint, whether it tries to connect first to the local and then to the remove address.

Issue #3414, opened 2023-06-05
Skipping local connections, when WiFi name cannot be read


dilyanpalauzov:
I have set up in the Local connections “Restrict access to these Wi-Fi networks”.

I do not understand what is the purpose of the “Restrict” toggle - if there are network names, then openHAB-Android shall use the local connection, only if the WiFi networks have one of the provided names.  Just the presence of any Wi-Fi network name implies restrcting connections to these names.

Apart from this, if the Location service is turned globally off, the applications cannot read the names of the Wi-Fi networks.  In these cases openhab-Android does not try to connect to the local connections and uses the global connection.

I think, when openhab-Android does not know the names of the WiFi networks, then it shall use try the local connection.

mueller-ma:
Entering a SSID makes the app switch the primary server when you connect to a specified Wi-Fi. There might be cases where servers are reachable over multiple SSIDs, e.g. router with 2.4 and 5 GHz with different names or two houses being connected over VPN.

> I think, when openhab-Android does not know the names of the WiFi networks, then it shall use try the local connection.

That could lead to a connection to a wrong (not-openhab) server, especially since local IP addresses aren't unique.

dilyanpalauzov:
My understand of this function is, that it works as speed-up.  If the application knows that in the current WLAN, based on its name, it cannot connect to the local instance, it does not try to the local system, and connects faster to the remove address.

If the application does not know the name of the Wi-Fi network, then the speed up does not apply: it first tryes to connect to the local instance, waits until it fails and the connects to the remote instance.

dilyanpalauzov:
DAVx5 has a setting [WiFi SSID restriction](https://manual.davx5.com/settings.html) described as:

> Use WiFi SSID restriction to restrict synchronization to one or more specific WiFi networks (requires location permission, including background location). List all allowed SSIDs as a comma-separated list, for instance SSID1,ssid2. Hidden networks and SSIDs with commas are not supported.

Recently I was in a public Wlan and started OpenHAB-Android.  It asked me to confirm a certificate.  As it turned out in this network on the ip and port I habe configured for local connection in OoenHAB-Android  another web server was listening in the public wlan.  So the only way to use OpenHAB-Android there was to disable the local connection.

It would be good if OpenHAB-Android has similar functionality to skip local connections, if the wlan name is not within a list.

mueller-ma:
[...]

PR #3570, opened 2024-01-15
Show local/remote connection only in debug mode

mueller-ma:
I don't see any use in these messages except for debugging, so show them only in debug mode.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -696,13 +696,13 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
             val hasLocalAndRemote =
                 ConnectionFactory.hasActiveLocalConnection && ConnectionFactory.hasActiveRemoteConnection
             val type = connection?.connectionType
-            if (hasLocalAndRemote && type == Connection.TYPE_LOCAL) {
+            if (hasLocalAndRemote && type == Connection.TYPE_LOCAL && prefs.isDebugModeEnabled()) {
                 showSnackbar(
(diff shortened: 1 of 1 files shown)


maniac103:
Personally I only have one use case: having a VPN connection active over a flaky mobile network. In that case the local server should be chosen, but due to the flaky mobile network it's sometimes not reachable. Maybe we should retry the connection check  in the background in that case.

mueller-ma:
I don't understand your use case. How does the snackbar influence the connection?

maniac103:
It doesn't, but it tells me the point in time where it's pointless to wait any longer for the content to load. I use myOH only for notifications, so the remote server won't give me contents either, this without the snack bar I won't be able to distinguish 'tries to establish local connection' from 'futile attempt to establish remote connection'.

dilyanpalauzov:
In my environment I have an internal private IP address and public static IP address.  For some reason, probably because the provider wants so and I cannot adjust the router to do otherwise, I cannot access the external IP address from the internal network.  I find it useful to see whether OpenHAB-Android tries to connect to the internal or to the external network, and which one fails.

In fact I asked at https://github.com/openhab/openhab-android/issues/3417 to spell whether local or remote connection is retried after failure.

I think that “Restricting the local connections based on Wi-Fi name” shall be used as speedup - https://github.com/openhab/openhab-android/issues/3414 .  In this case I consider it good, if it is spelled explicitly whether the connection is made to local or remote server.

mueller-ma:
What do you think about showing "Connecting to local/remote server" in the `ProgressFragment`?

dilyanpalauzov:
As far as this question is for me, I do not know what the `ProgressFragment` is.

mueller-ma:
> What do you think about showing "Connecting to local/remote server" in the ProgressFragment?

That's not possible any longer as the skeleton view is used while loading.

PR #3574, opened 2024-01-22
Fix vertical alignment and value spacing of compact-mode switch buttons

jimtng:
Before:
- The value appears to not be vertically centered against the button, especially against the labels inside the buttons)
- The value is too close to the buttons

Screenshot 2024-01-22 at 6 04 50 pm

After:
- Removed the top padding of the buttons, thus bringing them higher and centered against the value to its left
- Added some spacing between the value and the buttons
Screenshot 2024-01-22 at 6 20 34 pm

How it looks (after) with icons:
Screenshot 2024-01-22 at 6 20 12 pm

---

Before:
The value is too close to the on/off button
Screenshot 2024-01-22 at 6 14 44 pm

After:
Added some spacing between the value and the buttons
Screenshot 2024-01-22 at 6 18 26 pm


mueller-ma (approved):
LGTM!

PR #3583, opened 2024-01-25
Save input mode of TimePicker

mueller-ma:


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -767,14 +767,27 @@ class WidgetAdapter(
...
+                prefs.edit {
+                    putInt(PrefKeys.TIME_PICKER_INPUT_MODE, timePicker.inputMode)
+                }
+            }
+
             timePicker.addOnPositiveButtonClickListener {
                 sendUpdate(widget, date.withHour(timePicker.hour).withMinute(timePicker.minute))
+                saveInputType()
             }
+            timePicker.addOnCancelListener { saveInputType() }
```
maniac103, line 789: Cancellation should cause the dismiss listener to be called, thus the cancel listener should not be necessary.

mueller-ma: DismissListener is also called when the "OK" button is pressed. 👍

PR #3623, opened 2024-03-11
Consider mappings or options for displayed state

maniac103:
One use case for this are items bound to channels of type system.signal-strength. Those are number items with a default pattern of %0.f and options mapping the valid values 0..4 to descriptive strings. Show those descriptive strings (e.g. 'Excellent') if available instead of the plain number value ('4').

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
@@ -65,7 +65,11 @@ data class Widget(
     val rawInputHint: InputTypeHint?
 ) : Parcelable {
     val label get() = rawLabel.split("[", "]")[0].trim()
-    val stateFromLabel: String? get() = rawLabel.split("[", "]").getOrNull(1)?.trim()
+    val stateFromLabel: String? get() {
+        val value = rawLabel.split("[", "]").getOrNull(1)?.trim()
+        val optionLabel = mappingsOrItemOptions.find { it.value == value }?.label
+        return optionLabel ?: value
+    }
 
     val mappingsOrItemOptions get() = if (mappings.isEmpty() && item?.options != null) item.options else mappings
 


maniac103:
@lolodomo FYI for consideration to use the same approach in BasicUI. MainUI does exactly this when showing item details. Off-hand I don't see any downsides in doing so. The only problematic case would be a user wanting to display the raw number for some reason, but at the same being unable to remove the item options for whatever reason.

lolodomo:
Not sure to understand what you changed because options are already considered by Basic UI. The code is in core framework, not in Basic UI and we even have a unit test for that:
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.ui/src/test/java/org/openhab/core/ui/internal/items/ItemUIRegistryImplTest.java#L693

In case I do not understand your use case, please provide item and sitemap element so that I can test with Basic UI.

maniac103:
@lolodomo Item definition (sorry for screenshot, but code tab is empty?!):
grafik

This item doesn't have explicit metadata set.

Sitemap: `Text item=LawnMowerSignal`

Sitemap response for that widget:

```json
                                {
                                    "widgetId": "030501",
                                    "type": "Text",
                                    "visibility": true,
                                    "label": "Signalstärke [3]",
                                    "labelSource": "ITEM_LABEL",
                                    "icon": "qualityofservice",
                                    "staticIcon": false,
                                    "pattern": "%.0f",
                                    "unit": "",
                                    "mappings": [

                                    ],
                                    "item": {
                                        "link": "http://192.168.100.10:8080/rest/items/LawnMowerSignal",
                                        "state": "3",
                                        "stateDescription": {
                                            "minimum": 0,
                                            "maximum": 4,
                                            "step": 1,
                                            "pattern": "%.0f",
                                            "readOnly": true,
                                            "options": [
                                                {
                                                    "value": "0",
                                                    "label": "Kein Signal"
                                                },
                                                {
                                                    "value": "1",
                                                    "label": "Schwach"
                                                },
                                                {
                                                    "value": "2",
                                                    "label": "Durchschnittlich"
                                                },
                                                {
                                                    "value": "3",
                                                    "label": "Gut"
                                                },
                                                {
                                                    "value": "4",
                                                    "label": "Ausgezeichnet"
                                                }
                                            ]
                                        },
                                        "commandDescription": {
                                            "commandOptions": [
                                                {
                                                    "command": "0",
                                                    "label": "Kein Signal"
                                                },
                                                {
                                                    "command": "1",
                                                    "label": "Schwach"
                                                },
                                                {
                                                    "command": "2",
                                                    "label": "Durchschnittlich"
                                                },
                                                {
                                                    "command": "3",
                                                    "label": "Gut"
                                                },
                                                {
                                                    "command": "4",
                                                    "label": "Ausgezeichnet"
                                                }
                                            ]
                                        },
                                        "type": "Number",
                                        "name": "LawnMowerSignal",
                                        "label": "Signalstärke",
                                        "category": "QualityOfService",
                                        "tags": [
                                            "Point"
                                        ],
                                        "groupNames": [
                                            "LawnMower"
                                        ]
                                    },
                                    "widgets": [

                                    ]
                                },

```

From the unit test, I guess explicitly setting the state description pattern to `%s` instead of using the default `%.0f` will make it work as well. I wonder where that default comes from, though:
- it's not explicitly set on the item
- the item is linked to the `common#rfLinkLevel` channel of my mower, which is [of type `system:signal-strength`](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.gardena/src/main/resources/OH-INF/thing/thing-types.xml#L296)
- that channel type [does not specify any pattern](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java#L109-L123)

So where does the default come from? IMO we should do something about this either way, either handle that case in the client or not default to an unsuitable pattern.

maniac103:
I guess the `%.0f` state description pattern comes from [here](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelStateDescriptionProvider.java#L109), but I'm not sure yet what the exact call trace between the sitemap response and that method is. But obviously it triggers on an existing state description for number items without pattern, which exactly is the case here.

lolodomo:
Yes, that is the default pattern for a channel expecting a Number item. Note that I propose a change for that in openhab/openhab-core#4175

But I think the problem could be in the system channel type that does not provide any pattern
https://github.com/openhab/openhab-core/blob/c3ada84b77ca1d083164e97417a3441a48c3ba50/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java#L117

lolodomo:
But we can check what is implemented in ItemUIRegistryImpl when there is a matching option but the formatting with the pattern is failing.

lolodomo:
To be clear, you should have implemented nothing on UI side as the problem is the response from the REST API. In your example, the label field should have contained "Signalstärke [Gut]" instead of "Signalstärke [3]".

lolodomo:
The bug is here when catching the exception:
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java#L411

I am preparing a fix including a unit test.

lolodomo:
Fix is now merged in core framework. Your special handling is no more required.

maniac103:
> Your special handling is no more required.

That's not entirely true: it's still required for everybody not running 4.2, which will be a majority of our user base for some time coming.
@mueller-ma What's your opinion here? Revert now, revert later (some time after 4.2 release) or keep our client side handling?

mueller-ma:
IMO it should be fine to revert it when 4.2 has been released and the user that has this issue (you) updated.

PR #3639, opened 2024-04-10
Potential fix for #3637 / race condition sitemap vs events

seime:
See #3637

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.kt
@@ -423,7 +423,7 @@ class PageConnectionHolderFragment : Fragment(), CoroutineScope {
                         callback.onWidgetUpdated(url, updatedWidget)
                         return
                     }
-                } else if(lastWidgetList != null) {
+                } else if (lastWidgetList != null) {
                     // Either we didn't find the widget (possibly because the server didn't give us invisible widgets),
                     // or we couldn't update it because we couldn't trust the data, so reload the page.
                     // If we didn't have a widget list yet, simply ignore the event - we probably got it while loading the list,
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.kt
```diff
@@ -425,10 +425,15 @@ class PageConnectionHolderFragment : Fragment(), CoroutineScope {
                     }
                 }
```
maniac103, line 426: I think I'd rather do this here:

```
} else if (lastWidgetList != null) {
    // Either we didn't find the widget (possibly because the server didn't give us invisible widgets),
    // or we couldn't update it because we couldn't trust the data, so reload the page.
    // If we didn't have a widget list yet, simply ignore the event - we probably got it while loading the list.
    cancel()
    load()
}

Comment (maniac103) [reply mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.kt]: Almost, you just cut the last words of the comment ;-)

Comment (seime) [reply mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.kt]: 😇

File: mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.kt [hunk 2]
```diff
@@ -423,12 +423,14 @@ class PageConnectionHolderFragment : Fragment(), CoroutineScope {
                         callback.onWidgetUpdated(url, updatedWidget)
                         return
                     }
+                } else if(lastWidgetList != null) {
```
maniac103, line 426:
```suggestion
                } else if (lastWidgetList != null) {
```

(code linter will yell about this)

maniac103: @mueller-ma We don't run the linter as part of PR actions?

mueller-ma: No, we used https://github.com/apps/stickler-ci, but it seems that they shut down. I think we lost it when migrating to Kotlin.

I opened https://github.com/openhab/openhab-android/issues/3640

PR #3641, opened 2024-04-14
Support new Slider parameter: releaseOnly

mueller-ma:
See https://github.com/openhab/openhab-core/issues/3430


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -1659,15 +1659,17 @@ fun Widget.shouldUseDateTimePickerForInput(): Boolean {
         inputHint == Widget.InputTypeHint.Datetime
 }
 
-fun Item.shouldUseSliderUpdatesDuringMove(): Boolean {
+fun Widget.shouldUseSliderUpdatesDuringMove(): Boolean {
+    releaseOnly?.let { return !releaseOnly }
+    item ?: return false
```
maniac103, line 1664: Not sure if it's just me, but I think using old-school if statements for those two checks would be more readable.

PR #3644, opened 2024-04-14
Use gradle plugin for ktlint

mueller-ma:
and auto fix format issues.


mueller-ma:
@maniac103 What do you think about the auto fixes? Most of them seems good to me. But there are still tons of issues that cannot be auto fixed, so I'm not sure whether we want to run ktlint in the CI.

maniac103:
The amount of changes suggests we never used ktlint so far ... strange, I would've sworn this to be the case.
Probably the best idea is to work towards being ktlint clean first and enable it afterwards.

maniac103:
I'm working on the ktlint cleanup (auto fixes + manual postprocessing).

PR #3705, opened 2024-06-18
Allow configuration of MainUI start page

### LINKED ISSUE #3683: Add a default path option for Main UI
My home dashboard in Main UI is not located on the Main UI root, but rather on a tabbed page located at the path `/page/myHome`.
When opening the Android app, I always have to manually navigate to that page.

It would be great to have the option to specify a default path for Main UI, which is automatically opened when Main UI is opened inside the app. This functionality is already provided by the iOS app and is really great - I don't have to click me to my dashboard anymore.

Probably modifying Main UI to allow setting custom pages as the Main UI root, but that is a different topic and even though I don't know the Android app codebase and have never developed for Android before, seems like more effort than my suggestion above.

mueller-ma:
Closes  #3683

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,7 @@ captures/
 # Comment next line if keeping position of elements in Navigation Editor is relevant for you
 .idea/navEditor.xml
 .idea/workspace.xml
+.idea/deploymentTargetSelector.xml
 
 # Keystore files
 *.jks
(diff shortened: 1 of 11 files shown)


mueller-ma:
@florian-h05 The setting is just a text field. I'd like to a add a button "use current page as default" somewhere, but that requires #3706.

florian-h05:
Nice, thank you very much!!

PR #3706, opened 2024-06-18
WebViewUI: Open drawer instead of going back to Sitemaps

mueller-ma:
When pressing the 'back to app' button the drawer is opened instead of closing the UI. This way users can switch faster between e.g. MainUI and notifications.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
@@ -467,14 +467,8 @@ abstract class ContentController protected constructor(private val activity: Mai
         return false
     }
 
-    override fun closeFragment() {
-        if (temporaryPage != null) {
-            temporaryPage = null
-            activity.updateTitle()
-            updateActionBarState()
-            updateFragmentState(FragmentUpdateReason.PAGE_UPDATE)
-            updateConnectionState()
-        }
+    override fun openDrawer() {
+        activity.openDrawer()
     }
 
     override fun onPageUpdated(pageUrl: String, pageTitle: String?, widgets: List<Widget>) {
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -1368,7 +1368,11 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
     }
 
     fun setDrawerLocked(locked: Boolean) {
-        drawerLayout.isSwipeDisabled = locked
+        drawerLayout.swipeToOpenDisabled = locked
+    }
+
+    fun openDrawer() {
+        drawerLayout.open()
     }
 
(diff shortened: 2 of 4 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/activity/ContentController.kt
```diff
@@ -467,14 +467,8 @@ abstract class ContentController protected constructor(private val activity: Mai
...
     }
 
-    override fun closeFragment() {
-        if (temporaryPage != null) {
-            temporaryPage = null
-            activity.updateTitle()
-            updateActionBarState()
-            updateFragmentState(FragmentUpdateReason.PAGE_UPDATE)
-            updateConnectionState()
-        }
```
mueller-ma, line 477: This code needs to be restored otherwise the action bar doesn't show up when switching from MainUI to Sitemaps.

maniac103:
I'm not too sure I'm a fan of that, because 'exitToApp' exactly matches the icon/appearance of the button. That means with this proposal we'd have 2 buttons opening 2 different drawers in MainUI, which look completely different.
IMHO, if we want to expand from 'exitToApp', we should integrate some new API into MainUI, maybe even integrated into its drawer menu, and remove exitToApp from newer MainUI versions.

mueller-ma:
Main UI could be changed to display the same right drawer as in the browser, but make the entries configurable via the app.

digitaldan:
The IOS client does something similar to the Main UI, it calls `/rest/ui/tiles` to grab the enabled UI's on a user's openhab, and then displays them in the native side menu in its own section along with the other IOS menu items.

digitaldan:
In any case, it would be nice if the MainUI and SItemaps could both be the primary interface in Android, right now the Main UI UX is clunky with it exiting to a sitemap every time that button is pressed.

maniac103:
> Hmm, not sure i understand, what two buttons ? and what 2 drawers?

Hamburger button on the left -> opens MainUI drawer
Exit button on the right -> opens app drawer

digitaldan:
>Hamburger button on the left -> opens MainUI drawer
Exit button on the right -> opens app drawer

Gotcha, Understood.  Right now this works OK on IOS b/c its native menu comes in on the right side of the app, matching the MainUI button on the right side, could we maybe reuse the existing android side menu, just switch sides when in the main UI ?  I worry about trying to do much magic in mixing the MainUI and NativeUI menu together.

maniac103:
> could we maybe reuse the existing android side menu, just switch sides when in the main UI ?

I really don't think opening the drawer from different sides, depending from where/when it's opened (sitemap vs. Main UI), makes sense UX wise. Right hand side drawers are not really a thing in Android.

> I worry about trying to do much magic in mixing the MainUI and NativeUI menu together.

Well, at some point we'll need to decide whether we want seamless MainUI integration or not. If the answer is yes, there's a price to pay for that, which is some additional complexity...

digitaldan:
I would suggest as a stop gap until we can make a hybrid menu in the MainUI, we at least open the native drawer on the left when the right menu button is clicked.   While not totally ideal, its far superior to the experience right now , and gives us time to work on a longer term solution.

PR #3723, opened 2024-06-29
Support buttongrid's button widgets

### LINKED ISSUE #3682: support button widgets within Buttongrid
https://github.com/openhab/openhab-core/pull/4223

jimtng:
Resolve #3682


jimtng:
@mueller-ma am I on the right track here? Preliminary test shows that it works.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -54,6 +54,10 @@ data class Widget(
     private val rawMinValue: Float?,
     private val rawMaxValue: Float?,
     private val rawStep: Float?,
+    val row: Int,
+    val column: Int,
```
mueller-ma, line 58: I guess that should be of type `Int?`

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -342,6 +359,10 @@ fun Node.collectWidgets(parent: Widget?): List<Widget> {
         rawMinValue = minValue,
         rawMaxValue = maxValue,
         rawStep = step,
+        row = row,
+        column = column,
+        command = command,
+        releaseCommand = releaseCommand,
```
mueller-ma, line 365: Just set these to `null` and remove the changes from this function. When you're on it: You could add a comment to this function, that it's only used on openHAB versions with XML API, that is openHAB 1 only.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -393,6 +414,10 @@ fun JSONObject.collectWidgets(parent: Widget?): List<Widget> {
         rawMinValue = optFloatOrNull("minValue"),
         rawMaxValue = optFloatOrNull("maxValue"),
         rawStep = optFloatOrNull("step"),
+        row = optInt("row"),
+        column = optInt("column"),
```
mueller-ma, line 418: You need to create the function `optIntOrNull()` and use it here. Copy the implementation from `optFloatOrNull()` ;)

mueller-ma (changes requested):
I haven't looked into the API response when using the new `Button` element, but is it possible to parse it as `Widget.mapping` and extend the class `LabeledValue`? Then you can extend `ButtongridViewHolder` with only additions, as parsing is done in `Widget.kt`.

jimtng:
> I haven't looked into the API response when using the new `Button` element, but is it possible to parse it as `Widget.mapping` and extend the class `LabeledValue`?

The new buttons are provided as children widgets, with their additional properties very similar to other widgets, so it seems easier to convert things the other way around: mappings -> widget. 

Example sitemap:
```java
sitemap test {
    Buttongrid item=TestSwitch1 buttons=[1:1:"ON"="Label for on"="power", 1:2:"OFF"="Label for off"] {
      Button item=TestSwitch2 row=1 column=3 click="ON" icon="switch-off" release="OFF"
      Button item=TestSwitch3 row=1 column=4 click="OFF" icon="light" release="ON"
    }
}
```

API Response:
```javascript
{
  "name": "test",
  "link": "http://192.168.1.10:8484/rest/sitemaps/test",
  "homepage": {
    "id": "test",
    "link": "http://192.168.1.10:8484/rest/sitemaps/test/test",
    "leaf": false,
    "timeout": false,
    "widgets": [
      {
        "widgetId": "00",
        "type": "Buttongrid",
        "visibility": true,
        "label": "TestSwitch1",
        "labelSource": "ITEM_NAME",
        "icon": "buttongrid",
        "staticIcon": false,
        "unit": "",
        "mappings": [
          {
            "row": 1,
            "column": 1,
            "command": "ON",
            "label": "Label for on",
            "icon": "power"
          },
          {
            "row": 1,
            "column": 2,
            "command": "OFF",
            "label": "Label for off"
          }
        ],
        "item": {
          "link": "http://192.168.1.10:8484/rest/items/TestSwitch1",
          "state": "OFF",
          "type": "Switch",
          "name": "TestSwitch1",
          "tags": [],
          "groupNames": [
            "TestSwitches"
          ]
        },
        "widgets": [
          {
            "widgetId": "0000",
            "type": "Button",
            "visibility": true,
            "label": "ON",
            "labelSource": "SITEMAP_WIDGET",
            "icon": "switch-off",
            "staticIcon": false,
            "mappings": [],
            "row": 1,
            "column": 3,
            "command": "ON",
            "releaseCommand": "OFF",
            "stateless": false,
            "item": {
              "link": "http://192.168.1.10:8484/rest/items/TestSwitch2",
              "state": "OFF",
              "type": "Switch",
              "name": "TestSwitch2",
              "tags": [],
              "groupNames": [
                "TestSwitches"
              ]
            },
            "widgets": []
          },
          {
            "widgetId": "0001",
            "type": "Button",
            "visibility": true,
            "label": "OFF",
            "labelSource": "SITEMAP_WIDGET",
            "icon": "light",
            "staticIcon": false,
            "mappings": [],
            "row": 1,
            "column": 4,
            "command": "OFF",
            "releaseCommand": "ON",
            "stateless": false,
            "widgets": []
          }
        ]
      }
    ]
  }
}
```

jimtng:
I'm not familiar with Kotlin, but I'm wondering why we aren't using default arguments, e.g. for the Widget class, so when adding new fields, especially those that can be null, we can supply a default value (null) and won't have to update all the class initialisations everywhere with the new fields.

mueller-ma:
Looking at the API response you could add a new property `widgets: List<Widget>?` to class `Widget` and parse the Buttons in `Widget.kt`. The function `ButtongridViewHolder.onBind()` is executed whenever the buttongrid is shown, so currently you parse it multiple times.

> I'm not familiar with Kotlin, but I'm wondering why we aren't using default arguments, e.g. for the Widget class, so when adding new fields, especially those that can be null, we won't have to add the new fields to all the class initialisations everywhere

Currently (without your change) there are only 3 initialization of Widget: One for openHAB 1 (maybe remove that at some point? 🤔 ), one for openHAB >2 initial loading and one for updates. The properties must be set in at least the latter to cases, so it would only make it easier for the oh1 code. IMO it's better how it is currently as you don't forget to add the arguments in `updateFromEvent()`.

mueller-ma:
@maniac103 What do you think about my suggestions in https://github.com/openhab/openhab-android/pull/3723#issuecomment-2198485074?

jimtng:
> The function `ButtongridViewHolder.onBind()` is executed whenever the buttongrid is shown, so currently you parse it multiple times.

AFAIK, inside `ButtongridViewHolder.bind()` I'm only converting `mappings` into Widgets to normalise them all into widgets.  Is that what you're referring to as "parsing"? Agreed that this could be done once, perhaps in a constructor for `ButtongridViewHolder`?



> Looking at the API response you could add a new property `widgets: List<Widget>?` to class `Widget` and parse the Buttons in `Widget.kt`.

Yes, this can be done, but, won't they also get parsed in the normal course of recursive process of `collectWidgets()` into the main `WidgetAdapter.items`? Can I "swallow" them and remove them from the main json tree so collectWidgets() won't see them?

What I'm doing now is grab them from `WidgetAdapter.items` by finding parentid = buttongridid, and also ignoring them in `shouldShowWidget()`. As mentioned above, indeed this process could be moved into the constructor to avoid doing them repeatedly. wdyt?

jimtng:
I've just tried to move the code into constructor / initialisation stage, and found that I couldn't, because the widget ID that I need is given to the `bind` method. Now I see why you want to move it into the parsing stage!

jimtng:
@mueller-ma I've refactored it to what I think you described above. 

I still need to solve an issue with the release event (they're missing on every second click), plus the buttons don't provide any visual indication that they're being clicked at all.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -66,43 +73,48 @@ data class Widget(
     val rawInputHint: InputTypeHint?
 ) : Parcelable {
     val label get() = rawLabel.split("[", "]")[0].trim()
-    val stateFromLabel: String? get() {
```
maniac103, line 69: It would be good to get rid of the somewhat random reformatting here and below.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -229,6 +247,7 @@ data class Widget(
             item?.isOfTypeOrGroupType(Item.Type.Number) == true ||
                 item?.isOfTypeOrGroupType(Item.Type.NumberWithDimension) == true ->
                 state.toParsedState(item.state?.asNumber?.format)
+
```
maniac103, line 250: Random newline?

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -316,6 +336,7 @@ fun Node.collectWidgets(parent: Widget?): List<Widget> {
                 }
                 mappings.add(LabeledValue(mappingCommand, mappingLabel, null, 0, 0))
             }
+
```
maniac103, line 339: Random newline?

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -54,6 +55,12 @@ data class Widget(
     private val rawMinValue: Float?,
     private val rawMaxValue: Float?,
     private val rawStep: Float?,
+    // sub-widgets are for the buttons in buttongrids
+    val widgets: List<Widget>?,
```
maniac103, line 59: Why not name the member variable `buttons` to make the comment content more obvious?

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -373,6 +401,51 @@ fun JSONObject.collectWidgets(parent: Widget?): List<Widget> {
     val icon = optStringOrNull("icon")
     val staticIcon = optBoolean("staticIcon", false)
 
+    val widgets = if (type == Widget.Type.Buttongrid) {
+        val buttonWidgets = ArrayList<Widget>(mappings.map { mapping ->
```
maniac103, line 405: Does `map` not already yield a `List`?

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
```diff
@@ -120,14 +121,19 @@ fun RemoteViews.duplicate(): RemoteViews {
 }
 
 fun MaterialButton.setTextAndIcon(connection: Connection, mapping: LabeledValue) {
-    contentDescription = mapping.label
-    val iconUrl = mapping.icon?.toUrl(context, true)
+    setTextAndIcon(connection, mapping.label, mapping.icon)
```
maniac103, line 124: No need for this method overload, just give `icon` a default value below.

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
```diff
@@ -120,14 +121,19 @@ fun RemoteViews.duplicate(): RemoteViews {
...
+    val iconUrl = icon?.toUrl(context, true)
     if (iconUrl == null) {
-        icon = null
-        text = mapping.label
+        this.icon = null
+        text = label
         return
     }
     val iconSize = context.resources.getDimensionPixelSize(R.dimen.section_switch_icon)
+    val thisButton = this
```
maniac103, line 136: We usually do `this@MaterialButton` instead of using a variable.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -339,6 +340,7 @@ class WidgetAdapter(
                 widget.mappingsOrItemOptions.isNotEmpty() -> TYPE_SECTIONSWITCH
                 else -> TYPE_SWITCH
             }
+
```
maniac103, line 343: Random newlines here and below

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -859,8 +867,19 @@ class WidgetAdapter(
             }
         }
 
-        override fun onClick(view: View) {
-            connection.httpClient.sendItemCommand(boundWidget?.item, view.tag as String)
+        override fun onTouch(view: View, event: MotionEvent): Boolean {
```
maniac103, line 870: I think I'd prefer keeping `onClick` if there's no release command, since reacting to pointer down events directly is somewhat prone to mis-clicks. I understand it's not possible to do otherwise with release commands, but I imagine those to stay a niche use case.

maniac103 (commented):
@mueller-ma Yes, the idea behind not using default parameter values here was making sure to not forget to add any new fields, as well as forcing making a conscious decision about defaults in case of e.g. OH1.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
jimtng: This was done by Android Studio's default Kotlin formatting. Any tips what I should do in regards to this?

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
jimtng: ok

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
jimtng: Yes, but apparently it's an immutable one, so the subsequent attempt to add to the list results in an error.

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
jimtng: Kotlin noob here! Thanks for the tip!

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
jimtng: I'm not sure I understand what you meant. Would you mind elaborating on this please?

maniac103: You currently have two `setTextAndIcon` methods with different parameters, where one redirects to the other. You don't need the 2 parameter version, as you can implement the 3 parameter version with a default parameter for `icon`:
`fun MaterialButton.setTextAndIcon(connection: Connection, label: String, icon: IconResource? = null)`

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
maniac103: The more Kotlin-esque way to fix that is `.map { ... }.toMutableList()

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
maniac103: We're probably missing some settings file in the repository. Did this reformatting happen 'magically' or did you do something to trigger it?

jimtng: I am going to confess here. I edit the files using VSCode, then go to Android Studio to reformat it using Cmd+Option+L then build / test it. One of these days I need to learn how to use Android Studio properly!

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
jimtng: The 2 parameter version takes a `LabeledValue`, not a String. I can't see a clean way to combine them into one.

jimtng: I can go and change all the calls to setTextAndIcon to use the 3 parameter version. Perhaps that's better anyway.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
jimtng: I've manually restored the random formatting changes for now, but I'd love to get the automatic formatting done properly without messing up the established formatting.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
jimtng: I'm still trying to figure out this part. Even if we use onclick when no release command, I'd still have to implement click+release so they work properly.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
mueller-ma: Can you check this setting and make sure it's set to "Project"? 

grafik

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -354,6 +355,7 @@ class WidgetAdapter(
             Widget.Type.Mapview -> TYPE_LOCATION
             Widget.Type.Input -> if (widget.shouldUseDateTimePickerForInput()) TYPE_DATETIMEINPUT else TYPE_INPUT
             Widget.Type.Buttongrid -> TYPE_BUTTONGRID
+            Widget.Type.Button -> TYPE_BUTTON
```
mueller-ma, line 358:
```suggestion
            Widget.Type.Button -> TYPE_INVISIBLE
```

Then you can remove the unused `TYPE_BUTTON`.

jimtng: I am right in the middle of tracking down a weird problem related to this. 

Basically the new buttons from Buttongrid are actual widgets. This means that the server will send state updates about the button (widget), so they need to be included in the main list of widgets. Therefore I believe this needs to stay here because it's used in places. I'm still trying to understand the big picture of the app, didin't realise I'd have to delve into it this deep.

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
maniac103: > The 2 parameter version takes a LabeledValue, not a String

Oh, I missed that during review. Adjusting the calls is the right solution :+1:

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
maniac103: > I'd still have to implement click+release so they work properly.

What do you mean by that?

AFAICT, the logic could be (pseudo code):
```
if (has release command) {
    setOnClickListener(null)
    setOnTouchListener(this)
} else {
    setOnClickListener(this)
    setOnTouchListener(null)
}
````

The `onClick` implementation could stay as it is now, and you'd just *additionally* add the touch listener.

Comment (jimtng) [reply mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt:864]: It's a bit complicated. When handling the ontouch event, even after calling `view.performClick()` as per [the warning](https://googlesamples.github.io/android-custom-lint-rules/checks/ClickableViewAccessibility.md.html), the standard accesibility action (haptic feedback) didn't occur. 

I implemented a workaround that makes both with and without release command to work using a combination of both onclick and ontouch handlers.

In the case of no releasecommand, it still goes through onclick as normal, even though ontouch is also implemented. It's probably easier to understand what I'm trying to say by to looking at the code in the next commit.

File: mobile/src/main/java/org/openhab/habdroid/model/Widget.kt (hunk 4 above)
Comment (jimtng) [reply mobile/src/main/java/org/openhab/habdroid/model/Widget.kt]: Mine is also set to Project, but it's still adding those extra lines breaks

Comment (jimtng) [reply mobile/src/main/java/org/openhab/habdroid/model/Widget.kt]: image
image
image
image

File: mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt [hunk 14]
```diff
@@ -860,7 +867,24 @@ class WidgetAdapter(
...
 
         override fun onClick(view: View) {
-            connection.httpClient.sendItemCommand(boundWidget?.item, view.tag as String)
+            val button = view.tag as Widget
+            if (button.releaseCommand == null) {
+                // when there's a "releaseCommand", the normal "command" is sent in the onTouch event
+                button.command?.let { connection.httpClient.sendItemCommand(button.item, it) }
+            } else {
+                connection.httpClient.sendItemCommand(button.item, button.releaseCommand)
+            }
```
maniac103, line 873: Simplification suggestion:
```
// when there's a "releaseCommand", the normal "command" is sent in the onTouch handler
val command = button.releaseCommand
    ?: button.command
    ?: return
connection.httpClient.sendItemCommand(button.item, command)
```

jimtng: ohhhh some Kotliny goodness!

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -869,21 +869,28 @@ class WidgetAdapter(
...
+
+            when (event.action) {
+                MotionEvent.ACTION_DOWN -> {
+                    button.command?.let { connection.httpClient.sendItemCommand(button.item, it) }
+                }
+                MotionEvent.ACTION_UP -> {
+                    button.releaseCommand.let { connection.httpClient.sendItemCommand(button.item, it) }
+                }
+                else -> return false
+            }
```
maniac103, line 890: You're too fast for me ;-) In the spirit of above comment:

```
if (button.releaseCommand != null) {
    val command = when (event.action) {
        MotionEvent.ACTION_DOWN -> button.command
        MotionEvent.ACTION_UP -> button.releaseCommand
        else -> null
    }
    command?.let { connection.httpCLient.sendItemCommand(button.item, it) }
}
// Don't return true here!
// Even though we're handing this event, we want the click gesture to be handled normally
// so that we can capture the release command in the onClick event handler.
return false // tell the system that we didn't consume the event
```

jimtng: Nice! Thanks I really appreciate your help. Learning lots from here! I've committed this suggestion just now.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
jimtng: I _think_ in this case, it's nicer to keep the if condition just for releaseCommand. It makes the code easier to understand at a glance:

if (button.releaseCommand == null) {
  // do this here
}

..... inside onTouch:
if (button.releaseCommand != null) {
  // this is basically the "else" part of the above
}

wdyt?

maniac103: I personally think the comment should be sufficient to explain that, but I don't have an overly strong opinion on this either.

jimtng: I removed the comment several commits ago :( Also I changed the null check to `isNullOrEmpty()`. See the latest version/comment below.

jimtng:
This is now "functional", and supports text label, or icon, click and release commands. 

However, more work needs to be done to support stateful buttons, icon/text colours.

mueller-ma:
> However, more work needs to be done to support stateful buttons, icon/text colours.

Is this something you'd like to do here or in a different PR?

jimtng:
It's almost working on my local version, but it might not be the best way to achieve it. I'd love to get some feedback later once I'm ready to push the new changes.

jimtng:
@mueller-ma I've just pushed the latest version that supports dynamic buttons. I'd appreciate it if you could take a look and test this.

Here's my super messy sitemap that I'm using to test:

```java
sitemap test {
    Buttongrid label="main label" icon="garden" item=TestSwitch1 buttons=[5:1:ON="ON"] {
      // stateless button
      Button item=TestSwitch1 row=1 column=1 click="ON" stateless icon="switch" //release="OFF" // visibility=[TestSwitch1==ABC]
     
      // text label only with dynamic color
      Button item=TestSwitch1 row=1 column=2 click="ON" release="OFF" label="click/release" labelcolor=[TestSwitch1=="ON"="red", TestSwitch1=="OFF"="green"] // visibility=[TestSwitch1==ABC]
     
      Button item=TestSwitch1 row=1 column=3 click="OFF" iconcolor=[TestSwitch1==ON="red",TestSwitch1==OFF="yellow"] // release="RELEASE" // visibility=[TestSwitch1==ABC]
      Button item=TestSwitch1 row=2 column=3 click="OFF" icon="material:mic" iconcolor=[TestSwitch1==ON="red",TestSwitch1==OFF="yellow"] // release="RELEASE" // visibility=[TestSwitch1==ABC]
      Button item=TestSwitch1 row=2 column=2 click="ON" icon="material:mic" iconcolor=[TestSwitch1==ON="red"] // release="RELEASE" // visibility=[TestSwitch1==ABC]
    }
    Switch item=TestSwitch1 label="Test" icon="light" iconcolor=[TestSwitch1==ON="red",TestSwitch1==OFF="green"] valuecolor=[22="green"] //iconcolor=[22="green"]
}
```

Honestly I'm running blind when it comes to Android development. I created some helper maps to help me look things up. If the same information can be obtained from android's framework, that would be great. I tried poking around but couldn't find something I could use.

maniac103:
> but it might not be the best way to achieve it.

Indeed ;-) The `Widget` class must stay immutable, as keeping model classes immutable was one of the design decisions in the app design.

I'll look at potential improvement tomorrow or on Friday.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -178,6 +187,12 @@ data class Widget(
         return chartUrl.toString()
     }
```
maniac103, line 189: Is this actually called anywhere?

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -299,7 +307,12 @@ class WidgetAdapter(
         if (oldWidget.visibility != widget.visibility && items.any { w -> w.parentId == widget.id }) {
             notifyDataSetChanged()
         } else {
-            notifyItemChanged(position)
+            // update the parent Buttongrid if the updated widget is a button
+            if (widget.type == Widget.Type.Button && widget.parentId != null) {
+                (viewHoldersById[widget.parentId] as? ButtongridViewHolder)?.let { it.updateButton(widget) }
```
maniac103, line 312: Instead of this, just find the index of parentId in the list and call `notifyItemChanged(index)` ... this will cause onBind to be called again, where the updated data can be applied.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
jimtng: it's unrelated to this PR? A quick search shows toChartUrl() called in ChartWidgetActivity.kt and WidgetAdapter.kt

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -812,38 +828,74 @@ class WidgetAdapter(
...
+                buttonView.setTextAndIcon(
+                    connection = connection, 
+                    label = button.label, 
+                    icon = button.icon,
+                    labelColor = button.labelColor,
+                    iconColor = button.iconColor,
+                    mapper = colorMapper
+                )
+                buttonView.isChecked = button.item?.state?.asString == button.command
+                boundWidget?.buttons?.replaceAll { if (it.id == button.id) button else it }
```
maniac103, line 851: This should be done in the SSE event listener, like for other widgets.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -812,38 +828,74 @@ class WidgetAdapter(
...
+            }
+        }
 
         override fun bind(widget: Widget) {
+            if (waitingForTouchRelease) {
```
maniac103, line 856: This is a problem, as the app will potentially show stale data in that case. Need to think about it...

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
jimtng: Are you saying: find the parent (buttongrid) and call notifyItemChanged(buttongrid)? No that won't work, because 
- it messes up with click/release because Buttongrid's bind wipes out the tableview and recreates it while the button in the table is still trying to keep track of the mouse down, waiting for it to go up.
- The buttongrid widget doesn't have the updated Button widget state (color, item state)

This is why I made the Widget.buttons mutable, so it can be updated

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
jimtng: Depends on what "stale data" means. It handles button state updates through `ButtongridViewHolder.updateButton()`.

If however, during waiting, the whole buttongrid structure is changed (buttons added / removed), then possibly yes. I haven't tested this scenario but I don't think it's a normal operation.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
maniac103: > Sorry I still don't know my way around

Not a problem at all; the comments partially were meant as 'note to self' ;-)
I'm referring to [this code](https://github.com/openhab/openhab-android/blob/main/mobile/src/main/java/org/openhab/habdroid/ui/activity/PageConnectionHolderFragment.kt#L430)

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
maniac103: > The buttongrid widget doesn't have the updated Button widget state (color, item state)

Then we need to make it so it does have them ;-)
The flow basically should be
- SSE listener updates widget (here: button grid)
- SSE listener calls back into UI to notify about change
- Adapter uses that callback for rebinding

As for the release handling, I'll need to look into it for a bit before being able to comment properly. My gut feeling is that the view holder needs to keep track whether it's bound to the same or a different widget and if it's the same, keep touch tracking intact.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
maniac103: Sorry, the comment was meant to point to the newly added `updateButton` method.

jimtng:
@maniac103 I am working on yet a different approach. This time:
- Widget class stays immutable
- ButtongridViewHolder.bind() will be called on button update
- I'll work on your idea below:
> My gut feeling is that the view holder needs to keep track whether it's bound to the same or a different widget and if it's the same, keep touch tracking intact.

jimtng:
not sure what the customs are over here. I've just force pushed for minor changes to satisfy lint checks

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
jimtng: refactored so this is no longer done

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
jimtng: it now no longer keeps a separate copy of the Button widgets

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
jimtng: I've removed `updateButton`

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
jimtng: I'm still fighting with the linter. I tried running `ktlint --format` but it changed some existing code too.

jimtng:
It would be great if this could be merged before 4.2 release, and mention all the changes in the [release blog post](https://github.com/openhab/website/pull/475), since ios gets a mention there.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -825,42 +855,89 @@ class WidgetAdapter(
             labelView.isVisible = showLabelAndIcon
             iconView.isVisible = showLabelAndIcon
 
-            val mappings = widget.mappings.filter { it.column != 0 && it.row != 0 }
-            spareViews.addAll(table.children.map { it as? MaterialButton }.filterNotNull())
-            table.removeAllViews()
```
maniac103, line 830: @jimtng I'm currently looking at the code and am preparing some simplifications, and am wondering: was there a specific reason for this change (kill off `spareViews`, remove table children selectively)? I am asking because
- `spareViews` is an optimization for only creating new views if absolutely needed (it's a relatively expensive operation, as it involves e.g. XML parsing)
- removing table children selectively avoids a relatively cheap operation at the expense of much larger tracking overhead (the `buttonViews` map and its management isn't needed without it)

If there's no hard reason my proposal would be restoring the existing code. FWIW, my (currently untested) simplification proposal can be found [here](https://github.com/maniac103/openhab.android/commit/dfb20a06c342b1fa0ab45df54440688074ceec56)

jimtng: The whole aim of this exercise is to want to preserve the same buttonView object for the same grid location, and not calling `table.removeAllViews()`. When removeAllViews is called, the buttonView got wiped out, and that's why onTouch/Release tracking is gone. Touch down -> sends a command -> receives widget update -> calls bind() -> removeAllViews -> oops that buttonView waiting for the touch release just disappeared.

Thanks for your proposed code. I'll study it and get back to you on this.

maniac103: OK, got it, thanks. Updated proposal [here](https://github.com/maniac103/openhab.android/commit/efc14a3a263cb038d15262a3d01ee3047adfb986).

jimtng: I've had a look through your proposed code. I love that callback trick!

However, I think the removeAllViews will mess up with the touch release tracking. I tried your code and got the following exception

```
ACRA caught a IllegalStateException for org.openhab.habdroid.beta
java.lang.IllegalStateException: The specified child already has a parent. You must call removeView() on the child's parent first.
	at android.view.ViewGroup.addViewInner(ViewGroup.java:5275)
	at android.view.ViewGroup.addView(ViewGroup.java:5104)
	at android.view.ViewGroup.addView(ViewGroup.java:5076)
	at org.openhab.habdroid.ui.WidgetAdapter$ButtongridViewHolder.bind(WidgetAdapter.kt:898)
	at org.openhab.habdroid.ui.WidgetAdapter.onBindViewHolder(WidgetAdapter.kt:260)
	at org.openhab.habdroid.ui.WidgetAdapter.onBindViewHolder(WidgetAdapter.kt:128)
	at androidx.recyclerview.widget.RecyclerView$Adapter.onBindViewHolder(RecyclerView.java:7747)
	at androidx.recyclerview.widget.RecyclerView$Adapter.bindViewHolder(RecyclerView.java:7847)
	at androidx.recyclerview.widget.RecyclerView$Recycler.tryBindViewHolderByDeadline(RecyclerView.java:6646)
	at androidx.recyclerview.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:6917)
	at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:6757)
	at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:6753)
	at androidx.recyclerview.widget.LinearLayoutManager$LayoutState.next(LinearLayoutManager.java:2362)
	at androidx.recyclerview.widget.LinearLayoutManager.layoutChunk(LinearLayoutManager.java:1662)
	at androidx.recyclerview.widget.LinearLayoutManager.fill(LinearLayoutManager.java:1622)
	at androidx.recyclerview.widget.LinearLayoutManager.onLayoutChildren(LinearLayoutManager.java:687)
	at androidx.recyclerview.widget.RecyclerView.dispatchLayoutStep2(RecyclerView.java:4645)
	at androidx.recyclerview.widget.RecyclerView.dispatchLayout(RecyclerView.java:4348)
	at androidx.recyclerview.widget.RecyclerView.onLayout(RecyclerView.java:4919)
	at android.view.View.layout(View.java:24421)
	at android.view.ViewGroup.layout(ViewGroup.java:6440)
	at android.widget.FrameLayout.layoutChildren(FrameLayout.java:332)
	at android.widget.FrameLayout.onLayout(FrameLayout.java:270)
                                                                                                    	at android.view.View.layout(View.java:24421)
```

mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt
```diff
@@ -119,12 +121,20 @@ fun RemoteViews.duplicate(): RemoteViews {
     }
 }
 
-fun MaterialButton.setTextAndIcon(connection: Connection, mapping: LabeledValue) {
-    contentDescription = mapping.label
-    val iconUrl = mapping.icon?.toUrl(context, true)
+fun MaterialButton.setTextAndIcon(
```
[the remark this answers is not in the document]
jimtng: Thanks for the fix! Should I update my code with it?

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
jimtng: I commented before seeing your updated proposal!

jimtng:
@maniac103 I've cherry picked your commits and added some fixes.

jimtng:
My last issue with this is for optimisation, and can perhaps be done later on if it's difficult / not important.
- I think the ButtongridViewholder bind() gets called multiple times when there were multiple button widgets getting updated (due to changes in their states). 
How can we combine all the updates and only run the ButtongridViewholder bind() once - after all its widgets have finished updating? All I can think of is using some sort of delayed update, but that might make the UI feel laggy.

maniac103:
> How can we combine all the updates and only run the ButtongridViewholder bind() once - after all its widgets have finished updating? All I can think of is using some sort of delayed update, but that might make the UI feel laggy.

I don't think we meaningfully do that, since there's no way for the client to know whether and/or when the server will send more update events.

> I explored implementing ButtonViewHolder so that each would handle its own update, and then somehow update the view inside the Buttongrid's table. 

This sounds like an abuse of the view holder concept ;-) I think we should keep the 'one view holder corresponds to one row in the list' concept.

As long as there's no evidence of an actual problem here, my suggestion is to not pursue this optimization.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -154,6 +162,8 @@ class WidgetAdapter(
             widgets.size == items.size &&
             widgets.filterIndexed { i, widget -> getItemViewType(widget) != getItemViewType(items[i]) }.isEmpty()
 
+        Log.d(TAG, "update, forceFullUpdate: $forceFullUpdate, compatibleUpdate: $compatibleUpdate")
```
maniac103, line 165: Maybe remove this log line?

maniac103 (approved):
Apart from that one minor comment, code LGTM now :+1:

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
jimtng: yup!

jimtng:
hold off on merging!

jimtng:
@maniac103 this took way longer to figure out than I initially thought.

There were two unrelated problems:
- Because the buttonviews are now reused, the previous checked state can carry over to a stateless button in another sitemap. So a more complex logic is needed to update the checked state, as noted in the comments. Previously when the views were discarded, and a new one created, this wasn't an issue.
- The callback mechanism, as wonderful as it was, didn't work. This took me a while to figure out. What happens is the ButtongridViewHolder was only created once, perhaps because Android does the same thing with caching it rather than discarding / creating new ViewHolder. So it received the callback + reference to the widgetsByParentId object that belongs to the WidgetAdapter that initially created it. However, whenever we switch around to different sitemaps, new WidgetAdapter objects were created, and it was given the old ButtongridViewHolder to use. Therefore the callback referencing the old (wrong / stale) widgetsByParentId that was long discarded by the previous / different WidgetAdapter.

So I'm reverting back to the previous method of accessing widgetsByParentId. Let me know if you have a different/better idea to solve this.

As it is now, things are working properly as far as I can tell.

maniac103:
> The callback mechanism, as wonderful as it was, didn't work. This took me a while to figure out. What happens is the ButtongridViewHolder was only created once, perhaps because Android does the same thing with caching it rather than discarding / creating new ViewHolder. So it received the callback + reference to the widgetsByParentId object that belongs to the WidgetAdapter that initially created it. 

Damn, right, I completely forgot about this. The reason for that is [this](https://github.com/openhab/openhab-android/blob/main/mobile/src/main/java/org/openhab/habdroid/ui/WidgetListFragment.kt#L142) which causes `ViewHolders` to be shared among fragments. I'll give this another thought.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -825,42 +859,116 @@ class WidgetAdapter(
...
+                        //    because stateless == null means stateless
+                        //    so don't use `stateless == true` by rearranging the clauses below
+                        //
+                        // 2. Rule: isChecked cannot be changed when isCheckable is false
+                        //    and it will be stuck at whatever state it was before.
+                        //    So:
+                        //    before setting the checked state, set checkable to true
+                        //    consequently,
+                        //    before setting checkable to false, clear the checked state first
+                        //
```
maniac103, line 933: The comment seems a little bit verbose ;-) Rewording suggestion:

```
if (button.stateless == false) {
    // stateful button: make checkable and set checked state afterwards (isChecked can not be set if isCheckable is false)
    buttonView.isCheckable = true
    buttonView.isChecked = button.item?.state?.asString == button.command
} else {
    // stateless button: not checkable (unset isChecked before isCheckable for the reason outlined above)
    buttonView.isChecked = false
    buttonView.isCheckable = false
}
```

jimtng: Thanks! :)

maniac103:
> I'll give this another thought.

How about [this](https://github.com/maniac103/openhab.android/commit/3e6eaa8d5eb4ba7eca0355e54e11c94e05ec94d2)? It matches what we did for other cases of ViewHolder-to-Adapter callbacks.

jimtng:
> How about [this](https://github.com/maniac103/openhab.android/commit/3e6eaa8d5eb4ba7eca0355e54e11c94e05ec94d2)? It matches what we did for other cases of ViewHolder-to-Adapter callbacks.

Beautiful! I didn't know about that vhc! This works!

maniac103:
> I didn't know about that vhc!

I guess once upon a time I stumbled across the same issue (view holders being shared across fragments), + wanted to avoid long parameter lists, and this invented that solution. That was so long ago though that I completely forgot about it ;-)

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -825,42 +868,107 @@ class WidgetAdapter(
...
+                        newButton
+                    }
+
                     // Rows and columns start with 1 in Sitemap definition, thus decrement them here
-                    val mapping = mappings.firstOrNull { it.row - 1 == row && it.column - 1 == column }
+                    val button = buttons.firstOrNull { (it.row ?: 0) - 1 == row && (it.column ?: 0) - 1 == column }
                     // Create invisible buttons if there's no mapping so each cell has an equal size
-                    buttonView.isInvisible = mapping == null
-                    if (mapping != null) {
+                    buttonView.isInvisible = button == null
```
maniac103, line 917: Shouldn't this also check `button.visibility`, or is this checked elsewhere?

jimtng: This was the original code, I didn't pay too close attention to it

maniac103: Before buttons weren't widgets, this didn't have a visibility flag ;-)

jimtng: the call below (line 941) seems redundant

maniac103: Indeed, agreed.

jimtng: Actually, yes good catch! I'll push a fix.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -825,42 +867,108 @@ class WidgetAdapter(
...
+                        newButton
+                    }
+
                     // Rows and columns start with 1 in Sitemap definition, thus decrement them here
-                    val mapping = mappings.firstOrNull { it.row - 1 == row && it.column - 1 == column }
-                    // Create invisible buttons if there's no mapping so each cell has an equal size
-                    buttonView.isInvisible = mapping == null
-                    if (mapping != null) {
+                    val button = buttons.firstOrNull { (it.row ?: 0) - 1 == row && (it.column ?: 0) - 1 == column }
+                    if (button != null && button.visibility == true) {
```
maniac103, line 916: Since `visibility` isn't nullable, the `== true` part is redundant. (And if it were nullable, I think `!= false` would be more correct to make them visible by default)

jimtng: it isn't nullable, so I'll remove the == true part

mueller-ma:
@maniac103 @jimtng Is this ready to merge from your side?

jimtng:
Yes this is ready afaik.

maniac103:
@mueller-ma I have not personally tested it, but AFAICT code looks good.

mueller-ma (approved):

PR #3724, opened 2024-06-29
Implement new notification features

mueller-ma:
- [x] Parsing via FCM
- [x] Parsing from JSON using the new `payload` field
- [x] Image loading: URL, URL from Item state, base64 from Item state
- [x] Notification list: Title 
- [x] Notification list: Show Image
- [x] Notification: Title
- [x] Notification: Actions
- [x] Notification: onClick
- [x] Notification: Show Image
- [x] Rename severity to tag
- [x] Check docs if anything is missing
- [x] Hide notification
- [x] referenceId


mobile/src/main/java/org/openhab/habdroid/model/CloudNotification.kt
```diff
@@ -21,17 +23,40 @@ import java.util.TimeZone
...
+
+    suspend fun loadImage(connection: Connection, context: Context, size: Int): Bitmap? {
+        mediaAttachmentUrl ?: return null
+        //if (mediaAttachmentUrl.startsWith("item:")) {
+        //    val itemName = mediaAttachmentUrl.removePrefix("item:")
+        //    val item = ItemClient.loadItem(connection, itemName)
+        //
+        //}
+        val fallbackColor = context.getIconFallbackColor(IconBackground.APP_THEME)
+        return connection.httpClient
```
digitaldan, line 55: I know this is still in an early draft,  but i though i would mention that if the user puts in a relative path for media, so say `/static/foo.png` the client should try hitting that on the user's openhab (through local or remote settings).    I just made this change to IOS, and there is a PR for the cloud docs up right now to document it.   So the prefixes possible  are "http", "/" and "item" 👍   Other wise no other changes to what we discussed.

mueller-ma: The function https://github.com/openhab/openhab-android/blob/main/mobile/src/main/java/org/openhab/habdroid/util/HttpClient.kt#L94 is called internally by httpClient.get(), so relative paths should already work.

digitaldan: Thats awesome.  Thanks!

mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
```diff
@@ -94,6 +95,19 @@ class NotificationHelper(private val context: Context) {
         )
     }
 
+    private fun createActionIntent(action: CloudNotificationAction, notificationId: Int): PendingIntent {
+        val intent = Intent(context, NotificationActionReceiver::class.java)
+        intent.putExtra(NOTIFICATION_ID_EXTRA, notificationId)
+        intent.putExtra(NOTIFICATION_ACTION_LABEL, action.label)
```
maniac103, line 101: `action` is parcelable, so I'd suggest to use it as extra directly instead of splitting it in parts and reconstructing it later.

mobile/src/full/java/org/openhab/habdroid/core/FcmMessageListenerService.kt
```diff
@@ -42,15 +45,23 @@ class FcmMessageListenerService : FirebaseMessagingService() {
...
+                    id = data["persistedId"].orEmpty(),
+                    title = data["title"].orEmpty(),
+                    message = data["message"].orEmpty(),
                     // Older versions of openhab-cloud didn't send the notification generation
                     // timestamp, so use the (undocumented) google.sent_time as a time reference
                     // in that case. If that also isn't present, don't show time at all.
-                    data["timestamp"]?.toLong() ?: message.sentTime,
-                    data["icon"].toOH2IconResource(),
-                    data["severity"]
+                    createdTimestamp = data["timestamp"]?.toLongOrNull() ?: message.sentTime,
```
maniac103, line 59: While I understand it's a good thing to not crash the app on wrong timestamp formats, we probably wouldn't have noticed the openhab-cloud issue that way...

mobile/src/main/AndroidManifest.xml
```diff
@@ -279,6 +279,9 @@
         <receiver
             android:name=".core.NotificationDismissedReceiver"
             android:exported="false" />
+        <receiver
+            android:name=".core.NotificationActionReceiver"
+            android:exported="false" />
```
maniac103, line 284: Why not combine both notification receivers into one?

mobile/src/main/java/org/openhab/habdroid/background/BackgroundTasksManager.kt
```diff
@@ -428,6 +430,21 @@ class BackgroundTasksManager : BroadcastReceiver() {
             )
         }
 
+        fun enqueueNotificationAction(context: Context, action: CloudNotificationAction) {
+            val split = action.action.split(":", limit = 3)
```
maniac103, line 434: Ideally that kind of parsing should happen in `CloudNotificationAction` itself...maybe with a sealed class, I'll have a look at it.

maniac103: For your consideration (also dealing with the comment above): https://github.com/maniac103/openhab.android/commits/notification-enhancements/

maniac103:
> Notification: onClick

@digitaldan How is `onClickAction` meant to be handled? Executing the action is clear, but should that be done *in addition to* or *instead of* opening the notification content in the lst?

maniac103:
Also, how exactly is `referenceId` handled? From the cloud code I gather it's only used for `collapse_key`, but doesn't that only affect notifications which aren't delivered yet? That is, if we have shown a notification for reference ID X and another notification for X is sent via FCM, the collapse key won't change anything according to [the docs](https://firebase.google.com/docs/cloud-messaging/concept-options#collapsible_and_non-collapsible_messages):
`A collapsible message is a message that may be replaced by a new message if it has yet to be delivered to the device.`

AFAICT, that means we need to pass the reference ID via FCM payload, and need to do client side processing to remove older notifications for that reference ID.

maniac103:
Also, what about hiding notifications? If the user chooses to hide notifications by tag or reference ID, will the cloud code send `hideNotification` messages for each respective notification, or will we need to do this client side? If the latter, does the `hideNotification` message provide the `reference-id` or `tag` fields?

mobile/src/full/java/org/openhab/habdroid/core/FcmMessageListenerService.kt
mueller-ma: Maybe remove the fallback to `message.sentTime`? Then make sure that the timestamp is hidden from the notification it it's `null`.

mueller-ma:
I tried to fix the notification trampoline, but now clicking the button of a `ui:...` action does nothing, even no log :/

maniac103:
Since Android 12, the trampoline BroadcastReceiver -> Activity isn't allowed anymore, one must go to the activity from the notification directly. This means for UI actions you need a different `PendingIntent`. See docs [here](https://developer.android.com/about/versions/12/behavior-changes-12#notification-trampolines)

digitaldan:
> Also, how exactly is referenceId handled? From the cloud code I gather it's only used for collapse_key,

Its both set as the "collapse_key" as well as is included in the payload , so you should have access to it already.  I actually assumed it collapsed delivered notifications, but sounds like you can do that programmatically from the app? 

> Also, what about hiding notifications? 

See my comment here, https://github.com/openhab/openhab-addons/issues/16934#issuecomment-2201401787

> How is onClickAction meant to be handled? 

If the action is a `ui:` action, then it should open the app, `command:` actions should not open the app, just send the command.  `http` actions should open the app with an embedded browser for the link (or you could just open the user's browser, i guess up to you, i chose to keep everything in the app on iOS)

maniac103:
> I actually assumed it collapsed delivered notifications, but sounds like you can do that programmatically from the app?

It can't, because FCM isn't what posts those notifications, so the app must be responsible for that. Is this different on iOS?

mobile/src/main/java/org/openhab/habdroid/core/NotificationHandlingReceiver.kt
```diff
@@ -0,0 +1,93 @@
...
+        fun createDismissedIntent(context: Context, notificationId: Int): Intent {
+            return Intent(context, NotificationHandlingReceiver::class.java).apply {
+                action = ACTION_DISMISSED
+                putExtra(EXTRA_NOTIFICATION_ID, notificationId)
+            }
+        }
+
+        fun createActionIntent(context: Context, notificationId: Int, cna: CloudNotificationAction): Intent {
+            val cnaAction = cna.action
+            return if (cnaAction is CloudNotificationAction.Action.UiCommandAction) {
```
maniac103, line 77:
```
return when (val action = cna.action) {
    is CloudNotificationAction...
    ...
}
```

digitaldan:
>  Is this different on iOS?

It is,  the collapse key automatically removes / replaces the message with the same collapse id, although i might be able to do that strictly in the app if i wanted to, but this is what Apple recommends.   On iOS the app has the opportunity to be called before a non-background notification is posted to the message center, allowing it to mutate the message (so add media for example).  For  the "hide notifications" feature, we use background messages (no title or message) , which do not get posted to the notification center by default and only the app gets them.  This would be the equivalent to android, where the app would then be completely responsible for creating a visual notification based on meta data in the background message.

maniac103:
Thinking about the reference-id handling, I think it should be sufficient to just use it instead of persistedId for notification ID generation. This should give us the expected behavior with regards to notification replacement and removal.
For 'remove by tag', we can query the notifications and remove them by group key.

digitaldan:
>  I think it should be sufficient to just use it instead of persistedId for notification ID generation

Just FYi this field is only set if the rule or action sets it, so its not guaranteed right now to always be present.  I could change that on the cloud server side (so generate a uuid if its not set) 
if thats a problem.

maniac103:
> Just FYi this field is only set if the rule or action sets it, so its not guaranteed right now to always be present.

Yeah, that's understood. I should have been more precise; I meant to say 'use reference-id if present, else persistedId, for notification ID generation'.

mueller-ma:
Most things are implemented now, but need some testing.

maniac103:
I added reference-id handling and hiding by tag, together with some misc. cleanup, see latest 3 commits at https://github.com/maniac103/openhab.android/commits/notification-enhancements/

@digitaldan I noticed openhab-cloud doesn't honor reference-id itself, that is, notifications in the DB returned via `api/v1/settings/notifications` are not replaced by others with the same `reference-id`. Is this intended? If so, why? If not, can that behavior be changed? ;-) My personal use case: I have a notification reminding me about low battery state on devices. I don't need this notification to be persisted for every day ad infinitum; instead, it would be much better if I only get the latest one for it. 
In general, it would be great if the notification list would get some love in addition to the great enhancements done for 'live' notifications. See e.g. https://github.com/openhab/openhab-cloud/issues/383 or https://github.com/openhab/openhab-cloud/issues/178

> Most things are implemented now, but need some testing.

I'd love to do this, but I don't see a way to do any testing short of a local openhab-cloud setup, and even with that, I'm not sure what one needs to do to set up FCM on both sides :-/

mueller-ma:
I'm using a Docker container with the latest snapshot build and connect it to the official openhab-cloud for testing.

digitaldan:
>@digitaldan I noticed openhab-cloud doesn't honor reference-id itself, that is, notifications in the DB returned via api/v1/notifications are not replaced by others with the same reference-id. Is this intended? If so, why? 

Its intended in that i view the notification list as a journal of all notifications sent (except hide notifications), and let the clients work out what they want to display, not the server.  I am trying to  keep the cloud code a dumb as possible, so trying to blindly proxy messages, so as we add new payload values, we hopefully don't have to touch the server.  Obviously we do some data mutating, like the collapse key, but its very minimal.

The reference-id should be in the data/payload returned on the api , so you could pretty easily filter to only the last message with that id when showing in the client.   But, we don't record the hide notification message, so that would be nice to somehow represent, either by persisting it or trying to match the client logic, so update all messages in the DB as removed with the same ref-id or tag, but now we start making the cloud act as a client, and may have to coordinate logic to keep them  consistant (like if we add another type of field to delete on).  Only inserting and not having to update notifications in mongo  keeps things simple. 

This is not to say the cloud code can't change, i'm not opposed to it at all, my only preference would be to keep it as "dumb" as possible if we can.

maniac103:
> I'm using a Docker container with the latest snapshot build and connect it to the official openhab-cloud for testing.

I'm talking about app side changes - how do you test those?

maniac103:
> The reference-id should be in the data/payload returned on the api , so you could pretty easily filter to only the last message with that id when showing in the client.

OK, I'll make it do that.

mueller-ma:
> I think FCM will only send the messages to the app if it's signed with official keys.

No, you can use your debug keys as well and FCM works with that.

maniac103:
@mueller-ma While you're at picking things: https://github.com/maniac103/openhab.android/commit/d7f41fad1ef376ad874a51acc4502986e38a78c2

mobile/src/main/java/org/openhab/habdroid/core/NotificationHelper.kt
```diff
@@ -140,17 +147,38 @@ class NotificationHelper(private val context: Context) {
...
             .setContentIntent(contentIntent)
             .setDeleteIntent(deleteIntent)
             .setVisibility(NotificationCompat.VISIBILITY_PRIVATE)
             .setPublicVersion(publicVersion)
-            .build()
+
+        val messageImage = if (message.mediaAttachmentUrl != null) {
+            ConnectionFactory.waitForInitialization()
+            ConnectionFactory.primaryUsableConnection?.connection?.let {
+                message.loadImage(it, context, 99)
```
mueller-ma, line 164: @maniac103 I'm not sure about what size should be used here. Maybe the screen width?

maniac103:
> No, you can use your debug keys as well and FCM works with that.

Doesn't actually look like that's the case. I built the app with this diff:
```diff
diff --git a/mobile/src/full/java/org/openhab/habdroid/core/FcmRegistrationWorker.kt b/mobile/src/full/java/org/openhab/habdroid/core/FcmRegistrationWorker.kt
index e61b6d01..e0f92f58 100644
--- a/mobile/src/full/java/org/openhab/habdroid/core/FcmRegistrationWorker.kt
+++ b/mobile/src/full/java/org/openhab/habdroid/core/FcmRegistrationWorker.kt
@@ -131,6 +131,8 @@ class FcmRegistrationWorker(private val context: Context, params: WorkerParamete
                 connection.httpClient.get(registrationUrl).close()
             }
             Log.d(TAG, "FCM reg id success")
+        }.addOnFailureListener { e ->
+            Log.d(TAG, "Token failure", e)
         }
     }
 ```

which yielded this log:

```
07-10 17:37:00.438   995   995 D FcmRegistrationWorker: Token failure
07-10 17:37:00.438   995   995 D FcmRegistrationWorker: java.io.IOException: java.util.concurrent.ExecutionException: java.io.IOException: FIS_AUTH_ERROR
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.firebase.messaging.FirebaseMessaging.blockingGetToken(FirebaseMessaging.java:665)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.firebase.messaging.FirebaseMessaging.lambda$getToken$7$com-google-firebase-messaging-FirebaseMessaging(FirebaseMessaging.java:421)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.firebase.messaging.FirebaseMessaging$$ExternalSyntheticLambda11.run(D8$$SyntheticClass:0)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:462)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at java.util.concurrent.FutureTask.run(FutureTask.java:266)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:301)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.common.util.concurrent.zza.run(com.google.android.gms:play-services-basement@@18.3.0:2)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at java.lang.Thread.run(Thread.java:923)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker: Caused by: java.util.concurrent.ExecutionException: java.io.IOException: FIS_AUTH_ERROR
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.tasks.Tasks.zza(com.google.android.gms:play-services-tasks@@18.1.0:5)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.tasks.Tasks.await(com.google.android.gms:play-services-tasks@@18.1.0:9)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.firebase.messaging.FirebaseMessaging.blockingGetToken(FirebaseMessaging.java:663)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         ... 9 more
07-10 17:37:00.438   995   995 D FcmRegistrationWorker: Caused by: java.io.IOException: FIS_AUTH_ERROR
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.firebase.messaging.GmsRpc.handleResponse(GmsRpc.java:318)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.firebase.messaging.GmsRpc.lambda$extractResponseWhenComplete$0$com-google-firebase-messaging-GmsRpc(GmsRpc.java:329)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.firebase.messaging.GmsRpc$$ExternalSyntheticLambda0.then(D8$$SyntheticClass:0)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.tasks.zzc.run(com.google.android.gms:play-services-tasks@@18.1.0:3)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at androidx.media3.exoplayer.hls.offline.HlsDownloader$$ExternalSyntheticLambda0.execute(D8$$SyntheticClass:0)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.tasks.zzd.zzd(com.google.android.gms:play-services-tasks@@18.1.0:1)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.tasks.zzr.zzb(com.google.android.gms:play-services-tasks@@18.1.0:5)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.tasks.zzw.zzb(com.google.android.gms:play-services-tasks@@18.1.0:3)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.tasks.zzc.run(com.google.android.gms:play-services-tasks@@18.1.0:8)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.cloudmessaging.zzy.execute(Unknown Source:0)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.tasks.zzd.zzd(com.google.android.gms:play-services-tasks@@18.1.0:1)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.tasks.zzr.zzb(com.google.android.gms:play-services-tasks@@18.1.0:5)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.tasks.zzw.zzb(com.google.android.gms:play-services-tasks@@18.1.0:3)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.tasks.TaskCompletionSource.setResult(com.google.android.gms:play-services-tasks@@18.1.0:1)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.cloudmessaging.zzs.zzd(com.google.android.gms:play-services-cloud-messaging@@17.2.0:3)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.cloudmessaging.zzu.zza(com.google.android.gms:play-services-cloud-messaging@@17.2.0:2)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.google.android.gms.cloudmessaging.zzm.handleMessage(com.google.android.gms:play-services-cloud-messaging@@17.2.0:14)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at android.os.Handler.dispatchMessage(Handler.java:102)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at android.os.Looper.loop(Looper.java:223)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at android.app.ActivityThread.main(ActivityThread.java:7656)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at java.lang.reflect.Method.invoke(Native Method)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
07-10 17:37:00.438   995   995 D FcmRegistrationWorker:         at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
```

which to my eye strongly suggests Firebase isn't happy with my app signature? FWIW, this happens both in emulator as well as the device (Pixel 8) I have installed the app via Play Store on (in which I *do* get FCM messages).

**Comment by mueller-ma** [comment]
TODO: Notifications aren't replaced when having the same reference-id using NotificationPoller.


> which to my eye strongly suggests Firebase isn't happy with my app signature? FWIW, this happens both in emulator as well as the device (Pixel 8) I have installed the app via Play Store on (in which I do get FCM messages).

Strange. Works on my machine ™️ 
For now you can use the foss flavor as it shares some code with the full flavor.

**Comment by maniac103** [comment]
> TODO: Notifications aren't replaced when having the same reference-id using NotificationPoller.

See [this comment](https://github.com/openhab/openhab-android/pull/3724#issuecomment-2220807716) ;-)

**Comment by mueller-ma** [comment]
This PR is ready to merge from my side. Hiding notifications is tracked in #3686.

My debug cert has been added to the Firebase project. When you send me the  SHA-1 of your debug cert, I can add it to the app.

**Comment by maniac103** [comment]
> When you send me the SHA-1 of your debug cert, I can add it to the app.

Will do, thanks :+1: 

Should invoking a notification action hide the notification or not? I tend to say they should, but currently that's not the case.

**Comment by maniac103** [comment]
Some other thing I noticed: https://github.com/maniac103/openhab.android/commit/418b6f3fa77f02c59469b07de56658f6f6233d15

**Comment by mueller-ma** [comment]
> Should invoking a notification action hide the notification or not? I tend to say they should, but currently that's not the case.

Yes, probably it should hide the notification. 

> Do we actually need this endpoint? 

It's required that the app sends a "hide this notification" to other app installations of the same user.

**Comment by maniac103** [comment]
> Yes, probably it should hide the notification.

I've added a change for that to my repo, please pick.

> It's required that the app sends a "hide this notification" to other app installations of the same user.

Thanks, makes sense, didn't think of that.

**Comment by stefan-hoehn** [comment]
@mueller-ma 
@maniac103 

I am currently implementing the Blockly blocks for the notifications. Most of it works already (I'm on beta 3.15.0) but I ran into the following issue

- The notification was received
- But when I click on the notification it sends me to the notification screen which shows:

image

Ever since that with every notification that is received and I click on, this error screens pops up. Even closing the App doesn't fix it and it is therefore broken.

I guess I have seen some "wrong" notification which now has created a wrong notification history entry that breaks the App.

Any idea how we can debug that?

**Comment by maniac103** [comment]
What does the app log say? Additionally, what's the JS code which you used for creating that broken notification?

**Comment by stefan-hoehn** [comment]
I connected the mobile phone and locked into logcat with Android studio:

````
2024-07-14 15:09:40.499  5178-5178  CloudNotif...stFragment org.openhab.habdroid                 D  Notification response could not be parsed
     org.json.JSONException: No value for message
     	at org.json.JSONObject.get(JSONObject.java:398)
     	at org.json.JSONObject.getString(JSONObject.java:559)
     	at org.openhab.habdroid.model.CloudNotificationKt.toCloudNotification(Unknown Source:102)
     	at org.openhab.habdroid.ui.CloudNotificationListFragment$loadNotifications$1.invokeSuspend(Unknown Source:132)
     	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(Unknown Source:11)
     	at kotlinx.coroutines.DispatchedTask.run(Unknown Source:98)
     	at android.os.Handler.handleCallback(Handler.java:942)
     	at android.os.Handler.dispatchMessage(Handler.java:99)
     	at android.os.Looper.loopOnce(Looper.java:201)
     	at android.os.Looper.loop(Looper.java:288)
     	at android.app.ActivityThread.main(ActivityThread.java:7898)
     	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
     	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
````

Does that help?

How can I access the "applog"?

**Comment by maniac103** [comment]
> How can I access the "applog"?

At the bottom of the settings page ... but it's just logcat made accessible from the app, so what you posted is sufficient.

So the crash says you have a notification without message. How is that even possible?

**Comment by stefan-hoehn** [comment]
First of all, I have now built the project in Android Studio to be able to debug it directly.
To deploy it to my phone it told me that I had to remove the already installed version (which is the beta).
Now, unfortunately I don't get the error anymore :-(

> Looking at [the code](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java#L641) message is non-nullable and sent unconditionally?

I have an idea:

I implemented the hide notification as follows:

image

which creates the following code

> actions.notificationBuilder('').withReferenceId('refId111').hide().send();

As the message does not make sense in case of hiding a notification, I kept it empty, so maybe this causes the issue?

And guess what, as soon as I sent the above issue the error came up again! 

image

**Comment by maniac103** [comment]
I guess the actual bug/unexpected behavior is the cloud backend returning messages with type `hideNotification` when being asked for a list of notifications. `hideNotification` is expected to not have a message.
@digitaldan Is it expected for those messages to appear in the list returned by `api/V1/notifications`? If it is, we need to add some filtering code to both the list fragment and the notification poller.

**Comment by stefan-hoehn** [comment]
Additional information:

Let's say I send 

```
actions.notificationBuilder("hide notification").withReferenceId('refId113').withTag('info').withOnClickAction('command:AZ_Govee_Light_Strip_Power:OFF').hide().send();

```

Even IF I sent a message with the hide notification everything is omitted except 

> {"reference-id":"refId113","type":"hideNotification"}

So even the tag is missing here which is at least surprising.

Btw, I would't have expected to receive the action but it actually would say it makes sense even in a hide message to have it: for example a notification could turn on a light and hiding the notification could turn of that light...

maniac103:
The purpose of `hideNotification` is just to remove the notification from the device. This is why everything besides the ID is omitted: after all, there's no notification to show the additional content in. If there is, the client did something wrong ;-)

stefan-hoehn:
Keep in mind though, that tags are allowed

see https://www.openhab.org/addons/automation/jsscripting/#cloud-notification-actions

> .hide(): Hides notifications with the specified referenceId or tag.

maniac103:
Oh, right, indeed, tag should be kept.

digitaldan:
> @digitaldan Is it expected for those messages to appear in the list returned by api/V1/notifications? If it is, we need to add some filtering code to both the list fragment and the notification poller.

The cloud service logs all notification messages.  As of a few weeks ago, the API always returns the message payload for notifications, which in this case contains the `type: 'hideNotification'` as opposed to `'type: 'notification'` for normal messages, which i believe the android client is also using when the message is delivered live (vs the api) .

>actions.notificationBuilder("hide notification").withReferenceId('refId113').withTag('info').withOnClickAction('command:AZ_Govee_Light_Strip_Power:OFF').hide().send();


This is a helper function that @florian-h05  wrapped around the binding action for the JS library, but i believe if you use the 'hide' function there, its going to internally call `hideNotification` which will not include a message, title, commands, etc.... so those are being dropped in the helper library.

maniac103:
> The cloud service logs all notification messages. As of a few weeks ago, the API always returns the message payload for notifications, which in this case contains the type: 'hideNotification' as opposed to 'type: 'notification' for normal messages

OK, so this means the `hideNotification` message also gets written to the DB and thus gets its own `_id` like all other notification messages? That should be fine then and we should just filter it in the client.

> so those are being dropped in the helper library.

@stefan-hoehn Sounds like `tag` being dropped is worth a bug report in openhab-js then.

florian-h05:
Tag should not be kept. The JS Scripting docs are unclear at that point, but as the cloud connector add-on only provides methods to hide by tag and other methods to hide by reference id, they cannot be used together when hiding notifications.
The notification builder implementation in openhab-js prefers referenceId over tag when hiding notifications.

mueller-ma:
> If it is, we need to add some filtering code to both the list fragment and the notification poller.

Notification list is done here: https://github.com/openhab/openhab-android/pull/3750

mueller-ma:
One remaining issue: The notification list doesn't load any new notifications beyond the first 20. I guess the issue is caused by `CloudNotificationAdapter.addLoadedItems()` when it filters out items.

maniac103:
> The notification list doesn't load any new notifications beyond the first 20

I'm pretty sure I tried this in the scope of this PR, and it worked fine? The cloud only keeps the latest 50 notifications, are you sure you didn't run into that?

mueller-ma:
It didn't work in https://github.com/openhab/openhab-android/pull/3750 anymore, but that's fixed now.

PR #3731, opened 2024-07-02
Poll for notifications when in foreground

mueller-ma:


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -324,6 +325,18 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
...
         }
         handlePendingAction()
+        startNotificationPolling()
+    }
+
+    private fun startNotificationPolling() {
+        notificationPollingJob?.cancel()
+        if (CloudMessagingHelper.needsPollingForNotifications(this)) {
+            notificationPollingJob = launch {
+                CloudMessagingHelper.pollForNotifications(this@MainActivity)
```
maniac103, line 335: Polling every 30 seconds seems a little excessive, doesn't it? I'm not sure whether myopenhab will like the additional load.

mueller-ma: I increased it to 3 minutes. One reason I created this PR is, that the fetch is now done whenever the app is opened. As that's independent from the delay, a higher delay is also fine for me.

PR #3735, opened 2024-07-08
Update summary notification for any notificationId

mueller-ma:
There's no guarantee that the hash of the string id is greater than 0.


maniac103 (approved):

mueller-ma:
Done in #3724

PR #3738, opened 2024-07-11
Fix skeleton view breaks server discovery layout

### LINKED ISSUE #3736: Skeleton view breaks server discovery layout
### Actual behaviour
See screenshot:
Screenshot_20240710_173230

### Expected behaviour
Skeleton view not visible in that case, icon and text should be centered vertically

 
### Steps to reproduce
1. Clear app data
2. Restart app, ideally without server available in network

mueller-ma:
Closes #3736

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -523,7 +523,7 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
         when {
             newConnection != null -> {
                 handleConnectionChange()
-                controller.updateConnection(newConnection, null, 0)
+                controller.updateConnection(newConnection, null, 0, true)
             }
             failureReason is WrongWifiException -> {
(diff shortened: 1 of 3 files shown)


maniac103:
IMHO this makes things way too complicated and also doesn't provide the correct transitions. AFAICT we'd want
- normal: resolve connection (progress) -> load widgets (skeleton) -> done (widget list)
- special condition (no wifi or resolve server): resolve connection (progress) -> status (status)

What this PR (and/or previous code) are doing (AFAICT) is
- normal: resolve connection (skeleton) -> load widgets (skeleton) -> done (widget list) [good]
- special condition: resolve connection (skeleton) -> status (status) [bad]

Even if we want to show the skeleton while resolving the connection, we should IMO refactor the fragment logic:
- `ProgressFragment` doesn't show progress anymore -> remove
- instead, have a `WidgetListSkeletonFragment` shown while resolving connection and
- a `StatusFragment` consisting of text, icon, buttons, small progress spinner

The skeleton should IMO not part of any status fragment. Doing that refactoring will help in understanding the logic flow, and will prevent inherently prevent things like #3736.

PR #3767, opened 2024-07-18
Enable nonTransitiveRClass for better build perfomance

mueller-ma:


mueller-ma:
There isn't a real improvement in build time, but makes using the `R` class more complicated.

PR #3770, opened 2024-07-19
Fix stateFromLabel when label contains square brackets in pattern

### LINKED ISSUE #3756: Sitemaps Input widget strips off `[]` characters
### Actual behaviour

Item:
```
String TestString1
```

Sitemap:
```
sitemap test {
  Input item=TestString1
}
```

- Open that sitemap on android 3.15.0
- Enter `abcdefgh[]` into the text field, and click the checkmark (enter) on the Android's virtual keyboard
- The app will send a command `abcdefgh[]` to the item but then changed the text inside the input field to `abcdefgh` stripping away the `[]`
- The same thing would happen if `[]` is entered in the middle of the string, e.g. `abcdefg[]h` - > it strips off everything after `[]` so the text became `abcdefg` (note the `h` is also gone)
- But this "bug" doesn't happen, if `[]` is at the very beginning of the string, e.g. `[]abcdefgh` - the `[]` don't get stripped off, and neither do they get stripped off when I entered `[]abcdefgh[]`.

### Expected behaviour

It shouldn't strip off any text inputs

 
### Steps to reproduce
Described above

#### Client
* Android version: Android 14
* App version 3.15.0
* Build flavor Play Store
* Device language: English

#### Server
* Server version: 4.3.0-SNAPSHOT

jimtng:
Fix #3756

The problem was:
- StringItem's state contains `[]`
- Sitemap puts the item's (display) state into the label as `Label [formattedstate]`
- So now we have nested `[]` causing parsing errors when trying to extract the formattedstate from `[mytext with [] in it]`.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
@@ -73,7 +73,7 @@ data class Widget(
 ) : Parcelable {
     val label get() = rawLabel.split("[", "]")[0].trim()
     val stateFromLabel: String? get() {
-        val value = rawLabel.split("[", "]").getOrNull(1)?.trim()
+        val value = Widget.stateLabelRegex.find(rawLabel)?.groupValues?.getOrNull(1)?.trim()
         val optionLabel = mappingsOrItemOptions.find { it.value == value }?.label
         return optionLabel ?: value
     }
@@ -243,6 +243,8 @@ data class Widget(
                 state.toParsedState(item.state?.asNumber?.format)
             else -> state.toParsedState()
         }
+
+        internal val stateLabelRegex = Regex("\\[(.*)\\]$")
     }
 }
 
--- a/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
@@ -73,7 +73,7 @@ data class Widget(
 ) : Parcelable {
     val label get() = rawLabel.split("[", "]")[0].trim()
     val stateFromLabel: String? get() {
-        val value = Widget.stateLabelRegex.find(rawLabel)?.groupValues?.getOrNull(1)?.trim()
+        val value = Widget.stateLabelRegex.find(rawLabel)?.groupValues?.getOrNull(1)
         val optionLabel = mappingsOrItemOptions.find { it.value == value }?.label
         return optionLabel ?: value
     }
--- a/mobile/src/test/java/org/openhab/habdroid/model/WidgetTest.kt
+++ b/mobile/src/test/java/org/openhab/habdroid/model/WidgetTest.kt
@@ -46,7 +46,7 @@ class WidgetTest {
         assertEquals(sutXml.size, 2)
         assertEquals(sut1.size, 2)
         assertEquals(sut2.size, 1)
-        assertEquals(sut3.size, 4)
+        assertEquals(sut3.size, 5)
     }
 
     @Test
@@ -170,6 +170,7 @@ class WidgetTest {
         assertNull(sut1[0].stateFromLabel)
         assertNull(sut2[0].stateFromLabel)
         assertEquals("81 %", sut3[1].stateFromLabel)
+        assertEquals("Value [42]", sut3[4].stateFromLabel)
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


mueller-ma:
I think that deserves a test case in `mobile/src/test/java/org/openhab/habdroid/model/WidgetTest.kt`

jimtng:
> I think that deserves a test case in `mobile/src/test/java/org/openhab/habdroid/model/WidgetTest.kt`

It's a good idea, and whilst adding a few more test cases, I think we shouldn't even edit what's given by `stateFromLabel` because that could be wildly different to the actual item state, e.g. due to transformation.

Given this sitemap:

```
Sitemap test {
  Input item=StringItem1 label="Test [RB(| input == 'A' ? 'GOOD' : 'BAD'):%s]`
}
```

The item could contain state `A` or `B` the `stateFromLabel` will give us `GOOD` or `BAD`. When dealing with an input widget, we should be editing `A` / `B` (the actual state) not the label. So I'll push another change shortly.

jimtng:
@mherwege is it OK to remove this line?
https://github.com/openhab/openhab-android/blob/dc919bffa290a107c5741b05c0fab0a06a7e1638/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt#L703

See the comment right above this.

I'm just wondering under what circumstances would we need to edit the "formatted state" (ie. after it's being transformed) in an input widget, as opposed to editing the actual item state.

mherwege:
> I'm just wondering under what circumstances would we need to edit the "formatted state" (ie. after it's being transformed) in an input widget, as opposed to editing the actual item state.

QuantityTypes may not format properly through the line above, so you may end up with too many positions after the dot.

There is probably room to improve this, but you will have to test with all different data types, specifically QuantityType, DateTimeType (when not using any of the special formats for it).

I don't know the Android app codebase (and Kotlin) well, so difficult for me to say the exact consequence without a full test.

See some discussion on it here: https://github.com/openhab/openhab-android/pull/3337#issuecomment-1519083654
The decision there was to show the displayState, but expect the state as input, leaving the responsability to the user to create a proper state description. As transformations are defined in the same way as formatting state descriptions, it is a difficult thing to get right.

jimtng:
> @mherwege is it OK to remove this line?
> 
> https://github.com/openhab/openhab-android/blob/dc919bffa290a107c5741b05c0fab0a06a7e1638/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt#L703

I won't touch this in this PR. Indeed it's a bit complicated.

This PR is just about properly parsing the formatted state inside `[]`

However, if the label itself contains `[]` it would still mess this up. To truly fix that, the core would need to escape the characters `[]\` before sending that and mainui, basicui and mobile apps need to know to unescape it.

So do we want this PR merged, knowing that it helps, but not a full solution?

mherwege:
> However, if the label itself contains `[]` it would still mess this up.

I think that would be an issue in core and BasicUI as well. Core does some string parsing on this in some places if I remember well, so assumes the last [ ] pair contains the pattern and the rest goes in the label. Again, this is tricky in many ways and I spent a lot of time on this when implementing the input controls.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -243,6 +243,8 @@ data class Widget(
                 state.toParsedState(item.state?.asNumber?.format)
             else -> state.toParsedState()
         }
+
+        internal val stateLabelRegex = Regex("\\[(.*)\\]$")
```
mherwege, line 247: You could modify the regex to not match [ ] inside the [ ] pair. That would make sure you get the last pair for the pattern and the rest stays in the label. That assumes the pattern or transformation cannot contain [ or ].

jimtng: I'm not sure I understand. Are you suggesting something like `\\[[^\\[\\]]*\\]`? What if the "label" sent by the server is `Label [value that contains [] chars]` ?

mherwege: Indeed, that won’t work well either. There is always something.
FWIW in BasicUI, at initialization of the field, some analysis is done on the pattern (which is the content in the [ ]). Core has that available. I don’t remember if that is available in the Android app. It is not in the SSE events for sure. That allows avoiding the problem. The assumption is that there is just a formatting pattern in the square brackets, not a transformation. See the BasicUI fix PR that touches this code (https://github.com/openhab/openhab-webui/pull/2685). The java class is for the initialization, the js code works handles the updates and SSE events.
In core, a good starting point to get an idea of what is happening is in the getDisplayState method, creating the displayState for SSE.

jimtng: For the time being, we'll have to assume and expect that the "label" part must not contain `[]` characters. I had a look in core code, and it has the same assumption.

jimtng:
For the time being, we'll have to assume and expect that the "label" part must not contain `[]` characters. I had a look in [core](https://github.com/openhab/openhab-core/blob/e1574488b6f8815e55d8f9b0bff62c73b4532564/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java#L566), and it has the same assumption.

mueller-ma:
@jimtng FYI: when you add `Fix #3756` to the commit message of the first commit of a PR, I don't have to add it manually to the commit that goes in `main`.

PR #3782, opened 2024-07-27
Fix Buttongrid's layout changes when there is an invisible button

jimtng:


jimtng:
@mueller-ma can this still make it into 3.15?

mueller-ma:
Let's go for 3.16.

Issue #2804, opened 2022-01-21
Frame visibility in Basic Ui


wawa79:
Frames in sitemaps are not displayed in the app when a visibility parameter is set.
In the Web browser, frames with visibility parameter work properly.
OH3.2 app v2.19

maniac103:
Please provide an example sitemap file. What element is the visibility parameter set for, and what value is it set to?

wawa79:
> Please provide an example sitemap file. What element is the visibility parameter set for, and what value is it set to?

`.sitemap` for testing the issue:
```
sitemap test label="Test"{
	Frame label="Main frame" {
		Switch item=VISIBLE_item_switch
		Frame label="Visibility_test" visibility=[VISIBLE_item_switch==ON]{
			Switch item=TST_Switch
		}
	}
}

```
corresponding .`items` file:
```
Switch        VISIBLE_item_switch "Set frame visible"
Switch        TST_Switch		    "Test switch"

```

mueller-ma:
There was a fix for this kind of issue in 2.19.1-beta, so it's not in a stable version yet. Can you try the latest beta version?

wawa79:
Tested with 2.19.11: the issue is still there but I found that the visibility parameter is not taken into account for nested frames. As such, the following sitemap works properly contrarily to my prior example.

```
sitemap test label="Test"{
	Frame label="Main frame" {
		Switch item=VISIBLE_item_switch
	}
    Frame label="Visibility_test" visibility=[VISIBLE_item_switch==ON]{
        Switch item=TST_Switch
    }
}

```

Issue #420, opened 2017-10-20
Crash when chart refresh too low


mueller-ma:
I set the refresh to 10, no chart was shown and the app crashed after some time.
Version 1.8.0.8

lolodomo:
You are crazy to use such a small refresh frequency (10 ms) ;)
In Basic UI, we have considered a minimum value of 100 ms.
This should be considered in method https://github.com/openhab/openhab.android/blob/master/mobile/src/main/java/org/openhab/habdroid/util/MySmartImageView.java#L92
Minimal value should be adjusted after some testing. Maybe 100 ms is even too small.

PR #3785, opened 2024-07-28
Allow one level of nested frames in sitemaps

### LINKED ISSUE #324: Nested frames do not work
In the play store two people are complaining about nested frames are not working (frame in frame).

I could reproduce that.

jimtng:
image

```
sitemap nestedframe {
  Frame label="outer" {
    Frame label="inner1" {
      Text label="Text inside inner1"
    }
    Frame label="inner2" {
      Text label="Text inside inner2"
    }
  }
  Frame label="outer too" {
    Text label="Inside outer too"
  }
}
```

I learned that nested frames aren't "supposed" to be supported, but BasicUI and iOS renders them, whilst Android doesn't.

Here, I propose to render nested frames, and also provide a visual clue for the nesting structure by using `colorTertiaryContainer`, indenting the header text and reduce the header bg's height/padding, as suggested by @maniac103 

Resolve #324, #1639, #2804
Also [#420 ](https://github.com/openhab/openhab-webui/issues/420)

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/WidgetDataSource.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/WidgetDataSource.kt
@@ -41,13 +41,11 @@ class WidgetDataSource {
             .map { w -> w.id }
             .toSet()
         val secondLevelWidgetIds = allWidgets
-            .filter { w -> w.parentId in firstLevelWidgetIds }
+            .filter { w -> w.parentId in firstLevelWidgetIds && w.type in ALLOWED_SECOND_LEVEL_PARENTS }
             .map { w -> w.id }
             .toSet()
         return allWidgets.filter { w ->
-            w.parentId == null ||
-                w.parentId in firstLevelWidgetIds ||
-                w.parentId in secondLevelWidgetIds && w.type == Widget.Type.Button
+            w.parentId == null || w.parentId in (firstLevelWidgetIds + secondLevelWidgetIds)
         }
     }
 
@@ -70,5 +68,6 @@ class WidgetDataSource {
 
     companion object {
         private val TAG = WidgetDataSource::class.java.simpleName
+        private val ALLOWED_SECOND_LEVEL_PARENTS = setOf(Widget.Type.Buttongrid, Widget.Type.Frame)
     }
 }
--- a/mobile/src/main/java/org/openhab/habdroid/model/WidgetDataSource.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/WidgetDataSource.kt
@@ -45,7 +45,7 @@ class WidgetDataSource {
             .map { w -> w.id }
             .toSet()
         return allWidgets.filter { w ->
-            w.parentId == null || w.parentId in (firstLevelWidgetIds + secondLevelWidgetIds)
+            w.parentId == null || w.parentId in firstLevelWidgetIds || w.parentId in secondLevelWidgetIds
         }
     }
 
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -19,7 +19,6 @@ import android.content.Intent
 import android.content.SharedPreferences
 import android.content.res.ColorStateList
 import android.graphics.Color
-import android.graphics.drawable.GradientDrawable
 import android.text.InputType.TYPE_CLASS_NUMBER
 import android.text.InputType.TYPE_CLASS_TEXT
 import android.text.InputType.TYPE_NUMBER_FLAG_DECIMAL
@@ -53,8 +52,6 @@ import androidx.core.view.get
 import androidx.core.view.isGone
 import androidx.core.view.isInvisible
 import androidx.core.view.isVisible
-import androidx.core.view.marginStart
-import androidx.core.view.updatePadding
 import androidx.core.widget.ContentLoadingProgressBar
 import androidx.core.widget.doAfterTextChanged
 import androidx.fragment.app.DialogFragment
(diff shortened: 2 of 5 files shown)


maniac103:
Nice to see this implemented (even though nested frames still aren't fully supported, but only one level), but I think the visual representation can use some improvement. My suggestion would be using a separate view (and thus view holder) where
- the background still starts at the left edge
- the text is indented and
- the background is a bit more narrow (less padding)

mobile/src/main/java/org/openhab/habdroid/model/WidgetDataSource.kt
```diff
@@ -41,8 +41,13 @@ class WidgetDataSource() {
...
             .map { w -> w.id }
             .toSet()
-        return allWidgets
-            .filter { w -> w.parentId == null || w.parentId in firstLevelWidgetIds }
+        val secondLevelWidgetIds = allWidgets
+            .filter { w -> w.parentId in firstLevelWidgetIds && w.type in ALLOWED_SECOND_LEVEL_PARENTS }
+            .map { w -> w.id }
+            .toSet()
+        return allWidgets.filter { w ->
+            w.parentId == null || w.parentId in (firstLevelWidgetIds + secondLevelWidgetIds)
```
maniac103, line 49: I'd suggest two `in` checks instead of the concatenation to avoid the list reallocation.

jimtng: changed as suggested

jimtng:
> even though nested frames still aren't fully supported, but only one level

It's a tricky one. If we support unlimited nesting levels, we'd need to have a way to provide visual clues about the nesting levels. Simply indenting the "header" is going to look weird when the "content" of the frames will remain at the same left indentation. So to do that, we'd need to indent all the children too, which requires a bit of a restructure to the way Frames are currently done.

maniac103:
Just to be clear: I don't *want* to support unlimited nesting, precisely for the reasons you outlined: we'll run out of screen real estate pretty fast. Personally I'm not even convinced that two level nesting is a good idea, but I know there's quite a few people thinking otherwise.

jimtng:
> I think the visual representation can use some improvement. My suggestion would be using a separate view (and thus view holder) where
> 
> * the background still starts at the left edge
> * the text is indented and
> * the background is a bit more narrow (less padding)

I tried this
image

However, I think the difference is a bit too subtle. How can we improve it?

maniac103:
Different colors maybe? `colorPrimaryContainer` vs. `colorSecondaryContainer` or `colorTertiaryContainer`?

jimtng:
> Different colors maybe? `colorPrimaryContainer` vs. `colorSecondaryContainer` or `colorTertiaryContainer`?

Tried this. Only BasicUI Color Scheme has those two different. Other color schemes use the same color for them, and afaik `colorTertiaryContainer` doesn't exist.

image

mueller-ma:
How does it look with dynamic colors?

maniac103:
> and afaik colorTertiaryContainer doesn't exist.

It does, see [here](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/color/res/values/attrs.xml#L65)

jimtng:
Ahh, thanks! tertiary looks better, although we might need to tweak it? I totally suck at choosing colours though.

image

image

Basic UI

colorSecondaryContainer

image

Dynamic

image

mueller-ma (approved):
I can merge when you resolve the conflict caused by merging the other PR.

jimtng:
Rebased to main. Note the change from secondary to tertiary container color was also added.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -261,7 +265,8 @@ class WidgetAdapter(
             colorMapper,
             serverFlags,
             chartTheme,
-            { widgetsByParentId[widget.id] }
+            { widgetsByParentId[widget.id] },
```
jimtng, line 268: I'm still new to Kotlin and I don't understand why this one needs to be enclosed with braces, vs the one I've just added below doesn't (seem to) need braces. The line above was suggested by @maniac103

maniac103: This one is a function, while the one below isn't. It was just meant as an optimization so the list didn't need to be copied prematurely.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -610,6 +621,16 @@ class WidgetAdapter(
...
             labelView.isGone = widget.label.isEmpty()
+            val isNested = parentWidget?.type == Widget.Type.Frame
+
+            val paddingTop = if (isNested) 0 else originalPaddingTop
+            containerView.updatePadding(top = paddingTop, bottom = paddingTop)
+
+            backgroundColorMap[isNested]?.let { (containerView.background as GradientDrawable).setColor(it) }
+
+            labelView.layoutParams = (labelView.layoutParams as ViewGroup.MarginLayoutParams).apply {
+                marginStart = if (isNested) labelView.resources.dpToPixel(8f).toInt() else 0
```
maniac103, line 632: I don't know about @mueller-ma's stance on this, but personally I don't like manually messing with layout properties like this. I think having separate layouts for both type of frames, a common base class bundling the functionality and separate derived classes only differing in the constructor parameters (layout IDs) would be a bit more effort, but also be more maintainable in the long run.

jimtng: I agree

maniac103: I can have a look into the suggested implementation.

maniac103: @jimtng If you want to consider picking: https://github.com/maniac103/openhab.android/commit/b51e2a1899337c9c904bb5ad3482c69748761f6f

I also shrinked font size of secondary frames there.

Normal:
normal

Compact:
compact

maniac103: I just noticed I forgot the text indention in the compact layout. Fixed in https://github.com/maniac103/openhab.android/commit/59da677e95ee92521c3a0bcb203e13a10cd127de

jimtng: @maniac103 thanks!

maniac103 (approved):

Issue #3786, opened 2024-07-29
Icons misplaced


felixschndr:
Disclaimer: There are some open issues about icons but I am not sure weather any of these covers the following issue. If so please close this one.

### Actual behaviour
Since a recent version of the android app the icons seem to be misplaced sometimes:
![Screenshot_20240721_122802_openHAB](https://github.com/user-attachments/assets/6073ccce-20c8-4c60-a9ce-4459f939b096)
![Screenshot_20240721_122807_openHAB](https://github.com/user-attachments/assets/c7d6955f-7929-44ce-b3b5-4f6d8d958a88)
After scrolling in the window or reloading the icon seem to be placed where they belong.

Clearing the cache does not help.

In the following image I noticed that an icon got a background it should not have:
![Screenshot_20240721_122759_openHAB](https://github.com/user-attachments/assets/3722ef6d-3d37-4463-85a0-0e9f229516cb)
The original image has no white background.

On iOS the icons work as expected:

![IMG_0019](https://github.com/user-attachments/assets/278434a8-7469-43f7-9f8d-19b96b80ceae)

It does not matter whether the connection is local or via the cloud.

### Expected behaviour
The icons should all be aligned the same way.

 
### Steps to reproduce
See code below

#### Client
* Android version: 14
* Device model: Samsung Galaxy S22 Ultra
* App version 3.15.5 (I _think_ it happens since 3.15.0)
* Build flavor :  Full
* Device language: German

#### Server
* Server version: 4.2.0
* Reverse Proxy: Nginx
* Authentication method : None

#### App log
<details>
  <summary>Click to expand</summary>

[---------------.txt](https://github.com/user-attachments/files/16411933/---------------.txt)

</details>

Code for first image:
```txt
        Frame label="Saugen" {
            Switch item=Staubsauger_System_Staubsauger_Steuerung
                mappings=["vacuum"="Start", "pause"="Pause", "dock"="Dock"]
                visibility=[Staubsauger_System_Staubsauger_Status != "Charging"]
            Text item=Staubsauger_System_Staubsauger_Status
                visibility=[!= "Charging"]
            Selection item=Staubsauger_System_Staubsauger_Saugleistung
                mappings=[101="Leise", 102="Normal", 103="Turbo", 104="Maximal"]
            Text item=Staubsauger_System_Raum_Gruppe
                visibility=[Staubsauger_System_Staubsauger_Status == "Charging"]
            {
                Switch item=Staubsauger_System_Raum_16
[...]

PR #3787, opened 2024-07-29
Fix icon placement for section switches

### LINKED ISSUE #3773: 'Switch with mappings' icon misplaced
See screenshot from demo server:
Screenshot_20240719-191043

maniac103:
The way View.createSkeleton() works doesn't play nicely with ConstraintLayout as it causes constraints to be lost, thus insert a wrapper view to keep the constraints intact.

Fixes #3773


maniac103:
This is due to https://github.com/Faltenreich/SkeletonLayout/issues/47 ... I wonder whether we shouldn't think about a more generic solution though.

mueller-ma:
I had to create the skeleton manually in https://github.com/openhab/openhab-android/blob/main/mobile/src/main/res/layout/activity_chart.xml#L18-L22. 
WidgetImageView will check that first before creating the skeleton: https://github.com/openhab/openhab-android/blob/main/mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt#L80-L85

maniac103:
> WidgetImageView will check that first before creating the skeleton: https://github.com/openhab/openhab-android/blob/main/mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt#L80-L85

Yes, but that solution is weird as well (views usually don't control their parent), and doesn't help with the fact that automagic skeleton creation is fundamentally broken (see https://github.com/openhab/openhab-android/issues/3786#issuecomment-2256010924 and https://github.com/Faltenreich/SkeletonLayout/issues/47). As far as I can tell, there are 3 possible solutions:

- Add `SkeletonLayout` as parent to all `WidgetImageView`s, keep code as-is
- Make `WidgetImageView` derive from `SkeletonLayout` and make it have an internal `ImageView`
- Import the `SkeletonMask` related code into the app (it's not a large amount of code) and make `WidgetImageView` use it directly instead of using `SkeletonLayout`

I don't really like the first option (it's a lot of effort since we have a large number of layouts, and I still find it's somewhat of a hack as mentioned above), but I'm torn between the second and the third one currently.

maniac103:
Changed the approach and implemented option 2.

mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
```diff
@@ -45,377 +45,389 @@ import org.openhab.habdroid.util.getPrefs
...
-                setImageScalingType(ImageScalingType.entries[imageScalingType])
-            }
-            recycle()
-        }
-        // In some cases it's required to add the skeleton manually to XML.
-        // In these cases, set the parent as skeleton here, otherwise create it in applyProgressDrawable()
-        val parent = parent
-        if (parent is SkeletonLayout) {
-            skeleton = parent
-        }
```
maniac103, line 85: BTW, @mueller-ma, not sure whether you're aware of the syntax: `skeleton = parent as? SkeletonLayout` would have been shorter ;-)

mobile/src/main/java/org/openhab/habdroid/ui/widget/WidgetImageView.kt
```diff
@@ -426,5 +438,29 @@ class WidgetImageView(context: Context, attrs: AttributeSet?) : AppCompatImageVi
...
     companion object {
         private val TAG = WidgetImageView::class.java.simpleName
+        private fun createConfig(context: Context) = SkeletonConfig.default(context).apply {
+            maskColor = context.resolveThemedColor(R.attr.skeletonBackground, maskColor)
+            shimmerColor = context.resolveThemedColor(R.attr.skeletonShimmer, shimmerColor)
+        }
     }
 }
+
+internal interface WidgetImageViewIntf {
```
mueller-ma, line 448: What does the `Intf` mean?

maniac103: Interface. I can also spell it out if you prefer.

mueller-ma: Ok, I'm fine with that.

PR #3796, opened 2024-08-12
Increase memory allocation of Gradle daemon

maniac103:
With default values (512 MB) the daemon OOMs when building the app.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Configure Gradle
         run: |
           mkdir -p ~/.gradle/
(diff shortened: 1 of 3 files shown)


jimtng:
I've been keeping a stash on my system that sets this to 4g (2g would probably be enough), so this would certainly be very helpful to not have to keep applying my stash every time I checked something out. How have you been coping with this in the past?

maniac103:
> How have you been coping with this in the past?

The same way as you, hence the PR ;-)

The CI even [uses 8GB](https://github.com/openhab/openhab-android/blob/main/.github/workflows/deploy-apk.yml#L57), but that probably is overkill. Thinking about it, I maybe should adapt the CI scripts as well?

mueller-ma:
I wasn't aware that I set that in my personal gradle settings:

````
$ cat ~/.gradle/gradle.properties
org.gradle.jvmargs=-Xmx8192M
````

@maniac103 Can you remove the setting from CI completely? It should use the value from the project then.

maniac103:
> Can you remove the setting from CI completely? It should use the value from the project then.

Done.

PR #3799, opened 2024-08-14
Fix tools:src not working for WidgetImageView

maniac103:
This broke when WidgetImageView was changed to no longer be derived from ImageView in #3787. Since WidgetImageView passes all view attributes to its internal ImageView instance, let the framework know that this is the case.


mueller-ma:
It doesn't seem to work for me. The `tools:src` is still not displayed:

grafik
grafik

I cleaned the project and restarted Android Studio, but the issue persists.

maniac103:
Hmm, I tried one layout (not such anymore which specifically) and the icon was displayed there. I'll have another look.

BTW, solution comes from [here](https://stackoverflow.com/questions/43542758/androidsrc-attrs-for-custom-view)

mueller-ma:
Can you check it again which layout it was? I tried all and none displayed the `tools:src`. Maybe it's a caching bug on my machine.

maniac103:
Now it should work. Root cause was an NPE while constructing the view in edit mode.

PR #3817, opened 2024-09-17
Refactor connection handling for API 31+

maniac103:
Use non-deprecated functions, and restart connection check in `MainActivity.onStart()`.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
@@ -320,6 +320,10 @@ class MainActivity :
 
         updateDrawerServerEntries()
         onActiveConnectionChanged()
+        // Make sure the connection to be used is up-to-date. There can be scenarios where the current connection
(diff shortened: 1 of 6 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -318,6 +318,10 @@ class MainActivity : AbstractBaseActivity(), ConnectionFactory.UpdateListener {
 
         updateDrawerServerEntries()
         onActiveConnectionChanged()
+        // Make sure the connection to be used is up-to-data. There can be scenarios where the current connection
```
maniac103, line 321: Typo: up-to-date

maniac103:
@mueller-ma I think this should be ready for review.

mueller-ma:
The CI jobs time out after 6h.

maniac103:
> The CI jobs time out after 6h.

Thanks, I wasn't aware of this being related. Should be fixed now.

PR #3856, opened 2025-02-04
Disable controls for read-only items

mueller-ma:
See https://github.com/openhab/openhab-core/issues/4559

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
@@ -664,11 +664,15 @@ class WidgetAdapter(
             } else {
                 null
             }
+            switch.isEnabled = !widget.readOnly
 
             isBinding = false
         }
 
         override fun handleRowClick() {
+            if (boundWidget?.readOnly == true) {
+                return
+            }
             switch.toggle()
         }
     }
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -1916,6 +1949,10 @@ fun View.adjustForWidgetHeight(widget: Widget, fallbackRowCount: Int) {
     }
 }
 
+fun View.setReadOnly(widget: Widget) {
```
maniac103, line 1952: How about, instead of this extension, giving `Widget` a property?
`val isReadOnly get() = item?.readOnly == true`

That property can be used everywhere, unlike the extension method.

mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt
```diff
@@ -1951,10 +1988,12 @@ fun MaterialButton.setTextAndIcon(
     connection: Connection,
     label: String,
     iconRes: IconResource?,
+    readOnly: Boolean?,
     labelColor: String? = null,
     iconColor: String? = null,
     mapper: WidgetAdapter.ColorMapper? = null
 ) {
+    isEnabled = !(readOnly ?: false)
```
maniac103, line 1996: `isEnabled = readOnly != true` would be more understandable, I think.

lolodomo:
What rendering option did you finally choose ?

mueller-ma:
Mostly the components are grayed out. You can check it here: https://m3.material.io/components/sliders/specs#0ead3403-84ad-4d39-abb6-46400f9515a8 Note that the numbers are wrong there. Number 2 is disabled, not hover.

lolodomo:
And you applied that to switch only?

mueller-ma:
It has been applied to Switch, Input, Buttongrid, Slider, Setpoint, Color and Colortemperature.

PR #3875, opened 2025-03-17
Bump workmanager and fix NPE

mueller-ma:
https://github.com/openhab/openhab-android/pull/3871

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
@@ -410,6 +410,9 @@ fun Data.Builder.putValueWithInfo(key: String, value: ItemUpdateWorker.ValueWith
 }
 
 fun Data.getValueWithInfo(key: String): ItemUpdateWorker.ValueWithInfo? {
-    val array = getStringArray(key) ?: return null
-    return ItemUpdateWorker.ValueWithInfo(array[0], array[1], ItemUpdateWorker.ValueType.valueOf(array[2]))
+    @Suppress("UNCHECKED_CAST")
+    val array = this.keyValueMap[key] as? Array<String?> ?: return null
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/openhab/habdroid/background/ItemUpdateWorker.kt
```diff
@@ -410,6 +410,7 @@ fun Data.Builder.putValueWithInfo(key: String, value: ItemUpdateWorker.ValueWith
 }
 
 fun Data.getValueWithInfo(key: String): ItemUpdateWorker.ValueWithInfo? {
-    val array = getStringArray(key) ?: return null
-    return ItemUpdateWorker.ValueWithInfo(array[0], array[1], ItemUpdateWorker.ValueType.valueOf(array[2]))
+    @Suppress("UNCHECKED_CAST")
+    val array = this.keyValueMap[key] as Array<String?> ?: return null
+    return ItemUpdateWorker.ValueWithInfo(array[0]!!, array[1], ItemUpdateWorker.ValueType.valueOf(array[2]!!))
```
mueller-ma, line 415: @maniac103 The issue was, that the `Array<String>` had one element that was `null`, thus making it a `Array<String?>`. Not sure how to do all the casting nicely.

maniac103: Generally I try to avoid the `!!` operator as long as possible. Maybe something like this?

```
val array = this.keyValueMap[key] as? Array<String?> ?: return null
val value = array[0] ?: return null
val type = array[2] ?: return null
return ItemUpdateWorker.ValueWithInfo(value, array[1], ItemUpdateWorker.ValueType.valueOf(type))
```

One could also do `val value = requireNotNull(array[0])`, which has the same effect as the `!!` operator, being just a bit more explicit.

PR #3881, opened 2025-04-04
Avoid calling runBlocking from main thread

maniac103:
This can lead to ANRs, like the current myOH outage shows.


mobile/src/full/res/xml/automotive_app_desc.xml
mueller-ma, line 1: Shouldn't be here

mueller-ma (changes requested):
ktlint fails, but otherwise the PR looks good to me.

mobile/src/full/res/xml/automotive_app_desc.xml
```diff
@@ -0,0 +1,4 @@
+<automotiveApp>
```
maniac103, line 1: Indeed, I guess I accidentally `git add`ed this leftover from a previous experiment. I'll remove it.

PR #3889, opened 2025-04-20
Add interactive chart activity

maniac103:
This one uses the persistence REST API and replaces the one showing a larger server-side generated chart image if supported (that is, for OH >= 2). It uses client-side chart rendering, the advantage of which is the possibility of zooming into the chart data and getting tooltips for values.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/ui/ChartWidgetActivity.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/ui/ChartWidgetActivity.kt
@@ -220,6 +220,10 @@ class ChartWidgetActivity : AbstractBaseActivity() {
                 }
             }
             return@launch
+        } catch (e: StateParsingException) {
+            Log.w(TAG, "Could not load chart data", e)
+            showError(getString(R.string.chart_error_state_format, e.item.label ?: e.item.name), "")
+            return@launch
         }
 
         configureChartForData(data)
@@ -241,13 +245,14 @@ class ChartWidgetActivity : AbstractBaseActivity() {
         chart.isVisible = true
     }
 
-    private fun showError(message: CharSequence, retryButtonText: CharSequence, retryAction: () -> Unit) {
+    private fun showError(message: CharSequence, retryButtonText: CharSequence, retryAction: (() -> Unit)? = null) {
         chart.isVisible = false
         progressContainer.isVisible = false
         errorContainer.isVisible = true
         errorText.text = message
+        retryButton.isVisible = retryAction != null
         retryButton.text = retryButtonText
-        retryButton.setOnClickListener { retryAction() }
+        retryAction?.let { retryButton.setOnClickListener { it() } }
     }
 
     private fun goToChartImageActivity() {
@@ -262,7 +267,7 @@ class ChartWidgetActivity : AbstractBaseActivity() {
         onRefresh()
     }
 
-    @Throws(HttpClient.HttpException::class)
+    @Throws(HttpClient.HttpException::class, StateParsingException::class)
     private suspend fun loadData(
         connection: Connection,
         item: Item,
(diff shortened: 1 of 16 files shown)


mobile/src/main/java/org/openhab/habdroid/ui/ChartWidgetActivity.kt
```diff
@@ -104,105 +132,293 @@ class ChartWidgetActivity :
...
+    private suspend fun loadData(
+        connection: Connection,
+        item: Item,
+        period: Duration,
+        serviceId: String?,
+        progressCb: (itemName: String) -> Unit
+    ): ChartData {
+        val itemsForChart = if (item.type == Item.Type.Group) {
+            // sitemap response doesn't include member list, so re-fetch item to get members
+            ItemClient.loadItem(connection, item.name)?.members ?: listOf(item)
```
mueller-ma, line 211: This may throw `HttpClient.HttpException` and must be caught here.

mobile/src/main/java/org/openhab/habdroid/ui/ChartWidgetActivity.kt
```diff
@@ -104,105 +132,293 @@ class ChartWidgetActivity :
...
+    private suspend fun loadSeriesForItem(
+        connection: Connection,
+        item: Item,
+        startTime: Instant,
+        serviceId: String?
+    ): Series {
+        val uriBuilder = Uri.Builder()
+            .path("rest/persistence/items/${item.name}")
+            .appendQueryParameter("startTime", startTime.toEpochMilliseconds().toString())
+        serviceId?.let { uriBuilder.appendQueryParameter("serviceId", it) }
```
mueller-ma, line 334: `serviceId` might be an empty string and must not be added in that case.

mueller-ma (commented):
AFAIK the persistence api is only accessible for authenticated users, i.e. not when using this setting on the server and no username/password on the client:

grafik

Maybe add some help text for this case or fallback to image charts? Show a snackbar when falling back to image charts with a hint for interactive charts?

maniac103:
Note to self: should get default period from widget.

maniac103:
> Maybe add some help text for this case or fallback to image charts? Show a snackbar when falling back to image charts with a hint for interactive charts?

I guess I should add some text with a retry button as error handling in any case. For 401/403 the text could be adjusted and the button could lead to the image activity. WDYT?

mueller-ma:
> For 401/403 the text could be adjusted and the button could lead to the image activity. WDYT?

Sounds good.

Is it possible to load new data when swiping to the left in the chart until the start of the selected period? If not maybe load twice the period by default so you can change the period by zooming out?

maniac103:
> Is it possible to load new data when swiping to the left in the chart until the start of the selected period?

No.

> If not maybe load twice the period by default so you can change the period by zooming out?

I'm not entirely sure I'd like to add magic like that. Depending on change/update rate, the number of data points can be significant, which then causes both significant load time and data transfer. My chart with solar data includes 5 items that update every 5 seconds, so that's 5 * 24 * (3600 / 5) = 86400 data points per day. Loading that chart takes some seconds already; I don't want to needlessly double that time.
(BTW, that solar chart is the reason for both the progress message [that one won't notice/see with normal charts and short chart durations] and the dynamic limitation of available time spans. The chart library starts to perform poorly with too many data points.)

mobile/src/main/java/org/openhab/habdroid/ui/ChartWidgetActivity.kt
```diff
@@ -104,104 +146,368 @@ class ChartWidgetActivity :
...
-        showLegend = savedInstanceState.getBoolean(SHOW_LEGEND)
-        period = savedInstanceState.getString(PERIOD)!!
-        super.onRestoreInstanceState(savedInstanceState)
+    override fun onStart() {
+        super.onStart()
+        val data = loadedChartData
+        val dataIsOutdated = data != null &&
+            widget.refresh > 0 &&
+            Duration.between(data.timestamp, Instant.now()).toMillis() > widget.refresh
+        if (data == null || dataIsOutdated) {
```
maniac103, line 181: I wonder if I should consider data usage policy here?

mueller-ma:
> My chart with solar data includes 5 items that update every 5 seconds, so that's 5 * 24 * (3600 / 5) = 86400 data points per day. Loading that chart takes some seconds already; I don't want to needlessly double that time.

Ok, in that case I wouldn't add this logic.

mobile/src/main/java/org/openhab/habdroid/ui/ChartWidgetActivity.kt
mueller-ma: Ok, than I'd avoid refreshing when data saver is enabled.

PR #3901, opened 2025-05-07
Remove some non-determinism

licaon-kter:
Your app is not built reproducible in F-Droid, but we continuously test older versions on https://verification.f-droid.org/ and would like more and more apps to become repro

Looking at the latest app report:  https://verification.f-droid.org/packages/org.openhab.habdroid.beta/

This PR solves one part, the other is in https://github.com/openhab/openhab-android/blob/3.17.7-beta-fdroid/mobile/build.gradle#L57

could you make TIMESTAMP be based on the last commit date instead? eg. `git show --no-patch --format=%ct`

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -126,6 +126,7 @@ android {
 
 aboutLibraries {
     duplicationMode = DuplicateMode.MERGE
+    excludeFields = ['generated']
 }
 
 unMock {


mueller-ma:
I implemented using SOURCE_DATE_EPOCH in version https://github.com/openhab/openhab-android/releases/tag/3.17.11-beta. Can you check if the build is reproducible now?

licaon-kter:
make a quick local build and compared it with the last CI APK in https://gitlab.com/fdroid/checkupdates-bot-fdroiddata/-/jobs/10550694489 and looks fine :tada:

Issue #3747, opened 2024-07-13
Identical icons shown differently


dilyanpalauzov:
With 3.14.3 I see:

![Screenshot_20240713_175944](https://github.com/user-attachments/assets/5457374a-970f-41dc-9bab-952009eaca89)

Both items have the same icon, but OH-Android show them in different size.  If I pull down to refresh, the big icon gets smaller and then both icons have the same size.

mueller-ma:
Can you open app settings and select "Clear cache"?

dilyanpalauzov:
Clearing the cache does not make difference long term.

These are the items I have:

Number:Temperature termo1 "Хол температура [%.1f °C]" <tempair> ["Temperature", "Measurement"] {channel="mqtt:topic:termo1:termo", expire="10m,NULL"}
Number:Temperature termo2 "Горе температура [%.1f °C]" <tempair> ["Temperature", "Measurement"] {channel="mqtt:topic:termo2:termo", expire="10m,NULL"}

in Sitemap:
 Text item=termo1 visibility=[dht221_online==OPEN]
 Text item=termo2 visibility=[dht222_online==OPEN]

Sometimes both icons are shown large, sometimes both icons are shown small and sometimes one icon is large, and the other small.

tempair is a dynamic icon:

pi@raspberrypi:/etc/openhab $ ls $(find|grep tempair) -l
lrwxrwxrwx 1 openhab openhab 20  8 юни 2023 ./icons/classic/tempair-0.png -> temperature_cold.png
lrwxrwxrwx 1 openhab openhab 15  8 юни 2023 ./icons/classic/tempair-19.png -> temperature.png
lrwxrwxrwx 1 openhab openhab 19  8 юни 2023 ./icons/classic/tempair-27.png -> temperature_hot.png
lrwxrwxrwx 1 openhab openhab 15  8 юни 2023 ./icons/classic/tempair.png -> temperature.png

In all cases tempair-19 is utilized, as it is between 19 and 27 °C, so the same fil temperature.png is used.

dilyanpalauzov:
![Screenshot_20240718_113914.jpg](https://github.com/user-attachments/assets/56668b37-409d-4e84-8c29-06b17dc78251)

![Screenshot_20240718_113939.jpg](https://github.com/user-attachments/assets/e659f8f4-99d8-49c9-9116-9ddaabbfddee)

[...]

PR #3906, opened 2025-05-31
Fix icons sometimes not being loaded correctly

maniac103:
When scrolling through a widget list, an icon WidgetImageView sometimes goes through a lifecycle like this
1 onAttachedToWindow
2 onLayout
3 onDetachedFromWindow
4 onAttachedToWindow
5 onLayout

If step 2 saw a pending load job and launched a coroutine to execute it, it could happen that step 3 canceled the coroutine scope before the coroutine was executed. Make sure not to forget about the pending request, so it's reexecuted in step 5.


maniac103:
Looks like this still doesn't suffice. Icon loading is better with it, but still I sometimes get permanent loading indication (skeleton).
I'll investigate more.

mueller-ma:
> I'll investigate more.

Do you want to get this merged anyway?

maniac103:
It probably makes sense, yes. It's certainly better now, and I also believe the change itself is correct.

dilyanpalauzov:
Is this the same as https://github.com/openhab/openhab-android/issues/3747 for sitemaps?

dilyanpalauzov:
Can cancelling the coroutine⇔HTTP request in OH-Android lead to this error on the server: https://github.com/openhab/openhab-core/issues/4792?  I do see it pretty often.

PR #3920, opened 2025-07-18
Support new chart parameter: interpolation

holgerfriedrich:
Support new core feature openhab/openhab-core#4610

Partly fixes openhab/openhab-webui#3277

This is totally untested, I don't have proper dev setup at hand.

### DIFF
--- a/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
@@ -65,6 +65,7 @@ data class Widget(
     val legend: Boolean?,
     val forceAsItem: Boolean,
     val yAxisDecimalPattern: String?,
+    val interpolation: String?,
     val switchSupport: Boolean,
     val releaseOnly: Boolean?,
     val height: Int,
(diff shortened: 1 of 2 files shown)


mueller-ma (changes requested):
So `interpolation` is a new Sitemap chart parameter that just needs to be added to the URL when getting charts from the server? It case this change would be fine, however since the last release of the app, @maniac103 implemented client-side charts, which should also use this new parameter.

holgerfriedrich:
Yes, exactly, this is a new parameter to be appended to the URL.
I have done it similar to the `yAxisDecimalPattern` option, both described here in [docs](https://next.openhab.org/docs/ui/sitemaps.html#element-type-chart).
I thought I have added `interpolation` to all places, where `yAxisDecimalPattern` was used.


I don't know about then new client-side rendering. Maybe @maniac103 could give me some guidance.

maniac103:
> It seems that the client side chart library doesn't support interpolation

It does: there's `LineDataSet.setMode()`, which allows selecting `LINEAR` or `STEPPED`.

mobile/src/main/java/org/openhab/habdroid/model/Widget.kt
```diff
@@ -65,6 +65,7 @@ data class Widget(
     val legend: Boolean?,
     val forceAsItem: Boolean,
     val yAxisDecimalPattern: String?,
+    val interpolation: String?,
```
maniac103, line 68: I'd rather map this to an enumeration.

holgerfriedrich: Done

holgerfriedrich:
@mueller-ma Should we split this up and fix the server side charts first?

@maniac103  I don't see how I could implement the client-side charts. Would you be willing to guide me, or could you take over the implementation? We still have a few showstoppers for the OH5.0 release to resolve.

PR #3942, opened 2025-10-27
Ignore alarms from im.molly.app

sideshowbob84:
Molly is a relativley famous fork of Signal (org.thoughtcrime.securesms). It would be great, if Molly would not affect the alarms send openhab, too.


mueller-ma:
Don't worry, you didn't do anything wrong. The conflict was caused by merging https://github.com/openhab/openhab-android/pull/3938 that changed the same file as this PR.

PR #3984, opened 2026-04-24
Clear existing OSM markers before binding

### LINKED ISSUE #3978: Mapview can show two markers instead of one
With openHAB-Android 3.19.3-beta a `Mapview` element shows from time to time two markers.

This happens now and then, here is one way to reproduce:

I load a sitemap with Mapview object.  Then *☰* → *About*.  After a while the Location has probably changed.  When I close the *About* menu, the Mapview element is shown with two markers.

maniac103:
Makes sure markers of previous binding do not accumulate.

Fixes #3978

### DIFF
--- a/mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.kt
+++ b/mobile/src/foss/java/org/openhab/habdroid/ui/MapViewHelper.kt
@@ -83,6 +83,9 @@ object MapViewHelper {
         override fun bindAfterDataSaverCheck(widget: Widget) {
             super.bindAfterDataSaverCheck(widget)
             handler.post {
+                binding.mapview.overlays
+                    .filter { o -> o is Marker }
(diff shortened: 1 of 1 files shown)


maniac103:
@dilyanpalauzov Would you be able to build and test this?

dilyanpalauzov:
I cannot build.

maniac103:
I can build an APK, but it won't update an official version, so you'll need to uninstall that one before installing the APK. Is this OK for you nevertheless?

dilyanpalauzov:
I could install that build, but I cannot actually test.  The tracked item does not move currently, it might or might not move after a month. So I will have to send POST requests to openHAB to update the state of the `Location` item.  I will have to do it several times, and if it shows the old behaviour - then the problem is not fixed.  If it does not show the old behaviour, then maybe the problem is corrected, or maybe I am testing in a wrong way. 

For these reasons I suggest to include the change in the next regular release, without testing from my side.

PR #3989, opened 2026-05-05
Fix FCM registration not happening

maniac103:
The code to trigger this got lost during connection factory refactoring (#3972).


mueller-ma:
Currently I don't have a test setup for FCM, so feel free to merge this when you're confident with it.

PR #3993, opened 2026-05-07
Implement support for Android Auto

### LINKED ISSUE #3041: Openhab Android app compatability with AndroidAuto
It appears the Openhab Android App is not Android Auto compatible as it doesn't appear in available apps in the Android Auto Launcher.

I have a garage door opener configured in Openhab and I can open my garage door from the app on my phone but now that I have a car with Android Auto I would like to be able open it from the console in the car or even program a steering wheel button to open the door.

maniac103:
Fixes #3041


maniac103:
Implementation comments:
- Min SDK version bump is needed due to some transient dependency (probably pulled in by `lifecycle-service`) requiring it: `Manifest merger failed : uses-sdk:minSdkVersion 21 cannot be smaller than version 23 declared in library [androidx.savedstate:savedstate-ktx:1.4.0]`
- AA uses the primary connection
- Implemented are the following sitemap widgets: text, switch, switch-with-mappings, rollershutter, player, selection
- Maximum page link nesting depth is 5, including the additional screen for action selection (limitation of AA)
- Maximum number of sitemap widgets (== length of widget list) may or may not be limited by AA, exact limitation value can differ between cars and/or AA version
- The sitemap to be used is selected on first startup; for changing the selection one needs to clear the primary server's default sitemap

I'm open for improvement suggestions about
- default sitemap selection
- ways of conveying the AA limitations towards the user 

~~I've also thought about using `GridTemplate` instead of `ListTemplate`, but there's a couple of drawbacks:~~
- ~~grid items don't support inline switches~~
- ~~grid items kinda are based on images, but server side icons are not guaranteed to be present (also handling is complicated, as we'd need to pre-fetch them)~~
- ~~(widget count limit doesn't seem to be larger for grids either)~~

Edit: Changed my mind about this in the meantime. The point about server side images being unreliable still stands, but I've solved that by replacing them with in-app icons for the most likely to be used ones. This is the same approach as used by the iOS app for CarPlay.

Once this is accepted, we probably need a release on the beta track in Play Store, since sideloaded apps don't appear in the AA launcher.

maniac103:
Open issue I don't have a good idea for yet: Loading a different sitemap (e.g. after a server change) also counts against the 5 step limit documented [here](https://developers.google.com/cars/design/create-apps/apps-for-drivers/plan-task-flows#steps-refreshes). This means that after moving back and forth a few times, the app will consume more than 5 steps and crash. This problem is worsened by sitemap selection screen (which counts as a separate step each time), but even if we'd do the sitemap selection on the phone the problem is there and just takes a few toggles more to appear.
I have not yet found a way to restart the step counter.

One option to remedy this may be to not listen for server changes, so that after changing servers one needs to reconnect to AA to reflect the change in the AA app. Not sure if this is acceptable?

robnielsen:
> Loading a different sitemap (e.g. after a server change) also counts against the 5 step limit documented [here](https://developers.google.com/cars/design/create-apps/apps-for-drivers/plan-task-flows#steps-refreshes).

Maybe disable loading a different sitemap when moving? https://source.android.com/docs/automotive/driver_distraction/consume#DrivingState

maniac103:
> Maybe disable loading a different sitemap when moving? https://source.android.com/docs/automotive/driver_distraction/consume#DrivingState

From my understanding,
- that API is only available for AAOS, not for AA
- the 5 step limit takes effect no matter whether the car is parked or not, except
- if [adaptive task limits](https://developers.google.com/cars/design/create-apps/apps-for-drivers/plan-task-flows#task-limits) is active, in which case more than 5 steps may be possible when parked, but
- whether it's enabled is under OEM control, and app behavior on > 5 steps differs hugely between it being enabled or not (delayed display vs. crash)

Given all of those, I think we need to design our UX for the 5 step limit, no matter if parked or not.

mueller-ma:
We could:
- Sitemap selections on phone
- Disallow any sub-pages
- Don't refresh sitemaps when the configuration is changed on the server

That way we shouldn't hit the limit, but no sub-pages might be an issue for some people.

maniac103:
> We could:
> Sitemap selections on phone

Makes sense - any UI recommendations? The current flow follows what we're doing for the app side, and I don't have a good idea yet that doesn't involve some kind of popup.

> Disallow any sub-pages

Doesn't help much, see last point.

> Don't refresh sitemaps when the configuration is changed on the server

I think there's a misunderstanding here. My `Loading a different sitemap (e.g. after a server change)` doesn't refer to server side changes, but to changes in the (connection to the) server. That is, having multiple servers configured in the app and switching between them. We could allow selecting a specific server for AA in the app settings (instead of using the primary server), but that'd be yet another special case in ConnectionFactory, but maybe that's manageable (adding a method to 'ask' ConnectionFactory for a `Connection` for a specific `ServerConfiguration`)?

R0Wi:
@maniac103 this is an awesome feature, thank you! I would be highly interested in doing some beta testing and providing feedback. May I kindly ask if it would be possible to provide some debug APK or would I have to build it myself?

maniac103:
> May I kindly ask if it would be possible to provide some debug APK or would I have to build it myself?

Neither would work for getting it running in your car. For that it needs to be distributed in the Play Store. We can totally do that, I guess (a beta release, that is), but I don't think it makes a lot of sense as long as there's a high probability of crashing whenever the 5 step limit is hit (see above). We'll need some good idea to solve that one.

rkoshak:
One behavior I've noticed in several AA apps I've used is if you interact with the app a bunch of times (e.g. searching for a playlist in YT Music) it eventually pops up a pause preventing further interactions for a time before letting you continue interacting with the screen. I don't know for sure, but this feels realted to that 5 step limit. So maybe something similar would work here.

mhilbush:
> it eventually pops up a pause preventing further interactions for a time before letting you continue interacting with the screen.

I too have seen this after interacting with an app quite a lot. I forget which app it was... might've been Waze, or possibly one of the music apps, Pandora or Spotify.

I don't have many suggestions. One thought was to have a configuration where you can specify the sitemap that you want to use for AA. I expect the sitemap I use in AA will be a lot smaller (and different) than my other sitemaps. Of course if an AA-specific sitemap is not specified it would allow sitemap selection the way it does now.

robnielsen:
@maniac103 are you OK if I have Google Antigravity do a code review?

maniac103:
I don't mind, even though the larger problem I see is the sitemap selection UI question ;-)

mobile/src/main/java/org/openhab/habdroid/car/SelectionScreen.kt
```diff
@@ -0,0 +1,64 @@
...
+    carContext: CarContext,
+    private val title: String,
+    private val options: List<SelectionListItem>,
+    private val currentValue: String?,
+    private val onItemSelected: (item: SelectionListItem) -> Unit
+) : Screen(carContext) {
+    override fun onGetTemplate(): Template {
+        val selectedPosition = options.indexOfFirst { currentValue == it.command }
+        val itemsBuilder = ItemList.Builder()
+            .setOnSelectedListener { index -> onItemSelected(options[index]) }
```
robnielsen: When `currentValue` does not match any option (`selectedPosition < 0`), `noValueRow` is added at index 0 (line 42). However, `setOnSelectedListener` receives the index from `itemsBuilder`, which shifts all `options` by +1:
- Tapping the first option in the list (`index = 1`) selects `options[1]` (the second option).
- Tapping the last option (`index = options.size`) throws `IndexOutOfBoundsException: Index X out of bounds for length X`.
- If the dummy row at index 0 is tapped, it selects `options[0]`.

To fix this, if `selectedPosition < 0`, you either need to adjust for the +1 offset in the listener:
```kotlin
.setOnSelectedListener { index ->
    val optionIndex = if (selectedPosition < 0) index - 1 else index
    if (optionIndex in options.indices) {
        onItemSelected(options[optionIndex])
    }
}
```
Or avoid prepending an unselectable row when using `setOnSelectedListener`.

Also, after `onItemSelected` is called, `screenManager.pop()` should probably be invoked so the user returns to the grid rather than staying on the selection screen with no visible feedback.

mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt
```diff
@@ -0,0 +1,157 @@
...
+            carContext,
+            url,
+            id,
+            nestingDepth,
+            // Omit state portion of the label, as we can't update it anyway without it counting against the step limit
+            title.substringBefore("[").trim(),
+            onPageSelected = { page -> openWidgetListScreen(page, nestingDepth + 1) },
+            onWidgetCommand = { widget, command -> onSendWidgetCommand(widget, command, buildSourceId(id)) }
+        )
+        pageStack += screen
```
robnielsen, line 133: `WidgetGridScreen`s are appended to `pageStack` here, but they are never removed when the user navigates back (via the car back button or header back arrow).

This causes two issues:
1. **Memory & connection leak:** `pageUrls` keeps all previously opened subpages, keeping SSE and polling connections active in `CarService` indefinitely.
2. **Crash on `invalidate()`:** In `handlePageUpdate` and `handleWidgetUpdate`, `pageStack.firstOrNull { it.url == pageUrl }` matches the oldest (already popped and destroyed) screen instance. Calling `invalidate()` on a destroyed `Screen` in the Car App Library throws `IllegalStateException`.

A clean fix is to attach a lifecycle observer to remove the screen when destroyed:
```kotlin
screen.lifecycle.addObserver(object : DefaultLifecycleObserver {
    override fun onDestroy(owner: LifecycleOwner) {
        pageStack.remove(screen)
        onPageListChanged()
    }
})
```

mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt
```diff
@@ -0,0 +1,157 @@
...
+        return carContext.buildSitemapSourceId(sitemapName, pageId, "org.openhab.android.car")
+    }
+
+    private fun openWidgetListScreen(page: LinkedPage, nestingDepth: Int) {
+        Log.d(TAG, "Open widget list for page $page")
+        val screen = createWidgetListScreen(page.link, page.id, page.title, nestingDepth)
+        screen.screenManager.push(screen)
+    }
+
+    private fun ScreenManager.replaceRoot(screen: Screen) {
```
robnielsen, line 150: Regarding the 5-step limit and crashing on sitemap/server changes mentioned in the discussion:

The core reason `replaceRoot()` burns through the 5-step limit is that calling `push(screen)` counts as a step forward in the host's task flow, and `remove(oldRoot)` does not rewind the step counter. With `LoadingScreen` (step 1) -> `SitemapSelectionScreen` (step 2) -> `WidgetGridScreen` (step 3), 3 of the 5 allowed steps are consumed before any subpages are even visited. Any connection re-evaluation or server change calls `replaceRoot()` again, easily reaching 5 steps and crashing.

According to Google's [Car App Library Guidelines](https://developers.google.com/cars/design/create-apps/apps-for-drivers/plan-task-flows#steps-refreshes):
- **Template refreshes on the same `Screen` (via `invalidate()`) do NOT count against the step limit.**
- Popping back to the app's root landing screen resets the task quota.

Instead of swapping `Screen` instances at the root with `replaceRoot()`:
1. Use a single persistent root `Screen` returned in `onCreateScreen()`.
2. Inside its `onGetTemplate()`, dynamically return:
   - Loading `GridTemplate` (if still initializing)
   - `MessageTemplate` with retry (if error / no network)
   - `ListTemplate` (if selecting a sitemap)
   - Root `GridTemplate` (when sitemap is ready)
3. When sitemaps update or the connection reconnects, call `screenManager.popToRoot()` and `invalidate()` on the root screen.

This eliminates `replaceRoot()`, ensures every task flow starts from step 1 at the root screen, and prevents server/network reconnects from consuming steps.

mobile/src/main/java/org/openhab/habdroid/car/CarService.kt
```diff
@@ -0,0 +1,162 @@
...
+    @CallSuper
+    override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
+        dispatcher.onServicePreSuperOnStart()
+        return super.onStartCommand(intent, flags, startId)
+    }
+
+    @CallSuper
+    override fun onDestroy() {
+        dispatcher.onServicePreSuperOnDestroy()
+        connectionHolder.onStop()
```
robnielsen, line 80: `PageConnectionHolderFragment.onDestroy()` is where its coroutine `job.cancel()` is called. Calling only `connectionHolder.onStop()` here leaves background coroutines/SSE jobs inside `connectionHolder` running when `CarService` is destroyed.

Calling `connectionHolder.onDestroy()` here ensures proper coroutine cleanup.

mobile/src/main/java/org/openhab/habdroid/car/CarService.kt
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    override fun onCreateSession(sessionInfo: SessionInfo): Session {
+        dispatcher.onServicePreSuperOnBind()
+        val session = CarSession(
+            initDataFlow.map { result -> result?.map { it.props.sitemaps } },
+            onPageListChanged = this::updateConnections,
+            onSendWidgetCommand = this::sendWidgetCommand
+        )
+        sessions[sessionInfo.sessionId] = session
```
robnielsen, line 95: `sessions` stores active sessions, but destroyed sessions are never removed from this map. When Android Auto reconnects or creates new sessions over time, stale `CarSession` instances accumulate, receive page updates, and keep connections referenced.

Consider removing the session when its lifecycle is destroyed:
```kotlin
session.lifecycle.addObserver(object : DefaultLifecycleObserver {
    override fun onDestroy(owner: LifecycleOwner) {
        sessions.remove(sessionInfo.sessionId)
    }
})
```

mobile/src/main/java/org/openhab/habdroid/car/ErrorScreen.kt
```diff
@@ -0,0 +1,68 @@
...
+            .build()
+
+        val messageBuilder = MessageTemplate.Builder(actualMessage)
+            .setHeader(header)
+            .addAction(retryAction)
+
+        reason?.let {
+            messageBuilder
+                .setDebugMessage(carContext.getString(R.string.car_error_debug_header))
+                .setDebugMessage(it)
```
robnielsen, line 63: The second `.setDebugMessage(it)` call overwrites the debug header set on line 62.

To display both:
```kotlin
reason?.let {
    messageBuilder.setDebugMessage("${carContext.getString(R.string.car_error_debug_header)}\n${it.localizedMessage ?: it.toString()}")
}
```

mobile/src/main/java/org/openhab/habdroid/car/WidgetGridScreen.kt
```diff
@@ -0,0 +1,412 @@
...
+
+            else -> if (widget.linkedPage != null) {
+                WidgetPresentation.PageLinkPresentation(widget, widget.linkedPage)
+            } else {
+                WidgetPresentation.TextPresentation(widget)
+            }
+        }
+
+        val itemBuilder = GridItem.Builder()
+            .setTitle(widget.label)
```
robnielsen: If a widget has no label or an empty label (`""`), `GridItem.Builder.setTitle(widget.label)` throws `IllegalArgumentException: Title cannot be empty` in the Car App Library.

It's safer to provide a fallback:
```kotlin
.setTitle(widget.label.ifEmpty { widget.item?.name ?: "—" })
```

mobile/src/main/java/org/openhab/habdroid/car/WidgetGridScreen.kt
```diff
@@ -0,0 +1,412 @@
...
+            .setHeader(headerBuilder.build())
+
+        val widgetsToShow = widgets
+            ?.filter { shouldShowWidget(it) }
+        if (widgetsToShow == null) {
+            templateBuilder.setLoading(true)
+        } else {
+            val listBuilder = ItemList.Builder()
+            widgetsToShow.forEach { w ->
+                listBuilder.addItem(buildWidgetGridItem(w))
```
robnielsen, line 249: Car App Library templates enforce strict content limits via `ConstraintManager` (usually 6 to 12 items depending on host and screen). Adding more items to `ItemList` than the host allows will throw an uncatchable `IllegalArgumentException`.

You can query the host limit and cap the list to avoid crashing when users have large sitemaps:
```kotlin
val maxItems = carContext.getCarService(ConstraintManager::class.java)
    .getContentLimit(ConstraintManager.CONTENT_LIMIT_TYPE_GRID)
widgetsToShow.take(maxItems).forEach { w ->
    listBuilder.addItem(buildWidgetGridItem(w))
}
```
(The same applies to `SitemapSelectionScreen` with `CONTENT_LIMIT_TYPE_LIST`.)

mobile/src/main/java/org/openhab/habdroid/car/WidgetGridScreen.kt
```diff
@@ -0,0 +1,412 @@
...
+            Fan(true),
+            Alarm(true),
+            Garage(false),
+            Player(false),
+            Thermostat(false)
+        }
+
+        fun guessType(): WidgetType {
+            // First attempt: icon mapping
+            val iconToTypeMapping = mapOf(
```
robnielsen: `iconToTypeMapping` (map with 22 entries) and `tagToTypeMapping` (list of 25 pairs) are allocated on every call to `guessType()` for every widget on every draw pass.

These should be defined once as static constants in a `companion object` to avoid unnecessary GC churn.

robnielsen (commented):
Thank you for working on this! Adding Android Auto support is a fantastic addition to openHAB Android.

Overall the foundation and template mappings look very solid. Below is a detailed review with several important findings:
1. **Critical bug in `SelectionScreen`**: When `selectedPosition < 0`, options are offset by +1, causing wrong item selection and an `IndexOutOfBoundsException` on the last item.
2. **Screen & session leaks**: `CarSession.pageStack` and `CarService.sessions` do not remove destroyed instances, causing memory leaks and crashes when calling `invalidate()` on popped screens.
3. **Addressing the 5-step limit**: Explanation and proposed solution using a single persistent root screen with `invalidate()` rather than `replaceRoot()`.
4. **Automotive constraints**: Need for `ConstraintManager` checks and non-empty title guards to avoid `IllegalArgumentException`.

See the inline comments for details and suggested fixes!

robnielsen:
Regarding the sitemap selection UI, here is an approach that avoids awkward popups while keeping in-car navigation clean and within AA limits:

1. **Dedicated "Android Auto" section in Phone Settings:**
   - Similar to how **Tiles** and **Device Controls** are handled under *Preferences > Miscellaneous*, add an **Android Auto** settings fragment (`CarPreferencesFragment`).
   - This provides a clean place to configure:
     - **Server:** Defaults to Primary Server, but allows picking another configured server.
     - **Sitemap:** A dropdown with:
       - *Same as phone app* (Default)
       - Available sitemaps (populated from the server's cached `ServerProperties`)
   - This avoids any annoying popups on phone launch and lets users intentionally set a small, tailored sitemap for the car without requiring an active connection at preference creation time.

2. **In-Car Auto-Selection Hierarchy:**
   - When AA starts up:
     1. If an AA sitemap is chosen in settings $\rightarrow$ open it directly.
     2. Else if the server only has 1 sitemap $\rightarrow$ auto-select it.
     3. Else if the phone has a default sitemap $\rightarrow$ open that.
     4. Otherwise $\rightarrow$ show `SitemapSelectionScreen` as the initial root template.

3. **Switching Sitemaps from the Car:**
   - On the root sitemap grid (`nestingDepth == 0`), add an action in the header (or an action row): `"Switch Sitemap"`.
   - Clicking it resets the root screen state back to sitemap selection and calls `invalidate()` / `popToRoot()`.
   - This allows changing sitemaps directly from the dashboard whenever needed without hitting the 5-step limit or requiring the phone to clear preferences.

mobile/src/main/java/org/openhab/habdroid/car/ErrorScreen.kt
maniac103: I think that one is wrong. `setDebugMessage(String)` and `setDebugMessage(Throwable)` fill different members of the message builder.

maniac103:
> Regarding the sitemap selection UI, here is an approach that avoids awkward popups while keeping in-car navigation clean and within AA limits:

Suggestion 1 is interesting, albeit a bit inconsistent to how sitemaps are handled elsewhere (as per server setting). What to do if the selected sitemap is deleted from the server though? 

Suggestion 2 more or less is what is currently implemented.

I've thought about suggestion 3 as well, but I'm pretty sure every such invocation will burn through 2 steps.

mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt
maniac103: The guidelines you linked pretty clearly say that a new template increases the step count, so it's irrelevant whether the new template is part of the root screen or of a new one: the step count will be increased either way.

mobile/src/main/java/org/openhab/habdroid/car/ErrorScreen.kt
robnielsen: You're completely right, good catch! Verified that in `MessageTemplate.Builder`, `setDebugMessage(Throwable)` sets `mDebugCause` while `setDebugMessage(String)` sets `mDebugString`, so both are stored independently and preserved.

mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt
robnielsen: The key detail in the guidelines is how the host tracks steps and tasks:

1. `push(screen)` registers as a forward step towards the task limit.
2. Crucially, calling `remove(oldRoot)` does *not* decrement the step count—the host only decrements the count on `pop()`.
3. Because `replaceRoot()` does `push(screen)` followed by `remove(oldRoot)`, each invocation of `replaceRoot()` accumulates an extra forward step without any matching back operation. When network connectivity fluctuates while driving, `sitemapsFlow.collect` repeatedly triggers `replaceRoot()`, consuming a step on each reconnect until hitting 5 and crashing.
4. Furthermore, according to the guidelines, *"A new task begins when a user lands on (or returns to) the app's landing template"* and *"The step count decreases by 1 if the task returns to the previous view with the same content."* Normal forward and back navigation through subpages (`Grid` $\rightarrow$ `Subpage` $\rightarrow$ Back) increments and decrements the counter within a 2–3 step window. The crash happens when `replaceRoot()` consumes steps behind the scenes without the host recognizing a return to the landing screen.

robnielsen:
> Suggestion 1 is interesting, albeit a bit inconsistent to how sitemaps are handled elsewhere (as per server setting). What to do if the selected sitemap is deleted from the server though?

Good point on consistency! Rather than a separate top-level preference, it could live directly inside the per-server settings (`preferences_server.xml` / `ServerEditorFragment`) alongside the phone's default sitemap:
- **Default Sitemap:** (for phone)
- **Android Auto Sitemap:** (with choices: *Same as phone default*, or specific sitemaps from the server)

If the selected sitemap is deleted from the server, we can follow the exact fallback logic `MainActivity` already uses in `selectConfiguredSitemapFromList()`:
1. If the configured AA sitemap isn't found in the server's sitemap list, fall back to the phone's default sitemap.
2. If that's also missing, fall back to the first available sitemap, or the in-car selection screen.

The big advantage of having the AA sitemap selectable in the server settings is that it completely sidesteps the 5-step limit on the road: users pick their lightweight in-car sitemap once on their phone, and Android Auto boots straight into it without needing in-car switching flows.

Regarding `Suggestion 3` and burning steps: if the sitemap is pre-selected on the phone, switching sitemaps in the car becomes an edge case rather than the primary flow. If an in-car switch is initiated, landing on the newly selected sitemap sets it as the new task base, but avoiding in-car switching entirely via phone config is definitely the cleanest UX.

mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt
maniac103: With this diff:

```diff
diff --git a/mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt b/mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt
index 65d16f18a..f36719805 100644
--- a/mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt
+++ b/mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt
@@ -74,7 +74,9 @@ class CarSession(
         screenManager.replaceRoot(createErrorScreen(null, reason))
     }
 
-    override fun onCreateScreen(intent: Intent) = createScreenForCurrentSitemap(latestSitemapResult)
+    override fun onCreateScreen(intent: Intent) = createScreenForCurrentSitemap(latestSitemapResult).also {
+        Log.d(TAG, "Session screen is $it")
+    }
 
     private fun createScreenForCurrentSitemap(result: Result<List<Sitemap>>?): Screen = when {
         result == null -> LoadingScreen(carContext)
@@ -148,6 +150,7 @@ class CarSession(
     }
 
     private fun ScreenManager.replaceRoot(screen: Screen) {
+        Log.d(TAG, "Replacing root with $screen")
         popToRoot()
         // At this point only the root screen is left, which we want to replace with the screen
         // we're going to create. As there's no direct way to do that, we push the new screen to the top
```

I get the following log on startup:
```
09-23 10:00:23.602 26203 26203 D CarSession: Session screen is org.openhab.habdroid.car.LoadingScreen@d47f1bd
09-23 10:00:24.069 26203 26203 D CarSession: Got new sitemap result null
09-23 10:00:24.069 26203 26203 D CarSession: Replacing root with org.openhab.habdroid.car.LoadingScreen@ba9f1d3
09-23 10:00:24.081 26203 26203 D CarSession: Got new sitemap result Success([Sitemap(...), Sitemap(...), Sitemap(...)])
09-23 10:00:24.086 26203 26203 D CarSession: Replacing root with org.openhab.habdroid.car.WidgetGridScreen@507eced
```

So I called `replaceRoot()` twice. Yet, the debug overlay stil says I'm at step 1. As far as I can tell this directly contradicts all AI statements  made above:
- `remove(oldRoot) does not rewind the step counter`
- `LoadingScreen (step 1) -> SitemapSelectionScreen (step 2) -> WidgetGridScreen (step 3), 3 of the 5 allowed steps are consumed before any subpages are even visited` (at least for `LoadingScreen`)
- items 2 to 4 in the list above

The guidelines clearly say:
```
The step count is incremented whenever one of the following occurs:
    A new template is shown
    The same template is shown with new content – unless the new content is a refresh of existing content, as defined in What is a refresh?
```

This means the step tracking is done on the template level, not on the screen level. The `ScreenManager` likely is a client side concept only, not reaching the host. `LoadingScreen` uses a `GridTemplate` for that very purpose.
Having said that, there still seems to be an issue where the updates done in the sitemap flow collection in `CarSession` count against the step counter even if the sitemap (and thus the template result) is the same. It's easy to avoid that by checking if the sitemap actually changed, but I'd like to understand that problem first.

maniac103: OK, `LoadingScreen (step 1) -> SitemapSelectionScreen (step 2) -> WidgetGridScreen (step 3), 3 of the 5 allowed steps are consumed before any subpages are even visited` is correct, because `LoadingScreen` returns a `GridTemplate`, `SitemapSelectionScreen` a `ListTemplate` and `WidgetGridScreen` again a `GridTemplate`. This will be fixed by removing sitemap selection from the Auto UI.

robnielsen:
Nice work on the updates! Moving the icons/tags to companion constants, capping with `ConstraintManager`, and cleaning up `connectionHolder` and sessions look great.

I've left two minor follow-up comments directly in the review threads on `SelectionScreen.kt` and `CarSession.kt` regarding the offset calculation and callback invocation.

mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt
robnielsen: Exactly right on the template types! The host treats successive templates of the same type (`GridTemplate` $\rightarrow$ `GridTemplate`) as an in-place content refresh, which is why `LoadingScreen` $\rightarrow$ `WidgetGridScreen` stays on Step 1. The moment `ListTemplate` (`SitemapSelectionScreen`) intervened in between, it broke the refresh chain because the template class changed. Removing sitemap selection from the in-car UI preserves the `GridTemplate` refresh chain and starts at Step 1 every time.

Regarding why flow updates in `CarSession` might still increment the step counter:
In `CarSession`, `replaceRoot` does `push(newScreen)` and `remove(oldRoot)`. If the root is already displaying a `WidgetGridScreen`, pushing a newly instantiated `WidgetGridScreen` (which initially starts with `widgets = null` and `isLoading = true` before the widgets populate) introduces a transient loading template update. Checking whether the sitemap or connection actually changed before triggering `replaceRoot()`—or calling `invalidate()` on the existing screen when the sitemap URL is the same—will prevent that step from being consumed.

mobile/src/main/java/org/openhab/habdroid/car/SelectionScreen.kt
robnielsen: Careful with the offset calculation in commit 3628f355:
```kotlin
.setOnSelectedListener { index -> onItemSelected(options[index + firstOptionOffset]) }
```
When `selectedPosition < 0`, `noValueRow` is at index 0, so the options in the UI are shifted to indices 1, 2, ...:
- List index 1 is `options[0]`
- List index 2 is `options[1]`

With `index + firstOptionOffset`, clicking the first option (List index 1) computes `1 + 1 = 2`, selecting `options[2]` instead of `options[0]`. And clicking the last option will crash with `IndexOutOfBoundsException`.

The offset needs to be subtracted instead:
```kotlin
.setOnSelectedListener { index ->
    val optionIndex = index - firstOptionOffset
    if (optionIndex in options.indices) {
        onItemSelected(options[optionIndex])
    }
}
```

mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt
robnielsen: Thanks for adding this in commit 07155c79!

One small syntax fix needed on line 130:
```kotlin
screen.lifecycle.onDestroy {
    if (pageStack.remove(screen)) {
        onPageListChanged // missing () to invoke the callback
    }
}
```
Because `onPageListChanged` is a lambda property (`private val onPageListChanged: () -> Unit`), writing it without parentheses is an unused expression statement. It needs parentheses to actually fire: `onPageListChanged()`.

mobile/src/main/java/org/openhab/habdroid/car/SelectionScreen.kt
```diff
@@ -0,0 +1,65 @@
...
+    private val title: String,
+    private val options: List<SelectionListItem>,
+    private val currentValue: String?,
+    private val onItemSelected: (item: SelectionListItem) -> Unit
+) : Screen(carContext) {
+    override fun onGetTemplate(): Template {
+        val selectedPosition = options.indexOfFirst { currentValue == it.command }
+        val firstOptionOffset = if (selectedPosition < 0) 1 else 0
+        val itemsBuilder = ItemList.Builder()
+            .setOnSelectedListener { index -> onItemSelected(options[index + firstOptionOffset]) }
```
robnielsen: There is an offset calculation bug here:
```kotlin
.setOnSelectedListener { index -> onItemSelected(options[index + firstOptionOffset]) }
```
When `selectedPosition < 0`, `noValueRow` is at index 0, so the options in the UI are at indices 1, 2, ...:
- List index 1 is `options[0]`
- List index 2 is `options[1]`

With `index + firstOptionOffset`, clicking the first option (List index 1) computes `1 + 1 = 2`, which selects `options[2]` instead of `options[0]`. And clicking the last option will crash with `IndexOutOfBoundsException`.

The offset needs to be subtracted instead:
```kotlin
.setOnSelectedListener { index ->
    val optionIndex = index - firstOptionOffset
    if (optionIndex in options.indices) {
        onItemSelected(options[optionIndex])
    }
}
```

mobile/src/main/java/org/openhab/habdroid/car/CarSession.kt
```diff
@@ -0,0 +1,163 @@
...
+            id,
+            nestingDepth,
+            // Omit state portion of the label, as we can't update it anyway without it counting against the step limit
+            title.substringBefore("[").trim(),
+            onPageSelected = { page -> openWidgetListScreen(page, nestingDepth + 1) },
+            onWidgetCommand = { widget, command -> onSendWidgetCommand(widget, command, buildSourceId(id)) }
+        )
+        screen.lifecycle.onDestroy {
+            if (pageStack.remove(screen)) {
+                onPageListChanged
```
robnielsen: Missing parentheses on the callback invocation:
```kotlin
screen.lifecycle.onDestroy {
    if (pageStack.remove(screen)) {
        onPageListChanged // missing () to invoke the callback
    }
}
```
Because `onPageListChanged` is a lambda property (`private val onPageListChanged: () -> Unit`), writing it without parentheses is an unused expression statement. It needs `()` to actually be invoked:
```kotlin
onPageListChanged()
```

maniac103:
I've moved AA sitemap selection to the phone app settings now. Please let me know what you (not the AI :wink: ) think about that solution.

robnielsen:
> I've moved AA sitemap selection to the phone app settings now. Please let me know what you (not the AI 😉 ) think about that solution.

Sounds good to me

robnielsen (commented):
Left two follow-up comments on the latest commit regarding error handling in the new sitemap preference dialog and a minor defensive indexing check in `SelectionScreen`.

mobile/src/main/java/org/openhab/habdroid/ui/preference/widgets/AndroidAutoSitemapPreference.kt
```diff
@@ -0,0 +1,263 @@
...
+                        .let { loadSitemaps(it.conn!!) }
+
+                    sitemapsResult.onSuccess { sitemaps ->
+                        prepareSitemapRecyclerView(sitemaps)
+                        sitemapsLoaded = true
+                        updateVisibilities(true)
+                        updateOkButtonState()
+                    }
+                    sitemapsResult.onFailure { cause ->
+                        // TODO: error
```
robnielsen: If loading sitemaps fails (e.g. phone is offline, primary server unreachable, or SSL error), `sitemapsLoaded` remains `false`.

Because `updateVisibilities(true)` has:
```kotlin
binding.progress.isVisible = enabled && !sitemapsLoaded
```
the progress spinner will continue spinning indefinitely with no error message, and the positive "OK" button stays permanently disabled since `selectedSitemap` is null.

We should hide `binding.progress`, show an error message (or Toast/Snackbar), and/or reset `binding.enabled.isChecked = false` so the user isn't stuck in a loading state. Also, if `sitemaps` list is empty, it would be helpful to show a hint (like `R.string.error_empty_sitemap_list`) explaining why nothing is listed.

mobile/src/main/java/org/openhab/habdroid/car/SelectionScreen.kt
```diff
@@ -0,0 +1,65 @@
...
+    private val title: String,
+    private val options: List<SelectionListItem>,
+    private val currentValue: String?,
+    private val onItemSelected: (item: SelectionListItem) -> Unit
+) : Screen(carContext) {
+    override fun onGetTemplate(): Template {
+        val selectedPosition = options.indexOfFirst { currentValue == it.command }
+        val firstOptionOffset = if (selectedPosition < 0) 1 else 0
+        val itemsBuilder = ItemList.Builder()
+            .setOnSelectedListener { index -> onItemSelected(options[index - firstOptionOffset]) }
```
robnielsen: Good fix on the subtraction!

One small defensive suggestion here: when `selectedPosition < 0`, index 0 in the list is `noValueRow`. While it has `isEnabled = false`, if a rotary controller or DPAD navigation in a car head unit happens to trigger a selection on index 0, `0 - 1 = -1` will throw an `IndexOutOfBoundsException`.

Using safe indexing avoids any risk:
```kotlin
.setOnSelectedListener { index ->
    options.getOrNull(index - firstOptionOffset)?.let { onItemSelected(it) }
}
```

maniac103: Yes, but I'd consider the possibility of selecting a disabled item a major issue in the library, given preventing selection is the whole point of disabling items. I'd like to only add this if really necessary, is there any evidence of that?

robnielsen: Fair point. Relying on the host contract to enforce `isEnabled = false` makes sense unless we actually see host-side issues in practice. Happy to leave it as-is.

PR #3998, opened 2026-05-11
More connection factory fixups

### LINKED ISSUE #3997: After start oh-Android first waits, and then spells that it tries to connect
Since openHAB-Android 3.20.0-beta, including version 3.20.2-beta, when I start openHAB-Android it sometimes spells “This page doesn’t contain any widgets that are set to be visible”, as shown in the first image below.  Some seconds later it tries to connect locally and displays what is supposed to be displayed (sitemap “A2” - "A2" is in the first screenshot).

Sometimes, I do not know since when or if this always has been so, when I start openHAB-Android, it shows flacking lines, as if something is loading.  Instead of the sitemap name it shows “openHAB” - which might be the name of the configured server.  Later it shows as message “Connecting to local URL”.

* It would be good, if on start openHAB directly tries to connect, instead of first waiting and then connecting.

Image

Image

maniac103:
Fix some more regressions from #3972.

Should fix #3997

### DIFF
--- a/mobile/src/foss/java/org/openhab/habdroid/core/CloudMessagingHelper.kt
+++ b/mobile/src/foss/java/org/openhab/habdroid/core/CloudMessagingHelper.kt
@@ -41,7 +41,7 @@ object CloudMessagingHelper {
     private val TAG = CloudMessagingHelper::class.java.simpleName
 
     @Suppress("UNUSED_PARAMETER")
-    fun onConnectionUpdated(context: Context, connection: CloudConnection?) {}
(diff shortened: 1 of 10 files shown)


maniac103:
Sadly doesn't fix #3997 yet.

maniac103:
Should be OK now.

maniac103:
Haven't seen the issue in #3997 again with this. @mueller-ma OK to :ship: ?

mueller-ma:
I'm fine with that :+1:

PR #4026, opened 2026-07-30
Fix FCM registration

### LINKED ISSUE #4025: Cloud notifications not working on new install
My pixel 10 was replaced today and now I can't get notifications working. This was working fine earlier on my old Pixel 10 running the July update.

The connection to the cloud is working since I can view the Main UI. In the android app in settings under notifications it says "Notification registration is in progress". My new phone does not appear in the device list on myopenhab.org.

Is there anything I should be searching for in the log to help diagnose?

I'm running version 3.20.7 of the android app and openHAB version 5.1.4.

maniac103:
The changes in #3998 changed the registration trigger from flow listener to direct callback, but only did so for the 'active server is not the same as primary server' path. Move the callback to a place that's executed for any primary cloud server update.

Fixes #4025


maniac103:
@mueller-ma This might warrant a 3.20.8.

mueller-ma:
I pushed it as beta to Play Store.

PR #4027, opened 2026-07-31
Hide intro due to low contrast

mueller-ma:
The contrast between buttons and background is too low to be visible in some situations. Until fixed disable the intro for now.

https://community.openhab.org/t/unable-to-leave-welcome-screen/169896/10


maniac103:
I'd rather fix the colors instead -> #4050

PR #4044, opened 2026-09-18
Isolate WebView data per server

### LINKED ISSUE #4035: MainUI ist Not switching to selected Server while Remote URL / myopenhab.org
### Actual behaviour
Describe us what happens

I'm using two different OH-Server (OH 5.1.3 + 5.2)

When i Switch from one to the other, Main UI ist still connected to the one, which was choose while starting my Android Phone. 
Checked empty Cache but this do Not Help.
Need to restart my Phone after choosing.

### Expected behaviour
The choose Servers MainUI hast to be reachable after switching 

This ist only when I'm Not at Home and connect via myopenhab remotely.
At Home everything ist fine
 
### Steps to reproduce
1. Setup 2 ther Servers with different filled overview pages
2. Connect both to myopenhab
3. Switch remotely between Servers 

How can i use Demo mode?

#### Client
* Android version: 16
* Device model: Poco F7 pro
* App version 3.21.1
* Build flavor Full
* Device language: german

#### Server
* Server version: 5.1.3 / 5.2
* Reverse Proxy: no
* Authentication method : 

#### openHAB Cloud
* myopenhab.org: 

* Authentication method user and password

#### App log
<details>
  <summary>Click to expand</summary>

``
To Long to Post here
```

</details>


#### openHAB Server log
<details>
  <summary>Click to expand</summary>

```
Not required.
If you have installed openHAB via `apt-get`, the log is located under `/var/log/openhab2/openhab.log`
```

</details>

### PR DESCRIPTION
**Pull request by stamateviorel**
when two servers are reached over the same address (both over myopenhab, or the same local ip on different sites) the webview shares cache, saved login and cookies between them, so after switching server main ui keeps showing the previous one until the cache is cleared. this gives every server its own webview profile, on old webviews without profile support it clears the http cache and saved login when the server changes. androidx.webkit stays on 1.14.0 because 1.15 needs minSdk 23.

tested a debug build of this on my phone with two servers, after switching server main ui now shows the right one without clearing the cache.

Fixes #4035

### CODE REVIEW

File: mobile/src/main/java/org/openhab/habdroid/ui/ViewExtensions.kt [hunk 1]
```diff
@@ -54,6 +61,64 @@ fun WebView.setUpForConnection(connection: Connection) {
...
 
+private fun buildWebViewProfileName(serverId: Int) = "server_$serverId"
+
+/**
+ * Servers may share an origin (e.g. all myopenhab.org users, or the same local IP at different sites),
+ * so HTTP cache, HTTP auth cache, cookies and DOM storage must not be shared between them.
+ *
+ * Must be called before the WebView is used for anything else.
+ */
+fun WebView.isolateForServer(serverId: Int) {
```
maniac103: I wonder whether, instead of all those extensions, it wouldn't make more sense to have a manager object whose lifetime is tied to the activity or even the application. That manager object could then know about the active server (thus no need to store last known server to prefs) and the fragment could ask the manager to set up the WebView.

Thoughts about that?

stamateviorel: makes sense, moved it into a WebViewManager that lives with the application, the fragment just asks it to set up the webview. the pref is gone too, it was only there for old webviews without profile support because the http cache survives an app restart, there it now clears the cache on first use after app start instead. tested the new version on my phone with two servers.

maniac103:
I'd like to suggest some slight refactoring on top of the current state: https://github.com/openhab/openhab-android/commit/b9d8cdbdd7a2f9db43072b3c25084244d0f48350
Thoughts on this?

stamateviorel:
fine for me, cleaner like that. pushed your commit plus a small fix on top for the prefs listener, it was only weakly referenced and registered too late, details are in the commit message.

PR #4047, opened 2026-09-19
Fall back to web UI if there's no sitemap

stamateviorel:
when the server returns no sitemaps the app now opens main ui, or habpanel if main ui isn't there, instead of the empty sitemap list error. if something else was already requested (shortcut, start page) that still wins. without any web ui the old error and retry stay.


mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt
```diff
@@ -845,7 +845,14 @@ class MainActivity : AbstractBaseActivity() {
                 is ServerProperties.Companion.PropsSuccess -> {
                     serverProperties = result.props
                     updateDrawerServerEntries()
-                    if (result.props.sitemaps.isEmpty()) {
+                    val fallbackUi = listOf(WebViewUi.MAIN_UI, WebViewUi.HABPANEL)
+                        .firstOrNull { ui -> result.props.hasWebViewUiInstalled(ui) }
+                    if (result.props.sitemaps.isEmpty() && fallbackUi != null) {
```
maniac103: Maybe convert to `when`:

```
when {
    result.props.sitemaps.isNotEmpty() -> {
        ...
    }
    fallbackUi != null -> {
        ...
    }
    else -> {
        ...
    }
}
```

stamateviorel: done, changed it to when.

PR #4048, opened 2026-09-19
Add server and UI selection to toolbar

stamateviorel:
with more than one server i switch a lot between them and between sitemap and main ui, and every time that means leaving main ui, opening the drawer, picking the server and opening main ui again. this makes the toolbar title a button that opens a list with one row per server and ui (home · main ui, home · sitemap, office · main ui ...), the current one is highlighted. there is also an option to keep the toolbar visible in main ui and an option to hide the drawer, its entries then move to the toolbar menu, both are off by default.

the first commit is the same small fix as in #4046, i will rebase when that one is merged. i have it running as a test build on my phone, i can add screenshots. if you want it done in another way just tell me.


maniac103:
> this makes the toolbar title a button that opens a list with one row per server and ui

How does this help with having to close and reopen Main UI on switch given the toolbar isn't visible if Main UI is open?
Personal opinion: I think having redundant UI like this is not a good idea. The better solution IMO would be a way for the app to inject drawer items into Main UI's drawer. We discussed this some time in the past; will dig out the discussion once I have some time available.

stamateviorel:
there is an option in the pr, "show toolbar in main ui", with that on the toolbar stays visible while main ui is open so the button is always there. but i understand the point about redundant ui. adding the servers to main ui's own drawer would be cleaner, if you find that discussion i can have a look at what is needed on the app side.

maniac103:
One part of it is [here](https://github.com/openhab/openhab-android/pull/3706), but I'm sure we also discussed this elsewhere.

In any case, one would need to implement an injection JS interface in MainUI first, and once that is implemented, implement it here.

stamateviorel:
closing this one, makes sense to do the main ui side first like you said. i will look at an interface for main ui so the app can add its entries to the drawer and come back with that.

